PR #6, opened 2022-09-11
fix: insertUnexpander

pcpthm:


JamesGallicchio:
I'm not entirely sure what the bug originally was, but the unexpander for insert seems to work fine on most recent `main`. @pcpthm is this fix still necessary?

pcpthm:
Apparently it is already fixed in <https://github.com/leanprover/std4/commit/34d312dc4b2abd0c476d0405e41161f22abadd7a>. Closing.

PR #7, opened 2022-09-12
chore: update nightly 09-11

gebner:


Std/Data/Queue.lean
```diff
@@ -1,53 +0,0 @@
-/-
```
digama0, line 1: Why is this file being removed?

gebner: Because it's now in core.

gebner: I think I asked you about the naming conflict (because it's called `Std.Queue`), but you didn't complain so I assumed that you were okay with it.

digama0: Oh I see, it's in `Init` now. The docstrings should be relocated there though.

digama0: "this is only a temporary placeholder" should also probably be removed

digama0:
Is bors not running on this repo?

gebner:
> Is bors not running on this repo?

Only if you invite it.  I don't have any permissions here so I can't even check.

PR #10, opened 2022-09-16
chore(Std/Lean/Meta): fix repeat'

semorrison:
Per https://github.com/leanprover/std4/commit/ee3bcfce030ae6ae41016286402947c33faa9cca#r84122462


semorrison:
Oops, Mario did this in the meantime.

PR #11, opened 2022-09-18
chore: factor out definitions in guardExpr

mcdoll:
As discussed in [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/guardHyp/near/299386329) this PR makes it possible to define `guardExpr` and `guardHyp` for `conv` mode. For consistency, we apply the change also to `guardTarget`.


digama0:
The PR contents were merged manually in 678c117 along with the conv tactics themselves.

PR #14, opened 2022-10-11
feat: port List.Perm

Vtec234:


Std/Data/List/Lemmas.lean
```diff
@@ -1063,6 +1063,22 @@ theorem Pairwise.imp {α R S} (H : ∀ {a b}, R a b → S a b) :
   | _, .nil => .nil
   | _, .cons h₁ h₂ => .cons (H ∘ h₁ ·) (h₂.imp H)
 
+theorem pairwise_append_comm_of_symm {α R} (s : ∀ {a b}, R a b → R b a) {l₁ l₂ : List α} :
+    Pairwise R (l₁ ++ l₂) ↔ Pairwise R (l₂ ++ l₁) := by
+  have : ∀ l₁ l₂ : List α, (∀ x : α, x ∈ l₁ → ∀ y : α, y ∈ l₂ → R x y) →
+    ∀ x : α, x ∈ l₂ → ∀ y : α, y ∈ l₁ → R x y := fun l₁ l₂ a x xm y ym => s (a y ym x xm)
+  have and_left_comm : ∀ a b c, (a ∧ b ∧ c) = (b ∧ a ∧ c) := fun a b c => by
```
Vtec234, line 1070: @digama0 note that because we don't have the commutativity typeclasses, in particular we don't have [this lemma](https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Init/Logic.lean#L351) so I had to prove it inline.

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,168 @@
...
+  | cons  : ∀ (x : α) {l₁ l₂ : List α}, Perm l₁ l₂ → Perm (x::l₁) (x::l₂)
+  /-- Lists with the first two elements swapped are permutations of each other. -/
+  | swap  : ∀ (x y : α) (l : List α), Perm (y::x::l) (x::y::l)
+  /-- The `Perm` relation is transitive. -/
+  | trans : ∀ {l₁ l₂ l₃ : List α}, Perm l₁ l₂ → Perm l₂ l₃ → Perm l₁ l₃
+
+open Perm
+
+/-- Notation for two lists being permutations of each other. See also `Perm`. -/
+infixl:50 " ~ " => Perm
```
fgdorais, line 38: I don't think this should be a global notation.

fgdorais: Consider using `Setoid` instead?

digama0: I think it is fine to have it (and `<+`, `<+:` etc) be scoped in `List`.

Vtec234: Done.

Vtec234:
This is now included in #38.

PR #17, opened 2022-10-19
feat: basic filter lemmas

Vtec234:
~~Depends on #14 .~~


digama0:
what happened here? Are the lemmas superceded or no longer necessary?

Vtec234:
@digama0 they are included in #38, and I thought it might be easier to review in the context of that. Happy to reinstate this PR if you'd prefer that!

digama0:
No problem, just make sure to mention the replacement when you close a PR next time so future readers can find their way around.

PR #19, opened 2022-10-24
feat: Option.eq_some_of_isSome

semorrison:
This is used in mathlib3, and if we don't put it in std it will be orphaned in its own file in mathlib4, since if came from `init/data/option/instances.lean`


digama0:
:racing_car: by d1d14d6

PR #20, opened 2022-10-29
feat: iteration abstractions

JamesGallicchio:
- [ ] Typeclasses for foldl, foldr, and non-directional folds
- [ ] `Iterable` typeclass for collections that can produce a `Stream`-like iterator object. Used to implement `Std.Range` folds, and more generally for any collection where external iteration is the more natural iteration

### DIFF
--- a/Std/Classes/Collections/Iterable.lean
+++ b/Std/Classes/Collections/Iterable.lean
@@ -6,9 +6,9 @@ Authors: James Gallicchio
 
 import Std.Classes.Collections.Fold
 
-/-! ## External Iterators -/
+namespace Std.Collections
 
-namespace Std
+/-! ## External Iterators -/
 
 /--
 Exposes `toIterator`, which gives an object that can traverse
--- a/Std/Classes/Collections/Iterable.lean
+++ b/Std/Classes/Collections/Iterable.lean
@@ -49,23 +49,23 @@ An `Iterable` for which the iterator object has some well-ordered
 notion of size, which decreases at each `step`. This is used to
 automatically generate `Fold` implementations for `Iterable`s.
 -/
-class FinIterable (C : Type u) (τ : Type v) extends Iterable C τ where
+class WFIterable (C : Type u) (τ : Type v) extends Iterable C τ where
   /-- Iterator type `ρ` has a well-founded relation. -/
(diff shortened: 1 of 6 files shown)


Std/Classes/Collections/Iterable.lean
```diff
@@ -0,0 +1,174 @@
...
+Exposes `toIterator`, which gives an object that can traverse
+the collection one item at a time.
+-/
+class Iterable (C : Type u) (τ : outParam (Type v)) where
+  /-- The iterator object type. -/
+  ρ : Type w
+  /-- Convert a collection to its iteration object. -/
+  toIterator : C → ρ
+  /-- Step the iterator to its next value. -/
+  step : ρ → Option (τ × ρ)
```
kmill, line 23: Is there a reason to reinvent the Lean 4 core `ToStream` and `Stream` interfaces? https://github.com/leanprover/lean4/blob/bb8cc08de85f3daf67c6fb3feda245681c4b56fd/src/Init/Data/Stream.lean#L43

`ToStream` is the `toIterator` function, and `Stream` is the `step` function.

kmill: While it might be more intuitive for you, do you want to be duplicating pre-existing API? If `Iterable` exists, I think at the least it should be packaging up `ToStream` and `Stream` into a single bundle.

There's also the rule of thumb we've learned in mathlib, which is that type fields in structures/classes are usually not a great idea, though *maybe* it's justified here because you're specifically wanting to hide the iterator type.

JamesGallicchio: `Iterable` was intended to compete with the `ToStream` API, yeah. At one point I had some relationship between `Iterable` and `ToStream` (I think the former gave an instance of the latter?) which can of course be added back.

I think the reason why the type field works with `Iterable` is that they almost always appear as transitory, compile-time constants with relatively clean definitional equalities. Consumers of `Iterable` never care what the iterator type is, since the only thing you can do is apply `step`.

The `Fold` based API in leancolls is much more fleshed out than the `Iterable` API, so having early feedback here is helpful.

joehendrix:
Close for inactivity.  Reopen if needed.

PR #21, opened 2022-11-01
feat: Nat.sqrt

JamesGallicchio:
Unclear whether this should be an extern'd function in Lean core. But this implementation has fine enough performance for now.

### DIFF
--- a/Std/Data/Nat/Basic.lean
+++ b/Std/Data/Nat/Basic.lean
@@ -15,3 +15,19 @@ instance : Dvd Nat := ⟨fun a b => ∃ c, b = a * c⟩
 
 /-- Sum of a list of natural numbers. -/
 protected def sum (l : List Nat) : Nat := l.foldr (·+·) 0
+
+/--
+Integer square root function. Implemented via Newton's method.
+-/
+def sqrt (n : Nat) : Nat :=
+  let guess := n / 2
+  if guess = 0 then n else
+  let rec iter (guess : Nat) : Nat :=
+    let next := (guess + n / guess) / 2
+    if h : guess ≤ next then
+      guess
+    else
+      have : next < guess := Nat.gt_of_not_le h
+      iter next
+  iter guess
+termination_by iter guess => guess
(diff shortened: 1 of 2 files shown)


Std/Data/Nat/Basic.lean
```diff
@@ -15,3 +15,19 @@ instance : Dvd Nat := ⟨fun a b => ∃ c, b = a * c⟩
 
 /-- Sum of a list of natural numbers. -/
 protected def sum (l : List Nat) : Nat := l.foldr (·+·) 0
+
+/--
+Integer square root function. Implemented via Newton's method.
+-/
+def sqrt (n : Nat) : Nat :=
+  let guess := n / 2
+  if guess = 0 then n else
```
digama0, line 24:
```suggestion
  if n ≤ 1 then n else
  let guess := 1 <<< ((n - 1).log2 >>> 1 + 1)
```

Std/Data/Nat/Basic.lean
```diff
@@ -15,3 +15,19 @@ instance : Dvd Nat := ⟨fun a b => ∃ c, b = a * c⟩
...
+  let rec iter (guess : Nat) : Nat :=
+    let next := (guess + n / guess) / 2
```
digama0, line 26:
```suggestion
    let next := (guess + n / guess) >>> 1
```
Would be nice if the compiler automatically did strength reduction but it doesn't seem to looking at the IR.

Std/Data/Nat/Basic.lean
```diff
@@ -15,3 +15,19 @@ instance : Dvd Nat := ⟨fun a b => ∃ c, b = a * c⟩
...
+    if h : guess ≤ next then
+      guess
+    else
+      have : next < guess := Nat.gt_of_not_le h
+      iter next
```
digama0, line 31:
```suggestion
    if h : next < guess then iter next else guess
```

Std/Data/Nat/Basic.lean
hargoniX: I'll add that to the new code generator when the LetExpr refactoring is done.

digama0:
On second thought, I think the sqrt lemmas should be put in a separate PR. It was almost ready to merge before but now it looks like a bit more work.

PR #25, opened 2022-11-11
feat: port ext lemmas from mathlib3

mcdoll:
I don't know where the notation belongs, but I am pretty sure the `ext` lemmas have to be in `Std4`.


digama0:
Let's not have the notation in here. `Nat` and `Int` also don't have notation introduced in std.

PR #32, opened 2022-11-21
chore: bump to nightly-2022-11-21

semorrison:


digama0:
Duplicate of #31

Std/Data/RBMap/Basic.lean
```diff
@@ -976,7 +976,7 @@ smaller than or equal to `k`, if it exists.
 
 -- TODO: add proof
 private unsafe def modifyImpl (t : RBMap α β cmp) (k : α) (f : β → β) : RBMap α β cmp :=
-  t.modifyP (cmp k ·.1) (fun (a, b) => (a, f b)) (wf := lcProof)
+  @RBSet.modifyP _ _ t (cmp k ·.1) (fun (a, b) => (a, f b)) lcProof
```
semorrison, line 979: It seems this may be a regression in Lean 4, and we can't specify a trailing typeclass argument using `name := ...`. Reported as https://github.com/leanprover/lean4/issues/1868, and this can probably just wait on that.

digama0: Heh, this is also a duplicate: leanprover/lean4#1867

semorrison:
Oops... Sorry, limited interet still but should have checked.

PR #38, opened 2022-11-27
Prove correctness of `HashMap.findEntry?/insert/erase`

Vtec234:
- The PR includes a partial port of `List.Perm`. I was initially hoping to do a more principled port to mathport to begin with, but its dependency tree turned out to be a bit deep, so that might have to wait.

### DIFF
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -86,10 +86,7 @@ def empty (capacity := 8) : Imp α β :=
   empty' nBuckets (Nat.isPowerOfTwo_nextPowerOfTwo _)
 
 /-- Calculates the bucket index from a `hash` value. -/
-/- Remark: we use a C implementation because this function is performance critical. -/
-@[extern c inline "(size_t)(#2) & (lean_unbox(#1) - 1)"]
 def mkIdx {sz : Nat} (hash : UInt64) (h : sz.isPowerOfTwo) : {u : USize // u.toNat < sz} :=
-  -- TODO(WN): Use the (provably equivalent) bitwise `&&&` in the spec as well like libLean does
   ⟨hash.toUSize % sz, USize.modn_lt _ (Nat.pos_of_isPowerOfTwo h)⟩
 
 /--
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -20,34 +20,34 @@ namespace Imp
 The bucket array of a `HashMap` is a nonempty array of `AssocList`s.
 (This type is an internal implementation detail of `HashMap`.)
 -/
-def Bucket (α : Type u) (β : Type v) := {b : Array (AssocList α β) // 0 < b.size}
+def Buckets (α : Type u) (β : Type v) := {b : Array (AssocList α β) // 0 < b.size}
 
-namespace Bucket
+namespace Buckets
 
 /-- Construct a new empty bucket array with the specified capacity. -/
-def mk (buckets := 8) (h : 0 < buckets := by decide) : Bucket α β :=
+def mk (buckets := 8) (h : 0 < buckets := by decide) : Buckets α β :=
   ⟨mkArray buckets .nil, by simp [h]⟩
 
 /-- Update one bucket in the bucket array with a new value. -/
-def update (data : Bucket α β) (i : USize)
-    (d : AssocList α β) (h : i.toNat < data.1.size) : Bucket α β :=
+def update (data : Buckets α β) (i : USize)
+    (d : AssocList α β) (h : i.toNat < data.1.size) : Buckets α β :=
   ⟨data.1.uset i d h, (Array.size_uset ..).symm ▸ data.2⟩
 
 /--
 The number of elements in the bucket array.
 Note: this is marked `noncomputable` because it is only intended for specification.
 -/
-noncomputable def size (data : Bucket α β) : Nat := .sum (data.1.data.map (·.toList.length))
+noncomputable def size (data : Buckets α β) : Nat := .sum (data.1.data.map (·.toList.length))
 
 /-- Map a function over the values in the map. -/
-@[specialize] def mapVal (f : α → β → γ) (self : Bucket α β) : Bucket α γ :=
+@[specialize] def mapVal (f : α → β → γ) (self : Buckets α β) : Buckets α γ :=
   ⟨self.1.map (.mapVal f), by simp [self.2]⟩
 
 /--
 The well-formedness invariant for the bucket array says that every element hashes to its index
 (assuming the hash is lawful - otherwise there are no promises about where elements are located).
 -/
-structure WF [BEq α] [Hashable α] (buckets : Bucket α β) : Prop where
+structure WF [BEq α] [Hashable α] (buckets : Buckets α β) : Prop where
   /-- The elements of a bucket are all distinct according to the `BEq` relation. -/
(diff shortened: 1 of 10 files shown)


Std/Data/HashMap/Lemmas.lean
```diff
@@ -0,0 +1,346 @@
...
+  Pairwise.imp beq_nonsense_1 (Pairwise_bne_bucket bkts H i h)
+
+/-- It is a bit easier to reason about `foldl (append)` than `foldl (foldl)`, so we use this
+(less efficient) variant of `toList` as the mathematical model. -/
+def toListModel (bkts : Buckets α β) : List (α × β) :=
+  -- Note(WN): the implementation is  `bkts.foldl` rather than `bkts.data.foldl` because we need
+  -- to reason about array indices in some of the theorems.
+  bkts.val.foldl (init := []) (fun acc bkt => acc ++ bkt.toList)
+
+protected theorem foldl_cons_fn (l₁ l₂ : List α) :
```
Vtec234, line 50: I made these local `simp`s since they seemed a bit too specific to put in `List.Lemmas`.

Std/Data/HashMap/Lemmas.lean
```diff
@@ -0,0 +1,346 @@
...
+      haveI := mkIdx m.buckets.property a
+      m.buckets.exists_of_toListModel_update_WF hWF.right this.1
+        ((m.buckets.1[this.1.toNat]'this.2).replace a b) this.2
+    rw [hUpd, hTgt]
+    have hL₁ : ∀ ab ∈ l₁, ¬(ab.fst == a) := fun ab h hEq =>
+      Nat.ne_of_lt (LawfulHashable.hash_eq hEq ▸ hProp ab h) rfl
+    have ⟨p, hMem, hP⟩ := any_eq_true.mp (AssocList.contains_eq a _ ▸ hContains)
+    simp [eraseP_append_right _ hL₁,
+      eraseP_append_left (p := fun ab => ab.fst == a) hP _ hMem]
+    -- begin cursed manual proofs
```
Vtec234, line 264: This is the gnarliest proof. Golfing welcome!

Std/Data/HashMap/Basic.lean
```diff
@@ -88,26 +88,26 @@ def empty (capacity := 0) : Imp α β :=
     else ⟨nbuckets, Nat.zero_lt_of_ne_zero h⟩
   empty' n n.2
 
-/-- Calculates the bucket index from a hash value `u`. -/
-def mkIdx {n : Nat} (h : 0 < n) (u : USize) : {u : USize // u.toNat < n} :=
-  ⟨u % n, USize.modn_lt _ h⟩
+/-- Calculates the bucket index for the key `a`. -/
+def mkIdx [Hashable α] {n : Nat} (h : 0 < n) (a : α) : {u : USize // u.toNat < n} :=
+  ⟨(hash a |>.toUSize) % n, USize.modn_lt _ h⟩
```
digama0, line 93: This change is performance sensitive, I would prefer not to change it without benchmarking first (and we don't have the setup to do that right now, but at least some evidence that the generated IR is not adversely affected would be good). In particular, the previous function was not monomorphized but the new one is.

Also, I believe that lean4 has since changed this part of the `Lean.HashMap` implementation a bit, and we might want to consider porting that change to std as well. (I think it doesn't change the defeqs so hopefully the proofs are unaffected...)

Std/Data/HashMap/Lemmas.lean
digama0: They don't look like good global simp lemmas but I don't see any reason not to have them as regular theorems in `List.Lemmas`.

Std/Data/HashMap/Basic.lean
Vtec234: Indeed it was, but it [was found](https://github.com/leanprover/lean4/commit/25beba66242118eed7857cf3666c43f74ba7d7cb) to "make no difference in practice". Do you think it's still worth porting then?

Vtec234:
Ok @digama0, I backported the `extern` implementation of `mkIdx` using `&` but not the specification. I think this is justifiable because they are provably equal under the assumption that `sz.isPowerOfTwo`, but I left a TODO to fix the spec up.

Std/Data/HashMap/Basic.lean
```diff
@@ -148,15 +148,15 @@ where
       let target := es.foldl reinsertAux target
       go (i+1) source target
     else target
-termination_by _ i source _ => source.size - i
+  termination_by go i source _ => source.size - i
```
digama0, line 151:
```suggestion
termination_by go i source _ => source.size - i
```

Std/Data/HashMap/Basic.lean
```diff
@@ -77,37 +77,36 @@ A "load factor" of 0.75 is the usual standard for hash maps, so we return `capac
...
+def empty (capacity := 8) : Imp α β :=
+  let nBuckets := numBucketsForCapacity capacity |>.nextPowerOfTwo
+  empty' nBuckets (Nat.isPowerOfTwo_nextPowerOfTwo _)
 
-/-- Calculates the bucket index from a hash value `u`. -/
-def mkIdx {n : Nat} (h : 0 < n) (u : USize) : {u : USize // u.toNat < n} :=
-  ⟨u % n, USize.modn_lt _ h⟩
+/-- Calculates the bucket index from a `hash` value. -/
+/- Remark: we use a C implementation because this function is performance critical. -/
+@[extern c inline "(size_t)(#2) & (lean_unbox(#1) - 1)"]
```
gebner, line 90: I'd be extremely surprised if this makes a difference in practice, and wouldn't believe it without actual benchmarks.  Leo hasn't seen any performance improvements from this change in Lean 4 either.  (It's also incorrect, because `sz` might be boxed in general.)

The enormous practical downside of this change is of course that you can no longer use std4 without compiling it.  We didn't require that before.

(If you're concerned about performance, I'd be more concerned about using hashes directly.  The Hashable instances are not particularly clever and often return results divisible by powers of two.  See also https://github.com/leanprover/lean4/pull/1840)

Vtec234: > It's also incorrect, because `sz` might be boxed in general.

I'm confused by this point; there is a `lean_unbox` call, is that insufficient?

> The enormous practical downside of this change is of course that you can no longer use std4 without compiling it.

It's true! It is surprising to me that the interpreter does not fallback on the Lean implementation when it encounters an `extern` symbol with no code loaded. Maybe it should do this instead of crashing? Anyhow I will revert the change.

gebner: > I'm confused by this point; there is a `lean_unbox` call, is that insufficient?

I'm sorry, picked the wrong terminogy here.  The `sz` argument might not be a scalar number (directly stored in the argument word using tagging), but it could be a heap-allocated bignum (where the argument stores the pointer).  In that case `lean_unbox` would return the pointer (right-shifted by one).

Vtec234: That's true, we are assuming the map has fewer than `2^63` (or whatever the largest directly storable value is) entries, so perhaps `sz : USize` would be better.

Vtec234:
I added some higher-level lemmas which however need `List` theorems that are mostly already in mathlib. Rather than porting more stuff adhoc, I am marking this as draft since it may as well wait for `Data.List.Perm` to be mathported in full.

Vtec234:
This PR has rotten, I will readd hashmaps in a new one after #89.

PR #39, opened 2022-11-27
feat: add HashSet.insertMany

semorrison:
I've also sent this to Lean 4 core as https://github.com/leanprover/lean4/pull/1889, but I'd like this merged anyway so we can use aesop. I'll also need to PR to aesop to finish removing the collision. :-(

### DIFF
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -346,14 +346,3 @@ def ofListWith (l : List (α × β)) (f : β → β → β) : HashMap α β :=
     | some v => m.insert p.1 <| f v p.2
 
 end Std.HashMap
-
-namespace Lean.HashSet
-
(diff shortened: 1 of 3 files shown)


digama0:
Thid should be in `Std.Lean.HashMap` since it's about "the other hashmap".

Std/Data/HashMap/Basic.lean
```diff
@@ -344,3 +344,5 @@ def ofListWith (l : List (α × β)) (f : β → β → β) : HashMap α β :=
     match m.find? p.1 with
     | none   => m.insert p.1 p.2
     | some v => m.insert p.1 <| f v p.2
+
+end Std.HashMap
```
digama0, line 348:
```suggestion
```

semorrison:
Obsolete; this has landed in core now.

PR #42, opened 2022-12-01
feat: various functions for manipulating mvars

JLimperg:
---
These are utility functions which I need in Aesop (and which I think are also generally useful additions to the Lean API).


JLimperg:
Updated to fix linter issues.

PR #43, opened 2022-12-01
feat: add Lean.MVarId.tryClearMany'

JLimperg:
This is a variant of core's `tryClearMany` which sorts the given `FVarId`s by context order before trying to clear them. This makes sure we clear everything we can.

Depends on #42


JLimperg:
Updated to fix linter issues (sorry for the many problems; I've now learned how to run the linter locally).

PR #44, opened 2022-12-01
feat: add Lean.MVarId.assertHypotheses' 

JLimperg:
This is a variant of core's `assertHypotheses` which allows us to specify the `BinderInfo` and `LocalDeclKind` of the asserted hypotheses.

Depends on #42


JLimperg:
Updated to fix linter issues.

PR #46, opened 2022-12-01
feat: add Lean.Meta.getUnusedNames

JLimperg:
This is a variant of `LocalContext.getUnusedName` which (a) may or may not be more efficient and (b) allows us to get multiple unused names in one go.

### DIFF
--- a/Std/Data/Char.lean
+++ b/Std/Data/Char.lean
@@ -1,0 +1,21 @@
+/-
+Copyright (c) 2022 Jannis Limperg. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jannis Limperg
+-/
+
+import Std.Tactic.RCases
+
+namespace Char
+
+private theorem csize_eq (c) :
+    String.csize c = 1 ∨ String.csize c = 2 ∨ String.csize c = 3 ∨
+    String.csize c = 4 := by
+  simp only [String.csize, utf8Size]
+  repeat (first | split | (solve | simp))
+
+theorem csize_pos (c) : 0 < String.csize c := by
+  rcases csize_eq c with _|_|_|_ <;> simp_all
+
+theorem csize_le_4 (c) : String.csize c ≤ 4 := by
+  rcases csize_eq c with _|_|_|_ <;> simp_all
--- a/Std/Data/String.lean
+++ b/Std/Data/String.lean
@@ -1,0 +1,19 @@
+/-
+Copyright (c) 2022 Jannis Limperg. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jannis Limperg
+-/
+
+namespace String
+
+/--
+If `pre` is a prefix of `s`, i.e. `s = pre ++ t`, return the remainder `t`.
+-/
+def dropPrefix? (s : String) (pre : String) : Option Substring :=
+  let s := s.toSubstring
+  if s.take pre.length == pre.toSubstring then
+    s.drop pre.length
+  else
+    none
+
+end String
--- a/Std/Data/String.lean
+++ b/Std/Data/String.lean
@@ -4,16 +4,42 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Jannis Limperg
 -/
 
+import Std.Data.Char
+import Std.Data.Nat.Lemmas
+
+namespace Substring
+
+/--
+If `pre` is a prefix of `s`, i.e. `s = pre ++ t`, return the remainder `t`.
+-/
+def dropPrefix? (s : Substring) (pre : Substring) : Option Substring :=
+  go 0 (pre.stopPos - pre.startPos)
+where
+  /-- Auxiliary definition for `dropPrefix?`. -/
+  go (start : String.Pos) (stop : String.Pos) : Option Substring :=
+    if h : start ≥ stop then
+      some { s with startPos := s.startPos + start }
+    else
+      let cs := s.get start
+      let cp := pre.get start
+      if cs == cp then
+        have : start.byteIdx + String.csize cs ≤ stop.byteIdx + 4 :=
+          Nat.add_le_add (Nat.le_of_lt $ Nat.not_le.mp h) (Char.csize_le_4 _)
+        have : stop.byteIdx + 4 - (start.byteIdx + String.csize cs) <
+                stop.byteIdx + 4 - start.byteIdx :=
+          Nat.sub_add_lt_sub this (Char.csize_pos _)
+        go (start + cs) stop
+      else
+        none
+termination_by go start stop => stop.byteIdx + 4 - start.byteIdx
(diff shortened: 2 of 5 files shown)


Std/Data/String.lean
```diff
@@ -0,0 +1,19 @@
...
+
+/--
+If `pre` is a prefix of `s`, i.e. `s = pre ++ t`, return the remainder `t`.
+-/
+def dropPrefix? (s : String) (pre : String) : Option Substring :=
+  let s := s.toSubstring
+  if s.take pre.length == pre.toSubstring then
+    s.drop pre.length
+  else
+    none
```
digama0, line 17: This looks like it could easily be written to take `Substring` arguments as well. How about having both versions and implement the string version in terms of the substring version?

Std/Lean/Meta/UnusedNames.lean
```diff
@@ -0,0 +1,138 @@
...
+
+private def parseIndexSuffix (s : Substring) : Option Nat :=
+  if s.isEmpty then
+    none
+  else if s.front == '_' then
+    s.drop 1 |>.toNat?
+  else
+    none
+
+set_option linter.missingDocs false in
```
digama0, line 22: # :pleading_face:

Std/Data/String.lean
JLimperg: Done. Wasn't super easy, but the function is more efficient as a result.

Std/Lean/Meta/UnusedNames.lean
```diff
@@ -0,0 +1,142 @@
...
+    let acc := Array.mkEmpty n
+    match lctx.getUnusedUserNameIndex suggestion with
+    | none => loop (acc.push suggestion) (n - 1) 1
+    | some i => loop acc n i
+  where
+    /-- Auxiliary definition. -/
+    loop (acc : Array Name) (n i : Nat) : Array Name :=
+      match n with
+      | 0 => acc
+      | n + 1 => loop (acc.push $ suggestion.appendIndexAfter i) n (i + 1)
```
digama0, line 120:
```suggestion
where
  /-- Auxiliary definition for `getUnusedUserNames`. -/
  loop (acc : Array Name) (n i : Nat) : Array Name :=
    match n with
    | 0 => acc
    | n + 1 => loop (acc.push $ suggestion.appendIndexAfter i) n (i + 1)
```

Std/Data/Char.lean
```diff
@@ -0,0 +1,31 @@
...
+Authors: Jannis Limperg
+-/
+
+namespace Char
+
+theorem csize_eq (c) :
+    String.csize c = 1 ∨ String.csize c = 2 ∨ String.csize c = 3 ∨
+    String.csize c = 4 := by
+  simp only [String.csize, utf8Size]
+  repeat (first | split | (solve | simp))
```
digama0, line 13: Let's make this one private for now. I would like to rewrite it later to remove it in favor of the two following theorems.

Std/Data/Char.lean
```diff
@@ -0,0 +1,31 @@
...
+
+theorem csize_pos (c) : 0 < String.csize c := by
+  match csize_eq c with
+  | .inl eq => ?_1
+  | .inr (.inl eq) => ?_2
+  | .inr (.inr (.inl eq)) => ?_3
+  | .inr (.inr (.inr eq)) => ?_4
+  all_goals simp_all
```
digama0, line 21:
```suggestion
  | .inl eq | .inr (.inl eq) | .inr (.inr (.inl eq)) | .inr (.inr (.inr eq)) => simp_all
```

Std/Lean/Meta/UnusedNames.lean
JLimperg: Done (also for the other `where` clauses).

Std/Data/Char.lean
JLimperg: Done.

Std/Data/Char.lean
JLimperg: Yes! Much nicer.

Std/Lean/Meta/UnusedNames.lean
```diff
@@ -0,0 +1,142 @@
...
+  for ldecl in lctx do
+    let hypName := ldecl.userName
+    if hypName.hasMacroScopes then
+      continue
+    match ldecl.userName.matchUpToIndexSuffix suggestion with
+    | .exactMatch => minSuffix := updateMinSuffix minSuffix 1
+    | .noMatch => continue
+    | .suffixMatch i => minSuffix := updateMinSuffix minSuffix (i + 1)
+  minSuffix
+  where
```
digama0, line 83: You might have missed it in the other code suggestion but these where blocks should be dedented so that `def` lines up with `where` (and  `termination_by` when applicable).

Std/Data/Char.lean
```diff
@@ -0,0 +1,23 @@
...
+  simp only [String.csize, utf8Size]
+  repeat (first | split | (solve | simp))
+
+theorem csize_pos (c) : 0 < String.csize c := by
+  rcases csize_eq c with _|_|_|_ <;> simp_all
+
+theorem csize_le_4 (c) : String.csize c ≤ 4 := by
+  rcases csize_eq c with _|_|_|_ <;> simp_all
+
+end Char
```
digama0, line 23:
```suggestion
```
Std usually doesn't use `end` at the end of files.

Std/Data/String.lean
```diff
@@ -0,0 +1,47 @@
...
+      if cs == cp then
+        have : start.byteIdx + String.csize cs ≤ stop.byteIdx + 4 :=
+          Nat.add_le_add (Nat.le_of_lt $ Nat.not_le.mp h) (Char.csize_le_4 _)
+        have : stop.byteIdx + 4 - (start.byteIdx + String.csize cs) <
+                stop.byteIdx + 4 - start.byteIdx :=
+          Nat.sub_add_lt_sub this (Char.csize_pos _)
+        go (start + cs) stop
+      else
+        none
+  termination_by _ start stop => stop.byteIdx + 4 - start.byteIdx
```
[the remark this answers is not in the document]
JLimperg: Ah sorry, I thought this clause belongs to `go`, but it actually belongs to the whole `def`.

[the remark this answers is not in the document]
digama0: Yeah, that would be nice... currently you have to specify which function the termination_by applies to, that's what the `_` after `termination_by` is for.

Std/Data/Char.lean
JLimperg: Fixed (but I think enforcing this is not super useful).

digama0: It's not a strict rule. I prefer to use `end` iff the namespace being closed is a different one than the "main" one for the file, e.g. if the whole file is about `List` but there is some `Array.toList` definition which is pushed to the end so it's not in the same namespace block as everything else.

PR #47, opened 2022-12-02
feat: pairing heaps

negiizhao:


Std/Data/PairingHeap.lean
```diff
@@ -0,0 +1,372 @@
...
+def Heap.toArrayUnordered (s : Heap α) : Array α :=
+  s.foldTree id (fun a c s r => s (c (r.push a))) #[]
+
+/--
+The well formedness predicate for a heap node.
+It asserts that:
+* If `a` is added at the top to make the forest into a tree, the resulting tree
+  is a `le`-min-heap (if `le` is well-behaved)
+-/
+def Heap.nodeWellFormed (le : α → α → Bool) (a : α) : Heap α → Prop
```
digama0, line 214:
```suggestion
def Heap.NodeWF (le : α → α → Bool) (a : α) : Heap α → Prop
```
and `WF` for the other one

digama0 (approved):
Looks very good for a first PR! I have only minor comments.

digama0:
Oh, I almost forgot to mention: I did a bit of research on pairing heaps and now I'm wondering if we should rename this type to `PPairingHeap` for "persistent pairing heap", because this is actually the pure-functional version of pairing heaps that doesn't support the decrease-key operation. I'm not sure if we want to implement imperative pairing heaps, but we could, using an array to store the nodes and doing destructive updates.

negiizhao:
I think a heap with decrease-key operation is likely needed. But I'm not sure we should change the name to `PPairingHeap`. As I said in the documentation, a persistent pairing heap is slow in the worst case (a pairing heap has amortized O(log n) time, but when it is used as a persistent heap, you might repeat a slow operation).

PR #48, opened 2022-12-02
feat: add basic functions on Ord, Ordering

JLimperg:


Std/Data/Ord.lean
```diff
@@ -0,0 +1,99 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jannis Limperg
+-/
+
+namespace Ordering
+
+/--
+Check whether the ordering is 'equal'.
+-/
+def isEQ : Ordering → Bool
```
digama0, line 12:
```suggestion
def isEq : Ordering → Bool
```

Std/Data/Ord.lean
```diff
@@ -0,0 +1,99 @@
...
+def compareOn [ord : Ord β] (f : α → β) (x y : α) : Ordering :=
+  compare (f x) (f y)
+
+
+namespace Ord
+
+/--
+Derive a `BEq` instance from an `Ord` instance.
+-/
+def toBEq (ord : Ord α) : BEq α where
```
digama0, line 61: All of the functions here need to be `protected`.

Std/Data/Ord.lean
```diff
@@ -0,0 +1,99 @@
...
+Derive the lexicographic order on products `α × β` from orders for `α` and `β`.
+-/
+def lex (_ : Ord α) (_ : Ord β) : Ord (α × β) :=
+  lexOrd
+
+/--
+Create an order which compares elements first by `ord₁` and then, if this
+returns 'equal', by `ord₂`.
+-/
+def lex' (ord₁ ord₂ : Ord α) : Ord α where
```
digama0, line 98: how is this different from `lex`?

Std/Data/Ord.lean
JLimperg: Done, also for the other functions in this namespace.

Std/Data/Ord.lean
JLimperg: Done. What's the general rule wrt `protected`?

Std/Data/Ord.lean
JLimperg: `lex` is for pairs.

JLimperg:
I've also added another function, `compareLex'`, and generalised the type of `compareLex`.

Std/Data/Ord.lean
JLimperg: Oh, I just realised: the reason it was `isEQ` in the first place is that core has `isLE`. So this introduces an inconsistency.

JLimperg: Oh, I just realised: the reason it was `isEQ` is that core has `isLE`. So this introduces an inconsistency.

digama0: The suggestion is just for `Eq`, not `LE`. We have discussed this naming "inconsistency" in the mathlib4 stream and decided that it's not inconsistent, because `LE` is an acronym for less-equal and `Eq` is an abbreviation. (Just don't think too hard about `Ne`...)

JLimperg: Okay, I've changed `isGe` etc. back to `isGE`. I've also added `isNe`.

PR #55, opened 2022-12-06
feat: support Monad* classes in dangerousInstance linter

gebner:


digama0:
Would it be better to use an `@[init]` static hashmap instead?

gebner:
Yeah, this list should really be extensible.  I was actually thinking of adding a `semiOutParam` annotation to core, but that might be a hard sell.

digama0:
If the list is as short as it seems to be, I think it is fine just to have a meta function or `command` elab / attribute syntax that adds this to the linter state, and sell it as configuration for the dangerousInstances linter since that's all it is at the moment.

PR #58, opened 2022-12-07
feat: add mdata-safe variants of some Expr functions

JLimperg:
Also add `forallArity` and `lambdaArity`.


digama0:
I'm going to merge this, but maybe these should be upstreamed (replacing the originals) to core?

PR #64, opened 2022-12-13
chore: override Div Int instance to use Int.ediv

semorrison:


Std/Data/Int/Basic.lean
```diff
@@ -40,12 +40,10 @@ def toNat' : Int → Option Nat
 
 There are three main conventions for integer division,
 referred here as the E, F, T rounding conventions.
-All three pairs satisfy the identity x % y + (x / y) * y = x
-unconditionally.
+All three pairs satisfy the identity `x % y + (x / y) * y = x` when `y ≠ 0`,
```
digama0, line 43:
```suggestion
All three pairs satisfy the identity `x % y + (x / y) * y = x` unconditionally,
```
It's still true for `y = 0`.

semorrison: Oops.

Std/Data/Int/Basic.lean
```diff
@@ -109,6 +107,13 @@ This pair satisfies `div x y = round_to_zero (x / y)`.
 `Int.div` and `Int.mod` are defined in core lean.
 -/
 
+/--
+Core Lean provides instances using T-rounding division, i.e. `Int.div` and `Int.mod`.
+We override these here.
+-/
+instance (priority := 1001) : Div Int := ⟨Int.ediv⟩
+instance (priority := 1001) : Mod Int := ⟨Int.emod⟩
```
digama0, line 115:
```suggestion
instance : Div Int := ⟨Int.ediv⟩
instance : Mod Int := ⟨Int.emod⟩
```
We don't need higher priority here since you cannot import this file without importing the core file first, so this one will always win.

Std/Data/Int/DivMod.lean
```diff
@@ -204,48 +204,48 @@ theorem add_ediv_of_dvd_left {a b c : Int} (H : c ∣ a) : (a + b).ediv c = a.ed
...
+@[simp] protected theorem ediv_self {a : Int} (H : a ≠ 0) : a / a = 1 := by
   have := Int.mul_ediv_cancel 1 H; rwa [Int.one_mul] at this
 
 /-! ### mod -/
 
-theorem mod_def' (m n : Int) : m % n = mod m n := rfl
+theorem mod_def' (m n : Int) : m % n = emod m n := rfl
 
-@[simp, norm_cast] theorem ofNat_mod (m n : Nat) : ofNat (m % n) = m % n := rfl
+@[simp, norm_cast] theorem ofNat_mod (m n : Nat) : ofNat (m % n) = mod m n := rfl
```
digama0, line 229: This one should not be a simp/norm_cast lemma since it introduces `tmod`

Std/Data/Rat/Basic.lean
```diff
@@ -34,25 +34,28 @@ instance : Repr Rat where
 
 theorem Rat.den_pos (self : Rat) : 0 < self.den := Nat.pos_of_ne_zero self.den_nz
 
+-- Note: `Rat.normalize` uses `Int.div` internally,
+-- but we may want to refactor to use `/` (`Int.ediv`)
```
digama0, line 38: As you have presumably already discovered, the difference does not matter in this case because it is only used for exact division, where all conventions agree. I am inclined to use `.div` in the implementation and `/` in any definition unfolding simp lemmas and side conditions where possible.

PR #67, opened 2022-12-14
feat: Add nix flake

Anderssorby:
Add a nix flake setup for those who use lean through nix. This is a step in my project to integrate lake into nix.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,197 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1667395993,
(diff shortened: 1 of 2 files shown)


digama0:
@gebner Can you help test this and set up CI to make sure it works?

digama0:
As things are currently I don't think I can accept this PR because I can't make sure it is working nor keep it up to date. I think nix needs to find a way to do builds which does not involve making all lean 4 projects maintain two independent build pipelines.

Anderssorby:
@digama0 Yes, my project Nix.lean intends to fix that by providing native Nix support to lean which then can be added to lake. I guess we can postpone it until then, but I'm using this update to run Nix.lean (I should probably switch to using lake, but I like using nix so much).

joehendrix:
Closed per discussion above.  Reopen if needed.

PR #68, opened 2022-12-18
feat: add basics for Array.repeat

ammkrn:


Std/Data/Array/Basic.lean
```diff
@@ -29,3 +29,11 @@ def flatten (arr : Array (Array α)) : Array α :=
 /-- Turns `#[a, b]` into `#[(a, 0), (b, 1)]`. -/
 def zipWithIndex (arr : Array α) : Array (α × Nat) :=
   arr.mapIdx fun i a => (a, i)
+
+/-- Seeds an array with `n` copies of a `a`. Tail-recursive, reserves a capacity of `n` -/
+def «repeat» {α : Type u} (a : α) (n : Nat) : Array α :=
```
digama0, line 34: This is `Array.mkArray`

ammkrn: Sorry, I've been using https://leanprover-community.github.io/mathlib4_docs/ to find things and `List.repeat` was what lead me down the path. Is the choice of `Array.mkArray` instead of `Array.replicate` a conscious decision or historical accident?

digama0: It looks like an accident to me, especially the namespace stuttering.

PR #70, opened 2022-12-21
chore: bump to 2022-12-20

gebner:


Std/Data/AssocList.lean
```diff
@@ -163,7 +163,7 @@ with key equal to `a` to have key `a` and value `b`.
 
 @[simp] theorem replace_toList [BEq α] (a : α) (b : β) (l : AssocList α β) :
     (replace a b l).toList =
-    l.toList.replaceF (bif ·.1 == a then (a, b) else none) := by
+    l.toList.replaceF (bif ·.1 == a then some (a, b) else none) := by
```
gebner, line 166: I fear like this is fallout from changing the monad coercion code.

PR #74, opened 2022-12-28
feat: add `LinearCmp` class

fgdorais:
The main idea is to make Std.Classes.Order more usable.

* Add the `LinearCmp` class and some lemmas for `OrientedCmp` and `TransCmp`.

* Add class abbreviations `OrientedOrd`, `TransOrd`, `LinearOrd` for the case when `Ord.compare` is the comparator.

* Rename `OrientedOrd.symm` to `cmp_swap_symm`. I think "swap symmetry" is a better name for this property, this is analogous to "conjugate symmetry" and "skew symmetry" in linear algebra.

* Rename `cmp_eq_gt` to `cmp_gt_lt_symm`. This lemma was widely used but I couldn't figure out what `cmp_eq_gt` was supposed to mean. The new name makes more sense to me, hopefully nobody is strongly attached to the old name.

* Rename `cmp_refl` to `cmp_eq_refl` to avoid ambiguity.

* Add `cmp` prefixes everywhere to avoid name clashes. This allows opening the `OrientedOrd`, `TransOrd`, `LinearCmp` namespaces. (This was tested for RBMap, with no issues.)


fgdorais:
There was some discussion on Zulip about this PR. The outcome was unclear but I personally liked the idea of this PR being adapted and moved to the `Ord` namespace. (Or even ditched entirely for a better API in the `Ord` namespace.)

@JamesGallicchio I remember you had ideas about this. Would you like to discuss these further?

JamesGallicchio:
No particular thoughts. I think maybe adding LinearCmp class should be separated out from the rest of this PR (which seems to just be QOL improvements/fleshing out the existing Order file).

Do we have a use case yet for LinearCmp?

fgdorais:
I'm closing this for now but I still need the `TransOrd` and `LinearOrd` classes, so there will be some kind of modern revival in the future.

PR #80, opened 2023-01-08
fix: definitions of Rat.floor and Rat.ceil

hrmacbeth:
https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Rat.2Efloor


digama0:
We still want the first if statement.

hrmacbeth:
Will do, but how come?

digama0:
Because we don't want to do a bignum division if we don't have to.

PR #85, opened 2023-01-10
feat: add `LawfulAlternative` class

fgdorais:
Mostly for completeness, the main use case is `Option`. Another would be a `NonDet` monad.

I only included "linear" alternative laws (everything is evaluated the same number of times on both sides) since these are the only ones we can expect for a control-type monad. For data-type monads, there are additional distributive laws like `((a <|> b) >>= f) = ((a >>= f) <|> (b >>= f))` that might be expected to hold. These probably belong in a `MonadPlus` extension, or similar.

Migrated from closed PR [lean4#1999](https://github.com/leanprover/lean4/pull/1999)

### DIFF
--- a/Std/Classes/LawfulAlternative.lean
+++ b/Std/Classes/LawfulAlternative.lean
@@ -1,17 +1,21 @@
 import Std.Classes.LawfulMonad
+import Lean.Linter
 
-/-- combination class of `Monad` and `Alternative` with a common `Applicative` component -/
+/-- Combination class of `Monad` and `Alternative` with a common `Applicative` component -/
 class MonadAlternative (m : Type _ → Type _) extends Monad m, Alternative m
 
-/-- basic laws for `Alternative` -/
+-- Missing docs in core
+attribute [nolint docBlame] MonadAlternative.failure MonadAlternative.orElse
+
+/-- Basic laws for `Alternative` -/
 class LawfulAlternative (f : Type _ → Type _) [Alternative f] extends LawfulApplicative f : Prop where
-  /-- map failure is failure -/
+  /-- Map failure is failure -/
   map_failure (g : α → β) : g <$> (failure : f α) = failure
-  /-- map distributes over `orElse` -/
+  /-- Map distributes over `orElse` -/
   map_orElse (g : α → b) (x y : f α) : g <$> (x <|> y) = (g <$> x <|> g <$> y)
-  /-- failure is left identity for `orElse`  -/
+  /-- Failure is a left identity for `orElse`  -/
   failure_orElse (x : f α) : (failure <|> x) = x
-  /-- failure is right identity for `orElse` -/
+  /-- Failure is a right identity for `orElse` -/
   orElse_failure (x : f α) : (x <|> failure) = x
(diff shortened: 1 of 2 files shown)


Std/Classes/LawfulAlternative.lean
```diff
@@ -0,0 +1,73 @@
...
+  /-- Map distributes over `orElse` -/
+  map_orElse (g : α → b) (x y : f α) : g <$> (x <|> y) = (g <$> x <|> g <$> y)
+  /-- Failure is a left identity for `orElse`  -/
+  failure_orElse (x : f α) : (failure <|> x) = x
+  /-- Failure is a right identity for `orElse` -/
+  orElse_failure (x : f α) : (x <|> failure) = x
+  /-- `orElse` is associative -/
+  orElse_assoc (x y z : f α) : (x <|> (y <|> z)) = ((x <|> y) <|> z)
+  /-- `pure` always succeeds -/
+  pure_orElse (x : α) (y : f α) : (pure x <|> y) = pure x
```
semorrison, line 23: I've had occasion to want `Alternative List`, with operations `failure := []`, `<|> := ++` and `pure a := [a]`. This would satisfy all these axioms except for `pure_orElse`. 

There seems to be some debate in the Haskell community about the laws for `Alternative`: [1](https://en.wikibooks.org/wiki/Haskell/Alternative_and_MonadPlus#Alternative_and_MonadPlus_laws) [2](https://wiki.haskell.org/MonadPlus_reform_proposal).

I'd certainly like `List` to be `LawfulAlternative`!

fgdorais:
This is an old PR. It's evident that everyone has managed to live without it for a while. I think it needs a quick review to see whether it should be closed, revived, or whether something else might be better.

Std/Classes/LawfulAlternative.lean
```diff
@@ -0,0 +1,68 @@
...
+  failure_orElse     := by intros; rfl
+  orElse_failure x   := by cases x <;> rfl
+  orElse_assoc x _ _ := by cases x <;> rfl
+
+namespace ReaderT
+
+@[simp] theorem run_failure [Monad m] [Alternative m] (ctx : ρ) : (failure : ReaderT ρ m α).run ctx = failure := rfl
+
+@[simp] theorem run_orElse [Monad m] [Alternative m] (x y : ReaderT ρ m α) (ctx : ρ) : (x <|> y).run ctx = (x.run ctx <|> y.run ctx) := rfl
+
```
semorrison, line 39: I'd prefer these (and some declarations below that don't mention `LawfulAlternative` find files higher up to live in. (Perhaps even in the `Lean` directory.)

semorrison:
I'm in favour.

I would like to get to the point that we have a nice `NonDet` monad!

Std/Classes/LawfulAlternative.lean
fgdorais: Agreed. (This PR is currently low priority for me so it might take me a while to get back to this one. I eventually will though!)

fgdorais:
Closing due to age. Some parts are rescued in new PR.

PR #87, opened 2023-01-13
feat: add instances about `List`s

urkud:


JamesGallicchio:
LGTM. Do these instances exist in Mathlib? If so, does one of the linters catch duplicate instances, or do we need to be careful about deleting them when Mathlib is updated?

digama0:
there is no duplicate instance linter, you just have to watch out for that

PR #88, opened 2023-01-16
feat: add monad instance for `Thunk`

fgdorais:
This is missing from core because the old compiler doesn't know how to eliminate `pure _ >>= _` for `Thunk`. Nevertheless, I think this is a useful instance and the issue may be fixed in the new compiler.

See [related discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Why.20isn't.20Thunk.20a.20Monad.3F/near/321680164).

### DIFF
--- a/Std/Classes/LawfulMonad.lean
+++ b/Std/Classes/LawfulMonad.lean
@@ -5,6 +5,16 @@ Authors: Mario Carneiro
 -/
 import Std.Logic
 
+/-- Monad instance for `Thunk`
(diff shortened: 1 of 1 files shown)


fgdorais:
Is there a "depends on new compiler" tag?

fgdorais:
This is getting archaic... closing sounds good to me. What do you think @digama0 @joehendrix @semorrison ?

PR #89, opened 2023-01-19
chore: de-mathlib List.Perm

JamesGallicchio:
Depends on #100.


Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,1029 @@
...
+  | swap (x y : α) (l : List α) : Perm (y :: x :: l) (x :: y :: l)
+  /-- `Perm` is transitive. -/
+  | trans {l₁ l₂ l₃ : List α} : Perm l₁ l₂ → Perm l₂ l₃ → Perm l₁ l₃
+
+
+open Perm (swap)
+
+/-- `Perm l₁ l₂` or `l₁ ~ l₂` asserts that `l₁` and `l₂` are permutations
+  of each other. This is defined by induction using pairwise swaps. -/
+infixl:50 " ~ " => Perm
```
Vtec234, line 52: We may want to make this `scoped` as per [comment](https://github.com/leanprover/std4/pull/14#discussion_r1002771413).

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,1029 @@
...
+
+
+-- Porting note: used rec_on in mathlib3; lean4 eqn compiler still doesn't like it
+protected theorem Perm.symm {l₁ l₂ : List α} (p : l₁ ~ l₂) : l₂ ~ l₁ :=
+  p.rec
+    .nil
+    (fun x _ _ _ r₁ => .cons x r₁)
+    (fun x y l => .swap y x l)
+    (fun _ _ r₁ r₂ => .trans r₂ r₁)
+
```
Vtec234, line 67: Style-wise, is there any reason for double newlines between theorems? Afaict one-line is the norm (just because more fits on the screen then).

Std/Data/HashMap/WF.lean
```diff
@@ -321,7 +321,8 @@ theorem WF.filterMap {α β γ} {f : α → β → Option γ} [BEq α] [Hashable
     rw [← List.pairwise_map (R := (¬ · == ·))] at this ⊢
     exact this.sublist (H3 l.toList)
   · simp [Array.getElem_eq_data_get] at h ⊢
-    have := H.out.2.2 _ h; simp [AssocList.All] at this ⊢
+    have := H.out.2.2 _ h
+    simp [-List.mem_filterMap, AssocList.All] at this ⊢
```
Vtec234, line 325: This looks vaguely suspicious, perhaps a `simpNF` issue?

Std/Logic.lean
```diff
@@ -349,15 +349,18 @@ variable {p q : α → Prop} {b : Prop}
...
 theorem Exists.imp (h : ∀ a, p a → q a) : (∃ a, p a) → ∃ a, q a
   | ⟨a, hp⟩ => ⟨a, h a hp⟩
 
 theorem Exists.imp' {β} {q : β → Prop} (f : α → β) (hpq : ∀ a, p a → q (f a)) :
     (∃ a, p a) → ∃ b, q b
   | ⟨_, hp⟩ => ⟨_, hpq _ hp⟩
 
-@[simp] theorem exists_imp : ((∃ x, p x) → b) ↔ ∀ x, p x → b :=
-  ⟨fun h x hpx => h ⟨x, hpx⟩, fun h ⟨x, hpx⟩ => h x hpx⟩
+theorem exists_imp : ((∃ x, p x) → b) ↔ ∀ x, p x → b := by simp
```
Vtec234, line 363: Why no longer `@[simp]`?

Std/Data/List/Perm.lean
digama0: I assume it is a side effect of removing the `#align` lines. I agree the double newlines should be removed

Std/Data/HashMap/WF.lean
JamesGallicchio: Yeah. `List.mem_filterMap` is simp in mathlib, but I'm not familiar enough with this proof nor with the list simp normal form to diagnose this.

Std/Logic.lean
JamesGallicchio: I thought that lemmas provable by `simp` are generally not supposed to be marked `simp`?

Vtec234: My impression was that unless the proof that `by simp` comes up with is `same_lemma_different_name`, this can still be useful. It is putting a bunch of rewrite rules together into a compound one, a shortcut for the rewrite system. @digama0 should deny/confirm

digama0: Just because the proof is `by simp` doesn't mean that simp can do the same proof in context. And indeed it can also be useful as a shortcut like Wojciech says.

JamesGallicchio: In this case it can be proven by `apply forall_exists_index`. That is probably why the linter was complaining. So I think I should replace it with the `apply` proof for clarity, and leave the lemma no longer `@[simp]`?

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,889 @@
...
+  p.symm.eq_singleton.symm
+
+theorem singleton_perm_singleton {a b : α} : [a] ~ [b] ↔ a = b := by simp
+
+theorem perm_cons_erase [DecidableEq α] {a : α} {l : List α} (h : a ∈ l) : l ~ a :: l.erase a :=
+  let ⟨_l₁, _l₂, _, e₁, e₂⟩ := exists_erase_eq h
+  e₂.symm ▸ e₁.symm ▸ perm_middle
+
+@[elab_as_elim]
+theorem perm_induction_on {P : List α → List α → Prop} {l₁ l₂ : List α} (p : l₁ ~ l₂) (h₁ : P [] [])
```
Vtec234, line 198: We should restore the induction principle that was deleted [here](https://github.com/leanprover-community/mathlib4/commit/c5a07b2253b534effb902ccf5492d83480baed3a#diff-8cf784dd8c93e9893e31189155685d581e0e8915443d84c23756701010cc68b4L96), this one is not usable with `induction using`.

Std/Data/HashMap/WF.lean
JamesGallicchio: Okay, investigated -- the lemma is fine, the proof is the weird one here. The proof was trying to reason at the List level about filterMap instead of at the prop level using membership.

JamesGallicchio:
Splitting into smaller PRs (with contents identical to the PR as of closing this)

semorrison:
@JamesGallicchio sorry this has been so slow. Could you please merge `main`, and then set back to `awaiting-review`?(Please ping Mario on Zulip if you don't have permissions to change labels.)

Vtec234:
Hi @JamesGallicchio , I have resolved conflicts. Is still still a WIP or good to go?

JamesGallicchio:
I was planning to do a thorough code review and make a matching bump PR for mathlib (I was waiting for #100 to be finalized). I have time tomorrow if nobody gets around to it today!

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,908 @@
...
+    Perm.append_right _⟩
+
+theorem subperm_cons (a : α) {l₁ l₂ : List α} : a :: l₁ <+~ a :: l₂ ↔ l₁ <+~ l₂ :=
+  ⟨fun ⟨l, p, s⟩ => by
+    match s with
+    | .cons _ s' => exact (p.subperm_left.2 <| (sublist_cons _ _).subperm).trans s'.subperm
+    | .cons₂ _ s' => exact ⟨_, p.cons_inv, s'⟩
+  , fun ⟨l, p, s⟩ => ⟨a :: l, p.cons a, s.cons₂ _⟩⟩
+
+theorem cons_subperm_of_mem {a : α} {l₁ l₂ : List α} (d₁ : Nodup l₁) (h₁ : a ∉ l₁) (h₂ : a ∈ l₂)
```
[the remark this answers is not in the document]
JamesGallicchio: `Subperm.cons_left` at the end of the file

JamesGallicchio (commented):
ok there's a bunch of things i need to do, but i'm happy with most of the file.

(sorry for dropping this for so long!)

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,908 @@
...
+protected theorem Nodup.subperm (d : Nodup l₁) (H : l₁ ⊆ l₂) : l₁ <+~ l₂ :=
+  by
+  induction d with
+  | nil => exact ⟨nil, Perm.nil, nil_sublist _⟩
+  | cons h d IH =>
+    have ⟨H₁, H₂⟩ := forall_mem_cons.1 H
+    have := fun contra => h _ contra rfl
+    exact cons_subperm_of_mem d this H₁ (IH H₂)
+
+theorem perm_ext {l₁ l₂ : List α} (d₁ : Nodup l₁) (d₂ : Nodup l₂) :
```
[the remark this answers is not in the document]
Vtec234: Afaik ext lemmas are usually supposed to be of the form (extensional version -> equivalence), so in this case `(∀ a, a ∈ l₁ ↔ a ∈ l₂) → l₁ ~ l₂`. Other than that though, `ext` seems to be convention for lemmas of this form.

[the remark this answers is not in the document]
JamesGallicchio: It seems like it's meant for simplifying perm in the case where `Nodup l1` and `Nodup l2`, rather than being used as an ext lemma.

I'd prefer a name like `Perm.subsets` or `Perm.subsets_of_nodup`. I'll look into where it's used in Mathlib.

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,908 @@
...
+      exact p
+    case nil.cons =>
+      exact p.trans perm_middle
+    case cons.nil =>
+      exact perm_middle.symm.trans p
+    case cons.cons =>
+      exact (IH _ _ _ _ rfl rfl).cons _
+  · rcases l₁ with (_ | ⟨y, _ | ⟨z, l₁⟩⟩) <;> rcases l₂ with (_ | ⟨u, _ | ⟨v, l₂⟩⟩) <;>
+          dsimp at e₁ e₂ <;> injections <;> subst_vars
+    · subst_vars
```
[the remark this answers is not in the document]
JamesGallicchio: i tried my best. brutal proof.

Std/Data/List/Perm.lean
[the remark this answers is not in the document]
JamesGallicchio: It ends up being used in the proof of `Subperm.cons_left` indirectly. We could probably make this cleaner, but I'm tired.

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,908 @@
...
+  induction l₂ with
+  | nil => exact ⟨a :: l₁, by simp [List.diff]⟩
+  | cons b l₂ ih =>
+    rw [diff_cons, diff_cons, ←diff_erase, ←diff_erase]
+    refine Subperm.trans ?_ (erase_cons_subperm_cons_erase _ _ _)
+    apply Subperm.erase
+    exact ih
+
+theorem subset_cons_diff {a : α} {l₁ l₂ : List α} : (a :: l₁).diff l₂ ⊆ a :: l₁.diff l₂ :=
+  subperm_cons_diff.subset
```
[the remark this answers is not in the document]
JamesGallicchio: annoying to prove earlier

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,908 @@
...
+    have ⟨H₁, H₂⟩ := forall_mem_cons.1 H
+    have := fun contra => h _ contra rfl
+    exact cons_subperm_of_mem d this H₁ (IH H₂)
+
+theorem perm_ext {l₁ l₂ : List α} (d₁ : Nodup l₁) (d₂ : Nodup l₂) :
+    l₁ ~ l₂ ↔ ∀ a, a ∈ l₁ ↔ a ∈ l₂ :=
+  ⟨fun p _ => p.mem_iff, fun H =>
+    (d₁.subperm fun a => (H a).1).antisymm <| d₂.subperm fun a => (H a).2⟩
+
+theorem Nodup.sublist_ext {l₁ l₂ l : List α} (d : Nodup l) (s₁ : l₁ <+ l) (s₂ : l₂ <+ l) :
```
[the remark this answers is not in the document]
JamesGallicchio: it is used once (Mathlib/Data/Multiset/Powerset.lean)

Std/Data/List/Perm.lean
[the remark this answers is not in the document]
JamesGallicchio: It's used 7 times in mathlib. 5/7 uses are to construct an iff, 2/7 are rewriting.

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,908 @@
...
+      specialize H b
+      rw [(perm_cons_erase this).count_eq] at H
+      by_cases h : b = a <;> simp [h] at H ⊢ <;> assumption⟩
+
+theorem Subperm.cons_right {α : Type _} {l l' : List α} (x : α) (h : l <+~ l') : l <+~ x :: l' :=
+  h.trans (sublist_cons x l').subperm
+
+/-- The list version of `add_tsub_cancel_of_le` for multisets. -/
+theorem subperm_append_diff_self_of_count_le {l₁ l₂ : List α}
+    (h : ∀ x ∈ l₁, count x l₁ ≤ count x l₂) : l₁ ++ l₂.diff l₁ ~ l₂ :=
```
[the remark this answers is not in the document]
JamesGallicchio: attempted to remove dependency; couldn't think of a better approach.

JamesGallicchio:
Diff with current mathlib: https://gist.github.com/JamesGallicchio/a3338ee7269a8c71927e6f75498e40df

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+protected theorem Perm.refl : ∀ l : List α, l ~ l
+  | [] => Perm.nil
+  | x :: xs => (Perm.refl xs).cons x
+
+protected theorem Perm.symm {l₁ l₂ : List α} (p : l₁ ~ l₂) : l₂ ~ l₁ :=
+  p.rec
+    .nil
+    (fun x _ _ _ r₁ => .cons x r₁)
+    (fun x y l => .swap y x l)
+    (fun _ _ r₁ r₂ => .trans r₂ r₁)
```
fgdorais, line 58:
```suggestion
protected theorem Perm.symm {l₁ l₂ : List α} (h : l₁ ~ l₂) : l₂ ~ l₁ := by
  induction h with
  | nil => exact nil
  | cons _ _ ih => exact cons _ ih
  | swap => exact swap ..
  | trans _ _ ih₁ ih₂ => exact trans ih₂ ih₁
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+
+instance isSetoid (α) : Setoid (List α) :=
+  Setoid.mk (@Perm α) (Perm.eqv α)
+
+theorem Perm.mem_iff {a : α} {l₁ l₂ : List α} (p : l₁ ~ l₂) : a ∈ l₁ ↔ a ∈ l₂ :=
+  p.rec
+    Iff.rfl
+    (by simp (config := {contextual := true}) only [mem_cons, implies_true])
+    (by intros; simp only [mem_cons, or_left_comm])
+    (by simp (config := {contextual := true}))
```
fgdorais, line 80:
```suggestion
theorem Perm.mem_iff {a : α} {l₁ l₂ : List α} (p : l₁ ~ l₂) : a ∈ l₁ ↔ a ∈ l₂ := by
  induction p with
  | nil => rfl
  | cons _ _ ih => simp only [mem_cons, ih]
  | swap => simp only [mem_cons, or_left_comm]
  | trans _ _ ih₁ ih₂ => simp only [ih₁, ih₂]
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+
+theorem Perm.subset {l₁ l₂ : List α} (p : l₁ ~ l₂) : l₁ ⊆ l₂ :=
+  fun _ => p.mem_iff.mp
+
+theorem Perm.append_right {l₁ l₂ : List α} (t₁ : List α) (p : l₁ ~ l₂) : l₁ ++ t₁ ~ l₂ ++ t₁ :=
+  p.rec
+    (Perm.refl ([] ++ t₁))
+    (fun x _ _ _ r₁ => r₁.cons x)
+    (fun x y _ => swap x y _)
+    (fun _ _ r₁ r₂ => r₁.trans r₂)
```
fgdorais, line 90:
```suggestion
theorem Perm.append_right {l₁ l₂ : List α} (t₁ : List α) (p : l₁ ~ l₂) : l₁ ++ t₁ ~ l₂ ++ t₁ := by
  induction p with
  | nil => exact Perm.refl _
  | cons _ _ ih => exact cons _ ih
  | swap => exact swap ..
  | trans _ _ ih₁ ih₂ => exact trans ih₁ ih₂
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+  | a :: t, l₂ => (perm_append_comm.cons _).trans perm_middle.symm
+
+theorem concat_perm (l : List α) (a : α) : concat l a ~ a :: l := by simp
+
+theorem Perm.length_eq {l₁ l₂ : List α} (p : l₁ ~ l₂) : length l₁ = length l₂ :=
+  p.rec
+    rfl
+    (fun _x l₁ l₂ _p r => by simp [r])
+    (fun _x _y l => by simp)
+    (fun _p₁ _p₂ r₁ r₂ => Eq.trans r₁ r₂)
```
fgdorais, line 123:
```suggestion
theorem Perm.length_eq {l₁ l₂ : List α} (p : l₁ ~ l₂) : length l₁ = length l₂ := by
  induction p with
  | nil => rfl
  | cons _ _ ih => simp only [length_cons, ih]
  | swap => rfl
  | trans _ _ ih₁ ih₂ => simp only [ih₁, ih₂]
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+    fun h => h ▸ Perm.refl _⟩
+
+@[simp]
+theorem replicate_perm {n : Nat} {a : α} {l : List α} :
+    replicate n a ~ l ↔ replicate n a = l :=
+  (perm_comm.trans perm_replicate).trans eq_comm
+
+@[simp]
+theorem perm_singleton {a : α} {l : List α} : l ~ [a] ↔ l = [a] :=
+  @perm_replicate α 1 a l
```
fgdorais, line 167:
```suggestion
  perm_replicate (n:=1)
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+
+@[simp]
+theorem singleton_perm {a : α} {l : List α} : [a] ~ l ↔ [a] = l :=
+  @replicate_perm α 1 a l
```
fgdorais, line 171:
```suggestion
  replicate_perm (n:=1)
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+  perm_singleton.1 p
+
+theorem Perm.singleton_eq {a : α} {l : List α} (p : [a] ~ l) : [a] = l :=
+  p.symm.eq_singleton.symm
+
+theorem singleton_perm_singleton {a b : α} : [a] ~ [b] ↔ a = b := by simp
+
+theorem perm_cons_erase [DecidableEq α] {a : α} {l : List α} (h : a ∈ l) : l ~ a :: l.erase a :=
+  let ⟨_l₁, _l₂, _, e₁, e₂⟩ := exists_erase_eq h
+  e₂.symm ▸ e₁.symm ▸ perm_middle
```
fgdorais, line 183:
```suggestion
  match exists_erase_eq h with | ⟨_, _, _, e₁, e₂⟩ => e₂ ▸ e₁ ▸ perm_middle  
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+    {P : (l₁ : List α) → (l₂ : List α) → l₁ ~ l₂ → Prop} {l₁ l₂ : List α} (p : l₁ ~ l₂)
+    (nil : P [] [] .nil)
+    (cons : ∀ x l₁ l₂, (h : l₁ ~ l₂) → P l₁ l₂ h → P (x :: l₁) (x :: l₂) (.cons x h))
+    (swap : ∀ x y l₁ l₂, (h : l₁ ~ l₂) → P l₁ l₂ h →
+      P (y :: x :: l₁) (x :: y :: l₂) (.trans (.swap x y _) (.cons _ (.cons _ h))))
+    (trans : ∀ l₁ l₂ l₃, (h₁ : l₁ ~ l₂) → (h₂ : l₂ ~ l₃) → P l₁ l₂ h₁ → P l₂ l₃ h₂ →
+      P l₁ l₃ (.trans h₁ h₂)) : P l₁ l₂ p :=
+  have P_refl l : P l l (.refl l) :=
+    List.recOn l nil fun x xs ih => cons x xs xs (Perm.refl xs) ih
+  Perm.recOn p nil cons (fun x y l => swap x y l l (Perm.refl l) (P_refl l)) @trans
```
fgdorais, line 196: Move up?
```suggestion
@[elab_as_elim]
def Perm.recOnAux
    {motive : (l₁ : List α) → (l₂ : List α) → l₁ ~ l₂ → Sort _} {l₁ l₂ : List α} (p : l₁ ~ l₂)
    (nil : motive [] [] .nil)
    (cons : ∀ x l₁ l₂, (h : l₁ ~ l₂) → motive l₁ l₂ h → motive (x :: l₁) (x :: l₂) (.cons x h))
    (swap : ∀ x y l₁ l₂, (h : l₁ ~ l₂) → motive l₁ l₂ h →
      motive (y :: x :: l₁) (x :: y :: l₂) (.trans (.swap x y _) (.cons _ (.cons _ h))))
    (trans : ∀ l₁ l₂ l₃, (h₁ : l₁ ~ l₂) → (h₂ : l₂ ~ l₃) → motive l₁ l₂ h₁ → motive l₂ l₃ h₂ →
      motive l₁ l₃ (.trans h₁ h₂)) : motive l₁ l₂ p :=
  have motive_refl l : motive l l (.refl l) :=
    List.recOn l nil fun x xs ih => cons x xs xs (Perm.refl xs) ih
  Perm.recOn p nil cons (fun x y l => swap x y l l (Perm.refl l) (motive_refl l)) @trans
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+    · exact ⟨x :: y :: _, Perm.swap _ _ _, (s.cons₂ _).cons₂ _⟩
+  | trans _ _ IH₁ IH₂ =>
+    exact
+      let ⟨m₁, pm, sm⟩ := IH₁ s
+      let ⟨r₁, pr, sr⟩ := IH₂ sm
+      ⟨r₁, pr.trans pm, sr⟩
+
+theorem Perm.sizeOf_eq_sizeOf [SizeOf α] {l₁ l₂ : List α} (h : l₁ ~ l₂) :
+    sizeOf l₁ = sizeOf l₂ := by
+  induction h with -- hd l₁ l₂ h₁₂ h_sz₁₂ a b l l₁ l₂ l₃ h₁₂ h₂₃ h_sz₁₂ h_sz₂₃
```
fgdorais, line 267:
```suggestion
  induction h with
```

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+theorem Perm.nodup_iff {l₁ l₂ : List α} : l₁ ~ l₂ → (Nodup l₁ ↔ Nodup l₂) :=
+  Perm.pairwise_iff <| @Ne.symm α
+
+theorem Perm.join {l₁ l₂ : List (List α)} (h : l₁ ~ l₂) : l₁.join ~ l₂.join :=
+  Perm.recOn h (Perm.refl _) (fun x xs₁ xs₂ _ ih => ih.append_left x)
+    (fun x₁ x₂ xs => by
+      simp [join]
+      rw [←append_assoc, ←append_assoc]
+      refine perm_append_comm.append_right ?_)
+    @fun xs₁ xs₂ xs₃ _ _ => Perm.trans
```
fgdorais, line 827:
```suggestion
theorem Perm.join {l₁ l₂ : List (List α)} (h : l₁ ~ l₂) : l₁.join ~ l₂.join :=
  Perm.recOn h (Perm.refl _) (fun x xs₁ xs₂ _ ih => ih.append_left x)
    (fun x₁ x₂ xs => by
      simp [join]
      rw [←append_assoc, ←append_assoc]
      refine perm_append_comm.append_right ?_)
    @fun xs₁ xs₂ xs₃ _ _ => Perm.trans
```

fgdorais (changes requested):

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,852 @@
...
+-/
+
+
+open Nat
+
+universe uu vv
+
+namespace List
+
+variable {α : Type uu} {β : Type vv} {l₁ l₂ : List α}
```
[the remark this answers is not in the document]
JamesGallicchio: Removed this one. I left the `DecidableEq` variable since it does genuinely seem to be used for every definition in its section.

Std/Data/List/Perm.lean
JamesGallicchio: I think GH messed up this suggested change?

JamesGallicchio: (or at least the suggestion seems identical to the current diff?)

Std/Data/List/Perm.lean
JamesGallicchio: I think lets should be stylistically preferred to one-branch matches, no? At least when it's possible to write it as a let, which I think is whenever you're not doing dependent elimination or something.

(The symms are still unnecessary, though -- I'll change that in the current proof.)

Std/Data/List/Perm.lean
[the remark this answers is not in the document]
kmill: I think the usual name would be `ext_iff` rather than `ext` here. Maybe `perm_ext_iff_of_nodup`?

Std/Data/List/Perm.lean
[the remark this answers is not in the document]
kmill: `Nodup.perm_iff_eq_of_sublist`?

JamesGallicchio:
Based on the Mathlib patch ([here](https://github.com/leanprover-community/mathlib4/pull/8566)), ~all the breakage is from
- making `~` et al notations scoped to the `List` namespace
- lemmas using `\forall {a b}, R a b -> R b a` instead of `Symmetric R` (which is not in `Std`)

Std/Data/List/Perm.lean
fgdorais: Looks like copypasta. It probably was just another "use induction instead of rec" but it's lost now...

Std/Data/List/Perm.lean
fgdorais: Maybe keep both and define `perm_induction_on` in terms of `Perm.recOnAux`? I don't know how often `perm_induction_on` was used.

Std/Data/List/Perm.lean
fgdorais: I don't think it makes much of a difference for proofs, I just like one liners. But you're right, the main point wasn't the match but the symms.

Std/Data/List/Perm.lean
fgdorais: Found it! Fixed now.

Std/Data/List/Perm.lean
fgdorais: Because `f (by _) (by _)` is poor style when `by ..` works just as well.

Std/Data/List/Perm.lean
fgdorais: This use of `rec` is fine. Disregard if desired.

Std/Data/List/Perm.lean
fgdorais: Because `f (by _) (by _)` is poor style when `by ..` works just as well.

Std/Data/List/Perm.lean
fgdorais: This use of rec is fine. Disregard if desired.

fgdorais (changes requested):
Clarified style changes.

PS: I didn't expect gh to format these comments this way... Confusing!

Std/Data/List/Perm.lean
JamesGallicchio: It appears to not be used in mathlib. I'm thinking maybe `Perm.swap'_induction_on` to reflect how it is different from recOn.

(btw, your `Perm.recOnAux` doesn't seem to be a generalization, since `motive` has to be in `Prop` -- is this intentional?)

fgdorais: Using `Sort _` was a reflex, `Prop` is more accurate. The idea is that custom recursors preferably have `.rec*` names to make them identifiable and indicate that they should be used with `induction _ using ..` instead of `apply` and such. The original disguises itself as a normal theorem.

How about this version? It more closely matches `Perm.recOn` and uses `swap'`.
```lean
@[elab_as_elim]
def Perm.recOn'
    {motive : (l₁ : List α) → (l₂ : List α) → l₁ ~ l₂ → Prop} {l₁ l₂ : List α} (p : l₁ ~ l₂)
    (nil : motive [] [] .nil)
    (cons : ∀ (x) {l₁ l₂}, (h : l₁ ~ l₂) → motive l₁ l₂ h → motive (x :: l₁) (x :: l₂) (.cons x h))
    (swap' : ∀ x y {l₁ l₂}, (h : l₁ ~ l₂) → motive l₁ l₂ h →
      motive (y :: x :: l₁) (x :: y :: l₂) (.swap' _ _ h))
    (trans : ∀ {l₁ l₂ l₃}, (h₁ : l₁ ~ l₂) → (h₂ : l₂ ~ l₃) → motive l₁ l₂ h₁ → motive l₂ l₃ h₂ →
      motive l₁ l₃ (.trans h₁ h₂)) : motive l₁ l₂ p :=
  have motive_refl l : motive l l (.refl l) :=
    List.recOn l nil fun x xs ih => cons x (Perm.refl xs) ih
  Perm.recOn p nil cons (fun x y l => swap' x y (Perm.refl l) (motive_refl l)) @trans
```

JamesGallicchio: I like `Perm.recOnSwap'` if that sounds good to everyone.

Note that there isn't really a pattern established yet in Std. The other `elab_as_elim` recursors in Std:
```
Fin.succRec
Fin.succRecOn
Fin.induction
Fin.inductionOn
Fin.cases
Fin.reverseInduction
Fin.lastCases 
Fin.addCases
Nat.gcd.induction
Nat.recAux
Nat.recAuxOn
Nat.casesAuxOn
Nat.strongRec
Nat.strongRecOn
Nat.recDiagAux
Nat.recDiag
Nat.recDiagOn
Nat.casesDiagOn
WF.recC (private)
```
Presumably the ones with `induction` in the name were inherited from Mathlib? But I'm not convinced that `rec` is the most discoverable name for lemmas intended to be used in `induction using ...`

fgdorais: The `Nat.gcd.induction` is the only slightly strange one I see. They all have rec, cases, ind, or induction in there, same as in core.

fgdorais: They should also all be defs and not theorems, except for ind/induction ones which can go one way or the other. But do include a docstring even if it's a theorem.

JamesGallicchio: Is the difference between `recOn` and `inductionOn` just `Prop` vs `Sort u` as the motive?

fgdorais: Yes, but it's mostly used when the `Prop` version is significantly different (e.g. quotient types where propositional irrelevance simplifies things quite a bit).

fgdorais: PS: `Perm.recOnSwap'` sounds good to me!

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,863 @@
...
+  ⟨Perm.symm, Perm.symm⟩
+
+theorem Perm.swap' (x y : α) {l₁ l₂ : List α} (p : l₁ ~ l₂) : y :: x :: l₁ ~ x :: y :: l₂ :=
+  (swap _ _ _).trans ((p.cons _).cons _)
+
+theorem Perm.eqv (α) : Equivalence (@Perm α) :=
+  ⟨Perm.refl, Perm.symm, Perm.trans⟩
+
+theorem Perm.of_eq (h : l₁ = l₂) : l₁ ~ l₂ :=
+  h ▸ Perm.refl l₁
```
fgdorais, line 66: Move up after `.refl`.

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,863 @@
...
+
+@[simp]
+theorem singleton_perm {a : α} {l : List α} : [a] ~ l ↔ [a] = l :=
+  replicate_perm (n := 1)
+
+theorem Perm.eq_singleton {a : α} {l : List α} (p : l ~ [a]) : l = [a] :=
+  perm_singleton.1 p
+
+theorem Perm.singleton_eq {a : α} {l : List α} (p : [a] ~ l) : [a] = l :=
+  p.symm.eq_singleton.symm
```
fgdorais, line 173: Maybe use `alias`?

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,863 @@
...
+
+theorem singleton_perm_singleton {a b : α} : [a] ~ [b] ↔ a = b := by simp
+
+theorem perm_cons_erase [DecidableEq α] {a : α} {l : List α} (h : a ∈ l) : l ~ a :: l.erase a :=
+  let ⟨_l₁, _l₂, _, e₁, e₂⟩ := exists_erase_eq h
+  e₂ ▸ e₁ ▸ perm_middle
+
+/-- Similar to `Perm.recOn`, but the `swap` case is generalized to `Perm.swap'`,
+where the tail of the lists are not necessarily the same. -/
+@[elab_as_elim]
```
fgdorais, line 183: Move up?

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,863 @@
...
+theorem Perm.recOnSwap'
+    {motive : (l₁ : List α) → (l₂ : List α) → l₁ ~ l₂ → Prop} {l₁ l₂ : List α} (p : l₁ ~ l₂)
+    (nil : motive [] [] .nil)
+    (cons : ∀ x l₁ l₂, (h : l₁ ~ l₂) → motive l₁ l₂ h → motive (x :: l₁) (x :: l₂) (.cons x h))
+    (swap' : ∀ x y l₁ l₂, (h : l₁ ~ l₂) → motive l₁ l₂ h →
+      motive (y :: x :: l₁) (x :: y :: l₂) (.swap' _ _ h))
+    (trans : ∀ l₁ l₂ l₃, (h₁ : l₁ ~ l₂) → (h₂ : l₂ ~ l₃) → motive l₁ l₂ h₁ → motive l₂ l₃ h₂ →
+      motive l₁ l₃ (.trans h₁ h₂)) : motive l₁ l₂ p :=
```
fgdorais, line 191: Match `Perm.recOn` in terms of implicit parameters for each case.

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,863 @@
...
+
+theorem Subperm.filter (p : α → Bool) ⦃l l' : List α⦄ (h : l <+~ l') :
+    filter p l <+~ filter p l' := by
+  obtain ⟨xs, hp, h⟩ := h
+  exact ⟨_, hp.filter p, h.filter p⟩
+
+@[simp]
+theorem subperm_singleton_iff {α} {l : List α} {a : α} : [a] <+~ l ↔ a ∈ l :=
+  ⟨fun ⟨s, hla, h⟩ => by rwa [perm_singleton.mp hla, singleton_sublist] at h, fun h =>
+    ⟨[a], Perm.refl _, singleton_sublist.mpr h⟩⟩
```
fgdorais, line 334: Too much `fun`? :)

Std/Data/List/Perm.lean
```diff
@@ -0,0 +1,863 @@
...
+  p.countP_eq _
+
+theorem Subperm.count_le [DecidableEq α] {l₁ l₂ : List α} (s : l₁ <+~ l₂) (a) :
+    count a l₁ ≤ count a l₂ :=
+  s.countP_le _
+
+theorem Perm.foldl_eq' {f : β → α → β} {l₁ l₂ : List α} (p : l₁ ~ l₂)
+    (comm : ∀ x ∈ l₁, ∀ y ∈ l₁, ∀ (z), f (f z x) y = f (f z y) x)
+  : ∀ b, foldl f b l₁ = foldl f b l₂ := by
+  intro b
```
fgdorais, line 382: Better than `b`?
```suggestion
theorem Perm.foldl_eq' {f : β → α → β} {l₁ l₂ : List α} (p : l₁ ~ l₂)
    (comm : ∀ x ∈ l₁, ∀ y ∈ l₁, ∀ (z), f (f z x) y = f (f z y) x) (init) :
      foldl f init l₁ = foldl f init l₂ := by
```

fgdorais (approved):
Looks good to me!  Since I don't use `Perm`  much, a second set of eyes is recommended.

Just a few small style suggestions. None are required, I approve anyway.

digama0 (approved):

joehendrix (approved):

PR #91, opened 2023-01-22
feat: add `#compile inductive`

cppio:
Add a `#compile inductive` command to compile the recursors of an inductive type. Currently does not support nested/mututal inductive types. There are a few issues with some specific inductive types (most of the errors come from inductive propositions which feature large elimination), but 1151 out of 1175 inductive types defined in `Lean` and `Std` currently work.

### DIFF
--- a/Std.lean
+++ b/Std.lean
@@ -97,6 +97,7 @@ import Std.Tactic.Simpa
 import Std.Tactic.SqueezeScope
 import Std.Tactic.TryThis
 import Std.Tactic.Unreachable
+import Std.Util.CompileInductive
 import Std.Util.ExtendedBinder
 import Std.Util.LibraryNote
 import Std.Util.TermUnsafe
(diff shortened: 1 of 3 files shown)


digama0:
We discussed this PR in the lean dev meeting. This solution is a workaround for a missing feature in lean, and it is not what we would like to have long term. As such, I think it is better suited either as a mathlib4 addition if the purpose is to ease porting, or a lean core PR if the intention is to make lean do this work out of the box.

Implementation-wise, there is a fundamental problem with simply implementing recursors as functions which is that they have the wrong evaluation semantics: they evaluate from the bottom up, leading to a lot of performance pitfalls. So I don't think this approach will work as a proper solution even if it is done automatically as part of the `inductive` command. The better approach is to have the compiler recognize applications of these functions and implement them natively similar to other functions with nonstandard evaluation semantics like `if`.

cppio:
Understood, I'll move this PR to mathlib4.

PR #93, opened 2023-01-30
doc: adjust library_note example

rwbarton:
This is the format used throughout mathlib; I don't know if there are any other library_note users.

### DIFF
--- a/Std/Util/LibraryNote.lean
+++ b/Std/Util/LibraryNote.lean
@@ -26,8 +26,9 @@ initialize libraryNoteExt : SimplePersistentEnvExtension LibraryNoteEntry (Array
 open Lean Parser Command in
 /--
 ```
-/-- ... some explanation ... -/
-library_note "some tag"
+library_note "some tag"/--
+... some explanation ...
+-/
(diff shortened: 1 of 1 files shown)


digama0:
That's the format that mathport uses, but I'm not sure we want to make that the recommended style. In particular I think we should have a space before `/--`.

Std/Util/LibraryNote.lean
```diff
@@ -26,8 +26,9 @@ initialize libraryNoteExt : SimplePersistentEnvExtension LibraryNoteEntry (Array
 open Lean Parser Command in
 /--
 ```
-/-- ... some explanation ... -/
-library_note "some tag"
+library_note "some tag"/--
```
gebner, line 29:
```suggestion
library_note "some tag" /--
```
I think the documentation is my fault.. I just copied it from mathlib3 when I ported `library_note`.

rwbarton:
I only noticed this because emacs shows just the first line of the docstring in the minibuffer, and it was literally `/-- ... some explanation ... -/`.

PR #96, opened 2023-01-31
feat: more lemmas for proving pos

JamesGallicchio:
Adds some lemmas from mathlib4 for proving positivity.


fgdorais (approved):
The theorem names appear to agree with mathlib nomenclature. They are protected so there shouldn't be any downstream issues.

Std/Data/Nat/Lemmas.lean
```diff
@@ -80,6 +80,20 @@ protected theorem lt_add_of_pos_right {n k : Nat} (h : 0 < k) : n < n + k :=
...
+protected theorem pos_of_lt_add_left {n k : Nat} (h : n < k + n) : 0 < k :=
+  Nat.lt_of_add_lt_add_right (by rw [Nat.zero_add]; exact h)
+
+protected theorem lt_add_right_iff_pos {n k : Nat} : n < n + k ↔ 0 < k :=
+  ⟨Nat.pos_of_lt_add_right, Nat.lt_add_of_pos_right⟩
+
+protected theorem lt_add_left_iff_pos {n k : Nat} : n < k + n ↔ 0 < k :=
+  ⟨Nat.pos_of_lt_add_left, Nat.lt_add_of_pos_left⟩
+
+
```
digama0, line 95:
```suggestion
```

PR #98, opened 2023-01-31
chore: de-mathlib lots of Data/List/Lemmas.lean

JamesGallicchio:
I believe all the statements are pulled from mathlib4 identically.

Main weirdness is Std had a `mem_map` lemma with a conclusion that is in some sense backwards from what would generally be preferred. Mathlib has a `mem_map'` for porting, so I moved the `mem_map'` lemma here. We might want to deprecate `mem_map`.


digama0:
okay, I made mostly style and golfing changes, but also some renames. For `mem_map` I just replaced the original one as planned (plus one other related theorem). Some duplicate theorems were removed.

digama0 (approved):

digama0:
if the attributes can be added after the fact in mathlib (which I think is the case), then it's fine to remove the comments.

PR #99, opened 2023-01-31
chore: de-mathlib Data/List/Count.lean

JamesGallicchio:
Moves `Mathlib/Data/List/Count.lean` to Std. I believe there are no notable changes from mathlib, aside from dropping various lemmas about prod.

Depends on ~~#98 #96 #94 #97~~. EDIT: all merged :)

### DIFF
--- a/Std/Data/List/Count.lean
+++ b/Std/Data/List/Count.lean
@@ -31,15 +31,10 @@ protected theorem countP_go_eq_add (l) : countP.go p l n = n + countP.go p l 0 :
   | cons head tail ih =>
     unfold countP.go
     rw [ih (n := n + 1), ih (n := n), ih (n := 1)]
-    by_cases h : p head
-    · simp [h, Nat.add_assoc]
-    · simp [h]
+    if h : p head then simp [h, Nat.add_assoc] else simp [h]
 
 @[simp] theorem countP_cons_of_pos (l) (pa : p a) : countP p (a :: l) = countP p l + 1 := by
-  have : countP.go p (a :: l) 0 = countP.go p l 1 := by
-    suffices (bif _ then _ else _) = countP.go _ _ _ from this
-    rw [pa]
-    rfl
+  have : countP.go p (a :: l) 0 = countP.go p l 1 := show cond .. = _ by rw [pa]; rfl
   unfold countP
   rw [this, Nat.add_comm, List.countP_go_eq_add]
 
@@ -53,11 +48,12 @@ theorem length_eq_countP_add_countP (l) : length l = countP p l + countP (fun a 
   induction l with
   | nil => rfl
   | cons x h ih =>
-    by_cases h : p x
-    · rw [countP_cons_of_pos _ _ h, countP_cons_of_neg _ _ _, length, ih]
+    if h : p x then
+      rw [countP_cons_of_pos _ _ h, countP_cons_of_neg _ _ _, length, ih]
       · rw [Nat.add_assoc, Nat.add_comm _ 1, Nat.add_assoc]
       · simp only [h]
-    · rw [countP_cons_of_pos (fun a => ¬p a) _ _, countP_cons_of_neg _ _ h, length, ih]
+    else
+      rw [countP_cons_of_pos (fun a => ¬p a) _ _, countP_cons_of_neg _ _ h, length, ih]
       · rfl
       · simp only [h]
 
@@ -65,9 +61,9 @@ theorem countP_eq_length_filter (l) : countP p l = length (filter p l) := by
   induction l with
   | nil => rfl
   | cons x l ih =>
-    by_cases h : p x
-    · rw [countP_cons_of_pos p l h, ih, filter_cons_of_pos l h, length]
-    · rw [countP_cons_of_neg p l h, ih, filter_cons_of_neg l h]
+    if h : p x
+    then rw [countP_cons_of_pos p l h, ih, filter_cons_of_pos l h, length]
+    else rw [countP_cons_of_neg p l h, ih, filter_cons_of_neg l h]
 
 theorem countP_le_length : countP p l ≤ l.length := by
   simp only [countP_eq_length_filter]
@@ -139,10 +135,8 @@ variable [DecidableEq α]
 @[simp] theorem count_nil (a : α) : count a [] = 0 := rfl
 
 theorem count_cons (a b : α) (l : List α) :
-    count a (b :: l) = count a l + if a = b then 1 else 0 := by conv =>
-  simp [count, countP_cons]
-  lhs
-  simp only [eq_comm]
+    count a (b :: l) = count a l + if a = b then 1 else 0 := by
+  simp [count, countP_cons, eq_comm (a := a)]
 
 @[simp] theorem count_cons_self (a : α) (l : List α) : count a (a :: l) = count a l + 1 := by
   simp [count_cons]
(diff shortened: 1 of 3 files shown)


Std/Data/List/Count.lean
```diff
@@ -0,0 +1,261 @@
...
+theorem count_cons (a b : α) (l : List α) :
+    count a (b :: l) = count a l + if a = b then 1 else 0 := by conv =>
+  simp [count, countp_cons]
+  lhs
+  simp only [eq_comm]
+
+@[simp] theorem count_cons_self (a : α) (l : List α) : count a (a :: l) = count a l + 1 := by
+  simp [count_cons]
+
+@[simp] theorem count_cons_of_ne (h : a ≠ b) (l : List α) : count a (b :: l) = count a l := by
```
JamesGallicchio, line 146: I think (?) we could just mark `count_cons` simp, right?

Std/Data/List/Count.lean
```diff
@@ -0,0 +1,261 @@
...
+theorem Sublist.count_le (h : l₁ <+ l₂) (a : α) : count a l₁ ≤ count a l₂ :=
+  h.countp_le _
+
+theorem count_le_count_cons (a b : α) (l : List α) : count a l ≤ count a (b :: l) :=
+  (sublist_cons _ _).count_le _
+
+theorem count_singleton (a : α) : count a [a] = 1 := by
+  simp [count_cons]
+
+theorem count_singleton' (a b : α) : count a [b] = if a = b then 1 else 0 := by
```
JamesGallicchio, line 164: Not sure these need to be in Std, especially if `count_cons` is marked simp.

Std/Data/List/Count.lean
```diff
@@ -0,0 +1,261 @@
...
+      by_cases ha : a = b
+      · rw [← ha, eq_comm] at hc
+        rw [if_pos ha, if_neg hc, Nat.add_zero, Nat.add_zero]
+      · rw [if_neg ha, Nat.sub_zero, Nat.sub_zero]
+
+@[simp] theorem count_erase_self (a : α) (l : List α) : count a (List.erase l a) = count a l - 1 := by
+  rw [count_erase, if_pos rfl]
+
+@[simp] theorem count_erase_of_ne (ab : a ≠ b) (l : List α) : count a (l.erase b) = count a l :=
+  by rw [count_erase, if_neg ab, Nat.sub_zero]
```
JamesGallicchio, line 260: Same as `count_cons`, I feel like we could mark `count_erase` simp and delete these.

Std/Data/List/Count.lean
digama0: I don't think we should be putting an expression like that in the goal if people don't specifically ask for it.

Std/Data/List/Count.lean
digama0: These aren't simp lemmas in the first place. You are allowed to refer to these theorems directly and in that case it doesn't matter whether it is simp-provable.

JamesGallicchio:
Cleaning this up again; the list lemma files have quite a few conflicts because many of the list lemmas were added by someone at some point.

fgdorais (commented):
I noted a few questionable simp lemmas. The common thread is that they change apples to oranges, they preserve the basic nutritional value but not every apple recipe has an equivalent orange recipe. These can still be good simp lemmas in some contexts. For example, if everyone prefers oranges over apples.

JamesGallicchio:
I've made https://github.com/leanprover-community/mathlib4/pull/6575 so we can check where these suspicious simp lemmas are used.

Std/Data/List/Count.lean
```diff
@@ -0,0 +1,260 @@
...
+  countp_append _
+
+theorem count_concat (a : α) (l : List α) : count a (concat l a) = succ (count a l) := by simp
+
+@[simp] theorem count_pos {a : α} {l : List α} : 0 < count a l ↔ a ∈ l := by
+  simp only [count, countp_pos, beq_iff_eq, exists_eq_right]
+
+@[simp] theorem one_le_count_iff_mem {a : α} {l : List α} : 1 ≤ count a l ↔ a ∈ l := count_pos
+
+theorem count_eq_zero_of_not_mem {a : α} {l : List α} (h : a ∉ l) : count a l = 0 :=
```
JamesGallicchio, line 177: This was marked `simp` in mathlib, but in Std causes the `simpNF` linter to complain because the LHS of `count_nil` is no longer simp-normal. Thoughts on the right way to proceed here?

JamesGallicchio: Resolved by setting `count_nil` priority higher.

JamesGallicchio:
RE: questionable simp lemmas -- each lemma and where it is used in mathlib (for my own reference)
```
List.count_eq_length
 - List.count_replicate_self
List.count_eq_length
 - Std.Data.List.Count._auxLemma.15
List.count_eq_zero
 - List.count_bagInter
List.count_eq_zero
 - List.sum_map_count_dedup_filter_eq_countp
List.count_eq_zero
 - List.count_replicate
 - Std.Data.List.Count._auxLemma.14
List.count_pos
 - List.count_bagInter
List.count_pos
 - Nat.Prime.factorization_pos_of_dvd
List.count_pos
 - Std.Data.List.Count._auxLemma.12
List.count_pos
 - List.dedup_eq_cons
List.count_pos
 - List.count_eq_one_of_mem
List.count_pos
 - List.subperm_append_diff_self_of_count_le
 - List.one_le_count_iff_mem
List.count_pos
 - List.count_eq_zero_of_not_mem
List.count_pos
 - Mathlib.Data.List.Pairwise._auxLemma.38
 - List.perm_iff_count
List.count_pos
 - List.not_mem_of_count_eq_zero
List.countp_eq_length
 - List.count_eq_length
List.countp_eq_length
 - Std.Data.List.Count._auxLemma.7
List.countp_eq_zero
 - Std.Data.List.Count._auxLemma.6
List.countp_filter
 - Multiset.countp_eq_countp_filter_add
 - List.count_filter
List.one_le_count_iff_mem
 - Std.Data.List.Count._auxLemma.13
List.one_le_count_iff_mem
 - List.pairwise_of_forall_mem_list
```

fgdorais:
Outstanding research work! Sadly, I don't have time to look at these carefully right now but this is absolutely great!

Std/Data/List/Count.lean
```diff
@@ -0,0 +1,260 @@
...
+  simp only [countp_eq_length_filter, length_eq_zero, filter_eq_nil]
+
+@[simp] theorem countp_eq_length : countp p l = l.length ↔ ∀ a ∈ l, p a := by
+  rw [countp_eq_length_filter, filter_length_eq_length]
+
+theorem Sublist.countp_le (s : l₁ <+ l₂) : countp p l₁ ≤ countp p l₂ := by
+  simp only [countp_eq_length_filter]
+  apply s.filter _ |>.length_le
+
+@[simp] theorem countp_filter (l : List α) :
```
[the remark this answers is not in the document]
JamesGallicchio: It eliminates a nested function, and I think is a useful lemma, but I agree that it probably does not need to be a simp lemma. Mathlib seems to only use it as a simp lemma once.

Std/Data/List/Count.lean
```diff
@@ -0,0 +1,260 @@
...
+
+@[simp] theorem count_append (a : α) : ∀ l₁ l₂, count a (l₁ ++ l₂) = count a l₁ + count a l₂ :=
+  countp_append _
+
+theorem count_concat (a : α) (l : List α) : count a (concat l a) = succ (count a l) := by simp
+
+@[simp] theorem count_pos {a : α} {l : List α} : 0 < count a l ↔ a ∈ l := by
+  simp only [count, countp_pos, beq_iff_eq, exists_eq_right]
+
+@[simp] theorem one_le_count_iff_mem {a : α} {l : List α} : 1 ≤ count a l ↔ a ∈ l := count_pos
```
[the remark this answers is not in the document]
JamesGallicchio: `count_pos` and `one_le_count_iff_mem` are used widely in Mathlib, but only once or twice by simp. I sort of feel like we should remove `one_le_count_iff_mem`, rename `count_pos` to `count_pos_iff_mem`, and make it not simp.

JamesGallicchio:
I think this is good for review. The mathlib PR is helpful to see the changes from the mathlib version.

semorrison:
By the way, I have a tool that lists all lemmas used by other declarations, and can flag lemmas used by `simp`. If you need to do this research again and I haven't published that tool, please ping me. :-)

semorrison (approved):
LGTM

PR #95, opened 2023-01-31
feat: forall_exists_index

JamesGallicchio:
Adds `forall_exists_index`. Statement pulled from mathlib4.

PR #100, opened 2023-02-01
chore: de-mathlib Data/List/Pairwise.lean

JamesGallicchio:
All included statements are the same as mathlib4, but a good number of lemmas are dropped. See [diff](https://github.com/JamesGallicchio/std4/compare/TEMP_compare_pairwise..chore/move-Data/List/Pairwise.lean).

~~Depends on #99 #95.~~

### DIFF
--- a/Std/Data/List/Pairwise.lean
+++ b/Std/Data/List/Pairwise.lean
@@ -193,7 +193,7 @@ theorem pairwise_bind {R : β → β → Prop} {l : List α} {f : α → List β
   rw [←h]
   exact d _ g
 
-theorem pairwise_iff_forall_sublist [DecidableEq α] : l.Pairwise R ↔ (∀ {a b}, [a,b] <+ l → R a b) := by
+theorem pairwise_iff_forall_sublist : l.Pairwise R ↔ (∀ {a b}, [a,b] <+ l → R a b) := by
   induction l with
   | nil => simp
   | cons hd tl IH =>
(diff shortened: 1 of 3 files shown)


Std/Data/List/Pairwise.lean
```diff
@@ -0,0 +1,351 @@
...
+theorem pairwise_bind {R : β → β → Prop} {l : List α} {f : α → List β} :
+    List.Pairwise R (l.bind f) ↔
+      (∀ a ∈ l, Pairwise R (f a)) ∧ Pairwise (fun a₁ a₂ => ∀ x ∈ f a₁, ∀ y ∈ f a₂, R x y) l := by
+  simp [List.bind, pairwise_join, pairwise_map]
+  intro
+  refine ⟨fun h _ hx => h _ _ hx rfl, fun d e f g h => ?_⟩
+  rw [←h]
+  exact d _ g
+
+theorem pairwise_of_reflexive_on_dupl_of_forall_ne [DecidableEq α] {l : List α} {r : α → α → Prop}
```
JamesGallicchio, line 196: I don't love this name and I don't love this API. It is better encapsulated by `<+`, i.e. having a hypothesis `h : \forall a b, [a,b] <+ -> r a b`

The `<+` proof is quite clean, and this theorem can be derived straightforwardly. Seems like a worthwhile refactor.

I'll also deprecate this symbol, since the downstream usage (there is only one, in `Nodup.pairwise_of_forall_ne`) can go through the `<+` lemma.

JamesGallicchio: Oh, the `pairwise_iff_get` proof is also a specialization of the subseq theorem. I guess it also deserves the deprecation treatment.

JamesGallicchio:
Looks good to me. See https://github.com/leanprover-community/mathlib4/pull/6743 for the mathlib changes.

semorrison:
I think based on the Mathlib PR being painless, this should be safe to merge.

semorrison (approved):

Vtec234 (approved):

PR #101, opened 2023-02-02
feat: move `RatCast` into std4

thorimur:
This is used for leanprover-community/mathlib4#1707.

This also makes `Nat.cast`, `Int.cast`, and `Rat.cast` reducible.

We:
* move `RatCast` into std4
* define `Rat.cast` and give it coercion instances, in parallel to `Int.cast` and `Nat.cast`
* add reflexive instances for `NatCast`, `IntCast`, and `RatCast` (e.g. `NatCast Nat where natCast x := x`)
* add `@[reducible]` to (`Nat`/`Int`/`Rat`)`.cast`

### DIFF
--- a/Std/Classes/Cast.lean
+++ b/Std/Classes/Cast.lean
@@ -11,6 +11,8 @@ class NatCast (R : Type u) where
   /-- The canonical map `Nat → R`. -/
   protected natCast : Nat → R
 
+instance : NatCast Nat where natCast := id
+
 /-- Canonical homomorphism from `Nat` to a additive monoid `R` with a `1`.
 This is just the bare function in order to aid in creating instances of `AddMonoidWithOne`. -/
 @[coe, match_pattern] protected def Nat.cast {R : Type u} [NatCast R] : Nat → R := NatCast.natCast
@@ -25,6 +27,8 @@ instance [NatCast R] : CoeHTCT Nat R where coe := Nat.cast
 class IntCast (R : Type u) where
   /-- The canonical map `Int → R`. -/
   protected intCast : Int → R
+
+instance : IntCast Int where intCast := id
 
 /-- Canonical homomorphism from `Int` to a additive group `R` with a `1`.
 This is just the bare function in order to aid in creating instances of `AddGroupWithOne`. -/
--- a/Std/Classes/Cast.lean
+++ b/Std/Classes/Cast.lean
@@ -11,7 +11,7 @@ class NatCast (R : Type u) where
   /-- The canonical map `Nat → R`. -/
   protected natCast : Nat → R
 
-instance : NatCast Nat where natCast := id
+instance : NatCast Nat where natCast n := n
 
 /-- Canonical homomorphism from `Nat` to a additive monoid `R` with a `1`.
 This is just the bare function in order to aid in creating instances of `AddMonoidWithOne`. -/
--- a/Std/Classes/Cast.lean
+++ b/Std/Classes/Cast.lean
@@ -29,7 +29,7 @@ class IntCast (R : Type u) where
   /-- The canonical map `Int → R`. -/
   protected intCast : Int → R
 
-instance : IntCast Int where intCast := id
+instance : IntCast Int where intCast n := n
 
 /-- Canonical homomorphism from `Int` to a additive group `R` with a `1`.
 This is just the bare function in order to aid in creating instances of `AddGroupWithOne`. -/
--- a/Std/Classes/Cast.lean
+++ b/Std/Classes/Cast.lean
@@ -15,7 +15,8 @@ instance : NatCast Nat where natCast := id
 
 /-- Canonical homomorphism from `Nat` to a additive monoid `R` with a `1`.
 This is just the bare function in order to aid in creating instances of `AddMonoidWithOne`. -/
-@[coe, match_pattern] protected def Nat.cast {R : Type u} [NatCast R] : Nat → R := NatCast.natCast
+@[coe, reducible, match_pattern] protected def Nat.cast {R : Type u} [NatCast R] : Nat → R :=
+  NatCast.natCast
 
 -- see note [coercion into rings]
 instance [NatCast R] : CoeTail Nat R where coe := Nat.cast
@@ -32,7 +33,8 @@ instance : IntCast Int where intCast := id
 
 /-- Canonical homomorphism from `Int` to a additive group `R` with a `1`.
 This is just the bare function in order to aid in creating instances of `AddGroupWithOne`. -/
-@[coe, match_pattern] protected def Int.cast {R : Type u} [IntCast R] : Int → R := IntCast.intCast
+@[coe, reducible, match_pattern] protected def Int.cast {R : Type u} [IntCast R] : Int → R :=
+  IntCast.intCast
 
 -- see note [coercion into rings]
 instance [IntCast R] : CoeTail Int R where coe := Int.cast
--- a/Std/Classes/RatCast.lean
+++ b/Std/Classes/RatCast.lean
@@ -1,0 +1,22 @@
+/-
+Copyright (c) 2014 Robert Lewis. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Robert Lewis, Leonardo de Moura, Johannes Hölzl, Mario Carneiro, Gabriel Ebner
+-/
+import Std.Data.Rat.Basic
+
+/-- Type class for the canonical homomorphism `Rat → K`. -/
+class RatCast (K : Type u) where
+  /-- The canonical homomorphism `Rat → K`. -/
+  protected ratCast : Rat → K
+
+/-- Canonical homomorphism from `Rat` to a division ring `K`.
+This is just the bare function in order to aid in creating instances of `DivisionRing`. -/
+@[coe, match_pattern] protected def Rat.cast {K : Type u} [RatCast K] : Rat → K := RatCast.ratCast
+
+-- see note [coercion into rings]
+instance [RatCast K] : CoeTail Rat K where coe := Rat.cast
+
+-- see note [coercion into rings]
+instance [RatCast K] : CoeHTCT Rat K where coe := Rat.cast
+
--- a/Std/Classes/RatCast.lean
+++ b/Std/Classes/RatCast.lean
@@ -10,6 +10,8 @@ class RatCast (K : Type u) where
   /-- The canonical homomorphism `Rat → K`. -/
   protected ratCast : Rat → K
 
+instance : RatCast Rat where ratCast := id
+
 /-- Canonical homomorphism from `Rat` to a division ring `K`.
 This is just the bare function in order to aid in creating instances of `DivisionRing`. -/
 @[coe, match_pattern] protected def Rat.cast {K : Type u} [RatCast K] : Rat → K := RatCast.ratCast
--- a/Std/Classes/RatCast.lean
+++ b/Std/Classes/RatCast.lean
@@ -10,7 +10,7 @@ class RatCast (K : Type u) where
   /-- The canonical homomorphism `Rat → K`. -/
   protected ratCast : Rat → K
 
-instance : RatCast Rat where ratCast := id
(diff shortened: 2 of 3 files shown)


digama0:
LGTM, but I would also like @gebner to review this, specifically regarding the `OfScientific` / `RatCast` instance setup.

Std/Data/Rat/Basic.lean
```diff
@@ -108,7 +108,7 @@ protected def ofScientific (m : Nat) (s : Bool) (e : Nat) : Rat :=
   else
     (m * 10 ^ e : Nat)
 
-instance : OfScientific Rat where ofScientific := Rat.ofScientific
+-- Instance of `OfScientific Rat` is defined via `RatCast` in `Std.Data.Classes.RatCast`.
```
digama0, line 111: I don't think this is actually true, because there is no `RatCast Rat` instance in the other file.

thorimur: how about something like "Instance of `OfScientific Rat` is to be obtained via `RatCast`. See `Std.Data.Classes.RatCast`"? just thought it would be helpful to have some documented pointer to there from here, one way or another

thorimur: Also, *should* there be an instance of `RatCast Rat` in std? Currently you need mathlib4 to get one, I believe

Std/Classes/RatCast.lean
```diff
@@ -0,0 +1,23 @@
...
+import Std.Data.Rat.Basic
+
+/-- Type class for the canonical homomorphism `Rat → K`. -/
+class RatCast (K : Type u) where
+  /-- The canonical homomorphism `Rat → K`. -/
+  protected ratCast : Rat → K
+
+/-- Canonical homomorphism from `Rat` to a division ring `K`.
+This is just the bare function in order to aid in creating instances of `DivisionRing`. -/
+@[coe, match_pattern] protected def Rat.cast {K : Type u} [RatCast K] : Rat → K := RatCast.ratCast
```
gebner, line 15: :+1: on making this uniform with {Nat,Int}.cast.

But this is now inconsistent with mathlib3.  Do we want to backport this change, or do you think we can patch this over using `#align`?

Std/Classes/RatCast.lean
```diff
@@ -0,0 +1,23 @@
...
+
+-- see note [coercion into rings]
+instance [RatCast K] : CoeTail Rat K where coe := Rat.cast
+
+-- see note [coercion into rings]
+instance [RatCast K] : CoeHTCT Rat K where coe := Rat.cast
+
+instance [RatCast α] : OfScientific α where ofScientific m s e := (Rat.ofScientific m s e : α)
```
gebner, line 23:
```suggestion
instance [DivisionRing α] : OfScientific α where ofScientific m s e := (Rat.ofScientific m s e : α)
```
I thought this was the plan?

digama0: I think it was (is?). The trouble with this is that it leaves std without an `OfScientific Rat` / `RatCast Rat` instance since it doesn't have `DivisionRing`.

thorimur: It was indeed the plan; however, `DivisionRing` requires mathlib4 (`Mathlib.Algebra.Field.Defs`), so we'd have to either move `DivisionRing` into std or move this instance out of std.

I will say that I've just updated leanprover-community/mathlib4#1707 locally with these std changes to see if everything worked with this new instance, and it does. :)

On reflection, using `RatCast` for this instance also seems reasonable to me: if you know how to turn a rational number into an element of your type, you know how to turn a decimal expression into an element of your type. This new instance effectively says "decimals are simply rational numbers", even if you don't respect the algebraic properties of the rationals.

gebner: > This new instance effectively says "decimals are simply rational numbers"

The question when making this an instance is not whether it's a reasonable way.  But whether it's the *only* reasonable way.

For example, we probably want a `RatCast` instance for `Float`.  But without changing core, this will give rise to a different `OfScientific` instance for `Float`.  And then you end up in a situation where you have two `-0` literals that are not equal, which is something I'd really like to avoid.  (Note that `(0 : Float)` and `(-0 : Float)` (should) evaluate to two different numbers.)

This was the reason why I was okay to have the `[DivisionRing R] : OfScientific R` instance.  Because at least then there's only one reasonable choice for the instance.  (And the risk for non-defeq diamonds is lower since we're excluding floats.)

thorimur: > I think it is possible to do this with overlapping instances (e.g. `OfScientific Rat` here and `[DivisionRing α] : OfScientific α` in mathlib), as long as we can make them defeq at instance reducibility (which is what is needed for `norm_num` to work).

I haven't had much luck doing this myself. In a file where I can evaluate it, I get
```
attribute [-instance] instOfScientific

instance y : OfScientific Rat where ofScientific := Rat.ofScientific
instance x [DivisionRing α] : OfScientific α where ofScientific m s e := (y.ofScientific m s e : α)
instance x' [DivisionRing α] : OfScientific α where ofScientific m s e := (Rat.ofScientific m s e : α)

example : y = x := by with_reducible_and_instances rfl -- fails
example : y = x' := by with_reducible_and_instances rfl -- fails
```
Is there a better way?

gebner: I think it would be okay to keep the instance here as an `abbrev` and only use it to define an instance for division rings in mathlib4.  (And maybe for rationals already in std4.)

gebner: > I think it is possible to do this with overlapping instances (e.g. `OfScientific Rat` here and `[DivisionRing α] : OfScientific α` in mathlib), as long as we can make them defeq at instance reducibility (which is what is needed for `norm_num` to work).

What would that achieve?  The point here is that there shouldn't be a generic `[RatCast R] : OfScientific R` instance, because it causes diamonds with floats.  Adding an extra instance doesn't help with that.

Std/Data/Rat/Basic.lean
gebner: > Also, _should_ there be an instance of `RatCast Rat` in std? Currently you need mathlib4 to get one, I believe

I think `RatCast Rat` belongs in std4.  We also have `NatCast Nat` and `IntCast Int` here.

Std/Classes/RatCast.lean
thorimur: > The point here is that there shouldn't be a generic `[RatCast R] : OfScientific R` instance, because it causes diamonds with floats.

Does it? `#synth RatCast Float` fails for me. Is this expected to change in the future?

thorimur: Also I think Mario is proposing using `OfScientific Rat` and `[DivisionRing α] : OfScientific α` instead of `[RatCast α] : OfScientific α`, not in addition to it. But we would still need to solve the defeq problem somehow (assuming I'm testing it correctly).

thorimur: Would giving the `[RatCast α] : OfScientific α` instance a lower priority work? Or would that be too fragile even if it did work?

Std/Data/Rat/Basic.lean
thorimur: I actually can't find `NatCast Nat` or `IntCast Int` in std4...were they removed?

Std/Classes/RatCast.lean
digama0: It's not an issue of priority. You can get diamonds due to composing lemmas which each made the best available choice for the context in which they were stated.

Std/Classes/RatCast.lean
digama0: What's the inconsistency? Is it that there is no `RatCast` class in mathlib3? It seems relatively unlikely to break anything but a backport is the easiest way to be sure and avoid periodic questions from porters about it.

thorimur: There is an analogous `RatCast` class in mathlib; the current alignment is `#align has_rat_cast RatCast` in Mathlib.Algebra.Field.Defs. (As far as I can see there's no alignment for its field `rat_cast`.)

Std/Data/Rat/Basic.lean
gebner: Ah, we have `NatCast Int` though.  I think the missing "reflexive" instances are just an oversight.

Std/Classes/RatCast.lean
gebner: The difference to the existing `RatCast` class is that we don't have a helper `Rat.cast` definition but call `RatCast.ratCast` directly right now.  That's why I was concerned about the alignment because we're mapping 1 mathlib3 definition to 2 mathlib4 ones.

thorimur: I see, makes sense—should we e.g. prefer `Rat.cast` everywhere in mathlib4 then?

When we bump mathlib4's version of std4 after this PR, we could(?) `#align has_rat_cast.rat_cast Rat.cast` and do a find-replace in mathlib4 for `RatCast.ratCast` → `Rat.cast` (of which there aren't many occurrences right now, so we could make sure this really is the right thing to do in each of them). (If this PR is merged, we'll need to do that to fix at least one `norm_cast` lemma anyway, since a different thing has `@[coe]` now.)

Not sure of any subtle gotchas waiting for us here, though...

Std/Data/Rat/Basic.lean
thorimur: Oh, okay! Shall I add them all in, then?

Std/Classes/RatCast.lean
thorimur: I figured out a way to make the instances of `OfScientific Rat` and `[DivisionRing α] : OfScientific α` when `α` is `Rat` defeq: use `Rat.cast` in both.
```
instance : RatCast Rat where ratCast := id

instance x : OfScientific Rat where ofScientific m s e := Rat.cast (Rat.ofScientific m s e)

/-- mock division ring -/
class DivisionRing (α) extends RatCast α

instance : DivisionRing Rat where ratCast := id

instance y [DivisionRing α] : OfScientific α where
  ofScientific m s e := (Rat.ofScientific m s e : α)

example : x = y := by with_reducible_and_instances rfl -- works
```

digama0: There is no need for that, you can just avoid `id` in the `DivisionRing Rat` instance and it will also work:
```lean
instance x : OfScientific Rat where ofScientific m s e := Rat.ofScientific m s e
...
instance : DivisionRing Rat where ratCast x := x
...
example : x = y := by with_reducible_and_instances rfl -- works
```

Std/Classes/Cast.lean
```diff
@@ -11,9 +11,12 @@ class NatCast (R : Type u) where
   /-- The canonical map `Nat → R`. -/
   protected natCast : Nat → R
 
+instance : NatCast Nat where natCast := id
```
gebner, line 14:
```suggestion
instance : NatCast Nat where natCast n := n
```

`id` is not reducible.

Std/Classes/Cast.lean
```diff
@@ -26,9 +29,12 @@ class IntCast (R : Type u) where
   /-- The canonical map `Int → R`. -/
   protected intCast : Int → R
 
+instance : IntCast Int where intCast := id
```
gebner, line 32:
```suggestion
instance : IntCast Int where intCast n := n
```

Std/Classes/RatCast.lean
```diff
@@ -0,0 +1,24 @@
...
+Authors: Robert Lewis, Leonardo de Moura, Johannes Hölzl, Mario Carneiro, Gabriel Ebner
+-/
+import Std.Data.Rat.Basic
+
+/-- Type class for the canonical homomorphism `Rat → K`. -/
+class RatCast (K : Type u) where
+  /-- The canonical homomorphism `Rat → K`. -/
+  protected ratCast : Rat → K
+
+instance : RatCast Rat where ratCast := id
```
gebner, line 13:
```suggestion
instance : RatCast Rat where ratCast n := n
```

gebner:
LGTM.

PR #105, opened 2023-03-05
feat: Sum.inl, Sum.inr

JamesGallicchio:
More intuitive names than `.inl` and `.inr` for non-type-theorists.

See https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/No.20Either.20type

### DIFF
--- a/Std/Data/Sum.lean
+++ b/Std/Data/Sum.lean
@@ -1,0 +1,15 @@
+/-
+Copyright (c) 2023 James Gallicchio. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: James Gallicchio
+-/
+
+/-- Alias for `Sum.inl`. -/
+@[match_pattern] def Sum.left (x : α) : α ⊕ β := .inl x
+
+/-- Alias for `Sum.inr`. -/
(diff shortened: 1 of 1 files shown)


Std/Data/Sum.lean
```diff
@@ -0,0 +1,15 @@
...
+
+/-- Alias for `Sum.inl`. -/
+@[match_pattern] def Sum.left (x : α) : α ⊕ β := .inl x
+
+/-- Alias for `Sum.inr`. -/
+@[match_pattern] def Sum.right (x : β) : α ⊕ β := .inr x
+
+example : Nat ⊕ String := .left 0
+example (x : Nat ⊕ String) :=
+  match x with | .left n => n | .right _ => 0
```
digama0, line 15:
```suggestion
```
Examples go in test files

JamesGallicchio:
do these definitions need any of `reducible, inline, always_inline`?

digama0:
if you use `abbrev` they will get `reducible, inline`

eric-wieser:
It's possibly a bit confusing that `prod.left` is a projection but `sum.left` is a constructor.

fgdorais:
This WIP PR has been inactive for over a year. Should we close it?

PR #107, opened 2023-03-14
fix: prevent `Int.ofNat` being used instead of `Nat.cast`

eric-wieser:
Not tested, I'm hoping CI will do that.

### DIFF
--- a/Std/Classes/Cast.lean
+++ b/Std/Classes/Cast.lean
@@ -12,6 +12,7 @@ class NatCast (R : Type u) where
   protected natCast : Nat → R
 
 instance : NatCast Nat where natCast n := n
+instance : NatCast Int where natCast n := Int.ofNat n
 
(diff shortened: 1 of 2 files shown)


gebner:
FYI CI is failing.

eric-wieser (commented):
It looks like CI is failing because [`Int.instNatCastInt`](https://leanprover-community.github.io/mathlib4_docs/Std/Data/Int/Lemmas.html#Int.instNatCastInt) isn't available yet.

Should I move that instance to `Std/Classes/Cast`, or move this instance to `Std/Data/Int/Lemmas`?

gebner:
I would move both into `Std/Classes/Cast.lean`.  The instance doesn't have any dependencies, and forgetting to import `Std/Data/Int/Lemmas` seems like an easy mistake to make.

PR #109, opened 2023-03-21
feat: make `Rat.ofScientific` `@[irreducible]`

thorimur:
We make `Rat.ofScientific` irreducible to prevent it from being unfolded during defeq checks in which we'd rather unfold e.g. `RatCast.ratCast` on the rhs. See [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/finishing.20up.20ofScientific.20in.20norm_num).

### DIFF
--- a/Std/Data/Rat/Basic.lean
+++ b/Std/Data/Rat/Basic.lean
@@ -102,7 +102,8 @@ def divInt : Int → Int → Rat
 @[inherit_doc] scoped infixl:70 " /. " => Rat.divInt
 
 /-- Implements "scientific notation" `123.4e-5` for rational numbers. (This definition is
-`@[irreducible]` because you don't want to unfold it. Use `Rat.ofScientific_def` instead.) -/
+`@[irreducible]` because you don't want to unfold it. Use `Rat.ofScientific_true_def` or
+`Rat.ofScientific_false_def` instead.) -/
 @[irreducible] protected def ofScientific (m : Nat) (s : Bool) (e : Nat) : Rat :=
   if s then
     Rat.normalize m (10 ^ e) <| Nat.ne_of_gt <| Nat.pos_pow_of_pos _ (by decide)
--- a/Std/Data/Rat/Basic.lean
+++ b/Std/Data/Rat/Basic.lean
@@ -102,7 +102,7 @@ def divInt : Int → Int → Rat
 @[inherit_doc] scoped infixl:70 " /. " => Rat.divInt
 
 /-- Implements "scientific notation" `123.4e-5` for rational numbers. (This definition is
-`@[irreducible]` because you don't want to unfold it over e.g. `RatCast.ratCast`.) -/
+`@[irreducible]` because you don't want to unfold it. Use `Rat.ofScientific_def` instead.) -/
 @[irreducible] protected def ofScientific (m : Nat) (s : Bool) (e : Nat) : Rat :=
   if s then
     Rat.normalize m (10 ^ e) <| Nat.ne_of_gt <| Nat.pos_pow_of_pos _ (by decide)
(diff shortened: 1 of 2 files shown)


Std/Data/Rat/Basic.lean
```diff
@@ -101,8 +101,9 @@ def divInt : Int → Int → Rat
 
 @[inherit_doc] scoped infixl:70 " /. " => Rat.divInt
 
-/-- Implements "scientific notation" `123.4e-5` for rational numbers. -/
-protected def ofScientific (m : Nat) (s : Bool) (e : Nat) : Rat :=
+/-- Implements "scientific notation" `123.4e-5` for rational numbers. (This definition is
+`@[irreducible]` because you don't want to unfold it over e.g. `RatCast.ratCast`.) -/
```
digama0, line 105:
```suggestion
/-- Implements "scientific notation" `123.4e-5` for rational numbers. (This definition is
`@[irreducible]` because you don't want to unfold it. Use `Rat.ofScientific_def` instead.) -/
```
Wording copied from `Rat.mul`. (Yes, you should also add `ofScientific_def` in `Data.Rat.Lemmas`)

thorimur: (Got it, for some reason was under the impression the needed lemmas were autogenerated.) 

Does this mean we want something like `Rat.ofScientific_true_def` and `Rat.ofScientific_false_def`, or are we ok with that `if` being in the type?

thorimur: Actually the rhs of the equality even for the true version looks odd:
```
normalize m (10 ^ e) <| Nat.ne_of_gt <| Nat.pos_pow_of_pos _ (by decide)
```
...is there a better lemma we could state here?

thorimur: I pushed one possible solution, patterned off of `sub_def`, `sub_def'`; if it's not right we'll change it. :) (Also not sure if it should be `ofScientific_of_true_def` or not, or if brevity is better.)

digama0: you should use `mkRat` instead of `normalize` in both cases. I agree that it would also be better to have separate true and false cases, and it is optional whether you want to add `ofScientific_def` itself.

digama0: `ofScientific_true_def` and `ofScientific_false_def` seem fine to me as names.

thorimur: 👍 Ok, I've removed the `normalize` version of `ofScientific_true_def` and just left the `mkRat` version. (Note that `ofScientific_false_def` just has a `Nat` on the rhs.) Also added `ofScientific_def` for completeness' sake.

Std/Data/Rat/Lemmas.lean
```diff
@@ -298,3 +298,17 @@ theorem inv_def (a : Rat) : a.inv = a.den /. a.num := by
...
 
 theorem div_def (a b : Rat) : a / b = a * b.inv := rfl
+
+theorem ofScientific_true_def : Rat.ofScientific m true e = mkRat m (10 ^ e) := by
+  unfold Rat.ofScientific
+  exact normalize_eq_mkRat (Nat.ne_of_gt <| Nat.pos_pow_of_pos _ (by decide))
+
+theorem ofScientific_false_def : Rat.ofScientific m false e = (m * 10 ^ e : Nat) := by
+  unfold Rat.ofScientific
+  rfl
```
digama0, line 308:
```suggestion
  unfold Rat.ofScientific; apply normalize_eq_mkRat

theorem ofScientific_false_def : Rat.ofScientific m false e = (m * 10 ^ e : Nat) := by
  unfold Rat.ofScientific; rfl
```

thorimur: It didn't like `apply normalize_eq_mkRat`; the alternative along that direction is `unfold Rat.ofScientific; apply normalize_eq_mkRat ∘ Nat.ne_of_gt ∘ Nat.pos_pow_of_pos _; decide`, but I thought that was less clear than an `exact`.

digama0: no you shouldn't be providing the proof at all here, it is supposed to be found by unification on the existing term. did you try `rw [normalize_eq_mkRat]`?

thorimur: Ah gotcha. Yup, that followed by a `rfl` worked.

PR #113, opened 2023-04-09
feat: ext fails if there are too many patterns

semorrison:
This scenario has shown up several times in porting. It's easier to fail right away than have to diagnose this yourself.


Std/Tactic/Ext.lean
```diff
@@ -167,7 +167,12 @@ elab_rules : tactic
     let pats := RCases.expandRIntroPats pats
     let depth := n.map (·.getNat) |>.getD 1000000
     let gs ← extCore (← getMainGoal) pats.toList depth
-    replaceMainGoal <| gs.map (·.1) |>.toList
+    let unconsumed : List (TSyntax `rcasesPat) := match gs.map (·.2) |>.toList with
+    | [] => []
+    | h :: t => t.foldl (init := h) fun x y => x.filter fun p => y.any (p == ·)
+    match unconsumed with
+    | [] => replaceMainGoal <| gs.map (·.1) |>.toList
+    | _ => throwError "`ext` did not consume the patterns: {unconsumed}"
```
digama0, line 175: This should be a warning, as the input can still be interpreted sensibly.

semorrison: Fixed.

PR #115, opened 2023-04-16
feat: repeat1', which fails if no progress is made

semorrison:


Std/Lean/Meta/Basic.lean
```diff
@@ -306,32 +306,52 @@ def mkFreshIdWithPrefix [Monad m] [MonadNameGenerator m] («prefix» : Name) :
   pure r
 
 /--
-`repeat' f` runs `f` on all of the goals to produce a new list of goals,
-then runs `f` again on all of those goals, and repeats until `f` fails on all remaining goals,
-or until `maxIters` total calls to `f` have occurred.
+Implementation of `repeat'` and `repeat1'`.
```
digama0, line 309: This should include a description of the behavior. It can probably just be an adaptation of the original `repeat'` docstring.

semorrison: Fixed.

PR #120, opened 2023-04-20
feat: implement `ext?` and `ext!?`

joneugster:
Implement `ext?`, `ext1?` to show used extensionality lemmas, and `ext!?` to brute-force try all ext-lemmas.

- `ext?` displays a tactic replacement suggestion that could replace the `ext`. The tactic sequence comprises the tactics `apply`, `intros`, and `sorry`.
- `ext!?` ignores the type in the `DiscrTree` and goes through all lemmas but yields a warning when it applied a lemma that had the wrong type, suggesting to add a copy of the lemma to the library.


joneugster:
This is one of my first PRs to std4 (or doing meta-programming in general). I'd appreciate any feedback about the workflow or other things I didn't pick up on yet :)

Std/Tactic/Ext.lean
```diff
@@ -98,6 +104,8 @@ def applyExtLemma (goal : MVarId) : MetaM (List MVarId) := goal.withContext do
         let c ← mkConstWithFreshMVarLevels lem.declName
         let (_, _, declTy) ← withDefault <| forallMetaTelescopeReducing (← inferType c)
         guard (← isDefEq tgt declTy)
+      if tactic.ext.trace.get (← getOptions) then
+        logInfo s!"try: apply {lem.declName}"
```
digama0, line 108:
```suggestion
        logInfo m!"try: apply {c}"
```

joneugster, line 108: Is there a convenient way to bundle multiple log messages into one? Would need to be in `withExtN` and I don't know how to do that besides just adding the log string as a further argument to all the functions, which seems inconvenient.

Maybe using `Std.Tactic.TryThis` somehow?

Std/Tactic/Ext.lean
```diff
@@ -171,17 +181,29 @@ elab_rules : tactic
...
-/-- `ext? pat*` is like `ext pat*` but gives a suggestion on what pattern to use -/
-syntax "ext?" (colGt ppSpace rintroPat)* (" : " num)? : tactic
+/-- `ext1? pat*` is like `ext1 pat*` but gives a suggestion on what pattern to use.
+(you might need to call `intros` between the suggested lemmas and consider to which
+goal the lemmas should be applied.) -/
+syntax (name := ext1Trace) "ext1?" (colGt ppSpace rintroPat)* : tactic
+
+/-- `ext? pat*` is like `ext pat*` but gives a suggestion on what pattern to use.
+(you might need to call `intros` between the suggested lemmas and consider to which
+goal the lemmas should be applied.) -/
```
digama0, line 197: This sounds a bit awkward. Ideally we should produce exactly correct proof expressions.

joneugster: I haven't quite figured out how to track what `intros` does, but if you have a hint, I'll implement it.

I think `ext?` in lean3 did not create a working proof either, but might be wrong

Std/Tactic/Ext.lean
digama0: This depends on what the interface you are going for is, but yes I would expect this to involve passing the accumulated trace through `withExtN`. You can also do things with IO.Ref if you want to pass state through a combinator that doesn't support it but runs in a monad supporting IO.

Std/Tactic/Ext.lean
digama0: you don't need to, you can just say `try: intros`. Re: lean 3 not producing working proofs, I don't think that's an argument for reimplementing bugs.

Std/Tactic/Ext.lean
```diff
@@ -171,17 +181,29 @@ elab_rules : tactic
...
 
--- TODO
-/-- `ext1? pat*` is like `ext1 pat*` but gives a suggestion on what pattern to use -/
-syntax "ext1?" (colGt ppSpace rintroPat)* : tactic
-/-- `ext? pat*` is like `ext pat*` but gives a suggestion on what pattern to use -/
-syntax "ext?" (colGt ppSpace rintroPat)* (" : " num)? : tactic
+/-- `ext1? pat*` is like `ext1 pat*` but gives a suggestion on what pattern to use.
+(you might need to call `intros` between the suggested lemmas and consider to which
+goal the lemmas should be applied.) -/
+syntax (name := ext1Trace) "ext1?" (colGt ppSpace rintroPat)* : tactic
```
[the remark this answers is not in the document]
joneugster: done.

Std/Tactic/Ext.lean
joneugster: Good point, I'll give it a try!

joneugster: done.

joneugster:
I think now I'm happy with the functionality and should it be ready for a review.

Std/Tactic/Ext.lean
```diff
@@ -98,23 +109,41 @@ def applyExtLemma (goal : MVarId) : MetaM (List MVarId) := goal.withContext do
...
 
 /-- Apply a single extensionality lemma to the current goal. -/
-elab "apply_ext_lemma" : tactic => liftMetaTactic applyExtLemma
+elab "apply_ext_lemma" : tactic => do
+  usedExtTactics.modify fun _ => #[]
+  liftMetaTactic applyExtLemma
+  if tactic.ext.trace.get (← getOptions) then
+    let x ← usedExtTactics.get
+    let cmd ←`(tactic| · $x*)
+    logInfo m!"Try this:\n{cmd}"
```
joneugster, line 126: Is there a way to just print a sequence of tactics without any additional syntax?`` `(tactic| · $x*)`` adds the `·` and `` `(tactic| ($x*))`` adds parentheses around the block.

test/ext.lean
```diff
@@ -76,6 +76,68 @@ example (f : ℕ × (ℕ → ℕ)) : f = f := by
...
+  repeat admit
+
+end nested
+
+
+
+-- Note: This is a terrible `ext`-lemma that will cause all future calls to
+-- `ext` to fail by max. recursion depth.
+-- TODO: Why do we have this test?
 @[ext] theorem ext_intros {n m : Nat} (w : ∀ n m : Nat, n = m) : n = m := by apply w
```
joneugster, line 141: This is a horrible `ext`-lemma that crashes further tests and cost me some unecessary debugging time. Can we get rid of that test or what was it used for?

Std/Tactic/Ext.lean
```diff
@@ -106,25 +127,65 @@ def applyExtLemma (goal : MVarId) : MetaM (List MVarId) := goal.withContext do
...
       -- more useful in practice with dependently typed arguments of `@[ext]` lemmas.
       return ← goal.apply (cfg := { newGoals := .all }) (← mkConstWithFreshMVarLevels lem.declName)
     catch _ => s.restore
+  -- Fall back to try all `ext`-lemmas in case of failure
+  if tactic.ext.force.get (← getOptions) then
+    for (key, lemmas) in ← getAllExtLemmas do
+      -- TODO: `if key == ty then continue` to avoid going over the same key twice.
+      -- As far as I see, we need more API from
+      -- [Lean.Meta.DiscrTree] exposed, but everything there is private.
+      -- The first projection of `getMatchCore` for example
```
joneugster, line 143: This is a bit unfortunate, we might iterate again over the "good" ext lemmas we already tried. To avoid it, we need to compare `ty : Expr` with `key : DistrTree.Key`. Every option I found, like `(← getMatchCore d e).1` are shielded behind some `private` declarations in core...

Std/Tactic/Ext.lean
```diff
@@ -6,9 +6,30 @@ Authors: Gabriel Ebner, Mario Carneiro
...
+This option is used by `ext!?` -/
+register_option tactic.ext.force : Bool := {
+  defValue := false
+  descr    := "When enabled, calls to `ext` will try any `ext`-lemma ignoring the
+  `DistrTree` keys."
+}
+
+/-- Global variable tracking tactics used by `ext?`. This is a collection of
+`apply $lem`, `rintro $pat`, `intros`, and `sorry` statements. -/
+initialize usedExtTactics : IO.Ref (Array (Syntax.Tactic)) ← IO.mkRef #[]
```
eric-wieser, line 32: I'm a bit confuse why these are global settings; can't they just be arguments to `ext`?

test/ext.lean
```diff
@@ -76,6 +76,68 @@ example (f : ℕ × (ℕ → ℕ)) : f = f := by
...
+    (h'' :  c₁.c = c₂.c ) : c₁ = c₂ := by
+  ext?
+  -- Try this:
+  -- · apply extCtoB
+  --   intros
+  --   sorry
+  --   sorry
+  --   apply extBtoA
+  --   sorry
+  --   sorry
```
eric-wieser, line 111: Can we teach this to use `·` to group subgoals?

Std/Tactic/Ext.lean
joneugster: There are indeed two options here, either use options or pass the settings through all the nested functions. If you think the latter is better, that should not be a hard refactor.

I took inspiration from `simp?` which is implemented this way in core:
https://github.com/leanprover/lean4/blob/f1412ddb4550c22a4302542e4191c54af0cf758e/src/Lean/Elab/Tactic/Simp.lean#L247-L250

Same for `usedExtTactics`: You could also pass it back and forth through these nested function calls, I just thought that would be more complicated as it would change all the signatures of them. And it would mean the trace is even passed around in the 99.99% of times you just use `ext`

test/ext.lean
joneugster: With the current design of using a `IO.mkRef #[]` to keep track of the used lemmas, I believe this would be very hard. I tried once and figured it wasn't worth the effort back then.

test/ext.lean
semorrison: I don't remember if local `ext` attributes are allowed, but could you put this in a section? Or if not, move it to a separate test file so it can't hurt anything?

Std/Tactic/Ext.lean
semorrison: We're only exercising this code interactively, so non-ideal performance is not a big concern.

test/ext.lean
semorrison: Can we just use `show_term`, and print a `refine` statement with `?_` holes? Seems easier, and gives better structured output.

Std/Tactic/Ext.lean
semorrison: Maybe @digama0 can comment, but I suspect that `simp?` is not necessarily the model to follow: Mario's design considerations may have been in part dealing with the difficulties of getting changes to core refactored, so rather than replumbing anything he worked around the edges.

It would probably fit better for the Mathlib/Std style to avoid using options and `Ref`s to pass state for tactics.

The trace could be returned inside an `Option`, which is `none` when not needed?

test/ext.lean
semorrison: Interesting! Maybe we should decipher that first. Could you post an example on zulip and ping me?

Std/Tactic/Ext.lean
digama0: Using `IO.Ref` is not the same as using a global ref though. I would avoid global refs if possible, because they have issues in multithreaded situations.

PR #122, opened 2023-04-21
feat: add option for ext to flatten structure fields

joneugster:
This has come up in the mathlib port: [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/!4.233360.20new.20ext.20lemma.20generation/near/351104097), in particular in leanprover-community/mathlib4#3360

```
@[ext]
structure NonemptyInterval (α : Type _) [LE α] extends Prod α α where
  fst_le_snd : fst ≤ snd
```

here we would like the generated `ext`-lemma to be `x.toProd = y.toProd → x = y` instead of
`(x.toProd.fst = y.toProd.fst) → (x.toProd.snd = y.toProd.snd) → x = y`.

The suggested syntax is `@[ext (flat := false)]` for not flattening structure fields. The default value is `true` which keeps flattening them as it currently does.

### DIFF
--- a/Std/Tactic/Ext.lean
+++ b/Std/Tactic/Ext.lean
@@ -178,7 +178,6 @@ def extCore (g : MVarId) (pats : List (TSyntax `rcasesPat))
 * `ext pat*`: Apply extensionality lemmas as much as possible,
   using `pat*` to introduce the variables in extensionality lemmas like `funext`.
 * `ext`: introduce anonymous variables whenever needed.
-* `ext (flat := false) pat*`: Do not flatten structure fields.
 -/
 syntax "ext" (colGt ppSpace rintroPat)* (" : " num)? : tactic
 elab_rules : tactic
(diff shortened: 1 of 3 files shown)


Std/Tactic/Ext.lean
```diff
@@ -41,8 +45,15 @@ def withExtHyps (struct : Name)
...
-scoped elab "ext_type%" struct:ident : term => do
-  withExtHyps (← resolveGlobalConstNoOverloadWithInfo struct) fun params x y hyps => do
+scoped elab "ext_type%" struct:ident flat:extFlatOption ? : term => do
+  let mut flat' := true
+  match flat with
+  | none => pure ()
+  | some flat => match flat with
+    | `(extFlatOption| (flat := true)) => pure ()
+    | `(extFlatOption| (flat := false)) => flat' := false
+    | _ => throwUnsupportedSyntax
```
joneugster, line 55: It feels very cumbersome how I parse the option here. Can this be improved?

joneugster:
Implemented independently in 62f319c

PR #124, opened 2023-04-28
feat: move and add theorems on `String`

chabulhwi:
* Move helper `simp` theorems on `String.Pos` from Mathlib4 to Std4.
* Move `String.utf8GetAux.inductionOn` from Mathlib4 to Std4.
* Add new helper `simp` theorems on `String.Pos`.


Std/Data/String.lean
```diff
@@ -35,9 +35,66 @@ termination_by go start stop => stop.byteIdx + 4 - start.byteIdx
...
+@[simp]
+theorem Pos.zero_add_string (s : String) : (0 : Pos) + s = ⟨s.utf8ByteSize⟩ := by
+  simp
+
+@[simp]
+theorem Pos.le_iff {i₁ i₂ : Pos} : i₁ ≤ i₂ ↔ i₁.byteIdx ≤ i₂.byteIdx :=
+  ⟨id, id⟩
+
+@[simp]
+theorem Pos.lt_iff {i₁ i₂ : Pos} : i₁ < i₂ ↔ i₁.byteIdx < i₂.byteIdx :=
```
digama0, line 81: Lemmas go in a `Lemmas` file unless they are absolutely necessary for defining functions.

chabulhwi: I'll do it after I wake up.

Std.lean
```diff
@@ -45,6 +45,8 @@ import Std.Data.Rat
 import Std.Data.Rat.Basic
 import Std.Data.Rat.Lemmas
 import Std.Data.String
+import Std.Data.String.Basic
+import Std.Data.String.Lemmas
```
chabulhwi, line 49: I add these `import` lines to Std.

digama0:
I removed `@[simp]` from lemmas that unfold `Pos` without applying some kind of extensionality rule first. I don't think we want to just unfold this abstraction everywhere.

PR #125, opened 2023-04-29
feat: add APIs around Array splitting

hargoniX:


Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,102 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+protected def last? (xs : Array α) : Option α :=
+  if h : 0 < xs.size then
+    some (xs.get ⟨xs.size - 1, Nat.sub_lt h (by decide)⟩)
+  else
+    none
+
+/--
+Returns the last element of the array.
+-/
+protected def last (xs : Array α) (h : 0 < xs.size) : α :=
```
digama0, line 173: This is `Array.back`

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,102 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+Returns the first element of the array.
+-/
+protected def first (xs : Array α) (h : 0 < xs.size) : α :=
+  xs.get ⟨0, h⟩
+
+/--
+Split an array apart into two arrays at a given index. The first half
+will contain all values from `[0, mid)`, the second will contain `[mid, xs.size)`.
+-/
+protected def splitAt (xs : Array α) (mid : Fin xs.size) : Array α × Array α :=
```
digama0, line 195: The right bound here is `mid <= xs.size`

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,102 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+  let rhs := Subarray.mk xs mid.val xs.size h (Nat.le_refl _) |>.toArray
+  (lhs, rhs)
+
+/--
+Split an array apart into two arrays at a given index. The first half
+will contain all values from `[0, mid)`, the second will contain `[mid, xs.size)`.
+Since `mid` is not provably an index of `xs` this variant will return `none`
+in case it isn't.
+-/
+protected def splitAt? (xs : Array α) (mid : Nat) : Option (Array α × Array α) :=
```
digama0, line 207: I'm not sure an option-returning version of this function is useful. More useful would be to just return `(xs, #[])` when `mid` is out of range.

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,75 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+protected def first? (xs : Array α) : Option α :=
+  if h : 0 < xs.size then
+    some (xs.get ⟨0, h⟩)
+  else
+    none
+
+/--
+Returns the first element of the array.
+-/
+protected def first (xs : Array α) (h : 0 < xs.size) : α :=
```
digama0, line 173: This (and the others) should either be renamed to `front`, or we should add `last` as an alias for `back` or else it will be confusing.

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,75 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+/--
+Returns the first and all the rest of the elements of the array, or `none` if it is empty.
+-/
+protected def splitFront? (xs : Array α) : Option (α × Array α) :=
+  if h : 0 < xs.size then
+    let head := xs.get ⟨0, h⟩
+    let tail := Subarray.mk xs 1 xs.size (Nat.succ_le_of_lt h) (Nat.le_refl _) |>.toArray
+    some (head, tail)
+  else
+    none
```
fgdorais, line 138: No need to repeat code:
```lean
if h : < xs.size then
  some <| splitFront xs h
else
  none  
```

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,75 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+
+/--
+Returns the last and all the rest of the elements of the array, or `none` if it is empty.
+-/
+protected def splitLast? (xs : Array α) : Option (Array α × α) :=
```
fgdorais, line 143: Move after next def and deduplicate as above.

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,75 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+protected def splitLast (xs : Array α) (h : 0 < xs.size) : α × Array α :=
+  have h2 := Nat.sub_lt h (by decide)
+  let heads := xs.get ⟨xs.size - 1, h2⟩
+  let last := Subarray.mk xs 0 (xs.size - 1) (Nat.zero_le _) (Nat.le_of_lt h2) |>.toArray
+  (heads, last)
+
+/--
+Returns the first element of the array, or `none` if it is empty.
+-/
+protected def front? (xs : Array α) : Option α :=
```
fgdorais, line 164: Move after next def and deduplicate as above.

fgdorais (changes requested):
Code duplication.

fgdorais (approved):

hargoniX:
Are there still issues with this branch @digama0?

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,69 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+  else
+    none
+
+/--
+Returns the last and all the rest of the elements of the array.
+-/
+protected def splitLast (xs : Array α) (h : 0 < xs.size) : Array α × α :=
+  have h2 := Nat.sub_lt h (by decide)
+  let last := xs.get ⟨xs.size - 1, h2⟩
+  let heads := Subarray.mk xs 0 (xs.size - 1) (Nat.zero_le _) (Nat.le_of_lt h2) |>.toArray
```
digama0, line 143: This should use `pop` or its underlying machinery, because this will avoid making a copy if the input array is unshared

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,69 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+    none
+
+/--
+Split an array apart into two arrays at a given index. The first half
+will contain all values from `[0, mid)`, the second will contain `[mid, xs.size)`.
+Notably if `mid` is out of bounds for `xs` the function will return `(xs, #[])`.
+-/
+protected def splitAt (xs : Array α) (mid : Nat) : Array α × Array α :=
+  if h : mid ≤ xs.size then
+    let lhs := Subarray.mk xs 0 mid (Nat.zero_le _) h |>.toArray
```
digama0, line 177: Same here, the lhs could be a truncate of the input array

hargoniX: We don't have `Array.truncate` right now though. We do have `Array.shrink` but that just pops recursively and I wouldn't expect my splitAt function to be `O(n)`. That said I guess it would be reasonable to add a truncate C primitive as well? Do we want to have C code in std4?

digama0: > I wouldn't expect my splitAt function to be O(n)

It's already O(n), because you have to make a copy of the whole array (in two pieces). I think a truncate primitive should be added to core if needed.

semorrison:
I am tempted to suggest that these functions should be returning `Subarray`s, so they can be 0(1). 

However `Subarray` is rather underdeveloped, and has some issues, e.g. https://github.com/leanprover/lean4/issues/2360, so I'm not sure.

fgdorais:
Maybe Std should have its own `Subarray` with a better interface (leanprover/lean4#2164) and without a coercion to `Array` (leanprover/lean4#2360).

hargoniX:
So what is the call here? Merge like this (possibly until a Subarray API exists)? Or shall I try to develop a Subarray API for std and then re-implement this in terms of the std Subarray API.

fgdorais:
@hargoniX The smart call is: what do *you* actually need? It's definitely not a great for *you* to wait on a major array refactor. Do you need this now or can you wait another four or more months until something much better comes along?

semorrison:
My preference would be, assuming that this PR is actually needed somewhere and the performance of this PR is sufficient for that purpose, that we add a comment suggesting that in future when there is a usable `Subarray` API that these functions be refactored to use it. And otherwise proceed as is.

digama0:
We want the Subarray API to be similarly expansive to the Array API. But that doesn't affect this PR much, unless the author wants to write Subarray versions of these definitions.

hargoniX:
> @hargoniX The smart call is: what do _you_ actually need? It's definitely not a great for _you_ to wait on a major array refactor. Do you need this now or can you wait another four or more months until something much better comes along?

I'm just adding APIs that I think could be useful I don't use them (yet).

joehendrix:
I'd be inclined to not include these functions in stdlib.  The performance of the split variants is `O(n)` and so probably not a great choice for code where performance matters.  For theorem proving one would also want simplification theorems and related rules.

I'm separately partial to `head` instead of `front` since that's common in related libraries like the Haskell [`vector`](https://hackage.haskell.org/package/vector-0.13.1.0/docs/src/Data.Vector.html#head).

Expanding the subarray API (or theorems) would be great though.

hargoniX:
I'm not going to invest more time in it.

fgdorais:
Understood. I'm removing `awaiting-author` and adding `help-wanted` to signal that this PR is "for adoption".

PR #126, opened 2023-04-29
feat: Array  windows and chunks API from Rust

hargoniX:
Right now without the respective reverse chunks functions to check if this API design is fine before doing it wrong twice.

### DIFF
--- a/Std/Data/Array/Basic.lean
+++ b/Std/Data/Array/Basic.lean
@@ -5,6 +5,7 @@ Authors: Arthur Paulino, Floris van Doorn, Jannis Limperg
 -/
 import Std.Data.Array.Init.Basic
 import Std.Data.Ord
+import Std.Data.Nat.Basic
 
 /-!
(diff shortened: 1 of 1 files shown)


joehendrix:
Are there other projects using this?  I think we are potentially better served more abstract than this in `Std`, but I'm not quite sure how this is used.

hargoniX:
This PR (like my other Array based one) was inspired by the Rust API functions for `Vec` (https://doc.rust-lang.org/stable/std/vec/struct.Vec.html) since Lean often tends to copy Rust functionality I thought it might be a good idea to put it into our stdlib as well.

fgdorais:

hargoniX:
I'm not going to invest more time in it.

fgdorais:
Understood. I'm removing `awaiting-author` and adding `help-wanted` to signal that this PR is "for adoption".

fgdorais:

PR #127, opened 2023-04-30
feat: well founded stream class

digama0:
Based on discussions on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Porting.20just.20functionality/near/354657447). Still needs more love to make it polished.

### DIFF
--- a/Std/Classes/Stream.lean
+++ b/Std/Classes/Stream.lean
@@ -1,0 +1,101 @@
+import Std.WF
+import Std.Data.Nat.Lemmas
+
+class StreamAcc [Stream σ α] (s : σ) : Prop where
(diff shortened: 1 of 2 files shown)


fgdorais:
Just wondering what the status is here: is it deadlocked? on the backburner? abandoned?

digama0:
It's please-adopt. I'm not sufficiently invested to fix the remaining issues (or really recall what they are). There is some concern that this adds yet another way to write a for loop, which means we need more lemmas to relate the different options. Also the `StreamAccType` stuff is nasty, it needs core changes so that forIn can depend on the value and not just the type.

PR #128, opened 2023-05-03
feat: change and add theorems on `String`

chabulhwi:
* Rename the theorem `String.addString_eq` `String.addString_byteIdx`.
* Add new theorems to `Data.Nat.Lemmas` and `Data.String.Lemmas`.
* Move lemmas from `Mathlib.Data.String.Basic` to here.

### DIFF
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -47,7 +47,7 @@ theorem zero_addChar_byteIdx (c : Char) : ((0 : Pos) + c).byteIdx = csize c := b
 
 theorem zero_addChar_eq (c : Char) : (0 : Pos) + c = ⟨csize c⟩ := by rw [← zero_addChar_byteIdx]
 
-theorem lt_addChar (p : Pos) (c : Char) : p < p + c := Nat.lt_add_of_pos_right (String.csize_pos _)
+theorem lt_addChar (p : Pos) (c : Char) : p < p + c := Nat.lt_add_of_pos_right (csize_pos _)
 
 theorem ne_of_lt {i₁ i₂ : Pos} (h : i₁ < i₂) : i₁ ≠ i₂ := mt ext_iff.1 (Nat.ne_of_lt h)
 
@@ -71,7 +71,7 @@ theorem lt_iff {i₁ i₂ : Pos} : i₁ < i₂ ↔ i₁.byteIdx < i₂.byteIdx :
 end Pos
 
 private theorem add_csize_pos : 0 < i + csize c :=
-  Nat.lt_of_lt_of_le (String.csize_pos c) (Nat.le_add_left ..)
+  Nat.lt_of_lt_of_le (csize_pos c) (Nat.le_add_left ..)
 
 /-- The UTF-8 byte length of a list of characters. (This is intended for specification purposes.) -/
 @[inline] def utf8Len : List Char → Nat := utf8ByteSize.go
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -6,6 +6,7 @@ Authors: Bulhwi Cha
 import Std.Data.Char
 import Std.Data.Nat.Lemmas
 import Std.Tactic.Ext.Attr
+import Std.Tactic.Simpa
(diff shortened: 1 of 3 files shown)


Std/Data/String/Lemmas.lean
```diff
@@ -67,4 +72,9 @@ def utf8InductionOn {motive : List Char → Pos → Sort u}
       h ▸ eq c cs
     else ind c cs i h (utf8InductionOn cs (i + c) p nil eq ind)
 
+@[simp] theorem drop_empty {n : Nat} : "".drop n = "" := by
+  induction n with
+  | zero => rfl
+  | succ _ ih => exact ih
```
digama0, line 78: Can this be proved by applying extensionality and then a list lemma? I'm sure there are many more like this.

chabulhwi:
```lean
@[ext] theorem ext {s₁ s₂ : String} (h : s₁.data = s₂.data) : s₁ = s₂ :=
  show ⟨s₁.data⟩ = (⟨s₂.data⟩ : String) from h ▸ rfl

@[simp] theorem drop_empty {n : Nat} : "".drop n = "" := by
  apply ext
  show (drop "" n).data = []
  show [].drop n = ([] : List Char) -- Output: type mismatch
  exact List.drop_nil
```
@digama0 I don't think there's a way to do that.

digama0: I don't think it's a `show`, that's a rewrite / `simp`

chabulhwi:
```lean
ext] theorem ext {s₁ s₂ : String} (h : s₁.data = s₂.data) : s₁ = s₂ :=
  show ⟨s₁.data⟩ = (⟨s₂.data⟩ : String) from h ▸ rfl

theorem drop_eq : ∀ (s : String) (n : Nat), drop s n = ⟨s.toList.drop n⟩ := sorry

@[simp] theorem drop_empty {n : Nat} : "".drop n = "" := by
  apply ext
  rw [drop_eq]
  exact List.drop_nil
```
I used `String.drop_empty` to prove `String.drop_eq` in https://github.com/leanprover-community/mathlib4/pull/3789.

See https://github.com/leanprover-community/mathlib4/blob/5877ae2b8c35df2daf8ef993b937983edc1ea1e3/Mathlib/Data/String/Basic.lean#L172C1-L179.

digama0: Can that theorem be here?

chabulhwi: > Can that theorem be here?

Is there a way to make the following proof work in Std? If so, I can move more lemmas from `Mathilb.Data.String.Basic` to here.
```lean
example (a b c : Nat) (h : a = b) : a + c = b + c := by
  simp
  exact h
```

chabulhwi: > (If it can't be used to prove `drop_empty` then so be it, but it should be part of the general simp strategy expressed by the supplied `@[simp]` lemmas.)
> 
> "I'll do it in the next PR" is also a valid response to this request.

It can't because its proof already uses `drop_empty`.  Do you want me to add the `[simp]` attribute to `drop_nil` and [`drop_cons_succ`](https://github.com/leanprover-community/mathlib4/blob/bd11b7ca87c813093af51effb060e41ef8305560/Mathlib/Data/String/Basic.lean#L165-L170)?

chabulhwi: > Is there a way to make the following proof work in Std?

I think it'd be hard to do so.

> `drop_eq`'s proof already uses `drop_empty`.

@digama0, will you merge this PR? I don't know any other way to prove `drop_empty` than using induction on the natural numbers.

digama0: As I said, you don't need to change the proof of `drop_empty` but you should also put `drop_eq` (or rather, `toList_drop`) in this PR, unless there is a good reason not to, so that there is a confluent set of rewrites.

chabulhwi: @digama0 I moved lemmas including `drop_eq` from Mathlib to here.

PR #129, opened 2023-05-05
feat: move String.stripPrefix from mathlib

semorrison:
Two functions currently defined in mathlib, that I would like in a separate project.


Std/Data/String/Basic.lean
```diff
@@ -42,3 +42,13 @@ If `pre` is a prefix of `s`, i.e. `s = pre ++ t`, return the remainder `t`.
 -/
 def dropPrefix? (s : String) (pre : String) : Option Substring :=
   s.toSubstring.dropPrefix? pre.toSubstring
+
+/-- `s.stripPrefix p` will remove `p` from the beginning of `s` if it occurs there,
+or otherwise return `s`. -/
+def stripPrefix (s p : String) :=
```
digama0, line 48:
```suggestion
def stripPrefix (s p : String) : String :=
```

digama0: Also I think it would be better to return an option string here and leave the decision making for what to do when the prefix isn't present to the user via `getD` et al.

semorrison: mathlib4 also defines:

```
/-- `isPrefixOf? pre s` returns `some post` if `s = pre ++ post`.
If `pre` is not a prefix of `s`, it returns `none`. -/
def isPrefixOf? (pre s : String) : Option String :=
  if startsWith s pre then some <| s.drop pre.length else none

/-- `isSuffixOf? suff s` returns `some pre` if `s = pre ++ suff`.
If `suff` is not a suffix of `s`, it returns `none`. -/
def isSuffixOf? (suff s : String) : Option String :=
  if s.endsWith suff then some <| s.dropRight suff.length else none
```
for cases where you want to do the decision making yourself.

I find though that I want the "just get rid of it" functions frequently "inline" with more string processing functions, rather like you often see a `.trim` in the middle of a string pipeline.

Perhaps moving these `Option` values functions to `Std4` at the same time would be a good solution?

semorrison: Oh, `Std4` already has `dropPrefix? : String \to String \to Option Substring`, so I've pushed a commit using that.

We don't have `dropSuffix?` yet. I added an inefficient implementation, and if you'd be happy with this PR otherwise, I can replace it with the efficient one.

digama0: It seems like there should be a symmetric implementation for `dropSuffix?` since, assuming the string actually contains the substring, you know that `s.utf8Length - suff.utf8Length` is a valid `String.Pos` and can split there.

semorrison: Instead I factored out `Substring.eqAsString`, and implemented everything in terms of that. I *think* it is as performant, but you should be skeptical of that claim. :-) Happy to have a better name, too.

fgdorais: Isn't `Substring.eqAsString` basically a special case of [`String.substrEq`](https://github.com/leanprover/lean4/blob/master/src/Init/Data/String/Basic.lean#L481-L493) from core?

Std/Data/String/Basic.lean
```diff
@@ -9,29 +9,52 @@ import Std.Data.Nat.Lemmas
 
 namespace Substring
 
+/-- Check if two substrings are equal as strings. -/
+def eqAsString (s t : Substring) : Bool :=
+  if s.bsize ≠ t.bsize then
+    false
+  else
+    go s.bsize 0
```
fgdorais, line 17: `go s.bsize 0` is equivalent to `String.substrEq s.str s.startPos t.str t.startPos s.bsize`

Std/Data/String/Basic.lean
```diff
@@ -9,29 +9,52 @@ import Std.Data.Nat.Lemmas
...
+      none
+
+/--
+If `suff` is a suffix of `s`, i.e. `s = t ++ suff`, return the remainder `t`.
+-/
+def dropSuffix? (s : Substring) (suff : Substring) : Option Substring :=
+  if s.bsize < suff.bsize then
+    none
+  else
+    let p := s.stopPos + (suff.stopPos - suff.startPos)
```
digama0, line 53: The implementation of `dropPrefix?` is wrong because it uses `+` here instead of `-`:
```lean
#eval "ab".toSubstring.dropSuffix? "b".toSubstring -- none
```
But even after fixing that there is a more subtle issue which makes it still wrong:
```lean
def dropSuffix? (s : Substring) (suff : Substring) : Option Substring :=
  if s.bsize < suff.bsize then
    none
  else
    let p := s.stopPos - (suff.stopPos - suff.startPos)
    if { s with startPos := p }.eqAsString suff then
      some { s with stopPos := p }
    else
      none

#eval "←".toSubstring.dropSuffix? "A".toSubstring -- some "←".toSubstring
```
The issue is that `p` is not necessarily a UTF8 code point boundary of `s`, and this is a *precondition* of `String.substrEq` / `eqAsString`. Instead you have to walk backward from the end of the string, as the implementation of `s.endsWith suff` does.

fgdorais: May I recommend using these two functions to do the necessary comparisons:

```lean
protected theorem String.ne_zero_of_lt : {a b : Pos} → a < b → b ≠ 0
| _, _, hlt, rfl => Nat.not_lt_zero _ hlt

/-- Return the longest common prefix of two substrings. -/
def Substring.commonPrefix (s t : Substring) : Substring :=
  { s with stopPos := loop s.startPos t.startPos }
where
  loop spos tpos :=
    if h : spos < s.stopPos ∧ tpos < t.stopPos then
      if s.get spos == t.get tpos then
        have := Nat.sub_lt_sub_left h.1 (s.str.lt_next spos)
        loop (s.str.next spos) (t.str.next tpos)
      else
        spos
    else
      spos
termination_by loop => s.stopPos.byteIdx - spos.byteIdx

/-- Return the longest common suffix of two substrings. -/
def Substring.commonSuffix (s t : Substring) : Substring :=
  { s with startPos := loop s.stopPos t.stopPos }
where
  loop spos tpos :=
    if h : s.startPos < spos ∧ t.startPos < tpos then
      let spos' := s.str.prev spos
      let tpos' := t.str.prev tpos
      if s.get spos' == t.get tpos' then
        have : spos' < spos := s.str.prev_lt_of_pos spos (String.ne_zero_of_lt h.1)
        loop spos' tpos'
      else
        spos
    else
      spos
termination_by loop => spos.byteIdx
```

fgdorais: Actually, these are still incorrect. Comparing `bsize` is not sufficient because of [overlong encodings](https://en.wikipedia.org/wiki/UTF-8#Overlong_encodings).

fgdorais: Here is a correct version:
```lean
protected theorem String.ne_zero_of_lt : {a b : Pos} → a < b → b ≠ 0
| _, _, hlt, rfl => Nat.not_lt_zero _ hlt

private def Substring.commonPrefixAux (s t : Substring) : String.Pos × String.Pos :=
  loop s.startPos t.startPos
where
  loop spos tpos :=
    if h : spos < s.stopPos ∧ tpos < t.stopPos then
      if s.str.get spos == t.str.get tpos then
        have := Nat.sub_lt_sub_left h.1 (s.str.lt_next spos)
        loop (s.str.next spos) (t.str.next tpos)
      else
        (spos - s.startPos, tpos - t.startPos)
    else
      (spos - s.startPos, tpos - t.startPos)
termination_by loop => s.stopPos.byteIdx - spos.byteIdx

private def Substring.commonSuffixAux (s t : Substring) : String.Pos × String.Pos :=
  loop s.stopPos t.stopPos
where
  loop spos tpos :=
    if h : s.startPos < spos ∧ t.startPos < tpos then
      let spos' := s.str.prev spos
      let tpos' := t.str.prev tpos
      if s.str.get spos' == t.str.get tpos' then
        have : spos' < spos := s.str.prev_lt_of_pos spos (String.ne_zero_of_lt h.1)
        loop spos' tpos'
      else
        (spos - s.startPos, tpos - t.startPos)
    else
      (spos - s.startPos, tpos - t.startPos)
termination_by loop => spos.byteIdx

/-- Return the longest common prefix of two substrings. -/
def Substring.commonPrefix (s t : Substring) : Substring :=
  { s with stopPos := (commonPrefixAux s t).1 + s.startPos }

/-- Return the longest common suffix of two substrings. -/
def Substring.commonSuffix (s t : Substring) : Substring :=
  { s with startPos := (commonSuffixAux s t).1 + s.startPos }

/--
If `pre` is a prefix of `s`, i.e. `s = pre ++ t`, return the remainder `t`.
-/
def Substring.dropPrefix? (s : Substring) (pre : Substring) : Option Substring :=
  let (spos, tpos) := commonPrefixAux s pre
  if tpos.byteIdx = pre.bsize then
    some { s with startPos := spos + s.startPos }
  else
    none

/--
If `suff` is a suffix of `s`, i.e. `s = t ++ suff`, return the remainder `t`.
-/
def Substring.dropSuffix? (s : Substring) (suff : Substring) : Option Substring :=
  let (spos, tpos) := commonSuffixAux s suff
  if tpos = 0 then
    some { s with stopPos := spos + s.startPos }
  else
    none
```

digama0: > Actually, these are still incorrect. Comparing `bsize` is not sufficient because of [overlong encodings](https://en.wikipedia.org/wiki/UTF-8#Overlong_encodings).

Overlong encodings are not encodings (the name is misleading). Those bit patterns are not legal UTF-8, and you are guaranteed not to see them in `String`s. A `a` character is always encoded in UTF-8 in one byte.

semorrison: Okay, switched to @fgdorais' implementation, thank you!

fgdorais (approved):
Looks good! Maybe a quick comment that `commonPrefix`/`commonSuffix` return a substring of the first since that is an implicit assumption for `dropPrefix?`/`dropSuffix?`. Also fix the missing docstrings for `loop`.

PR #130, opened 2023-05-08
feat: 'Try this' support

digama0:
Adds proper widget / code action support for "try this" prompts in tactics. Still open: merging the widget into the message itself, depends on leanprover/lean4#2064 .

### DIFF
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -85,7 +85,7 @@ The parameters are:
 * `ref`: the span of the info diagnostic
 * `suggestion`: the replacement syntax
 * `suggestionForMessage?`: the message to display in the info diagnostic (only).
-  The widget message uses only `suggestion`. If not provided, it uses `suggestion` in both places.
+  The widget message uses only `suggestion`. If not provided, `suggestion` is used in both places.
 * `origSpan?`: a syntax object whose span is the actual text to be replaced by `suggestion`.
   If not provided it defaults to `ref`.
 * `extraMsg`: an extra piece of message text to apply to the widget message (only).
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -19,38 +19,6 @@ namespace Std.Tactic.TryThis
 
 open Lean Elab Elab.Tactic PrettyPrinter Meta Server Lsp RequestM
 
-/-- An info-tree data node corresponding to an application of the "Try this" command. -/
-structure TryThisInfo where
-  /-- The suggested replacement for this syntax, usually the rendering of another tactic syntax. -/
-  suggestion : String
-  /-- This is the span to replace with `suggestion`. If not supplied it will default to
-  the span of the syntax on which this info node is placed. -/
-  span? : Option (String.Pos × String.Pos) := none
-  deriving TypeName
-
-/--
-This is a code action provider that looks for `TryThisInfo` nodes and supplies a code action to
-apply the replacement.
--/
-@[codeActionProvider] def tryThisProvider : CodeActionProvider := fun params snap => do
-  let doc ← readDoc
-  let startPos := doc.meta.text.lspPosToUtf8Pos params.range.start
-  let endPos := doc.meta.text.lspPosToUtf8Pos params.range.end
-  pure <| snap.infoTree.foldInfo (init := #[]) fun _ctx info result => Id.run do
-    let .ofCustomInfo info := info | result
-    let some tti := info.value.get? TryThisInfo | result
-    let some (head, tail) := (tti.span? <|> return (← info.stx.getPos?, ← info.stx.getTailPos?))
-      | result
-    unless head ≤ endPos && startPos ≤ tail do return result
-    result.push {
-      eager.title := "Apply 'Try this'"
-      eager.kind? := "refactor"
-      eager.edit? := WorkspaceEdit.ofTextEdit params.textDocument.uri {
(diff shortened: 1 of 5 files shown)


Std/Tactic/TryThis.lean
```diff
@@ -1,42 +1,126 @@
...
+export default function(props) {
+  const editorConnection = React.useContext(EditorContext)
+  function onClick() {
+    editorConnection.api.applyEdit({
+      changes: { [props.pos.uri]: [{ range: props.range, newText: props.suggestion }] }
+    })
+  }
+  return e('div', {className: 'ml1'}, e('pre', {className: 'font-code pre-wrap'}, [
+    'Try this: ',
+    e('a', {onClick, className: 'link pointer dim', title: 'Apply suggestion'}, props.suggestion),
```
fpvandoorn, line 73: Could you add `style: { cursor: 'pointer' }` to the `onClick` structure, so that the mouse becomes a pointer on hover? (suggested by Adam Topaz in the Zulip thread)

fpvandoorn:
Two more things I noticed: 
* The `tryThisAt` in your original code snippet doesn't produce a blue squiggle under the tactic. Is it possible to add this (currently I also use a `logInfoAt` to produce the squiggle)
* In the implementation of `recommend using h` I wrote, I put the `logInfoAt` and `tryThisAt` both on the word `recommend`. If I have my pointer at the end of the line, I still see the `logInfoAt` message that was created on the same line (without opening "All messages"). However, to see the tactic replacements, I need to have my pointer exactly on the word `recommend`. Ideally it would show it whenever you are on the same line.
image

Std/Tactic/TryThis.lean
```diff
@@ -1,42 +1,126 @@
...
 
-/-- Add a "try this" suggestion. (TODO: this depends on code action support) -/
-def addSuggestion [Monad m] [MonadLog m] [AddMessageContext m] [MonadOptions m]
-    (origStx : Syntax) (suggestion : Syntax) : m Unit :=
-  logInfoAt origStx m!"{suggestion}"
+/-- Add a "try this" suggestion. -/
+def addSuggestion (origStx : Syntax) {kind : Name} (suggestion : TSyntax kind)
+    (suggestionForMessage : Option MessageData := none)
+    (ref? : Option Syntax := none)
+    (extraMsg : String := "") : MetaM Unit := do
```
gebner, line 93: I think it would be clearer to swap (the names) `origStx` and `ref?` here.  Usually we use `ref` to specify where the error message should go.  While "original syntax" sounds like it should be the original tactic/term that we want to replace using the suggestion.

digama0: I reorganized the names a bit and added parameter documentation.

digama0:
@fpvandoorn See whether the new version addresses your concerns. The widget hover (and code action) now happens as long as your cursor is over any of the lines containing the message, same as the messages display.

Std/Tactic/TryThis.lean
```diff
@@ -1,42 +1,144 @@
...
+    { start := map.lineStart (map.toPosition stxRange.start).line
+      stop := map.lineStart ((map.toPosition stxRange.stop).line + 1) }
+    let range := map.utf8RangeToLspRange range
+    let json := Json.mkObj [("suggestion", text), ("range", toJson range), ("info", extraMsg)]
+    Widget.saveWidgetInfo ``tryThisWidget json (.ofRange stxRange)
+
+/-- Add a `exact e` or `refine e` suggestion.
+
+The parameters are:
+* `ref`: the original syntax to be replaced
```
Vtec234, line 113:
```suggestion
* `ref`: the span of the info diagnostic
```
Right?

Std/Tactic/TryThis.lean
```diff
@@ -1,42 +1,144 @@
...
+  addSuggestion ref tac (suggestionForMessage? := msg)
+    (origSpan? := origSpan?) (extraMsg := extraMsg)
 
-/-- Add a term suggestion. (TODO: this depends on code action support) -/
-def addTermSuggestion (origTerm : Syntax) (e : Expr) : TermElabM Unit := do
-  addSuggestion origTerm (← delabToRefinableSyntax e)
+/-- Add a term suggestion.
+
+The parameters are:
+* `ref`: the original syntax to be replaced
```
Vtec234, line 136:
```suggestion
* `ref`: the span of the info diagnostic
```

Std/Tactic/TryThis.lean
```diff
@@ -1,42 +1,144 @@
...
+  the suggestion
+* A code action `Apply 'Try this'` is added, which will apply the suggestion.
 
-/-- Add a `exact e` or `refine e` suggestion. (TODO: this depends on code action support) -/
-def addExactSuggestion (origTac : Syntax) (e : Expr) : TermElabM Unit := do
+The parameters are:
+* `ref`: the span of the info diagnostic
+* `suggestion`: the replacement syntax
+* `suggestionForMessage?`: the message to display in the info diagnostic (only).
+  The widget message uses only `suggestion`. If not provided, it uses `suggestion` in both places.
```
Vtec234, line 88:
```suggestion
  The widget message uses only `suggestion`. If not provided, `suggestion` is used in both places.
```
Otherwise it sounds like "the widget message uses `suggestion` in both places" which is presumably not what was meant.

Vtec234 (approved):
Looks good, all functional issues with the PR that we identified are to be fixed in core rather than here.

PR #133, opened 2023-05-09
feat: `#guard_expr` command and adjust handling of expected types for `guard_expr` tactics

kmill:
This adds `#guard_expr`, which is a command version of the `guard_expr` tactic

Furthermore, this adjusts the handling of expected types so that when synthetic metavariables are synthesized then instances are more likely to be correct. This affects numeric literals for example, which otherwise might become `Nat`s due to the default instance.

Lastly, adds a `#guard` command that uses the evaluator to check if the given expression evaluates to true.

This PR is moving commands defined in a testing file from [mathlib4#3427](https://github.com/leanprover-community/mathlib4/pull/3427) to std4.


kmill:
@digama0 I got around to moving the `#guard_expr` command to std and tidying things up. Just pinging you to let you know.

PR #134, opened 2023-05-10
fix: rename theorems on `String.csize`

chabulhwi:
Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/.60Char.2Ecsize_pos.60.20and.20.60String.2Ecsize_pos.60.20are.20the.20same

### DIFF
--- a/Std/Data/Char.lean
+++ b/Std/Data/Char.lean
@@ -19,3 +19,5 @@ theorem csize_pos (c) : 0 < csize c := by
 
 theorem csize_le_4 (c) : csize c ≤ 4 := by
   rcases csize_eq c with _|_|_|_ <;> simp_all
+  
+end String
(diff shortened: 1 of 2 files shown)


Std/Data/Char.lean
```diff
@@ -6,16 +6,16 @@ Authors: Jannis Limperg
...
+  simp only [csize, Char.utf8Size]
   repeat (first | split | (solve | simp))
 
-theorem csize_pos (c) : 0 < String.csize c := by
+theorem csize_pos (c) : 0 < csize c := by
   rcases csize_eq c with _|_|_|_ <;> simp_all
 
-theorem csize_le_4 (c) : String.csize c ≤ 4 := by
+theorem csize_le_4 (c) : csize c ≤ 4 := by
   rcases csize_eq c with _|_|_|_ <;> simp_all
```
digama0, line 21:
```suggestion
  rcases csize_eq c with _|_|_|_ <;> simp_all
  
end String
```
so that other theorems in this file aren't accidentally added to the wrong namespace. (Alternatively, move the theorems to Data.String.)

PR #135, opened 2023-05-11
tactic ext tooltip: Mention `: n` parameter

nomeata:


nomeata:
This would have helped me a bit today.

PR #136, opened 2023-05-12
fix: clean up `Data.String.Lemmas`

chabulhwi:
* Clean up code.
* Move `Nat.add_pos_left` and `Nat.add_pos_right` from `Mathlib.Data.Nat.Order.Basic` to `Std.Data.Nat.Lemmas`.
* Add lemmas to `Data.String.Lemmas`.
* Rename `String.extract.go₁_zero_endPos` `String.extract.go₁_zero_utf8Len`.
* Fix the statements of `Substring.next_cons_zero`, `Substring.next_cons_addChar`, and `Substring.nextn_cons_addChar`.

### DIFF
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -70,7 +70,7 @@ theorem lt_iff {i₁ i₂ : Pos} : i₁ < i₂ ↔ i₁.byteIdx < i₂.byteIdx :
 
 end Pos
 
-private theorem add_csize_pos {i : Nat} {c : Char} : 0 < i + csize c :=
+private theorem add_csize_pos : 0 < i + csize c :=
   Nat.add_pos_right _ (csize_pos c)
 
 private theorem ne_add_csize_add_self {i n : Nat} {c : Char} : i ≠ n + csize c + i :=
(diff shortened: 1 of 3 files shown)


digama0:
Note that I have some pending changes to Data.String.Lemmas that may more significantly change these proofs, introducing a notion of "position is valid for string" and "substring is valid" which is required for several lemmas. I won't block this on that work though. I don't really like `nextn_cons_addChar`, it's very tailor-made for the theorem it is a lemma for and there are much more general things that can be said about `nextn` that subsume this theorem.

Std/Data/String/Lemmas.lean
```diff
@@ -70,14 +70,22 @@ theorem lt_iff {i₁ i₂ : Pos} : i₁ < i₂ ↔ i₁.byteIdx < i₂.byteIdx :
 
 end Pos
 
-private theorem add_csize_pos : 0 < i + csize c :=
-  Nat.lt_of_lt_of_le (csize_pos c) (Nat.le_add_left ..)
+private theorem add_csize_pos {i : Nat} {c : Char} : 0 < i + csize c :=
```
digama0, line 73:
```suggestion
private theorem add_csize_pos : 0 < i + csize c :=
```
Std uses auto implicits where appropriate, so unless this is required to get the right types or binder info the preferred style is to omit implicit binders.

digama0 (approved):

PR #137, opened 2023-05-13
feat: MetaM exact tactic

hrmacbeth:


Std/Tactic/Exact.lean
```diff
@@ -0,0 +1,19 @@
+/-
+Copyright (c) 2023 Mario Carneiro. All rights reserved.
```
hargoniX, line 2: Is this the right copyright owner + author if you wrote the file?

Std/Tactic/Exact.lean
```diff
@@ -0,0 +1,19 @@
...
+
+/-! # `exact` tactic (`MetaM` version) -/
+
+open Lean Meta
+
+/--
+`MetaM` version of `Lean.Elab.Tactic.evalExact`: add `mvarId := x` to the metavariable assignment.
+This method wraps `Lean.MVarId.assign`, checking whether `mvarId` is already assigned, and whether
+the expression has the right type. -/
+def Lean.MVarId.myExact (g : MVarId) (e : Expr) : MetaM Unit := do
```
hargoniX, line 16: `myExact` sounds more like a tutorial tactic to me that we would e.g. write in the meta programming book as a show case, is there a reason you didn't go for just `exact`?

Std/Tactic/Exact.lean
hrmacbeth: Mario wrote this but asked me to PR it!

Std/Tactic/Exact.lean
hrmacbeth: And indeed the name is a placeholder, Mario wants to change it but didn't have an idea yet.

kmill: How about `Lean.MVarId.assign'` so that it pops up in the autocomplete for `assign`? Or `Lean.MVarId.assignIfDefeq`?

By the way, the first line of the definition could be `guard <| ← isDefEq (← g.getType) (← inferType e)`.

Std/Tactic/Exact.lean
```diff
@@ -0,0 +1,19 @@
...
+  let .true ← isDefEq (← g.getType) (← inferType e) | failure
```
hrmacbeth, line 17:
```suggestion
  guard <| ← isDefEq (← g.getType) (← inferType e)
```

Std/Tactic/Exact.lean
hrmacbeth:
```suggestion
def Lean.MVarId.assignIfDefeq (g : MVarId) (e : Expr) : MetaM Unit := do
```

Std/Tactic/Exact.lean
```diff
@@ -0,0 +1,19 @@
...
+
+open Lean Meta
+
+/--
+`MetaM` version of `Lean.Elab.Tactic.evalExact`: add `mvarId := x` to the metavariable assignment.
+This method wraps `Lean.MVarId.assign`, checking whether `mvarId` is already assigned, and whether
+the expression has the right type. -/
+def Lean.MVarId.assignIfDefeq (g : MVarId) (e : Expr) : MetaM Unit := do
+  guard <| ← isDefEq (← g.getType) (← inferType e)
+  g.checkNotAssigned `myExact
```
digama0, line 18:
```suggestion
  g.checkNotAssigned `assignIfDefeq
```

PR #141, opened 2023-05-14
fix: add `local` modifier to `simp` attribute

chabulhwi:
Restrict the scope of the `simp` attribute to `Std.Data.String.Lemmas`.

### DIFF
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -243,7 +243,7 @@ end Substring
 
 namespace String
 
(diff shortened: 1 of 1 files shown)


chabulhwi:
This PR will debug the errors raised by the linter in https://github.com/leanprover-community/mathlib4/pull/3966.

digama0:
as the comment says, this is deliberate. You should change the mathlib simp lemma as suggested by the linter, and remove or un-simp the `(singleton c).toList` lemma.

digama0:
(Arguably it would be better if `String.data` didn't exist in the first place, and the projection was called `toList`, but this requires changes in core.)

PR #143, opened 2023-05-22
feat: add a by? macro

alexjbest:
Just a cute little feature for quickly replacing tactic proofs with proof terms

also a few other unrelated doc tweaks

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -1125,7 +1125,7 @@ instance nodupDecidable [DecidableEq α] : ∀ l : List α, Decidable (Nodup l) 
 /-- `eraseDup l` removes duplicates from `l` (taking only the first occurrence).
 Defined as `pwFilter (≠)`.
 
-    `eraseDup [1, 0, 2, 2, 1] = [0, 2, 1]` -/
+    eraseDup [1, 0, 2, 2, 1] = [0, 2, 1] -/
 @[inline] def eraseDup [DecidableEq α] : List α → List α := pwFilter (· ≠ ·)
 
 /-- `range' start len step` is the list of numbers `[start, start+step, ..., start+(len-1)*step]`.
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -1123,7 +1123,7 @@ instance nodupDecidable [DecidableEq α] : ∀ l : List α, Decidable (Nodup l) 
   instDecidablePairwise
 
 /-- `eraseDup l` removes duplicates from `l` (taking only the first occurrence).
-  Defined as `pwFilter (≠)`.
+Defined as `pwFilter (≠)`.
 
     `eraseDup [1, 0, 2, 2, 1] = [0, 2, 1]` -/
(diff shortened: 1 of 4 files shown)


Std/Data/List/Basic.lean
```diff
@@ -1125,7 +1125,7 @@ instance nodupDecidable [DecidableEq α] : ∀ l : List α, Decidable (Nodup l)
 /-- `eraseDup l` removes duplicates from `l` (taking only the first occurrence).
   Defined as `pwFilter (≠)`.
 
-    eraseDup [1, 0, 2, 2, 1] = [0, 2, 1] -/
+    `eraseDup [1, 0, 2, 2, 1] = [0, 2, 1]` -/
```
digama0, line 1128: this is supposed to be an indented code block, it should not have backticks around it (or it should use triple backtick before and after)

Std/Tactic/ShowTerm.lean
```diff
@@ -30,3 +30,8 @@ elab (name := showTerm) tk:"show_term " t:term : term <= ty => do
   Term.synthesizeSyntheticMVarsNoPostponing
   addTermSuggestion tk (← instantiateMVars e).headBeta (origSpan? := ← getRef)
   pure e
+
+syntax (name := by?) "by?" tacticSeq : term
+macro_rules
+  | `(term| by? $t) => `(show_term by $t)
```
digama0, line 36:
```suggestion
  | `(term| by?%$tk $t) => `(show_term%$tk by%$tk $t)
```
double check that the message and the errors from `by` both end up on the `by?` token

Std/Data/List/Basic.lean
alexjbest: Ok in the docs it appears to have the wrong font for me (and no links / coloured background). Is this a docgen bug or are you just saying i should triple backtick it?

digama0: I think this doc was copied from lean 3, it is possible docgen4 doesn't handle indent code blocks. Assuming adding more indentation doesn't fix it, you should just triple backtick it (but don't put the `-/` on the same line as the end backticks, this triggers a comment bug in vscode).

digama0: just took a look, it appears the issue is the indentation of the `Defined as` line, that should be flush left or else lean will pull 2 spaces off the whole doc comment and break the indent code block

Std/Tactic/ShowTerm.lean
alexjbest: Seems to work fine (it seemed to also work ok before tbh, but I appreciate that this is probably an improvement)

digama0: keep in mind that you have to test with `by?` on the same line as `tac`, or else you might get a misleading impression of the end of the message span

Std/Data/List/Basic.lean
digama0: the backticks are still there

Std/Tactic/ShowTerm.lean
alexjbest: Yeah I did check a couple of different things like that, multiline, nested by's, syntax errors later and everything seemed to behave as by did

PR #144, opened 2023-05-24
feat: make linter output errors using errorformat

alexjbest:
This prints linter errors in the following form:
```
./build/bin/runLinter
-- Found 1 error in 4260 declarations (plus 16946 automatically generated ones) in Std with 12 linters

/- The `synTaut` linter reports:
THE FOLLOWING DECLARATIONS ARE SYNTACTIC TAUTOLOGIES. This usually means that they are of the form `∀ a b ... z, e₁ = e₂` where `e₁` and `e₂` are identical expressions. We call declarations of this form syntactic tautologies. Such lemmas are (mostly) useless and sometimes introduced unintentionally when proving basic facts using `rfl`, when elaboration results in a different term than the user intended. You should check that the declaration really says what you think it does. -/
-- Std.Tactic.Lint.Frontend
./Std/Tactic/Lint/Frontend.lean:137:1: error: Std.Tactic.Lint.h LHS equals RHS syntactically
make: *** [lint] Error 1
```
the advantage of this is that errors printed in this form are recognised by a number of tools (vscode ctrl+click from terminal, vim's default quickfix errorformat, and can be matched by github actions problem matchers https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md).

We will use this to implement github annotations easily for mathlib (this will be a 3 line change in the workflow), this should make it easier to tell at a glance what is wrong with a given PR from the changes tab (which should speed up porting).

Currently this PR turns the option useErrorFormat on by default for std also for the `runLinter` program, with the logic that people running runLinter from command line are more likely to find this useful than #check lines producing a valid lean file.

This setting when activated will always group linter errors by file, as this lead to a clean implementation and seemed more useful.

### DIFF
--- a/Std/Tactic/Lint/Frontend.lean
+++ b/Std/Tactic/Lint/Frontend.lean
@@ -3,6 +3,7 @@ Copyright (c) 2020 Floris van Doorn. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Floris van Doorn, Robert Y. Lewis, Gabriel Ebner
 -/
+import Lean.Util.Paths
 import Std.Tactic.Lint.Basic
 
 /-!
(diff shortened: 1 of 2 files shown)


scripts/runLinter.lean
```diff
@@ -63,7 +63,7 @@ unsafe def main (args : List String) : IO Unit := do
       let failed := results.any (!·.2.isEmpty)
       if failed then
         let fmtResults ←
-          formatLinterResults results decls (groupByFilename := true)
+          formatLinterResults results decls (groupByFilename := true) (useErrorFormat := True)
```
digama0, line 66:
```suggestion
          formatLinterResults results decls (groupByFilename := true) (useErrorFormat := true)
```

Std/Tactic/Lint/Frontend.lean
```diff
@@ -113,19 +113,28 @@ def sortResults (results : HashMap Name α) : CoreM <| Array (Name × α) := do
...
 
 /-- Formats a map of linter warnings using `print_warning`, sorted by line number. -/
-def printWarnings (results : HashMap Name MessageData) : CoreM MessageData := do
+def printWarnings (results : HashMap Name MessageData) (filepath : System.FilePath := default)
+  (useErrorFormat : Bool := False) : CoreM MessageData := do
   (MessageData.joinSep ·.toList Format.line) <$>
-    (← sortResults results).mapM fun (declName, warning) => printWarning declName warning
+    (← sortResults results).mapM
+      fun (declName, warning) => printWarning declName warning (useErrorFormat := useErrorFormat)
+        (filepath := filepath)
```
digama0, line 130:
```suggestion
    (← sortResults results).mapM fun (declName, warning) =>
      printWarning declName warning (useErrorFormat := useErrorFormat) (filepath := filepath)
```

Std/Tactic/Lint/Frontend.lean
```diff
@@ -134,23 +143,24 @@ def groupedByFilename (results : HashMap Name MessageData) : CoreM MessageData :
   let grouped' := grouped.toArray.qsort fun (a, _) (b, _) => toString a < toString b
   (MessageData.joinSep · (Format.line ++ Format.line)) <$>
     grouped'.toList.mapM fun (mod, msgs) => do
-      pure m!"-- {mod}\n{← printWarnings msgs}"
+      let fp := modToFilePath ⟨"."⟩ mod "lean"
```
digama0, line 146: This is definitely wrong when the executable isn't being run from the project root, or when another package is being linted. It needs to use `getPackageDir`

alexjbest:
Thanks @digama0, I tried to follow your suggestion but couldn't quite get it to work yet, so it's nice that you pushed the right fix

digama0:
Note that now it is best to run `runLinter` via `lake exe runLinter` so that it will get the right search path, although I tried to give it a reasonable fallback using just `.` if you run `build/bin/runLinter` directly.

PR #145, opened 2023-05-24
feat: add explicit and implicit getAppArgs functions

mcdoll:


digama0:
I question the utility of these functions. What are they being used for?

mcdoll:
I use `.getExplicitArgs` in `congrm` to match the pattern. I have no strong feelings about these, I just they might come in handy.

digama0:
I'm going to close this, we can revisit if another use case arises. (Elaborating applications is complicated, this is at best an approximation.)

PR #146, opened 2023-05-24
feat: add function for refl and apply

mcdoll:


Std/Lean/Meta/Basic.lean
```diff
@@ -274,6 +274,20 @@ where
           modify (·.insert pendingMVarId)
         go pendingMVarId
 
+/-- Try to close the goal using `refl`. If `refl` fails return the `mvarId` unchanged and
+return `none` otherwise.
+
+This formulation is especially useful to `filterMapM` a list of goals. -/
+def notRefl? (mvarId : MVarId) : MetaM (Option MVarId) := do
```
digama0, line 281: I think this one would be better off as `MetaM Bool` reporting success or not, since it just returns the input unchanged on failure, but at that point it's not much better than just open-coding the whole thing. I would like to see the use case for this definition.

Also the name `notRefl?` is pretty bad, `tryRefl?` or even `refl?` is better.

mcdoll: My use-case is that I want to close all goals that can be proved by `refl`. In my case I just want to filter through a list of metavariables, but one could also think of adding an option for Kyle's `congr!` so that it only uses `refl` but not `assumption`, see https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Tactic/Congr!.lean#L818
I suspect this to be more convenient in the negated form, that is if `refl` succeeds then return `false` so that you can `filterM` a list (or `filterMapM` in the current version). But you have way more experience in this stuff, so if you prefer that reports success then that is fine with me as well.

kmill: I think the bool-valued one that returns whether it succeeds is the one that would fit in the best with the rest of the Lean library.
```
def refl? (mvarId : MVarId) : MetaM Bool := do
  try mvarId.refl; return true catch _ => return false
```
I'm not sure it's necessary to use `observing?` unless `isDefEq` has any side effects when it fails.

For your use case, it's not so bad to negate the result to filter out when it succeeds:
```
def mapRefl (mvarIds : List MVarId) : MetaM (List MVarId) :=
  mvarIds.filterM (not <$> refl? ·)
```

semorrison: Don't have a strong opinion about the fate of this function, but noting the related `Lean.MVarId.rfl : MVarId \to MetaM Unit` in [`Mathlib.Tactic.Relation.Rfl`](https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Tactic/Relation/Rfl.lean#L49).

Std/Lean/Meta/Basic.lean
```diff
@@ -274,6 +274,15 @@ where
           modify (·.insert pendingMVarId)
         go pendingMVarId
 
+/-- Try to close the goal using `refl`. -/
+def refl? (mvarId : MVarId) : MetaM Bool := do
+  try mvarId.refl; return true catch _ => return false
+
+/-- Apply a lemma with fresh metavariable levels. -/
+def applyWithFreshMVarLevels (mvarId : MVarId) (lemmaName : Name) :
```
semorrison, line 282: Should support passing an `ApplyConfig`.

kmill:
The apply function is in core now as `Lean.MVarId.applyConst`, but I don't think `refl?` exists in core.

Std/Lean/Meta/Basic.lean
```diff
@@ -274,6 +274,15 @@ where
...
+    MetaM (List MVarId) := mvarId.withContext do
+  mvarId.apply (← mkConstWithFreshMVarLevels lemmaName)
```
fgdorais, line 284: Now in Lean core as `Lean.MVarId.applyConst`
```suggestion
```

fgdorais:

PR #147, opened 2023-05-25
feat: `#where` and `#guard_msgs` commands

kmill:
Adds a mathlib3-style `#where` command that pretty prints the current top-level scope, including the namespace, any `open` namespaces, the universe levels, the `variable` command, and any options with non-default values.

To test this, also adds a `#guard_msgs` command that checks the messages produced by another command.


digama0:
Note for the future: I tried out this implementation which describes more precisely the interleaving of `section`, `section Foo` and `namespace`:
```lean
inductive NamespaceKind
  | ns : Name → NamespaceKind
  | sect : (nc : Bool) → Name → NamespaceKind
  deriving Repr

def describeNamespaces (n : Nat) (nc : Bool) : List Scope → List NamespaceKind → List NamespaceKind
  | [], nks => dbg_trace repr nks.reverse; nks.reverse
  | sc::rest, nks =>
    let n' := sc.currNamespace.getNumParts
    dbg_trace (n, n')
    let nc' := sc.isNoncomputable
    if n == n' then
      let ncSelf := !nc && nc'
      describeNamespaces n' nc' rest <|
        match nks with
        | .sect nc2 name :: rest =>
          if name == `«» || sc.header == `«» || ncSelf then
            .sect ncSelf sc.header :: nks
          else
            .sect nc2 (name.str sc.header) :: rest
        | _ => .sect ncSelf sc.header :: nks
    else
      describeNamespaces n' nc' rest <|
        match nks with
        | .ns name :: rest => NamespaceKind.ns (name.str sc.header) :: rest
        | _ => .ns sc.header :: nks
        
-- in the #where impl
  let mut nsMsg := #[]
  for nk in describeNamespaces 0 false (← getScopes).reverse.tail! [] do
    match nk with
    | .ns name =>
      nsMsg := nsMsg.push m!"namespace {name}"
    | .sect nc name =>
      let nc := if nc then "noncomputable " else ""
      nsMsg := nsMsg.push <| if name == `«» then m!"{nc}section" else m!"{nc}section {name}"
  if !nsMsg.isEmpty then
    msg := msg.push m!"{MessageData.joinSep nsMsg.toList "\n"}"
```

This describes the scoping precisely enough that you can figure out what is legal to type after `end` at this point, as well as how many scopes you have to close to make things computable and what namespace you would be in after closing N scopes.

I decided against it though, because it's somewhat involved, and I felt that it was not summarized enough, e.g. it is not immediately obvious what namespace you are actually in without reading this whole header. Instead, I think the command should focus on the state in *this* scope and not the ones above it. To that end I pushed a modification that simplifies the output a bit to focus on that information.

digama0:
I think I'm addicted to code actions now. I pushed a code action which replaces the `#guard_msgs` doc string with the command's actual output.

kmill:
Cool, having a code action crossed my mind, but I wasn't sure how to write one.

PR #148, opened 2023-05-29
feat: `#instances` command to find applicable instances

kmill:
This adds a command similar to `#synth` that finds all applicable instances. For example, `#instances Add _` looks for all instances whose conclusion unifies with `Add _`. Trailing underscores can be omitted, so `#instances Add` works as well. Another example is `#instances OfNat _ (nat_lit 0)`, which shows all instances of `OfNat` for `0`.

### DIFF
--- a/Std/Tactic/Instances.lean
+++ b/Std/Tactic/Instances.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Kyle Miller
 -/
 import Lean.Elab.Command
+import Lean.PrettyPrinter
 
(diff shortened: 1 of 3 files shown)


Std/Tactic/Instances.lean
```diff
@@ -0,0 +1,47 @@
+/-
+Copyright (c) 2023 Kyle Miller. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kyle Miller
+-/
+import Lean
```
digama0, line 6: std uses precise imports FYI

PR #152, opened 2023-06-01
feat: function to find inverse for extended binders

kmill:
Adds `getBinderPredOfProp` to query for inverse of `satisfies_binder_pred%` and extends `binder_predicate` notation to support this.

This can be applied to creating pretty printers that make use of extended binders.

### DIFF
--- a/Std/Util/ExtendedBinder.lean
+++ b/Std/Util/ExtendedBinder.lean
@@ -24,7 +24,33 @@ declare_syntax_cat binderPred
 -/
 syntax "satisfies_binder_pred% " term:max binderPred : term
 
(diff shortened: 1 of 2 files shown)


kmill:
I think `extBinders` needs to be changed in a deep way to make it be more consistent with `fun` binder syntax, so the inverse-finding of this PR is not relevant anymore. There are still improvements to the `binder_predicate` command that might be worth merging -- I will probably open a new PR for that, but I'll keep this open for now so I don't lose track of these changes.

joehendrix:
Based on the last comment @kmill, I am migrating this to awaiting-author.

fgdorais:
Convert to draft?

PR #154, opened 2023-06-16
Basic pow lemmas for Nat

fgdorais:
The corresponding lemmas in Mathlib are based on a monoid class (Mathlib/Algebra/GroupPower/Basic.lean) so they won't be down-ported to Std. Core has very limited need for powers, so they are unlikely to appear there.

### DIFF
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -230,6 +230,9 @@ theorem sub_eq_sub_min (n m : Nat) : n - m = n - min n m := by
 protected theorem mul_right_comm (n m k : Nat) : n * m * k = n * k * m := by
   rw [Nat.mul_assoc, Nat.mul_comm m, ← Nat.mul_assoc]
 
+protected theorem mul_cross_comm (a b c d : Nat) : (a * b) * (c * d) = (a * c) * (b * d) := by
+  rw [Nat.mul_assoc, Nat.mul_assoc, Nat.mul_left_comm b]
+
 protected theorem mul_two (n : Nat) : n * 2 = n + n := by simp [Nat.mul_succ]
 
 protected theorem two_mul (n : Nat) : 2 * n = n + n := by simp [Nat.succ_mul]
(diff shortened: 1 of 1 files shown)


Std/Data/Nat/Lemmas.lean
```diff
@@ -230,6 +230,9 @@ theorem sub_eq_sub_min (n m : Nat) : n - m = n - min n m := by
 protected theorem mul_right_comm (n m k : Nat) : n * m * k = n * k * m := by
   rw [Nat.mul_assoc, Nat.mul_comm m, ← Nat.mul_assoc]
 
+protected theorem mul_cross_comm (a b c d : Nat) : (a * b) * (c * d) = (a * c) * (b * d) := by
```
digama0, line 233: Is this what the theorem is called in mathlib? I would have guessed it is something more like `mul_mul_mul_comm`.

Std/Data/Nat/Lemmas.lean
```diff
@@ -791,6 +794,8 @@ theorem mul_mod (a b n : Nat) : a * b % n = (a % n) * (b % n) % n := by
 theorem add_mod (a b n : Nat) : (a + b) % n = ((a % n) + (b % n)) % n := by
   rw [add_mod_mod, mod_add_mod]
 
+/- pow -/
```
digama0, line 797:
```suggestion
/-! ### pow -/
```

Std/Data/Nat/Lemmas.lean
fgdorais: You're right, it is called `mul_mul_mul_comm`.

digama0:
oops, forgot to fix the commit message (should have a `feat:` in front). Oh well.

fgdorais:
That's funny, I'm pretty sure my first commit had `feat:`. I must have accidentally edited it out.

digama0:
It does have `feat:`, but the PR title doesn't and the name of the squash merge comes from the PR title

PR #155, opened 2023-06-17
feat: additional config for `congr`

Komyyy:
The `MetaM` version of `congr` (`Lean.MVarId.congrN`) has 2 configs: `closePre` and `closePost`.
Especially, `closePost` is significant to prevent timeout.
This file adds a config for the `congr` & `rcongr` tactic.


digama0:
Can you add a test to `test/congr.lean`?

PR #156, opened 2023-06-17
feat: use whnf in ext, more like Lean3 behaviour

semorrison:
We restore Lean3-like behaviour to ext, by keeping track of two discrimation trees. The first is the current behaviour of `ext`, while the second indexes using `whnf`.

See https://github.com/leanprover-community/mathlib4/pull/5191 for examples of how this works out in mathlib4. (That PR is far from exhaustive.)

Obviously this will be slower, but I hope it's considered worth it for the simplifications available in mathlib4. Note this is still much faster than the mathlib3 behaviour (where there wasn't any discrimation tree at all, just a single big bucket of lemmas). `apply` is fast.

### DIFF
--- a/Std/Tactic/Ext.lean
+++ b/Std/Tactic/Ext.lean
@@ -204,3 +204,5 @@ attribute [ext] funext propext
 
 @[ext] protected theorem PUnit.ext (x y : PUnit) : x = y := rfl
 protected theorem Unit.ext (x y : Unit) : x = y := rfl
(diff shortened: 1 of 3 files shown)


semorrison:
Per discussion on [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/power-up.20for.20ext), I don't want to pursue this for now.

PR #159, opened 2023-06-18
chore: fix bug in ext when introducing multiple patterns per lemma

semorrison:


test/ext.lean
```diff
@@ -75,3 +75,16 @@ example (f : ℕ × (ℕ → ℕ)) : f = f := by
...
+
+example : 3 = 7 := by
+  ext : 1
+  rename_i n m
+  guard_target = n = m
+  admit
+
+example : 3 = 7 := by
+  ext n m : 1
+  guard_target = n = m
```
semorrison, line 89: Before this change, this step failed, as the goal was `\forall m, n = m`.

eric-wieser:
A [related Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/ext.20tactic.20intros.20too.20few.20variables/near/342651528)

PR #160, opened 2023-06-20
feat: `case _ : t => tac` tactic to search for goals by type

kmill:
This supplements the built-in `case` tactic, which searches for a goal by tag, with a `case _ : t => tac` that searches for a goal having type `t` by unification. The syntax it provides is a strict superset of the core `case` tactic, and more specifically it gives the following three extensions: rather than just `tag n1 ... n2` for searching by tag and renaming inaccessibles, it provides `tag n1 ... n2 : t` for also filtering by unification, it allows `case ... := e` for `case ... => exact e`, and it also allows `case ...` with no body for lifting goals, making it act like Lean 3 `show`.

This also extends the `case'` tactic in a similar way, but it does not provide `case' ... := e` or `case' ...` since these are both handled by `case`.


digama0:
> I'm not sure I follow what you're saying about `_`/`?_` in place of a tactic block. I added a no-tactic-block version for bringing goals up to being the first goal(s), but if this isn't doing what you're suggesting then let's move this discussion over to that PR.

The syntax for `induction` takes not `tacticSeq` but rather `tacticSeq <|> hole <|> syntheticHole`, and the latter cases are treated as deferring the goal until after the tactic (i.e. not nested under the tactic itself).

kmill:
Since this tactic only deals with a single tactic block, I'm not sure how useful it is to allow `hole` and `syntheticHole`, rather than just having a no-`=>` version, but I added it.

You can get into some weird shenanigans with named holes, like re-using previously assigned named metavariables (though it seems better to do `=> exact ?m` when you mean it rather than `=> ?m`).

Std/Tactic/Case.lean
```diff
@@ -0,0 +1,148 @@
...
+  `case _ : t₁ | _ : t₂ | ...` makes the matched goals be the first goals in the given order.
+
+* `case _ : t := _` and `case _ : t := ?m` are the same as `case _ : t` but in the `?m` case the
+  goal tag is changed to `m`.
+  In particular, the goal becomes metavariable `?m`.
+-/
+-- Low priority so that type-free `case` doesn't conflict with core `case`,
+-- though it should be a drop-in replacement.
+syntax (name := casePatt) (priority := low)
+  "case " sepBy1(casePattArg, " | ") (casePattBody)? : tactic
```
digama0, line 54:
```suggestion
  "case " sepBy1(casePattArg, " | ") (casePattTac <|> casePattExpr)? : tactic
```
where `casePattTac := " => " (hole <|> syntheticHole <|> tacticSeq)` and `casePattExpr := " := " colGt t:term`. I think this will behave better wrt the backtracking behavior of the parser compared to having two tactics with a long common prefix.

test/case.lean
```diff
@@ -0,0 +1,126 @@
...
+    guard_target =ₛ z = w
+    exact h'
+  case left =>
+    exact h
+
+example (x y z : α) (h : x = y) (h' : y = z) : x = z := by
+  apply Eq.trans
+  case _ : α := y
+  -- Note: `case` inserts a `let_fun` due to the fact it's implemented with `show`
+  · guard_target =ₛ x = let_fun this := y; this
```
digama0, line 63: This seems unfortunate, especially in the example given here where you wanted to fill the third goal specifically by `y`. Maybe we should just not introduce a `let_fun` if the types are reducibly defeq after unification?

kmill: Yeah, it was unfortunate. Your idea seems reasonable, and I added a defeq detector.

Std/Tactic/Case.lean
```diff
@@ -0,0 +1,191 @@
...
+
+Returns the found goal, goals caused by elaborating `patt`, and the remaining goals. -/
+def findGoalOfPatt (gs : List MVarId)
+    (tag : TSyntax ``binderIdent) (patt? : Option Term) (renameI : TSyntaxArray `Lean.binderIdent) :
+    TacticM (MVarId × List MVarId × List MVarId) :=
+  Term.withoutErrToSorry do
+    let gs ←
+      match tag with
+      | `($tag:ident) => filterTag gs tag.getId
+      | _ => pure gs
```
kmill, line 107: Oops, this `let` shouldn't shadow `gs` since we need to return the remaining goals. I'll review this tomorrow.

PR #162, opened 2023-06-21
fix: remove `simp` attribute from getElem_fin/getElem?_fin/getElem!_fin

alexkeizer:
Having these as simp lemmas is making it more difficult to use the `getElem` notation as the simp normal form of indexing a `Vector` with a `Fin n`.


alexkeizer:
Closing this, since the consensus on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/simp.20normal.20form.20for.20.60Vector.60.20indexing.3A.20.60get.60.20vs.20.60.2E.5B.2E.5D.60) seems that we don't want to rely on `getElem` for `Vector`s simp-normal-form

PR #163, opened 2023-06-23
chore: bump to nightly-2023-06-22

semorrison:


Std/Data/Rat/Lemmas.lean
```diff
@@ -5,6 +5,11 @@ import Std.Data.Rat.Basic
 
 namespace Rat
 
+@[simp] theorem zero_num : (0 : Rat).num = 0 := rfl
+@[simp] theorem zero_den : (0 : Rat).den = 1 := rfl
```
semorrison, line 9: This one is not actually needed below but is a natural member of this family.

digama0: Yep, leo mentioned that we would need
```lean
@[simp] theorem zero_num : (0 : Rat).num = 0 := rfl
@[simp] theorem one_num : (1 : Rat).num = 1 := rfl
@[simp] theorem one_den : (1 : Rat).den = 1 := rfl
```
in the next bump

PR #168, opened 2023-06-28
feat: DArray.{strongSet, *[mM]odifyM*}

JamesGallicchio:


Std/Data/Function/Basic.lean
```diff
@@ -0,0 +1,124 @@
...
+             (λ _ : a = a' =>  b) := funext h2
+   let f := λ x => dite (a = a') x (λ (_: ¬ a = a') => (f a))
+   exact congrArg f h3
+
+@[simp]
+theorem update_same (a : α) (v : β a) (f : ∀ a, β a) : update f a v a = v :=
+  dif_pos rfl
+
+@[simp]
+theorem update_noteq {a a' : α} (h : a ≠ a') (v : β a') (f : ∀ a, β a) : update f a' v a = f a :=
```
fgdorais, line 37: `update_ne` would be just as clear. If there is a conflict I'm not seeing, consider `update_other`.

Std/Data/Function/Basic.lean
```diff
@@ -0,0 +1,124 @@
...
+    update f a' b a = if a = a' then b else f a :=
+by have h2 : (h : a = a') → Eq.rec (motive := λ _ _ => β) b h.symm = b :=
+     by intro h; cases h; simp
+   have h3 : (λ h : a = a' => Eq.rec (motive := λ _ _ => β) b h.symm) =
+             (λ _ : a = a' =>  b) := funext h2
+   let f := λ x => dite (a = a') x (λ (_: ¬ a = a') => (f a))
+   exact congrArg f h3
+
+@[simp]
+theorem update_same (a : α) (v : β a) (f : ∀ a, β a) : update f a v a = v :=
```
fgdorais, line 33: I would use `update_self`? `update_same` sounds more like:
```
theorem update_eq (a a' : α) (v : β a') (f : ∀ a, β a) (heq : a = a') : update f a' v a = heq ▸ v := sorry
```

Std/Data/Function/Basic.lean
JamesGallicchio: These were the names in Mathlib. I agree that `update_ne` and `update_eq` are more consistent with other functions with similar statements, so we should change it IMO.

PR #172, opened 2023-06-30
feat: S[ubs]tring.{posOf,contains}S[ubs]tr

JamesGallicchio:
Adds functions for finding or checking existence of strings in strings and substrings in substrings and those things in between.

Worst-case `O(nm)` runtime on all new functions, where `n = |s|, m = |pattern|`.

### DIFF
--- a/Std/Data/String/Basic.lean
+++ b/Std/Data/String/Basic.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2022 Jannis Limperg. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Jannis Limperg
+Authors: Jannis Limperg, James Gallicchio
 -/
 
 import Std.Data.Char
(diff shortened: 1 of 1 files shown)


fgdorais:
#178 now incorporates this PR with worst case O(n+m) runtime on all new functions, where n = |s|, `m = |pattern|.

fgdorais:
I'm marking this as WIP since it's incorporated into #178. (At least partly, I think I modified some bits.) If you want to close it, that's fine too. Remember that you can always comment on #178 if the edits I made are inappropriate for your needs.

semorrison:
I think it's fine to just close. @JamesGallicchio is probably happy with the outcome of a faster implementation, and can take the credit for nerd-snipping @fgdorais into making the #178 PR. :-) Of course can be re-opened etc if I've missed something.

JamesGallicchio:
At first I was thinking the naive implementation should be kept around as the reference implementation against which we verify correctness, but then I decided that I do not care about the functional correctness of string matching :-)

digama0:
the naive implementation isn't naive enough to be usable as a model for functional correctness. I would prefer `<:+:` on the underlying lists for the model.

PR #118, opened 2023-04-18
feat: add Nat.mul_lt_mul₀

sorawee:
This is a Nat variant of
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Order/WithZero.html#mul_lt_mul%E2%82%80

PR #177, opened 2023-07-04
feat: cleanup some Nat lemmas

fgdorais:
Depends on #182 

Potentially breaking changes:

* Changed `Nat.mul_lt_mul` (1 use in Mathlib). See comments below.
* Made two parameters implicit in `Nat.lt_add_right` (4 uses in Mathlib)
* Made parameters implicit for `Nat.add_le_add_iff_left` (no uses in Mathlib)
* Made parameters implicit for `Nat.add_le_add_iff_right` (no uses in Mathlib)
* Made parameters implicit for `Nat.add_lt_add_iff_left` (no uses in Mathlib)
* Made parameters implicit for `Nat.add_lt_add_iff_right` (no uses in Mathlib)
* Made parameters implicit for `Nat.dvd_iff_mod_eq_zero` (16 uses in Mathlib)
* Renamed `Nat.max_le` (no uses in Mathlib) to `Nat.max_le_iff` and added new `Nat.max_le` (Mathlib conformance)
* Renamed `Nat.le_min` (no uses in Mathlib)  to `Nat.le_min_iff`  and added new `Nat.le_min` (Mathlib conformance)
* Protected `Nat.zero_le` (unknown uses in Mathlib; at least one hiding clause)
* Protected `Nat.one_pos` (unknown uses in Mathlib)
* Protected `Nat.two_pos` (no uses in Mathlib)
* Deprecated `Nat.lt_min` (no uses in Mathlib)
* Deprecated `Nat.add_le_to_le_sub` (no uses in Mathlib)
* Deprecated `Nat.succ_add_eq_succ_add` (4 uses in Mathlib)
* Deprecated `Nat.mul_le_mul_of_nonneg_left` (4 uses in Mathlib)
* Deprecated `Nat.mul_le_mul_of_nonneg_right` (3 uses in Mathlib)
* Deprecated `Nat.mul_lt_mul'` (2 uses in Mathlib)
* Deprecated `Nat.zero_max` (1 use in Mathlib)
* Deprecated `Nat.zero_min` (no uses in Mathlib)
* Deprecated `Nat.max_zero` (no uses in Mathlib)
* Deprecated `Nat.min_zero` (no uses in Mathlib)

Deprecated lemmas all have recommended replacements.

### DIFF
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -106,33 +106,28 @@ theorem lt_succ : m < succ n ↔ m ≤ n :=
 theorem lt_succ_of_lt (h : a < b) : a < succ b := le_succ_of_le h
 
 theorem succ_ne_self : ∀ n, succ n ≠ n
-  | 0,   h => absurd h (succ_ne_zero 0)
-  | n+1, h => succ_ne_self n (Nat.noConfusion h id)
+  | _+1, h => succ_ne_self _ (succ.inj h)
 
 theorem succ_pred_eq_of_pos : ∀ {n}, 0 < n → succ (pred n) = n
-  | succ _, _ => rfl
+  | _+1, _ => rfl
 
-theorem eq_zero_or_eq_succ_pred (n) : n = 0 ∨ n = succ (pred n) := by
-  cases n <;> simp
+theorem eq_zero_or_eq_succ_pred : ∀ n, n = 0 ∨ n = succ (pred n)
+  | 0 => .inl rfl
+  | _+1 => .inr rfl
 
-theorem exists_eq_succ_of_ne_zero (H : n ≠ 0) : ∃ k, n = succ k :=
-  ⟨_, (eq_zero_or_eq_succ_pred _).resolve_left H⟩
+theorem exists_eq_succ_of_ne_zero : ∀ {n}, n ≠ 0 → ∃ k, n = succ k
+  | _+1, _ => ⟨_, rfl⟩
 
-theorem succ_eq_one_add (n) : succ n = 1 + n := by
-  rw [Nat.succ_eq_add_one, Nat.add_comm]
+theorem succ_eq_one_add (n) : succ n = 1 + n := Nat.add_comm _ 1
 
 theorem succ_inj' : succ n = succ m ↔ n = m :=
   ⟨succ.inj, congrArg _⟩
 
 theorem pred_inj : ∀ {a b}, 0 < a → 0 < b → pred a = pred b → a = b
-| a+1, b+1, _,  _, h => by rw [show a = b from h]
-| a+1, 0,   _, hb, _ => absurd hb (Nat.lt_irrefl _)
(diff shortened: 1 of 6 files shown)


Std/Data/Nat/Lemmas.lean
```diff
@@ -99,63 +100,55 @@ theorem succ_le {n m : Nat} : succ n ≤ m ↔ n < m := .rfl
...
-theorem exists_eq_succ_of_ne_zero {n : Nat} (H : n ≠ 0) : ∃ k, n = succ k :=
-  ⟨_, (eq_zero_or_eq_succ_pred _).resolve_left H⟩
+theorem exists_eq_succ_of_ne_zero : ∀ {n : Nat}, n ≠ 0 → ∃ k, n = succ k
+  | _+1, _ => ⟨_, rfl⟩
 
 theorem succ_eq_one_add (n : Nat) : n.succ = 1 + n := by
   rw [Nat.succ_eq_add_one, Nat.add_comm]
 
-theorem succ_inj' : succ n = succ m ↔ n = m :=
+theorem succ_inj' {n m : Nat} : succ n = succ m ↔ n = m :=
```
digama0, line 121: What is the reason for this change? Style-wise, std prefers auto-implicits, although it's fine to include types if you think it is clearer or you need to manipulate argument order. I would prefer to leave them off in most cases here, it's pretty obvious everything is about Nat.

fgdorais: Easy fix!

fgdorais: Should I do the reverse and remove implicit parameters that could be auto-implicit?

digama0: That's fine by me. Sometimes it makes a small difference with order of parameters, but I think we want to normalize to the ordering you get with autoimplicits anyway.

fgdorais:
This is a good stopping point. I did only a few small additions. The main one are some custom recursors in Std.Data.Nat.Basic. I'm happy to hear reviews from anyone who has time.

fgdorais:
Regarding #118: I'm tempted to rename `Nat.mul_lt_mul₀` to just `Nat.mul_lt_mul`. There are some issues:

- On the one hand, Mathlib has `mul_lt_mul₀` in `Mathlib.Algebra.Order.WithZero` so it makes sense to use the same name. On the other hand, like every theorem in that file, the subscript zero indicates `WithZero` which doesn't make much sense as a distinction in `Nat`.
- On the one hand, there is a deprecated `Nat.mul_lt_mul` with one use in Mathlib. But the old `Nat.mul_lt_mul` was a bad theorem with unnecessarily strong hypotheses, so it's a good riddance.

fgdorais:
I finally opted for `Nat.mul_lt_mul` over `Nat.mul_lt_mul₀`. What tipped the scale for me is that the old `Nat.mul_lt_mul` was just a bad theorem. Existing uses can be replaced by `Nat.mul_lt_mul_of_lt_of_le` after adapting the hypotheses.

fgdorais:
Now broken up into smaller pieces.

PR #178, opened 2023-07-07
feat: KMP matcher

fgdorais:
- [x] Depends on #219 

A versatile implementation of the Knuth-Morris-Pratt matching algorithm.

### DIFF
--- a/Std/Data/Array/Match.lean
+++ b/Std/Data/Array/Match.lean
@@ -9,9 +9,13 @@ import Std.Data.String.Basic
 
 namespace Array
 
-/-- Prefix table for the Knuth-Morris-Pratt matching algorithm -/
+/-- Prefix table for the Knuth-Morris-Pratt matching algorithm
+
+  This is an array of the form `t = [(x₀,n₀), (x₁,n₁), (x₂, n₂), ...]` where for each `i`, `nᵢ` is
+  the length of the longest proper prefix of `xs = [x₀,x₁,...,xᵢ]` which is also a suffix of `xs`.
+-/
 structure PrefixTable (α : Type _) extends Array (α × Nat) where
-  /-- Validity condition to help termination proofs -/
+  /-- Validity condition to help with termination proofs -/
   valid : (h : i < toArray.size) → toArray[i].2 ≤ i
 
 instance : Inhabited (PrefixTable α) where
@@ -23,7 +27,13 @@ abbrev PrefixTable.size (t : PrefixTable α) := t.toArray.size
 /-- Returns the pattern array of the prefix table -/
 abbrev PrefixTable.pattern (t : PrefixTable α) : Array α := t.toArray.map Prod.fst
 
-/-- Transition function for the KMP matcher -/
+/-- Transition function for the KMP matcher
+
+  Assuming we have an input `xs` with a suffix that matches the pattern prefix `t.pattern[:len]`
+  where `len : Fin (t.size+1)`. Then `xs.push x` has a suffix that matches the pattern prefix
+  `t.pattern[:t.step x len]`. If `len` is as large as possible then `t.step x len` will also be
+  as large as possible.
+-/
 def PrefixTable.step [BEq α] (t : PrefixTable α) (x : α) : Fin (t.size+1) → Fin (t.size+1)
 | ⟨k, hk⟩ =>
   if hx : some x == t.pattern[k]? then
@@ -46,7 +56,10 @@ def PrefixTable.step [BEq α] (t : PrefixTable α) (x : α) : Fin (t.size+1) →
     ⟨0, Nat.zero_lt_succ _⟩
 termination_by _ k => k.val
 
-/-- Extend prefix table by one element -/
+/-- Extend a prefix table by one element
+
+  If `t` is the prefix table for `xs` then `t.extend x` is the prefix table for `xs.push x`.
+-/
 def PrefixTable.extend [BEq α] (t : PrefixTable α) (x : α) : PrefixTable α where
   toArray := t.toArray.push (x, t.step x ⟨t.size, Nat.lt_succ_self _⟩)
   valid := by
@@ -63,7 +76,7 @@ def PrefixTable.extend [BEq α] (t : PrefixTable α) (x : α) : PrefixTable α w
       rw [heq, ← Nat.lt_succ]
       exact Fin.isLt ..
 
-/-- Make prefix table from array -/
+/-- Make prefix table from a pattern array -/
 def mkPrefixTable [BEq α] (xs : Array α) : PrefixTable α :=
   if h : xs.size = 0 then ⟨#[], fun .⟩ else
     have : xs.size-1 < xs.size := Nat.pred_lt h
@@ -72,16 +85,20 @@ termination_by _ xs => xs.size
 
 /-- KMP matcher structure -/
 structure Matcher (α) where
-  /-- prefix table for the pattern -/
+  /-- Prefix table for the pattern -/
   table : PrefixTable α
-  /-- current longest matching prefix -/
+  /-- Current longest matching prefix -/
   state : Fin (table.size+1) := 0
 
-/-- Initialize KMP matcher -/
-def Matcher.init [BEq α] (pattern : Array α) : Matcher α where
-  table := mkPrefixTable pattern
+/-- Initialize KMP matcher for a given pattern -/
+def Matcher.init [BEq α] (pat : Array α) : Matcher α where
+  table := mkPrefixTable pat
 
-/-- Find next match from a given stream -/
+/-- Find next match from a given stream
+
+  Runs the stream until it reads a sequence that mathes the sought pattern, then returns the stream
+  state at that point and an updated matcher state.
+-/
 partial def Matcher.next? [BEq α] [Stream σ α] (m : Matcher α) (stream : σ) :
   Option (σ × Matcher α) :=
   match Stream.next? stream with
@@ -108,6 +125,6 @@ where
     | none => ps
     | some (s, m) => loop m s (ps.push (s.startPos - pattern.endPos))
 
-#eval "abba".findAll "babbababbaabaa"
+#eval "abba".findAll "babbababbabbaabbbaaababba"
 
 end Test
--- a/Std/Data/Array/Match.lean
+++ b/Std/Data/Array/Match.lean
@@ -68,11 +68,7 @@ def PrefixTable.extend [BEq α] (t : PrefixTable α) (x : α) : PrefixTable α w
     · next h => exact Nat.le_trans (Nat.lt_succ.1 <| Fin.isLt ..) (Nat.not_lt.1 h)
 
 /-- Make prefix table from a pattern array -/
-def mkPrefixTable [BEq α] (xs : Array α) : PrefixTable α :=
-  if h : xs.size = 0 then default else
-    have : xs.size-1 < xs.size := Nat.pred_lt h
-    (mkPrefixTable xs.pop).extend xs[xs.size - 1]
-termination_by _ => xs.size
+def mkPrefixTable [BEq α] (xs : Array α) : PrefixTable α := xs.foldl (·.extend) default
 
 /-- Make prefix table from a pattern stream -/
 partial def mkPrefixTableOfStream [BEq α] [Stream σ α] (stream : σ) : PrefixTable α :=
--- a/Std/Data/Array/Match.lean
+++ b/Std/Data/Array/Match.lean
@@ -99,7 +99,7 @@ structure Matcher (α) where
   /-- Current longest matching prefix -/
   state : Fin (table.size+1) := 0
 
-/-- Make a KMP matcher for a given a pattern array -/
+/-- Make a KMP matcher for a given  pattern array -/
 def Matcher.ofArray [BEq α] (pat : Array α) : Matcher α where
   table := mkPrefixTable pat
 
--- a/Std/Data/Array/Match.lean
+++ b/Std/Data/Array/Match.lean
@@ -109,7 +109,7 @@ def Matcher.ofStream [BEq α] [Stream σ α] (pat : σ) : Matcher α where
 
 /-- Find next match from a given stream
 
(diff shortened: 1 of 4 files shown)


fgdorais:
Now integrates #172

fgdorais:
Good idea! Do you think `@[specialize]` might be enough? Or do you absolutely need a reusable matcher type as in the Array API?

fgdorais:
I added a few things that you might find useful. This API would be a lot simpler if there was a coercion from `String` to `Substring`.

nomeata:
Great, thanks!

> This API would be a lot simpler if there was a coercion from `String` to `Substring`.

Could there be one?

Std/Data/String/Basic.lean
```diff
@@ -1,15 +1,63 @@
...
 -/
 
 import Std.Data.Char
 import Std.Data.Nat.Lemmas
+import Std.Data.Array.Match
 
-protected theorem String.Pos.ne_zero_of_lt : {a b : Pos} → a < b → b ≠ 0
+namespace String
+
+instance : Coe String Substring := ⟨toSubstring⟩
```
digama0, line 13: This should be a separate PR, which also removes most uses of `toSubstring` in the rest of the code at the same time

Std/Data/String/Basic.lean
```diff
@@ -1,15 +1,63 @@
...
+  loop s #[]
+where
+  /-- Accumulator loop for `String.Matcher.findAll` -/
+  loop (s : Substring) (occs : Array Substring) : Array Substring :=
+    match m.next? s with
+    | none => occs
+    | some (s, _) =>
+      loop s <| occs.push {s with
+        startPos := ⟨s.startPos.byteIdx - m.patternSize⟩
+        stopPos := s.startPos}
```
digama0, line 48:
```suggestion
      loop s <| occs.push { s with
        startPos := ⟨s.startPos.byteIdx - m.patternSize⟩
        stopPos := s.startPos }
```

Std/Data/String/Basic.lean
```diff
@@ -72,10 +120,41 @@ def dropSuffix? (s : Substring) (suff : Substring) : Option Substring :=
   else
     none
 
+/--
+Returns all the substrings of `s` that martch `pattern`.
```
digama0, line 124:
```suggestion
Returns all the substrings of `s` that match `pattern`.
```

Std/Data/String/Basic.lean
```diff
@@ -1,15 +1,63 @@
...
 
-protected theorem String.Pos.ne_zero_of_lt : {a b : Pos} → a < b → b ≠ 0
+namespace String
+
+instance : Coe String Substring := ⟨toSubstring⟩
+
+protected theorem Pos.ne_zero_of_lt : {a b : Pos} → a < b → b ≠ 0
 | _, _, hlt, rfl => Nat.not_lt_zero _ hlt
 
+/-- KMP Matcher -/
```
digama0, line 18: KMP needs to be explained here (it should be okay on functions as long as the type explains it, but generally you shouldn't assume too much context in docstrings)

digama0: Actually this is probably a good place to put the overall documentation on how to use this type effectively, maybe put a paragraph and a code block showing basic usage.

Std/Data/String/Basic.lean
fgdorais: #219

Std/Data/String/Basic.lean
fgdorais: I found an embarrassing bug so I added a basic test suite as well.

fgdorais: What do you think about the description? Should I add something about how the basic working of the algorithm? (I don't think it's necessary for users to know how it works but this is a neat one.)

digama0: The docs here can just reference the algorithm, possibly including a link to e.g. the wikipedia page. For a more detailed description of the algorithm you should use the module doc on `Data.Array.Match`.

Std/Data/Array/Match.lean
```diff
@@ -0,0 +1,126 @@
...
+abbrev PrefixTable.pattern (t : PrefixTable α) : Array α := t.toArray.map Prod.fst
+
+/-- Transition function for the KMP matcher
+
+  Assuming we have an input `xs` with a suffix that matches the pattern prefix `t.pattern[:len]`
+  where `len : Fin (t.size+1)`. Then `xs.push x` has a suffix that matches the pattern prefix
+  `t.pattern[:t.step x len]`. If `len` is as large as possible then `t.step x len` will also be
+  as large as possible.
+-/
+def PrefixTable.step [BEq α] (t : PrefixTable α) (x : α) : Fin (t.size+1) → Fin (t.size+1)
```
semorrison, line 36: Could `[BEq \a]` be a variable?

Std/Data/Array/Match.lean
```diff
@@ -0,0 +1,126 @@
...
+    | some (x, stream) => loop (t.extend x) stream
+
+/-- KMP matcher structure -/
+structure Matcher (α) where
+  /-- Prefix table for the pattern -/
+  table : PrefixTable α
+  /-- Current longest matching prefix -/
+  state : Fin (table.size+1) := 0
+
+/-- Make a KMP matcher for a given a pattern array -/
```
semorrison, line 102:
```suggestion
/-- Make a KMP matcher for a given  pattern array -/
```

Std/Data/Array/Match.lean
```diff
@@ -0,0 +1,126 @@
...
+def Matcher.ofArray [BEq α] (pat : Array α) : Matcher α where
+  table := mkPrefixTable pat
+
+/-- Make a KMP matcher for a given a pattern stream -/
+def Matcher.ofStream [BEq α] [Stream σ α] (pat : σ) : Matcher α where
+  table := mkPrefixTableOfStream pat
+
+/-- Find next match from a given stream
+
+  Runs the stream until it reads a sequence that mathes the sought pattern, then returns the stream
```
semorrison, line 112:
```suggestion
  Runs the stream until it reads a sequence that matches the sought pattern, then returns the stream
```

Std/Data/Array/Match.lean
fgdorais: I'm not sure what you mean. Just making it an implicit parameter?

Std/Data/String/Basic.lean
```diff
@@ -74,10 +140,41 @@ def dropSuffix? (s : Substring) (suff : Substring) : Option Substring :=
...
+Returns true iff `pattern` occurs as a substring of `s`.
+-/
+@[inline] def containsSubstr (s pattern : Substring) : Bool :=
+  s.findSubstr? pattern |>.isSome
+
 end Substring
 
 namespace String
 
+@[inherit_doc Substring.findSubstr?]
```
semorrison, line 166:
```suggestion
@[inherit_doc Substring.findAllSubstr]
```

Std/Data/String/Basic.lean
```diff
@@ -74,10 +140,41 @@ def dropSuffix? (s : Substring) (suff : Substring) : Option Substring :=
...
+abbrev findAllSubstr (s pattern : Substring) : Array Substring :=
```
semorrison, line 167: `s : String`

test/kmp_matcher.lean
```diff
@@ -0,0 +1,23 @@
...
+#guard_msgs in
+#eval Option.isSome <| m.find? "AbbabbA" -- false
+
+/-- info: true -/
+#guard_msgs in
+#eval Option.isSome <| m.find? "aabbaa" -- true
+
+/-- info: 2 -/
+#guard_msgs in
+#eval Array.size <| m.findAll "abbabba" -- 2
```
semorrison, line 19:
```suggestion
#guard (Array.size <| m.findAll "abbabba") = 2
```
etc for the other tests.

fgdorais: Will do.

test/kmp_matcher.lean
```diff
@@ -0,0 +1,23 @@
+import Std.Tactic.GuardMsgs
+import Std.Data.String.Basic
+
+/-! Tests for Knuth-Morris-Pratt matching algorithm -/
```
semorrison, line 4: Wouldn't hurt for the tests to directly exercise `String.findSubstr?`, given that's the main entrypoint.

Std/Data/String/Basic.lean
```diff
@@ -1,17 +1,83 @@
...
+  size := m.patternSize
+  /-- Accumulator loop for `String.Matcher.findAll` -/
+  loop (s : Substring) (m : Matcher) (occs : Array Substring) : Array Substring :=
+    match m.next? s with
+    | none => occs
+    | some (s, m) =>
+      loop s m <| occs.push { s with
+        startPos := ⟨s.startPos.byteIdx - size⟩
+        stopPos := s.startPos }
+
```
semorrison, line 69: Not a particularly serious suggestion, but you could produce a lazy list of matches here. :-)

Std/Data/Array/Match.lean
semorrison: Sorry: write `variable [BEq \a]` at the top of the file, so it isn't in every type signature.

Std/Data/Array/Match.lean
```diff
@@ -0,0 +1,126 @@
...
+def Matcher.ofStream [BEq α] [Stream σ α] (pat : σ) : Matcher α where
+  table := mkPrefixTableOfStream pat
+
+/-- Find next match from a given stream
+
+  Runs the stream until it reads a sequence that matches the sought pattern, then returns the stream
+  state at that point and an updated matcher state.
+-/
+partial def Matcher.next? [BEq α] [Stream σ α] (m : Matcher α) (stream : σ) :
+  Option (σ × Matcher α) :=
```
semorrison, line 116:
```suggestion
    Option (σ × Matcher α) :=
```
I suspect the same indenting rules apply here as in Mathlib: when the type signature continue over multiple lines it is indented 4 spaces.

Std/Data/String/Basic.lean
fgdorais: Good question! This is a definite possibility using `MLList`. I'll think about it. If it turns out to be straightforward I will include it here but a separate PR might be a better venue.

Std/Data/Array/Match.lean
digama0: @semorrison I try to avoid `variable` wherever possible in std. In this case I think it's fine as is.

fgdorais: I just reverted the change but I must admit having no personal preference either way.

Std/Data/String/Basic.lean
fgdorais: I thought about it for a bit. I like the idea but I think this would belong in a later PR. There seems to be a few choices available (`MLList` and `Stream` are the top two). Also, `Array.Match.next?` is already there as a low-level infrastructure to generate matches lazily. Since you said this was not particularly serious, I will politely pass.

test/kmp_matcher.lean
```diff
@@ -0,0 +1,27 @@
...
+
+/-- Matcher for pattern "abba" -/
+def m := String.Matcher.ofString "abba"
+
+#guard Option.isSome (m.find? "AbbabbA") = false
```
semorrison, line 9:
```suggestion
#guard ! (m.find? "AbbabbA").isSome
```
etc?

semorrison (approved):
Otherwise, LGTM now.

Std/Data/String/Basic.lean
```diff
@@ -1,17 +1,83 @@
...
+@[inline] def Matcher.ofString (pattern : String) : Matcher :=
+  Matcher.ofSubstring pattern
+
+/-- The string pattern for the matcher -/
+def Matcher.pattern (m : Matcher) : String :=
+  m.table.foldl (fun s (c,_) => s.push c) ""
+
+/-- The byte size of the string pattern for the matcher -/
+def Matcher.patternSize (m : Matcher) : Nat :=
+  String.Pos.byteIdx <| m.table.foldl (fun n (c,_) => n + c) 0
```
digama0, line 53: It is surprising that this is not O(1), and I believe this leads to an unexpected overhead in `findAll` below, which calls it in every iteration. (You shouldn't use `let rec` for precomputed state! It's just another recursive function.)

fgdorais: Yes, that's much better!

Std/Data/Array/Match.lean
```diff
@@ -0,0 +1,126 @@
...
+  valid : (h : i < toArray.size) → toArray[i].2 ≤ i
+
+instance : Inhabited (PrefixTable α) where
+  default := ⟨#[], fun .⟩
+
+/-- Returns the size of the prefix table -/
+abbrev PrefixTable.size (t : PrefixTable α) := t.toArray.size
+
+/-- Returns the pattern array of the prefix table -/
+abbrev PrefixTable.pattern (t : PrefixTable α) : Array α := t.toArray.map Prod.fst
```
digama0, line 27: This operation is not O(1), which I think messes up the asymptotic order of the algorithm because it is called in `PrefixTable.step`. Maybe the numbers array should just be a parallel array instead of being interleaved with the pattern?

Std/Data/Array/Match.lean
```diff
@@ -35,24 +35,19 @@ abbrev PrefixTable.pattern (t : PrefixTable α) : Array α := t.toArray.map Prod
...
     | k + 1 =>
       have h2 : k < t.size := Nat.lt_of_succ_lt_succ hk
       let k' := t.toArray[k].2
       have hk' : k' < k + 1 := Nat.lt_succ_of_le (t.valid h2)
       step t x ⟨k', Nat.lt_trans hk' hk⟩
+  if hsz : k < t.size then
+    if x == t.toArray[k].1 then
+      ⟨k+1, Nat.succ_lt_succ hsz⟩
+    else cont ()
+  else cont ()
```
fgdorais, line 47: This is cool! I didn't realize Lean would automatically inline `cont` so that this is still tail recursive.

PR #179, opened 2023-07-07
feat: ne_nil equivalence for take and drop

mo271:


mo271:
https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F

Std/Data/List/Lemmas.lean
```diff
@@ -1763,6 +1763,25 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+  List.drop_eq_nil_of_le <| h ▸ List.length_nil ▸ zero_le i
+
+theorem take_eq_nil_of_eq_nil.{u} {α : Type u} {as : List α} {i : Nat}
+    (h: as = []) : as.take i = []  :=
+  List.eq_nil_iff_forall_not_mem.mpr (fun a ha =>
+  List.eq_nil_iff_forall_not_mem.mp h a (List.mem_of_mem_take ha))
+
+theorem ne_nil_of_drop_ne_nil.{u} {α : Type u} {as : List α} {i : Nat}
+    (h: as.drop i ≠ []) : as ≠ [] :=
+    Exists.casesOn (List.exists_mem_of_ne_nil (List.drop i as) h) fun _ ha =>
```
fgdorais, line 1777: Simpler proof: `mt drop_eq_nil_of_eq_nil h`

Same for the next theorem.

fgdorais (changes requested):

Std/Data/List/Lemmas.lean
```diff
@@ -1763,6 +1763,25 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
       refine ⟨fun h => h₀ _ (mem_of_mem_drop h) rfl, ?_⟩
       exact disjoint_take_drop h₁ (Nat.le_of_succ_le_succ h)
 
+theorem drop_eq_nil_of_eq_nil.{u} {α : Type u} {as : List α} {i : Nat}
+    (h: as = []) : as.drop i = []  :=
```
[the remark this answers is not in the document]
mo271: done

Std/Data/List/Lemmas.lean
mo271: done

Std/Data/List/Lemmas.lean
```diff
@@ -1763,6 +1763,19 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
       refine ⟨fun h => h₀ _ (mem_of_mem_drop h) rfl, ?_⟩
       exact disjoint_take_drop h₁ (Nat.le_of_succ_le_succ h)
 
+theorem drop_eq_nil_of_eq_nil {as : List α} {i : Nat} (h: as = []) : as.drop i = [] :=
+  List.drop_eq_nil_of_le <| h ▸ List.length_nil ▸ zero_le i
```
fgdorais, line 1767: Simpler still:
```lean
theorem drop_eq_nil_of_eq_nil : ∀ {as : List α} {i}, as = [] → as.drop i = []
  | _, _, rfl => drop_nil
```
Same could be done for `take_eq_nil_of_eq_nil` after adding the missing `take_nil` lemma.

fgdorais (changes requested):
Please move to take and drop section that starts on line 693.

Std/Data/List/Lemmas.lean
mo271: Great! I also like the other golf with `bt`.
I added the `take_nil` lemma and was about to open a pr, but then I read the [contribution guidelines](https://github.com/leanprover/lean4/blob/c268d7e97bb075427a8412cceecc4e483379ac25/CONTRIBUTING.md)
and now I'm not sure if I should ask on zulip first. I think it will probably take 1minute of the reviewers time, but still.. What do you think?

https://github.com/mo271/lean4/commit/05273e9eed5f3680f5e2f199506b76f3764d256f

mo271: [Converting to draft, since the current way would depend on the propsed `take_nil` lemma.

fgdorais: Those are guidelines for lean4 core not for std4! Just add `take_nil` in this PR and it should be fine.

mo271: Yes, I just thought it would be nice to have `take_nil` in the same place as `drop_nil`, but I guess we can also put it in std4...

mo271: Added `take_nil` as suggested.

fgdorais: Adding things to core that aren't actually needed in core is a no-go... Who knows, maybe `drop_nil` will eventually be moved to std4!

fgdorais (approved):

JamesGallicchio (approved):

fpvandoorn (approved):

PR #182, opened 2023-07-09
feat: cleanup Nat lemmas

fgdorais:
Split from #177. Just cleaning and golfing. No breaking changes. Added five handy lemmas: `le_mul_of_pos_left`, `le_mul_of_pos_right`, `shiftLeft_eq'`, `one_dvd`, `two_pos`.


fgdorais:
No longer needed.

PR #183, opened 2023-07-10
feat: `Bool` lemmas

fgdorais:
- [x] Depends on #200 
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/tree/fgdorais-bool


Std/Data/Bool.lean
```diff
@@ -0,0 +1,159 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: F. G. Dorais
+-/
+
+import Std.Tactic.Alias
+
+/-- Boolean exclusive or -/
+abbrev xor : Bool → Bool → Bool := bne
+
+@[inherit_doc] infixl:30 " ^^ " => xor
```
fgdorais, line 12: I'm not entirely sure about making this a global notation. It's a natural companion to the other `Bool` operators `!`, `&&`, `||`.

fgdorais:
I'm working on a [Mathlib patch](https://github.com/leanprover-community/mathlib4/tree/fgdorais-bool) but I'm hitting a strange issue at Mathlib.Analysis.BoxIntegral.Basic that is hard to track down.

Std/Data/Bool.lean
urkud: Should we reuse `Xor`? It has a correct type signature in this case.

digama0: I don't think we should have xor at all. If mathlib wants it, okay, but it's just a synonym for an existing definition with a notation already.

Std/Data/Bool.lean
```diff
@@ -0,0 +1,166 @@
...
+
+section
+variable (xs ys : List Bool)
+
+attribute [local simp] and_assoc or_assoc xor_assoc List.map List.join
+
+/-- Boolean conjunction (`and`) of a list -/
+abbrev all : Bool := xs.all id
+
+@[simp] theorem all_nil : all [] = true := rfl
```
urkud, line 127: Can't `simp` prove this using `simp` lemmas for `List.all`?
Do we need an abbreviation for a special case of `List.all`?

Std/Data/Bool.lean
fgdorais: The abbreviation is still useful since `bne` is polymorphic. I moved `xor` in the `Bool` namespace and removed the notation. Mathlib can add `export Bool (xor)` somewhere, if necessary.

Std/Data/Bool.lean
```diff
@@ -0,0 +1,166 @@
...
+/-- Boolean exclusive or -/
+abbrev xor : Bool → Bool → Bool := bne
+
+@[inherit_doc] infixl:30 " ^^ " => xor
+
+namespace Bool
+variable (x y z : Bool)
+
+/- Boolean truth table tactic -/
+local syntax "btt" (&"using" tactic)? (colGt term:max)* : tactic
```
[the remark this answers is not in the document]
fgdorais: Good point! I updated the `btt` macro as a convenient local shorthand for `revert $vars*; decide`.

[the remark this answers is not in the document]
digama0: `revert_all; decide` is also available in mathlib as `decide!`

Std/Data/Bool.lean
fgdorais: That was a remnant from when I was defining `List.all` as `Bool.all (List.map _ _)` in my personal libraries. It's no longer needed.

Std/Data/Bool.lean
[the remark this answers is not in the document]
fgdorais: But `revert_all` also doesn't exist...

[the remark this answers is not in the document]
digama0: `decide!` was also apparently never ported...

fgdorais:
Updated to use new alias command from #200

Std/Data/Bool.lean
```diff
@@ -0,0 +1,129 @@
...
+  | isTrue ht, isTrue hf => isTrue fun | true => ht | false => hf
+
+instance (p : Bool → Prop) [inst : DecidablePred p] : Decidable (∃ x, p x) :=
+  match inst true, inst false with
+  | isTrue ht, _ => isTrue ⟨_, ht⟩
+  | _, isTrue hf => isTrue ⟨_, hf⟩
+  | isFalse ht, isFalse hf => isFalse fun | ⟨true, h⟩ => absurd h ht | ⟨false, h⟩ => absurd h hf
+
+/-- Convenience truth table tactic -/
+local macro "btt " vars:(colGt ident)* : tactic => `(tactic| revert $vars* <;> decide)
```
semorrison, line 34: The remaining non-`btt` proofs below are also just `by btt`.

semorrison (approved):

Std/Data/Bool.lean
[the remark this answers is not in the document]
semorrison: @digama0, what do you want done here?
1. Nothing
2. A note saying to replace `btt` with `decide!` when it exists
3. To wait on `decide!`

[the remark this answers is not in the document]
digama0: if you manually do the revert by moving the colon back, can't you just use `decide` here?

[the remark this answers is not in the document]
semorrison: Much better!

Std/Data/Bool.lean
```diff
@@ -0,0 +1,251 @@
...
+  decide
+
+theorem xor_comm : ∀ (x y : Bool), xor x y = xor y x := by
+  decide
+
+theorem xor_left_comm : ∀ (x y z : Bool), xor x (xor y z) = xor y (xor x z) := by
+  decide
+
+theorem xor_right_comm : ∀ (x y z : Bool), xor (xor x y) z = xor (xor x z) y := by
+  decide
```
joehendrix, line 158: Could you say why `xor_left_comm` and `xor_right_comm` are useful?  Is it for normalization purposes?

Std/Data/Bool.lean
```diff
@@ -0,0 +1,251 @@
...
+
+theorem and_or_inj_right' :
+    ∀ {m x y : Bool}, (x && m) = (y && m) ∧ (x || m) = (y || m) ↔ x = y := by
+  decide
+
+theorem and_or_inj_left : ∀ {m x y : Bool}, (m && x) = (m && y) → (m || x) = (m || y) → x = y := by
+  decide
+
+theorem and_or_inj_left' : ∀ {m x y : Bool}, (m && x) = (m && y) ∧ (m || x) = (m || y) ↔ x = y := by
+  decide
```
joehendrix, line 249: I'm probably missing context, but could you say why `and_or_inj_right` and `and_or_inj_right'` and it's left counterparts are useful?

Std/Data/Bool.lean
fgdorais: These are "traditional" (cf. `Nat.add_right_comm`, `Nat.mul_right_comm`, etc.) They save some juggling parentheses here and there. Long ago, some of these were marked simp as a safer way to use commutativity in simp. This is not the case anymore but they are probably still used in some normalization procedures. @digama0 probably knows the history better than I do.

Std/Data/Bool.lean
fgdorais: This is from Boolean Algebra theory: neither `&&` nor `||` have a cancellation property, but they do have a joint cancellation property which is this one. It's mostly just a "wheel" that doesn't need perpetual reinventing. That said, I'm not a fan of the `'` notation, so I'd be happy with different names. (The names I used existed somewhere at some time in code, unfortunately I never thought it useful to record where and when.)

Std/Data/Bool.lean
joehendrix: Thanks!

Std/Data/Bool.lean
joehendrix: Thanks.  How about renaming `and_or_inj_left'` to `and_or_inj_left_iff`?  That's consistent with `or_eq_true_iff` above.

Std/Data/Bool.lean
```diff
@@ -0,0 +1,251 @@
...
+
+end Bool
```
joehendrix, line 251: I think for consistency, let's remove the `end Bool`.

This could eventually be a lint check or something an auto-formatter resolves.

joehendrix (changes requested):

joehendrix (approved):

Std/Data/Bool.lean
```diff
@@ -0,0 +1,256 @@
...
+instance (p : Bool → Prop) [inst : DecidablePred p] : Decidable (∃ x, p x) :=
+  match inst true, inst false with
+  | isTrue ht, _ => isTrue ⟨_, ht⟩
+  | _, isTrue hf => isTrue ⟨_, hf⟩
+  | isFalse ht, isFalse hf => isFalse fun | ⟨true, h⟩ => absurd h ht | ⟨false, h⟩ => absurd h hf
+
+instance : LE Bool := leOfOrd
+instance : LT Bool := ltOfOrd
+instance : Max Bool := maxOfLe
+instance : Min Bool := minOfLe
```
eric-wieser, line 35: These change the defeq of the boolean order to be (IMO) worse. Was this a deliberate change?

fgdorais: There was a hope long ago that core would fixup `Ord` before this PR landed. Without that fixup, this is indeed a bad choice. My preferred versions are:
```lean
le x y := !x || y
lt x y := !x && y
max x y :=  x || y
min x y := x && y
```
These lead to the most efficient code in each case. I'm not sure if you really prefer your version for `le`? I will submit a small follow-up PR fixing this as soon as I have a chance.

digama0:
I think this PR still needs changes, but I didn't have time to review it and things are a bit more complicated now. I told Scott to work around this for now for the [mathlib bump](https://github.com/leanprover-community/mathlib4/pull/7982), and I will try to make a follow up PR when I have the time.

PR #184, opened 2023-07-10
feat: add alias tactic

fgdorais:
Move from Mathlib


Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,204 @@
...
+/-- Like `++`, except that if the right argument starts with `_root_` the namespace will be
+ignored.
+```
+appendNamespace `a.b `c.d = `a.b.c.d
+appendNamespace `a.b `_root_.c.d = `c.d
+```
+
+TODO: Move this declaration to a more central location.
+-/
+def appendNamespace (ns : Name) : Name → Name
```
fpvandoorn, line 69: Maybe this should be moved to a central location in Std? Other than that, LGTM

fpvandoorn (approved):

Std/Tactic/Alias.lean
digama0: or in lean... I know this is reimplemented at least 3 times in lean core

fgdorais: I'm happy to change this given a pointer to an implementation in core.

fgdorais: I couldn't find this exact thing in core but I did simplify the definition a bit.

fgdorais: There is now a PR in core: [lean4#2341](https://github.com/leanprover/lean4/pull/2341).

fgdorais:
I added support for protected aliases and I added a Mathlib patch.

fgdorais:
Obsoleted by #200

PR #185, opened 2023-07-12
feat: make the unreachable tactic linter complain if tactics are called with no goals remaining

alexjbest:


fgdorais (approved):

digama0:
The unreachableTactic linter already has a whitelist for tactics that are supposed to be unreachable (like the `unreachable` tactic, or the body of the `stop` tactic). Things like `tada` and `done` would similarly be whitelisted from the noGoals linter.

fgdorais:

PR #191, opened 2023-07-21
feat: golf proof of `Int.eq_ofNat_of_zero_le`

chabulhwi:


Std/Data/Int/Lemmas.lean
```diff
@@ -581,7 +581,7 @@ protected theorem le_total (a b : Int) : a ≤ b ∨ b ≤ a :=
 theorem ofNat_zero_le (n : Nat) : 0 ≤ (↑n : Int) := ofNat_le.2 n.zero_le
 
 theorem eq_ofNat_of_zero_le {a : Int} (h : 0 ≤ a) : ∃ n : Nat, a = n := by
-  have t := le.dest_sub h; simp [Int.sub_eq_add_neg] at t; exact t
+  have t := le.dest_sub h; rw [Int.sub_zero] at t; exact t
```
fgdorais, line 584: You can also use `rwa`: `have t := le.dest_sub h; rwa [Int.sub_zero] at t`

chabulhwi: @fgdorais Thanks for the heads-up!

fgdorais (approved):

PR #193, opened 2023-07-23
feat: custom recursors for `Nat`

fgdorais:
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/6188

Added recursors that use `0` and `_+1` instead of `Nat.zero` and `Nat.succ`.

Added strong recursors.

Added diagonal recursors for two variables.

Added lemmas for these recursors.

### DIFF
--- a/Std/Data/Nat/Basic.lean
+++ b/Std/Data/Nat/Basic.lean
@@ -8,6 +8,100 @@ import Std.Classes.Dvd
 namespace Nat
 
 /--
+  Recursor identical to `Nat.rec` but uses notations `0` for `Nat.zero` and `·+1` for `Nat.succ`
+-/
+@[elab_as_elim]
+protected def recAux {motive : Nat → Sort _} (zero : motive 0) (succ : ∀ n, motive n → motive (n+1)) :
+    (t : Nat) → motive t
+  | 0 => zero
+  | _+1 => succ _ (Nat.recAux zero succ _)
+
+/--
+  Recursor identical to `Nat.recOn` but uses notations `0` for `Nat.zero` and `·+1` for `Nat.succ`
+-/
+@[elab_as_elim]
+protected def recAuxOn {motive : Nat → Sort _} (t : Nat) (zero : motive 0)
+  (succ : ∀ n, motive n → motive (n+1)) : motive t := Nat.recAux zero succ t
+
+/--
+  Recursor identical to `Nat.casesOn` but uses notations `0` for `Nat.zero` and `·+1` for `Nat.succ`
+-/
+@[elab_as_elim]
+protected def casesAuxOn {motive : Nat → Sort _} (t : Nat) (zero : motive 0)
+  (succ : ∀ n, motive (n+1)) : motive t := Nat.recAux zero (fun n _ => succ n) t
+
+/--
+  Strong recursor for `Nat`
+-/
+@[elab_as_elim]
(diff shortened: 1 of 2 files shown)


Std/Data/Nat/Lemmas.lean
```diff
@@ -10,7 +10,109 @@ import Std.Data.Nat.Basic
 
 namespace Nat
 
-/-! ## le/lt -/
+/-! ### rec/cases -/
+
+section recAux
+variable {motive : Nat → Sort _} (zero : motive 0) (succ : ∀ n, motive n → motive (n+1))
```
digama0, line 16: `variable` should be avoided unless absolutely necessary, especially if you start shadowing them like below. For these theorems you can use either autoimplicits or `(zero succ)` binders in each theorem.

digama0 (commented):
LGTM modulo suggestions

Std/Data/Nat/Lemmas.lean
fgdorais: `elab_as_elim` got in the way of using autoimplicits but I did remove all `variable` commands.

Std/Data/Nat/Lemmas.lean
```diff
@@ -10,7 +10,127 @@ import Std.Data.Nat.Basic
 
 namespace Nat
 
-/-! ## le/lt -/
+/-! ### rec/cases -/
+
+@[simp] theorem recAux_zero {motive : Nat → Sort _} (zero : motive 0)
+  (succ : ∀ n, motive n → motive (n+1)) :
```
digama0, line 16:
```suggestion
    (succ : ∀ n, motive n → motive (n+1)) :
```
standard style is 4 spaces indent for everything in the header

Std/Data/Nat/Basic.lean
```diff
@@ -7,6 +7,100 @@ import Std.Classes.Dvd
...
+    (m n : Nat) → motive m n := Nat.recDiagAux left right succ_succ
+where
+  /-- Left leg for `Nat.recDiag` -/
+  left : ∀ n, motive 0 n
+  | 0 => zero_zero
+  | _+1 => zero_succ _ (left _)
+  /-- Right leg for `Nat.recDiag` -/
+  right : ∀ m, motive m 0
+  | 0 => zero_zero
+  | _+1 => succ_zero _ (right _)
```
eric-wieser, line 78: I assume there's a reason this is preferred to:
```
@[elab_as_elim]
def recDiag {motive : Nat → Nat → Sort _}
  (zero_zero : motive 0 0)
  (zero_succ : ∀ n, motive 0 n → motive 0 (n+1))
  (succ_zero : ∀ m, motive m 0 → motive (m+1) 0)
  (succ_succ : ∀ m n, motive m n → motive (m+1) (n+1)) :
    (m n : Nat) → motive m n
  | 0, 0 => zero_zero
  | 0, _+1 => zero_succ _ (recDiag zero_zero zero_succ succ_zero succ_succ _ _)
  | _+1, 0 => succ_zero _ (recDiag zero_zero zero_succ succ_zero succ_succ _ _)
  | _+1, _+1 => succ_succ _ _ (recDiag zero_zero zero_succ succ_zero succ_succ _ _)
  termination_by recDiag m n => m + n
```

fgdorais: Yes, a similar version made some simp lemmas non-trivial. Also, I often find myself preferring `Nat.recDiagAux` over `Nat.recDiag` in practice.

PR #194, opened 2023-07-23
feat: more order lemmas for `Nat`

fgdorais:
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/8077

There are lots and lots of synonyms here! I didn't remove any but I did try to group them together.

### DIFF
--- a/Std/Data/Nat/Init/Lemmas.lean
+++ b/Std/Data/Nat/Init/Lemmas.lean
@@ -60,10 +60,6 @@ protected theorem pow_two_pos (w : Nat) : 0 < 2^w := Nat.pos_pow_of_pos _ (by de
 @[simp] protected theorem not_lt {a b : Nat} : ¬ a < b ↔ b ≤ a :=
   ⟨Nat.ge_of_not_lt, flip Nat.not_le_of_gt⟩
 
---protected theorem not_lt_of_le {n m : Nat} (h₁ : m ≤ n) : ¬ n < m := (Nat.not_le_of_gt · h₁)
-
---protected theorem le_of_not_le {a b : Nat} : ¬ a ≤ b → b ≤ a := (Nat.le_total a b).resolve_left
-
 protected theorem le_min_of_le_of_le {a b c : Nat} : a ≤ b → a ≤ c → a ≤ min b c := by
   intros; cases Nat.le_total b c with
   | inl h => rw [Nat.min_eq_left h]; assumption
--- a/Std/Data/Nat/Init/Lemmas.lean
+++ b/Std/Data/Nat/Init/Lemmas.lean
@@ -54,9 +54,15 @@ protected theorem le_max_right (a b : Nat) : b ≤ max a b := Nat.max_comm .. �
 
 protected theorem pow_two_pos (w : Nat) : 0 < 2^w := Nat.pos_pow_of_pos _ (by decide)
 
-protected theorem not_lt_of_le {n m : Nat} (h₁ : m ≤ n) : ¬ n < m := (Nat.not_le_of_gt · h₁)
+@[simp] protected theorem not_le {a b : Nat} : ¬ a ≤ b ↔ b < a :=
+  ⟨Nat.gt_of_not_le, Nat.not_le_of_gt⟩
 
-protected theorem le_of_not_le {a b : Nat} : ¬ a ≤ b → b ≤ a := (Nat.le_total a b).resolve_left
+@[simp] protected theorem not_lt {a b : Nat} : ¬ a < b ↔ b ≤ a :=
+  ⟨Nat.ge_of_not_lt, flip Nat.not_le_of_gt⟩
+
+--protected theorem not_lt_of_le {n m : Nat} (h₁ : m ≤ n) : ¬ n < m := (Nat.not_le_of_gt · h₁)
+
+--protected theorem le_of_not_le {a b : Nat} : ¬ a ≤ b → b ≤ a := (Nat.le_total a b).resolve_left
 
 protected theorem le_min_of_le_of_le {a b c : Nat} : a ≤ b → a ≤ c → a ≤ min b c := by
   intros; cases Nat.le_total b c with
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -335,8 +335,6 @@ theorem le_pred_of_lt (h : n < m) : n ≤ pred m := (le_pred_iff_lt (Nat.zero_lt
 
 /-! ### add -/
 
-theorem add_one_ne_zero (n) : n + 1 ≠ 0 := Nat.succ_ne_zero _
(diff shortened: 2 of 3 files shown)


fgdorais (commented):
Because of all the redundancies, this should make use of alias #200

fgdorais:
I'm not sure whether to do extra effort to avoid split here.

digama0:
> I'm not sure whether to do extra effort to avoid split here.

Let's not consider avoiding `split`  until it is clear that it will not be fixed.

Std/Data/Nat/Lemmas.lean
```diff
@@ -133,85 +133,170 @@ theorem recDiagOn_succ_succ {motive : Nat → Nat → Sort _} (zero_zero : motiv
...
+protected theorem lt_iff_le_and_ne {a b : Nat} : a < b ↔ a ≤ b ∧ a ≠ b :=
+  ⟨fun h => ⟨Nat.le_of_lt h, Nat.ne_of_lt h⟩, fun h => Nat.lt_of_le_of_ne h.1 h.2⟩
+
+protected theorem le_antisymm_iff {a b : Nat} : a = b ↔ a ≤ b ∧ b ≤ a :=
+  ⟨fun | rfl => ⟨Nat.le_refl _, Nat.le_refl _⟩, fun ⟨hle, hge⟩ => Nat.le_antisymm hle hge⟩
+
+protected alias eq_iff_le_and_ge := Nat.le_antisymm_iff
 
-protected theorem not_lt_of_le {n m : Nat} (h₁ : m ≤ n) : ¬ n < m := (Nat.not_le_of_gt · h₁)
+protected theorem lt_connex {a b : Nat} : a ≠ b → a < b ∨ b < a := by
```
YaelDillies, line 175: This is a weird name. Is there any precedent?

digama0: This is "connectedness" of a relation, but I agree it is not used in the lean naming convention. I would just call it `lt_or_gt_of_ne` (and ignore the part where it says `gt` even though it uses `lt`).

fgdorais: Yes, there's a million synonyms in this section. So, in this case, the consensus is to just keep `lt_or_gt_of_ne` and delete both `lt_connex` and `lt_or_lt_of_ne`?

fgdorais:
> As in #203 I find it a bit sad to duplicate mathlib material in Std for the purpose of specialising it to `ℕ`. Can we do some subset of the following suggestions?
> 
> * Not duplicate basic order lemmas in Std
> * Put in the docstring of duplicate `Nat.le_something` that `le_something` exists in mathlib
> * Deprecate mathlib-wise (is that currently possible?) each `Nat.le_something` in favor of `le_something`
> * Put a big warning in `Std.Data.Nat.Lemmas` that if people want a lot of theory about `≤` and `<` on `ℕ`, then they might be better off using mathlib instead of PRing more lemmas here

This was addressed in #203 and then #237

Std/Data/Nat/Lemmas.lean
```diff
@@ -134,85 +134,167 @@ theorem recDiagOn_succ_succ {motive : Nat → Nat → Sort _} (zero_zero : motiv
...
 
 @[simp] protected theorem not_le {a b : Nat} : ¬ a ≤ b ↔ b < a :=
   ⟨Nat.gt_of_not_le, Nat.not_le_of_gt⟩
 
 @[simp] protected theorem not_lt {a b : Nat} : ¬ a < b ↔ b ≤ a :=
   ⟨Nat.ge_of_not_lt, flip Nat.not_le_of_gt⟩
 
-theorem le_lt_antisymm {n m : Nat} (h₁ : n ≤ m) (h₂ : m < n) : False :=
-  Nat.lt_irrefl n (Nat.lt_of_le_of_lt h₁ h₂)
+protected alias ⟨lt_of_not_ge, _⟩ := Nat.not_le
```
adomani, line 145:
```suggestion
protected alias ⟨lt_of_not_ge, _⟩ := Not.le
```
I'm not sure that this is really desirable, but I think that this would allow dot-notation...

YaelDillies: Damiano, do you understand the new `alias` syntax? The above line is equivalent to
```
/-- Alias of the forward direction of `Nat.not_le`. -/
protected theorem lt_of_not_ge := Nat.not_le.1
```

adomani: Ah, apparently I do not understand the syntax!

Anyway, I'm not sure that having a declaration `Not.le` around is necessarily a good idea!

fgdorais: Where is `Not.le`? I can't find it.

YaelDillies: Damiano thought you were `Nat.not_le` an alias for `Nat.lt_of_not_ge`, not the other way around.

fgdorais: OK. I'll mark this resolved then.

digama0 (approved):
Approved with modifications:

* `lt_iff_compare_eq_lt` (et al) was reversed and renamed to just `compare_eq_lt`
* `compare_swap`
* `lt_or_lt_of_ne` was undeprecated, the current pattern of aliases allows referring to flipped `<` as either `gt` or `lt`
*  `compare_swap_symm` -> `compare_swap`
* Assorted proof shortening

Std/Data/Nat/Init/Lemmas.lean
```diff
@@ -54,9 +54,15 @@ protected theorem le_max_right (a b : Nat) : b ≤ max a b := Nat.max_comm .. 
...
+@[simp] protected theorem not_le {a b : Nat} : ¬ a ≤ b ↔ b < a :=
+  ⟨Nat.gt_of_not_le, Nat.not_le_of_gt⟩
 
-protected theorem le_of_not_le {a b : Nat} : ¬ a ≤ b → b ≤ a := (Nat.le_total a b).resolve_left
+@[simp] protected theorem not_lt {a b : Nat} : ¬ a < b ↔ b ≤ a :=
+  ⟨Nat.ge_of_not_lt, flip Nat.not_le_of_gt⟩
+
+--protected theorem not_lt_of_le {n m : Nat} (h₁ : m ≤ n) : ¬ n < m := (Nat.not_le_of_gt · h₁)
+
+--protected theorem le_of_not_le {a b : Nat} : ¬ a ≤ b → b ≤ a := (Nat.le_total a b).resolve_left
```
digama0, line 65: move these back to `Data.Nat.Lemmas`

fgdorais: They're already there, but I did forget to fully delete them.

digama0:
conflicts with other PR

fgdorais:
Fixed conflicts.

PR #195, opened 2023-07-23
feat: more succ/pred lemmas for `Nat`

fgdorais:
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/6203

There are two potentially breaking changes here:

* `Nat.le_pred_of_lt` was changed to use `pred` instead of `_-1` and `Nat.le_sub_one_of_lt` was added as a replacement.
* Implicit parameters for `Nat.eq_zero_of_nonpos`.

### DIFF
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -249,7 +249,7 @@ theorem succ_lt_succ_iff : succ a < succ b ↔ a < b := ⟨lt_of_succ_lt_succ, s
 theorem pred_inj : ∀ {a b}, 0 < a → 0 < b → pred a = pred b → a = b
   | _+1, _+1, _, _ => congrArg _
 
-theorem pred_ne_self : ∀ {a}, 0 ≠ a → pred a ≠ a
+theorem pred_ne_self : ∀ {a}, a ≠ 0 → pred a ≠ a
   | _+1, _ => (succ_ne_self _).symm
 
 theorem pred_lt_self : ∀ {a}, 0 < a → pred a < a
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -212,7 +212,7 @@ protected theorem one_pos : 0 < 1 := Nat.zero_lt_one
 
 protected theorem two_pos : 0 < 2 := Nat.zero_lt_two
 
-protected theorem eq_zero_of_nonpos (h : ¬0 < n) : n = 0 :=
+protected theorem eq_zero_of_not_pos (h : ¬0 < n) : n = 0 :=
   Nat.eq_zero_of_le_zero (Nat.not_lt.1 h)
 
 /-! ## succ/pred -/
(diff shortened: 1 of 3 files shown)


Std/Data/Nat/Lemmas.lean
```diff
@@ -199,77 +199,96 @@ protected theorem le_iff_lt_or_eq {n m : Nat} : n ≤ m ↔ n < m ∨ n = m :=
...
+protected theorem one_lt_two : 1 < 2 := Nat.succ_lt_succ Nat.zero_lt_one
 
-/-! ### succ/pred -/
+protected theorem one_pos : 0 < 1 := Nat.zero_lt_one
 
-attribute [simp] succ_ne_zero lt_succ_self Nat.pred_zero Nat.pred_succ
+protected theorem two_pos : 0 < 2 := Nat.zero_lt_two
 
-theorem succ_le {n m : Nat} : succ n ≤ m ↔ n < m := .rfl
+protected theorem eq_zero_of_nonpos (h : ¬0 < n) : n = 0 :=
```
digama0, line 218:
```suggestion
protected theorem eq_zero_of_not_pos (h : ¬0 < n) : n = 0 :=
```
nonpos is used for `n <= 0`

Std/Data/Nat/Lemmas.lean
```diff
@@ -199,77 +199,96 @@ protected theorem le_iff_lt_or_eq {n m : Nat} : n ≤ m ↔ n < m ∨ n = m :=
...
-  | a+1, b+1, _,  _, h => by rw [show a = b from h]
-  | a+1, 0,   _, hb, _ => absurd hb (Nat.lt_irrefl _)
-  | 0,   b+1, ha, _, _ => absurd ha (Nat.lt_irrefl _)
-  | 0,   0,   _,  _, _ => rfl
+theorem pred_inj : ∀ {a b}, 0 < a → 0 < b → pred a = pred b → a = b
+  | _+1, _+1, _, _ => congrArg _
 
-theorem pred_lt_pred : ∀ {n m : Nat}, n ≠ 0 → n < m → pred n < pred m
-  | 0,   _,   h, _ => (h rfl).elim
+theorem pred_ne_self : ∀ {a}, 0 ≠ a → pred a ≠ a
```
digama0, line 255:
```suggestion
theorem pred_ne_self : ∀ {a}, a ≠ 0 → pred a ≠ a
```

Std/Data/Nat/Lemmas.lean
```diff
@@ -991,3 +1012,8 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
   | k + 1 => by
     rw [shiftRight_add, shiftRight_eq_div_pow m k]
     simp [Nat.div_div_eq_div_mul, ← Nat.pow_succ]
+
+/-! ## deprecated -/
+
+@[deprecated Nat.one_add]
+theorem succ_eq_one_add (n) : succ n = 1 + n := (one_add _).symm
```
digama0, line 1019: remove the section, but also I don't think this needs to be deprecated

digama0 (approved):

PR #196, opened 2023-07-23
feat: more min/max lemmas for `Nat`

fgdorais:
- [x] Depends on #193 
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/8074

### DIFF
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -453,47 +453,185 @@ protected theorem sub_add_lt_sub {n m k : Nat} (h₁ : m + k ≤ n) (h₂ : 0 < 
 
 /-! ## min/max -/
 
+theorem min_succ_succ (x y) : min (succ x) (succ y) = succ (min x y) := by
+  simp [Nat.min_def, succ_le_succ_iff]; split <;> rfl
+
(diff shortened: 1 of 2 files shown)


Std/Data/Nat/Lemmas.lean
```diff
@@ -971,3 +1117,17 @@ protected theorem dvd_of_mul_dvd_mul_right (kpos : 0 < k) (H : m * k ∣ n * k)
 
 @[simp] theorem sum_append : Nat.sum (l₁ ++ l₂) = Nat.sum l₁ + Nat.sum l₂ := by
   induction l₁ <;> simp [*, Nat.add_assoc]
+
+/-! ## deprecated -/
+
+@[deprecated Nat.max_zero_left]
```
digama0, line 1123: Why the rename? I don't see any rationale given,  and TBH I prefer the original name.

digama0:
> Deprecated Nat.zero_min, Nat.min_zero, Nat.zero_max, Nat.max_zero because min/max are not infix operators.

Naming is not based on the notation of operators. min/max are binary operators, so this is a perfectly valid way to name lemmas.

fgdorais:
Updated to enforce naming conventions. Mathlib patch now in CI.

PR #197, opened 2023-07-23
feat: more add lemmas for `Nat`

fgdorais:
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/6202

Potentially breaking changes here are mostly fixes to implicit parameters for `Nat.lt_add_right`, `Nat.add_le_add_iff_left`, `Nat.add_le_add_iff_right`, `Nat.add_lt_add_iff_left`, `Nat.add_lt_add_iff_right`. Also `Nat.succ_add_eq_succ_add` was deprecated in favor of `Nat.succ_add_eq_add_succ`.


semorrison (approved):

Std/Data/Nat/Lemmas.lean
```diff
@@ -333,84 +333,96 @@ theorem lt_of_le_pred (h : 0 < m) : n ≤ pred m → n < m := (le_pred_iff_lt h)
...
-/-! ### add -/
+/-! ## add -/
 
-protected theorem eq_zero_of_add_eq_zero_right : ∀ {n m : Nat}, n + m = 0 → n = 0
-  | 0,   m => by simp [Nat.zero_add]
-  | n+1, m => fun h => by
-    rw [add_one, succ_add] at h
-    cases succ_ne_zero _ h
+protected theorem add_add_add_comm (a b c d : Nat) : (a + b) + (c + d) = (a + c) + (b + d) := by
+  rw [Nat.add_assoc, Nat.add_assoc, Nat.add_left_comm b]
```
semorrison, line 339: I couldn't guess the theorem from the name.

Is this actually useful? I feel this one might be beyond the point we want to stop with associativity/commutativity lemmas.

semorrison (approved):
Otherwise looks fine and hopefully we can get in this time!

Std/Data/Nat/Lemmas.lean
fgdorais: This one is not necessarily for human consumption, but it occurs all the time in associativity/commutativity normalization algorithms (e.g. the general use pattern is similar to proving that the two ways to distribute `(a + b) * (c + d)` are equal).

I'm not a fan of the name either (I normally call this cross commutativity) but this is the traditional Lean/Mathlib name.

digama0 (approved):
merge conflict

PR #198, opened 2023-07-23
feat: more mul lemmas for `Nat`

fgdorais:
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/6204

The major changes here are deprecating lemmas with obscure names: `Nat.mul_le_mul_of_nonneg_left`, `Nat.mul_le_mul_of_nonneg_right`, `Nat.mul_lt_mul`, `Nat.mul_lt_mul'`.

This PR also finally implements #118 as `Nat.mul_lt_mul_of_lt_of_lt`.


Std/Data/Nat/Lemmas.lean
```diff
@@ -991,3 +997,21 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
...
+protected theorem mul_le_mul_of_nonneg_left {a b c : Nat} : a ≤ b → c * a ≤ c * b :=
+  Nat.mul_le_mul_left c
+
+@[deprecated Nat.mul_le_mul_right]
+protected theorem mul_le_mul_of_nonneg_right {a b c : Nat} : a ≤ b → a * c ≤ b * c :=
+  Nat.mul_le_mul_right c
+
+@[deprecated Nat.mul_lt_mul_of_lt_of_le]
+protected theorem mul_lt_mul (hac : a < c) (hbd : b ≤ d) (hb : 0 < b) : a * b < c * d :=
+  Nat.mul_lt_mul_of_lt_of_le hac hbd (Nat.lt_of_lt_of_le hb hbd)
```
digama0, line 1013: I don't really like the idea of not having a theorem called `mul_lt_mul`, but I take your point regarding the ambiguity. But `mul_lt_mul_of_lt_of_le` is not really a good replacement for this theorem, since the proof needs to use `hbd` twice. You should make this a deprecated alias for an undeprecated theorem, perhaps `mul_lt_mul_of_lt_of_le'`?

Std/Data/Nat/Lemmas.lean
```diff
@@ -503,43 +503,49 @@ protected theorem mul_right_comm (n m k : Nat) : n * m * k = n * k * m := by
...
+
+protected theorem mul_lt_mul_of_le_of_lt (hac : a ≤ c) (hbd : b < d) (hc : 0 < c) :
+    a * b < c * d :=
+  Nat.lt_of_le_of_lt (Nat.mul_le_mul_right _ hac) (Nat.mul_lt_mul_of_pos_left hbd hc)
+
+protected theorem mul_lt_mul_of_lt_of_lt {a b c d : Nat} (hac : a < c) (hbd : b < d) :
+    a * b < c * d :=
+  Nat.mul_lt_mul_of_le_of_lt (Nat.le_of_lt hac) hbd (Nat.zero_lt_of_lt hac)
+
+theorem succ_mul_succ_eq (a b) : succ a * succ b = a * b + a + b + 1 := by
```
digama0, line 543:
```suggestion
theorem succ_mul_succ (a b) : succ a * succ b = a * b + a + b + 1 := by
```

Std/Data/Nat/Lemmas.lean
```diff
@@ -991,3 +997,21 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
   | k + 1 => by
     rw [shiftRight_add, shiftRight_eq_div_pow m k]
     simp [Nat.div_div_eq_div_mul, ← Nat.pow_succ]
+
+/-! ## deprecated -/
```
digama0, line 1001: remove the section

Std/Data/Nat/Lemmas.lean
```diff
@@ -5,6 +5,7 @@ Authors: Leonardo de Moura, Jeremy Avigad, Mario Carneiro
 -/
 import Std.Logic
 import Std.Tactic.Basic
+import Std.Tactic.Alias
```
eric-wieser, line 8: This import is now unused, right?

digama0: used on L548

Std/Data/Nat/Lemmas.lean
```diff
@@ -779,22 +788,41 @@ protected theorem mul_left_cancel_iff {n: Nat} (p : 0 < n) (m k : Nat) : n * m =
...
+@[deprecated] protected alias mul_lt_mul' := Nat.mul_lt_mul_of_le_of_lt
 
-protected theorem mul_lt_mul' (h1 : a ≤ c) (h2 : b < d) (h3 : 0 < c) : a * b < c * d :=
-  Nat.lt_of_le_of_lt (Nat.mul_le_mul_of_nonneg_right h1) (Nat.mul_lt_mul_of_pos_left h2 h3)
+protected theorem mul_lt_mul_of_lt_of_lt {a b c d : Nat} (hac : a < c) (hbd : b < d) :
+    a * b < c * d :=
+  Nat.mul_lt_mul_of_le_of_lt (Nat.le_of_lt hac) hbd (Nat.zero_lt_of_lt hac)
 
+theorem succ_mul_succ (a b) : succ a * succ b = a * b + a + b + 1 := by
+  rw [succ_mul, mul_succ]; rfl
```
semorrison, line 825: This seems strange, mixing `succ` and `+1`. Can't we declare this out of scope because you should replace `succ` first?

digama0: note that succ and +1 are reducibly defeq, so this would trigger either way as a simp LHS.

semorrison: Okay!

Still I'm a bit hesitant that this belongs in this file. I feel like by the time I would ever want this lemma I might just have `ring` available anyway.

digama0: I can't imagine where else it would go. I certainly use theorems like this routinely, and of course people using std without mathlib won't have `ring` available.

fgdorais: This was already there under the name `succ_mul_succ_eq`. I don't think it's used anywhere so I'm fine with deleting it if necessary.

semorrison:
(Joe asked that I merge this after reviewing; I'm happy now, so about to hit merge.)

PR #200, opened 2023-07-25
feat: new alias command

fgdorais:
WIP - help welcome!

Implementation of the revised alias command discussed on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/alias.20command.20ordering/near/378351642).

Supersedes #184


digama0:
I think you should rebase this on top of #184's commits, so it is easier to review what changed and ensure that modifications from that PR are reflected here (which does seem to be the case)

Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,108 @@
...
+  /- alias doesn't trigger the missing docs linter so we add a default -/
+  if (← findDocString? (← getEnv) declName).isNone then
+    if mp then
+      addDocString declName s!"**Alias** for the forward direction of {thm.name}"
+    else
+      addDocString declName s!"**Alias** for the reverse direction of {thm.name}"
+
+@[inherit_doc «alias»]
+elab (name := aliasLR) mods:declModifiers "alias "
+    "⟨" aliasFwd:binderIdent ", " aliasRev:binderIdent "⟩" " := " name:ident : command =>
```
fgdorais, line 98: Maybe we can use parentheses here instead of angles? That would make the notation just ASCII and more accessible for [blind users](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.7C-/near/370481492).

digama0: No, I think it should be angle brackets for consistency with anonymous constructors. If you want to add an ASCII alternative I encourage you to take it up with core first, std will try to stay consistent with whatever is done upstream.

digama0: A possible ASCII alternative would be `{ mp := aliasFwd, mpr := aliasRev ]`

fgdorais:
The original Mathlib alias had a comment to add an `alias` attribute. I'm not sure what that attribute was intended to do other than "I'm an alias of". I think even that basic functionality makes some sense but is it needed?

digama0:
The attribute was intended to be usable so that third party code could programmatically determine that a declaration is an alias of another.

fgdorais:
Should that be implemented as an attribute or as an environment extension? I'm thinking the latter since I don't see much need for `@[alias ...]`.

fgdorais:
Yes, it supports all modifiers and attributes the same way as `def` and `theorem` do!

alexjbest:
> Should that be implemented as an attribute or as an environment extension? I'm thinking the latter since I don't see much need for `@[alias ...]`.

At some point I wrote a linter for duplicate lemmas and the attribute was useful as a way of skipping intentional duplicates. The interaction with `to_additive`  was a little subtle here and it was important for that application that `to_additive` could copy the alias attribute to an additive version of an existing alias.

Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,165 @@
...
+      | Lean.ConstantInfo.inductInfo c
+      | Lean.ConstantInfo.axiomInfo c
+      | Lean.ConstantInfo.opaqueInfo c
+      | Lean.ConstantInfo.ctorInfo c
+      | Lean.ConstantInfo.recInfo c =>
+        .defnDecl { c with
+          name := declName
+          value := mkConst resolved (c.levelParams.map mkLevelParam)
+          hints := .regular 0 -- FIXME
+          safety := if declMods.isUnsafe then .unsafe else .safe
```
digama0, line 91:
```suggestion
          safety := if declMods.isUnsafe then .unsafe else c.safety
```
I don't think it should be necessary to use `unsafe alias` to alias an unsafe declaration, it can just infer the right thing.

Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,165 @@
...
+        }
+    checkNotAlreadyDeclared declName
+    if declMods.isNoncomputable then
```
digama0, line 94:
```suggestion
    if declMods.isNoncomputable || const.isNoncomputable then
```
same deal here, we can infer `noncomputable`. (this function doesn't exist though, some research is needed to determine how to check if a constant has compiled code)

Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,165 @@
...
+      selectionRange := ← getDeclarationRange alias
+    }
+    addDocString' declName declMods.docString?
+    Term.applyAttributes declName declMods.attrs
+    let info := match ← getAliasInfo const.name with
+      | some i => i
+      | none => AliasInfo.plain const.name
+    setAliasInfo info declName
+    /- alias doesn't trigger the missing docs linter so we add a default -/
+    if (← findDocString? (← getEnv) declName).isNone then
```
digama0, line 109:
```suggestion
    if declMods.docString?.isNone then
```
Seems a bit roundabout when we just set the docstring.

Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,165 @@
...
+      addDocString declName info.toString
```
digama0, line 110:
```suggestion
      let mut doc := info.toString
      if let some origDoc ← findDocString? (← getEnv) resolved then
        doc := s!"{doc}\n\n---\n\n{origDoc}"
      addDocString declName doc
```
The alias documentation is nice, but as a user I usually want to see the original docs instead.

Std/Tactic/Alias.lean
```diff
@@ -0,0 +1,165 @@
...
+    }
+  addDocString' declName declMods.docString?
+  Term.applyAttributes declName declMods.attrs
+  let info := match ← getAliasInfo thm.name with
+    | some (.plain name) => if mp then AliasInfo.forward name else AliasInfo.reverse name
+    | _ => if mp then AliasInfo.forward thm.name else AliasInfo.reverse thm.name
+  setAliasInfo info declName
+  /- alias doesn't trigger the missing docs linter so we add a default -/
+  if (← findDocString? (← getEnv) declName).isNone then
+    addDocString declName info.toString
```
digama0, line 141: ditto

test/alias.lean
```diff
@@ -0,0 +1,95 @@
...
+set_option linter.unusedVariables false
+
+open Lean Meta
+namespace A
+
+/-- doc string for foo -/
+theorem foo : 1 + 1 = 2 := rfl
+
+/-- doc string for `alias foo` -/
+alias foo1 := foo
```
digama0, line 13: hovering `foo1` does not show the type, unlike in `def foo1 := 1`. I think you are missing a `ofTermInfo` info tree node placed over the declaration span

Std/Tactic/Alias.lean
digama0: Oh nvm, this is about `inherit_doc` handling, isn't it.

fgdorais: Yes it is.

Std/Tactic/Alias.lean
fgdorais: My thought process here and below was to use the same behavior as `def` so that users "instinctively" know how to use `alias` but I'm happy to change that.

Std/Tactic/Alias.lean
fgdorais: I'm a bit puzzled here: How would a user know this is an alias? I don't understand how they would come to want to see the original docs. Of course, if you happen to know it's an alias you can just use the original if that's what you want, no?

In any case, I'm happy to include the original docs. What should I do if there's an `inherit_doc`? Should I dig to see if it's an alias of an alias, or should I just pile on the docs?

test/alias.lean
fgdorais: Ah! That's what was missing! Thanks!

Std/Tactic/Alias.lean
digama0: When you make an alias to a definition, a user of that alias is generally using it because the new name is "better". It is very much deliberate that there is no linter to warn if you use an alias by its new name - if you want this effect you use `@[deprecated]` instead. Mathlib has plenty of examples of this, for example `lt_trans` is aliased to `LT.lt.trans` so that people can write `h1.trans h2` instead of `lt_trans h1 h2`.

When you hover over `.trans` there, it is useful to report that this is an alias of `lt_trans`, but if `lt_trans` itself was complicated and confusing and had a big docstring explaining what it means, I think a user would want to see that docstring too on `.trans`.

> In any case, I'm happy to include the original docs. What should I do if there's an inherit_doc? Should I dig to see if it's an alias of an alias, or should I just pile on the docs?

Aliases of aliases should already be resolved. That is, `alias B := A; alias C := B` should have the same effect as `alias B := A; alias C := A`, the logic of which is at least implemented in the `aliasExt`. This does have a slightly unfortunate effect if `B` has a custom docstring since `C` will skip it and show A's docstring instead, but docstrings on aliases are rare anyway. Either `inherit_doc` or a manual docstring should cancel all automated docstring behavior and just show the docstring that was requested; this is included in the code snippet I gave since it is inside the `if docs.isNone then` branch.

Std/Tactic/Alias.lean
digama0: the idea is that `alias A := B` should always "just work" for a wide variety of choices of `B`, this is one of the advantages of `alias` over `def` (you can get alias-like behavior from def using `def A := @B`).

Std/Tactic/Alias.lean
fgdorais: That clears up my confusion. Thanks!

Note that `aliasExt` does implement that logic but it is not used to resolve the alias yet. Should I implement that as well?

test/alias.lean
fgdorais: I just realized that I don't know the proper way to do this. Any pointers?

Std/Tactic/Alias.lean
fgdorais: I'll fix the resolution step to use `aliasExt` and get to the root alias.

test/alias.lean
fgdorais: Found it! Weird day... it turns out it was at the *first* place I looked but somehow completely missed it. 🤣

PR #201, opened 2023-07-28
chore: bump to nightly-2023-07-24

semorrison:


eric-wieser:
LGTM, but I don't think I have bors powers here.

digama0:
std doesn't use bors (yet?).

PR #202, opened 2023-07-28
chore: bump to nightly-2023-07-25

semorrison:
The extra rewrites required now are due to https://github.com/leanprover/lean4/pull/2317


semorrison:
There will be lots of these rewrite changes needed in Mathlib, I starting making a few in https://github.com/leanprover-community/mathlib4/pull/6201, but I'm not too happy that we have to do this.

PR #203, opened 2023-07-28
feat: more sub lemmas for `Nat`

fgdorais:
- [x] Mathlib patch: https://github.com/leanprover-community/mathlib4/pull/6216

There's a bunch of deprecated lemmas, all of which have better alternatives except for one which appears to be a single purpose lemma that lost its purpose.


semorrison:
@YaelDillies, do you have an opinion here, or are able to review?

fgdorais:
This is the last in a series. They're mostly independent but it might be easier to start with #194.

YaelDillies (commented):
I'm happy with the deprecations. However I find it a bit sad to duplicate all this `tsub` material in Std. Can we do some subset of the following suggestions?
* Not duplicate `tsub` lemma in Std
* Put in the docstring of duplicate `Nat.sub_something` that `tsub_something` exists in mathlib
* Deprecate mathlib-wise (is that currently possible?) each `Nat.sub_something` in favor of `tsub_something`
* Put a big warning in `Std.Data.Nat.Lemmas` that if people want a lot of theory about truncated subtraction, then they should use `Mathlib.Algebra.Order.Sub.Canonical` instead of PRing more lemmas here

digama0:
I'd rather we did none of the above. The tsub lemmas are more complex to understand, and elaborate. I think there should be no particular preference in mathlib for one over the other. Certainly there won't be a warning about not adding any more lemmas on Nat: tsub is only relevant if you care about other types with truncated subtraction like NNReal.

YaelDillies:
If we had to only do one of my suggestions, I would like it to be
> * Put in the docstring of duplicate `Nat.sub_something` that `tsub_something` exists in mathlib
because I want mathlib users to be directed towards the API that's useful to them, avoiding ad hoc Std lemmas.

fgdorais:
@YaelDillies Note that `Nat` lemmas in Std are primarily to reason about sizes of objects, array indices and such. There's no intent to duplicate or clash with Mathlib. My impression is that most users would use these lemmas for their intended purpose, so I don't quite see the need for the first three suggestions.

However, I am happy with the fourth suggestion:

* Put a big warning in Std.Data.Nat.Lemmas that if people want a lot of theory about truncated subtraction, then they should use Mathlib.Algebra.Order.Sub.Canonical instead of PRing more lemmas here

I will add a module doc to clarify the intended purpose of these lemmas and a pointer to Mathlib alternatives.

I would like to hear what others think before implementing the first three suggestions.

fgdorais:
@YaelDillies Please see #237.

semorrison:
I agree that module-docs are the way to go to explain the additional theory available in Mathlib.

Any results that are needed or helpful for non-mathematical uses of Nat need to be here in Std, without expecting such users to know about the general theory of truncated subtraction.

I am ambivalent about the "Mathlib-specific deprecation" suggestion, but don't think we should mention `tsub` in Std.

@fgdorais, as far as I'm concerned I think you can mark this back as `awaiting-review`.

Std/Data/Nat/Lemmas.lean
```diff
@@ -991,3 +1002,36 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
...
+@[deprecated Nat.add_le_of_le_sub']
+protected theorem add_le_of_le_sub_left {n k m : Nat} (h : m ≤ k) : n ≤ k - m → m + n ≤ k :=
+  Nat.add_le_of_le_sub' h
+
+@[deprecated Nat.sub_lt_of_pos_le]
+protected theorem sub_lt_self (h₀ : 0 < a) (h₁ : a ≤ b) : b - a < b :=
+  Nat.sub_lt_of_pos_le h₀ h₁
+
+@[deprecated]
+theorem sub_lt_succ (a b) : a - b < succ a :=
```
digama0, line 1032: Why is this deprecated?

Std/Data/Nat/Lemmas.lean
```diff
@@ -991,3 +1002,36 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
...
+  lt_succ_of_le (sub_le a b)
+
+@[deprecated]
+theorem sub_one_sub_lt (h : i < n) : n - 1 - i < n := by
```
digama0, line 1036: Why is this deprecated?

digama0 (approved):
Approved with modification:

* `succ_sub_one`, `sub_lt_self`, `sub_one_sub_lt` were undeprecated
* deprecated theorems were moved after their respective undeprecated versions instead of in a separate section (if we want this it should be implemented in docgen)

Std/Data/Nat/Lemmas.lean
fgdorais: This theorem didn't appear to have any uses. I marked it as deprecated mostly so I would notice if it came back. It hasn't reappeared for a while now, so I think it's safe to delete.

Std/Data/Nat/Lemmas.lean
fgdorais: Same.

Std/Data/Nat/Lemmas.lean
digama0: I mean, it seems like a reasonable enough theorem, I don't see any reason to delete it

Std/Data/Nat/Lemmas.lean
digama0: This one looks relevant for indexing into arrays

Std/Data/Nat/Lemmas.lean
fgdorais: OK. I'm probably sweeping too hard.

Std/Data/Nat/Lemmas.lean
fgdorais: Agreed. I suspect it has a better name somewhere else but I never tried looking for it.

Std/Data/Nat/Lemmas.lean
YaelDillies: It is the composition of two simpler lemmas: `sub_le_self` and `lt_succ`.

digama0: It is, but certain simple combinations of lemmas are nevertheless useful, especially ones that might end up in theorem statements like this one (because it shows up in `Fin (succ n)` terms).

YaelDillies: Sure yeah

joehendrix: Seeing the head pattern `x < succ y`  seems suspect to me.  Why not make `Nat.lt_succ` be a simp rule and then convert this to `Nat.sub_le_self` (which doesn't seem defined on `Nat`)?

semorrison (approved):
Looks good to me. @digama0 has already approved, and @fgdorais has resolved the merge conflicts, so I hope we can merge this round!!

PR #204, opened 2023-07-29
New conv tactic: equals

nomeata:
Given a goal `a * b`, I find myself wanting for a concise way of specifying the position of `b` (without repeating it), saying that I want to prove it to be equal to `c`, then doing the proof of `b = c`, and end up with `a = c`. Of course the terms can be larger and more deeply nested, possibly in binders.

So here is a tactic that does that.

I think it is a very natural way of interactive proving. Stare at the goal, notice a subexpression where you can make progress, focus, say what you want to rewrite it to, prove the rewriting, all with DRY.
It’s basically the less pretty; less verbose style version of a nice `calc` proof, without the repetition and the `:= by congr 3; …` before the actual proof steps.
Or the equivalent of the pen-and-paper style of putting curly braces under some subexpression to state that it’s equal to something else.

I also found it useful in a lemma dealing with series. If I try to do it with `calc` and `∑' _ = _,` then I have all the `Summable` side conditions, and have to show some transformations twice. But if I state the result using `HasSum _ _` the summability is part of the goal, and I can use this tactic to freely shuffle things under the sum, or in the value of the sum, without dealing with side conditions.

I went back and forth between `by` and `=>` (like with `tactic =>`) half a dozen times. Right now it’s using `=>`.

See discussion in https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Define.20RHS.20in.20conv.20mode/near/379304747

### DIFF
--- a/Std/Tactic/Basic.lean
+++ b/Std/Tactic/Basic.lean
@@ -146,11 +146,11 @@ macro (name := Conv.exact) "exact " t:term : conv => `(conv| tactic => exact $t)
  to the given expression, and proves this claim using the given tactic.
 ```
 example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
-  conv in (_ - _) => equals 0 by
+  conv in (_ - _) => equals 0 =>
     -- current goal: ⊢ n - n = 0
     apply Nat.sub_self
   -- current goal: P (fun n => 0)
 ```
 -/
-macro (name := Conv.equals) "equals " t:term " by " tac:tacticSeq : conv =>
-  `(conv| tactic => next => show (_ = $t); ($tac))
+macro (name := Conv.equals) "equals " t:term " => " tac:tacticSeq : conv =>
+  `(conv| tactic => show (_ = $t); next => $tac)
--- a/Std/Tactic/Basic.lean
+++ b/Std/Tactic/Basic.lean
@@ -153,4 +153,4 @@ example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
 ```
 -/
 macro (name := Conv.equals) "equals " t:term " by " tac:tacticSeq : conv =>
- `(conv|tactic => show (_ = $t); ($tac) )
+  `(conv| tactic => next => show (_ = $t); ($tac))
--- a/Std/Tactic/Basic.lean
+++ b/Std/Tactic/Basic.lean
@@ -141,3 +141,16 @@ macro (name := triv) "triv" : tactic =>
 
 /-- `conv` tactic to close a goal using an equality theorem. -/
 macro (name := Conv.exact) "exact " t:term : conv => `(conv| tactic => exact $t)
+
+/-- The `conv` tactic `equals` claims that the currently focused subexpression is equal
+ to the given expression, and proves this claim using the given tactic.
+```
+example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
+  conv in (_ - _) => equals 0 by
+    -- current goal: ⊢ n - n = 0
+    apply Nat.sub_self
+  -- current goal: P (fun n => 0)
+```
+-/
+macro (name := Conv.equals) "equals " t:term " by " tac:tacticSeq : conv =>
+ `(conv|tactic => show (_ = $t); ($tac) )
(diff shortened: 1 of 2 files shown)


Std/Tactic/Basic.lean
```diff
@@ -141,3 +141,16 @@ macro (name := triv) "triv" : tactic =>
...
+```
+example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
+  conv in (_ - _) => equals 0 by
+    -- current goal: ⊢ n - n = 0
+    apply Nat.sub_self
+  -- current goal: P (fun n => 0)
+```
+-/
+macro (name := Conv.equals) "equals " t:term " by " tac:tacticSeq : conv =>
+ `(conv|tactic => show (_ = $t); ($tac) )
```
kmill, line 156: I think we want to be sure the tactic sequence closes the goal, otherwise it might be confusing to a user of this that in `conv`-mode they might get non-`conv` goals.
```suggestion
  `(conv| tactic => next => show (_ = $t); ($tac))
```

Regarding the syntax, using `by` like this is unique. Usually it's `=>` to indicate a tactic sequence (examples: `next => ...`, `case t => ...`, `induction` match arms, `conv => ...`, `focus => ...`, etc.), and there are some exceptions like `repeat` or `classical` that don't use an indicator at all. I think we should use `=>` for consistency.

In addition to this, we could have the following variant:
```lean
macro (name := Conv.equals') "equals " t:term " := " v:term : conv =>
  `(conv| tactic => exact ($v : _ = $t))
```
and you could write `equals 0 := by apply Nat.sub_self` if you'd rather see `by`. Having such a variant is consistent with tactics like `have`.

nomeata: I like that variant, even more composable. Will update the PR later (and also look for how to write a few small tests for this.)

nomeata: Hmm, that doesnt' work so nicely; it seems one then has to indent the proof a lot, like this:
```
example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
  conv in (_ - _) => equals 0 := by
                      -- current goal: ⊢ n - n = 0
                      apply Nat.sub
```
This makes logical sense, but isn't very helpful here.

kmill: I think you might have to temper expectations about doing things on one line here. At least there's
```lean
example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
  conv in (_ - _) =>
    equals 0 := by
      apply Nat.sub
```
or
```
example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
  conv in (_ - _) =>
  equals 0 := by
  apply Nat.sub
```

nomeata: Maybe … but with `" => " tacticSeq` as it stands right now, the indentation is more flexible, so I might have a slight preference for this variant after all (and two character less to type)

test/conv_equals.lean
```diff
@@ -0,0 +1,41 @@
...
+import Std.Tactic.GuardMsgs
+
+
+-- The example from the doc string
+
+/-- warning: declaration uses 'sorry' -/
+#guard_msgs in
+example (P : (Nat → Nat) → Prop) : P (fun n ↦ n - n) := by
+  conv in (_ - _) => equals 0 =>
+    -- current goal: ⊢ n - n = 0
```
kmill, line 17: You might need to import `Std.Tactic.GuardExpr`, but then you can do
```suggestion
    guard_target = n - n = 0
```
and before your sorry, `guard_target = P (fun n => 0)`

nomeata: Added! I left the comments in so that one can easily copy the test into the docstring.

kmill (approved):
It seems good to me. I'm not a std4 maintainer, but I can at least approve this PR.

alexjbest:
It might be nice to unify the name of this conv tactic, and `convert_to` in mathlib? It seems they do sortof the same thing?

kmill:
@alexjbest The relationship seems to be that `convert_to e; ...` is roughly `conv => equals e => congr!; ...`. The `convert_to` tactic I'd say is mostly about how it applies `congr!` after setting up this equality.

Maybe there could be a `convert_to e => ...` conv tactic (it takes a tactic block since it wouldn't create conv goals but rather normal tactic goals). Then `equals e => ...` is to `convert_to e => ...` as `change e` is to `convert_to e` as `exact x` is to `convert x`.

PR #206, opened 2023-08-03
chore: lower simp priority for List.mem_map

semorrison:
This is a slightly ugly simplification, and lowering its priority allows nicer lemmas like 
```
theorem mem_map_of_injective {f : α → β} (H : Injective f) {a : α} {l : List α} :
    f a ∈ map f l ↔ a ∈ l
```
to fire first, when `Injective f` is available to the simplifier.

### DIFF
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -156,7 +156,7 @@ theorem mem_iff_append {a : α} {l : List α} : a ∈ l ↔ ∃ s t : List α, l
 
 theorem map_singleton (f : α → β) (a : α) : map f [a] = [f a] := rfl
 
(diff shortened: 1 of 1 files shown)


digama0:
any particular reason you want to handle this by lowering the priority of `mem_map` instead of raising the priority of `mem_map_of_injective`?

semorrison:
My thinking was that `mem_map` is "ugly", i.e. its not particularly obvious the RHS is actually simpler than the LHS (it involves an existential, and it has "changed API level"). Thus it should be a low priority lemma.

In the other direction, there is nothing special about `mem_map_of_injective` that makes it look better than your average lemma.

digama0:
> My thinking was that `mem_map` is "ugly", i.e. its not particularly obvious the RHS is actually simpler than the LHS (it involves an existential, and it has "changed API level"). Thus it should be a low priority lemma.

I'm not sure I agree. We have `mem_image`, don't we? The definition is an existential, and the unfolding is "triggered" by using it in a `x \in` expression. This is just the list version of that lemma.

PR #208, opened 2023-08-07
chore: union_def and inter_def lemmas

semorrison:
The main purpose here is to provide `union_def` and `inter_def`, so no one has to call simp with the auto-generated instance name. (Supporting https://github.com/leanprover/lean4/issues/2343.)

I've also restated the lemmas about inter and union in terms of the notation, but I'm happy to roll that part back if preferred.

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -424,6 +424,8 @@ will be retained (but order will otherwise be preserved).
 
 instance [DecidableEq α] : Union (List α) := ⟨List.union⟩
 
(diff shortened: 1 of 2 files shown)


Std/Data/List/Basic.lean
```diff
@@ -424,6 +424,8 @@ will be retained (but order will otherwise be preserved).
 
 instance [DecidableEq α] : Union (List α) := ⟨List.union⟩
 
+theorem union_def [DecidableEq α] (l₁ l₂ : List α)  : l₁ ∪ l₂ = foldr .insert l₂ l₁ := rfl
```
digama0, line 427: Move it to `Lemmas`, `Basic` is only for definitions and lemmas required for definitions

semorrison: Done.

PR #210, opened 2023-08-08
feat: upstream addHaveSuggestion and addRewriteSuggestion

semorrison:
Upstreams two more "Try this:" builders.


Std/Tactic/TryThis.lean
```diff
@@ -147,3 +148,52 @@ def addTermSuggestion (ref : Syntax) (e : Expr)
...
+    let loc ← loc?.mapM fun loc => do `(location| at $(← delab loc):term)
+    `(tactic| rw [$rules_stx,*] $(loc)?)
+
+  let mut tacMsg :=
+    let rulesMsg := MessageData.sbracket <| MessageData.joinSep
+      (rules.map fun ⟨e, symm⟩ => (if symm then "← " else "") ++ m!"{e}") ", "
+    if let some loc := loc? then
+      m!"rw {rulesMsg} at {loc}"
+    else
+      m!"rw {rulesMsg}"
```
semorrison, line 193:
```suggestion
  let mut tacMsg := m!"{tac}"
```

digama0: The reason for duplicating the structure is so that the message data contains embedded `Expr`s instead of just `Syntax`. (Yes, this is a hack and ideally there would be some better way to attach elaboration info to `Syntax` in a `MessageData`, but alas.)

semorrison: Thanks. I've revert this change and added a comment explaining why we're rebuilding the message.

fgdorais (approved):

semorrison:
WIP

semorrison:
awaiting-review

Std/Tactic/TryThis.lean
```diff
@@ -147,3 +148,57 @@ def addTermSuggestion (ref : Syntax) (e : Expr)
...
+    let t ← delabToRefinableSyntax e
+    if symm then `(rwRule| ← $t:term) else `(rwRule| $t:term)
+  let tac ← do
+    let loc ← loc?.mapM fun loc => do `(location| at $(← delab loc):term)
+    `(tactic| rw [$rules_stx,*] $(loc)?)
+
+  -- We don't simply write `let mut tacMsg := m!"{tac}"` here
+  -- but instead rebuild it, so that there are embedded `Expr`s in the message,
+  -- thus giving more information in the hovers.
+  -- Perhaps in future we will have a better way to attached elaboration information to
```
semorrison, line 190:
```suggestion
  -- Perhaps in future we will have a better way to attach elaboration information to
```

Std/Lean/Syntax.lean
```diff
@@ -0,0 +1,29 @@
...
+/-!
+# Helper functions for working with typed syntaxes.
+-/
+
+namespace Lean
+
+/--
+Applies the given function to every subsyntax.
+
+Like `Syntax.replaceM` but for typed syntax.
```
digama0, line 17: This should mention that it doesn't ensure type-correctness

semorrison:
```suggestion
Like `Syntax.replaceM` but for typed syntax. (Note there are no guarantees of type correctness here.)
```

Std/Tactic/TryThis.lean
```diff
@@ -523,3 +524,57 @@ def addTermSuggestions (ref : Syntax) (es : Array Expr)
...
+def addHaveSuggestion (ref : Syntax) (t? : Option Expr) (e : Expr)
+  (origSpan? : Option Syntax := none) : TermElabM Unit := do
+  let estx ← delabToRefinableSyntax e
+  let prop ← isProp (← inferType e)
+  let tac ← if let some t := t? then
+    let tstx ← delabToRefinableSyntax t
+    if prop then
+      `(tactic| have : $tstx := $estx)
+    else
+      `(tactic| let this : $tstx := $estx)
```
digama0, line 540: does this put `this†` in the output?

semorrison: Indeed, it does, and there's a FIXME in Mathlib's `test/propose.lean` about this.

digama0: it should use ``$(mkIdent `this)`` instead

semorrison: Okay, I've fixed the behaviour of Mathlib's `have?` and `observe?` in https://github.com/leanprover-community/mathlib4/pull/9454, and will modify this PR accordingly.

joehendrix (approved):
I don't have context of how these are used in Mathlib, but the operations themselves seem useful.

PR #211, opened 2023-08-08
chore: upstream pickle

semorrison:


fgdorais (approved):
I've been thinking about doing this one, so I think this is a good idea! The API from Mathlib is clean and clear, no need for additional changes from my perspective.

PR #212, opened 2023-08-08
chore: upstream MLList.whileAtLeastHeartbeatsPercent

semorrison:
I appreciate this may seem slightly esoteric, but it is used in `apply?`, to prefer giving a "best available" answer to the user in situations where `apply?` runs for a long time, rather than crashing out with a timeout.

### DIFF
--- a/Std/Data/MLList/Heartbeats.lean
+++ b/Std/Data/MLList/Heartbeats.lean
@@ -1,0 +1,24 @@
+/-
+Copyright (c) 2023 Scott Morrison. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Scott Morrison
+-/
+import Std.Data.MLList.Basic
+import Std.Lean.CoreM
+
+/-!
+# Truncate a `MLList` when running out of available heartbeats.
+-/
+
+open Lean.Core (CoreM)
+
+/-- Take an initial segment of a monadic lazy list,
+trying to leave at least `percent` of the remaining allowed heartbeats. -/
+def MLList.whileAtLeastHeartbeatsPercent [Monad m] [MonadLiftT CoreM m]
+    (L : MLList m α) (percent : Nat := 10) : MLList m α :=
+  MLList.squash fun _ => do
+    if (← getMaxHeartbeats) = 0 then do
+      return L
+    let initialHeartbeats ← getRemainingHeartbeats
+    pure <| L.takeWhileM fun _ => do
+      return .up <| (← getRemainingHeartbeats) * 100 / initialHeartbeats > percent
--- a/Std/Data/MLList/Heartbeats.lean
+++ b/Std/Data/MLList/Heartbeats.lean
@@ -13,12 +13,20 @@ import Std.Lean.CoreM
 open Lean.Core (CoreM)
 
 /-- Take an initial segment of a monadic lazy list,
-trying to leave at least `percent` of the remaining allowed heartbeats. -/
+stopping when there is less than `percent` of the remaining allowed heartbeats.
+
+If `getMaxHeartbeats` returns `0`, then this passes through the original list unmodified.
+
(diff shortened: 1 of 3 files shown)


Std/Data/MLList/Heartbeats.lean
```diff
@@ -0,0 +1,24 @@
...
+import Std.Lean.CoreM
+
+/-!
+# Truncate a `MLList` when running out of available heartbeats.
+-/
+
+open Lean.Core (CoreM)
+
+/-- Take an initial segment of a monadic lazy list,
+trying to leave at least `percent` of the remaining allowed heartbeats. -/
```
digama0, line 16: This description is not so good. When is the heartbeat check run, and what will be in the list? The doc should give a precise specification without terms like "trying to".

semorrison: Excellent criticism! The code from Mathlib was actually discarding the last produced element if it went over the time limit, for no good reason.

I've changed it to use `takeUpToFirstM` rather than `takeWhileM`, and expanded the doc-string.

Std/Lean/CoreM.lean
```diff
@@ -0,0 +1,43 @@
...
+
+open Lean
+
+/-- Count the number of heartbeats used during a monadic function. -/
+-- See also `Lean.withSeconds`
+def Lean.withHeartbeats [Monad m] [MonadLiftT BaseIO m] (x : m α) : m (α × Nat) := do
+  let start ← IO.getNumHeartbeats
+  let r ← x
+  let finish ← IO.getNumHeartbeats
+  return (r, (finish - start) / 1000)
```
digama0, line 20: The division by 1000 seems inappropriate here

semorrison: Ah, okay. There's this annoying question of at what point to divide by 1000. Since even after dividing by 1000 its reasonably fine grained (i.e. any tactic invocation however small will take at least 10s?), and there is potential confusion between the user facing "kiloheartbeats" (as used in options) and the internal heartbeats, my instinct had been to minimise(?) confusion by dividing as soon as possible.

semorrison, line 20:
```suggestion
  return (r, finish - start)
```

Std/Lean/CoreM.lean
```diff
@@ -0,0 +1,43 @@
...
+-/
+import Lean.CoreM
+
+/-!
+# Additional functions using `CoreM` state.
+-/
+
+open Lean
+
+/-- Count the number of heartbeats used during a monadic function. -/
```
semorrison, line 14:
```suggestion
/--
Count the number of heartbeats used during a monadic function.

Remember that user facing heartbeats (e.g. as used in `set_option maxHeartbeats`) differ from the internally tracked heartbeats by a factor of 1000, so you need to divide the results here by 1000 before comparing with user facing numbers.
-/
```

Std/Lean/CoreM.lean
```diff
@@ -0,0 +1,47 @@
...
+/-!
+# Additional functions using `CoreM` state.
+-/
+
+open Lean
+
+/--
+Count the number of heartbeats used during a monadic function.
+
+Remember that user facing heartbeats (e.g. as used in `set_option maxHeartbeats`) differ from the internally tracked heartbeats by a factor of 1000, so you need to divide the results here by 1000 before comparing with user facing numbers.
```
semorrison, line 17:
```suggestion
Remember that user facing heartbeats (e.g. as used in `set_option maxHeartbeats`) 
differ from the internally tracked heartbeats by a factor of 1000, 
so you need to divide the results here by 1000 before comparing with user facing numbers.
```

Std/Lean/CoreM.lean
digama0: My rationale is that there is nothing about the name of this function that suggests it would be in different units than `getNumHeartbeats`. It's in moon units anyway, I don't see why not to use it as is.

PR #213, opened 2023-08-08
chore: upstream the once-per-file Cache for tactics

semorrison:
This is another component of `exact?` that will need to be upstreamed.

### DIFF
--- a/Std/Util/Cache.lean
+++ b/Std/Util/Cache.lean
@@ -1,0 +1,179 @@
+/-
+Copyright (c) 2021 Gabriel Ebner. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Gabriel Ebner
+-/
+import Std.Lean.Meta.DiscrTree
+
+/-!
+# Once-per-file cache for tactics
+
+This file defines cache data structures for tactics
+that are initialized the first time they are accessed.
+Since Lean 4 starts one process per file,
+these caches are once-per-file
+and can for example be used to cache information
+about the imported modules.
+
+The `Cache α` data structure is
+the most generic version we define.
+It is created using `Cache.mk f`
+where `f : MetaM α` performs
+the initialization of the cache:
+```
+initialize numberOfImports : Cache Nat ← Cache.mk do
+  (← getEnv).imports.size
+
+-- (does not work in the same module where the cache is defined)
+#eval show MetaM Nat from numberOfImports.get
+```
+
+The `DeclCache α` data structure computes
+a fold over the environment's constants:
+`DeclCache.mk empty f` constructs such a cache
+where `empty : α` and `f : Name → ConstantInfo → α → MetaM α`.
+The result of the constants in the imports is cached
+between tactic invocations,
+while for constants defined in the same file
+`f` is evaluated again every time.
+This kind of cache can be used e.g.
+to populate discrimination trees.
+-/
+
+open Lean Meta
+
+namespace Mathlib.Tactic
+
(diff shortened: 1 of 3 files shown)


Std/Util/Cache.lean
```diff
@@ -0,0 +1,179 @@
...
+open Lean Meta
+
+namespace Mathlib.Tactic
+
+/-- Once-per-file cache. -/
+def Cache (α : Type) :=
+  IO.Ref <| Sum (MetaM α) <|
+    Task <| Except Exception α
+
+instance : Nonempty (Cache α) :=
```
fgdorais, line 52: Using `Nonempty` instead of `Inhabited` here flags a probably unnecessary use of choice somewhere. Using `Nonempty` is useful to avoid diamonds but I'm skeptical it's necessary here.

fgdorais (commented):
Using choice is mostly harmless but it is best to avoid it since Std is intended for computation. In core, it's mostly used to get excluded middle, which only affects `Prop` so it's essentially harmless for computation. The use here is in an actual type so potentially not so harmless.

Std/Util/Cache.lean
fgdorais: Actually, it doesn't appear to be used except for deriving `Nonempty` at `DeclCache`. Might as well remove both and wait to see if an application needs it.

semorrison: I don't understand, sorry. We certainly need either `instance (α : Type) : Nonempty (CachedData α)` or `instance (α : Type) : Inhabited (CachedData α)` downstream, in order to be able to call `initialize`.

Either work, so if you would like to change this to `Inhabited` please do so, but I don't see any reason to.

semorrison: Note
```
#synth Nonempty (IO.Ref Unit)
```
points to [`ST.instNonemptyRef`](https://github.com/leanprover/lean4/blob/master/src/Init/System/ST.lean#L52). I'm not sure where you would get an `Inhabited` here except by using choice.

fgdorais: Using `Inhabited` is better if you need to reason about it but `Nonempty` is better to avoid diamond issues. I think the former is better unless a diamond is expected to happen. Preemptively avoiding diamonds is a Mathlib thing which is not that necessary for Std.

fgdorais: If `Inhabited` is not an option, as it seems to be the case here, please add a comment to that effect.

semorrison: Added comment.

fgdorais (approved):

nomeata:
Great! Another dependency of `#find` in std.

I had a PR against that file in mathlib4, at https://github.com/leanprover-community/mathlib4/pull/7384. Should I port this against std4 now, or is that PR unwanted anyways?

Std/Util/Cache.lean
```diff
@@ -0,0 +1,160 @@
...
+`DeclCache.mk empty f` constructs such a cache
+where `empty : α` and `f : Name → ConstantInfo → α → MetaM α`.
+The result of the constants in the imports is cached between tactic invocations,
+while for constants defined in the same file `f` is evaluated again every time.
+This kind of cache can be used e.g. to populate discrimination trees.
+-/
+
+open Lean Meta
+
+namespace Mathlib.Tactic
```
nomeata, line 36: Did you intentionally keep using the `Mathlib` namespace even in std, @semorrison ?

semorrison: Uh, no, sorry, this was an oversight.

PR #215, opened 2023-08-10
feat: Try these

thorimur:
This PR introduces `Try these:` functionality to the `tryThis` widget and associated definitions, such as `addSuggestions`, which renders an array of replacement suggestions as a bulleted list in the infoview preceded by "Try these:".
An image of a bulleted list of tactic replacements in the infoview.

Currently, tactics in Mathlib which return many different "Try this:" suggestions, such as `apply?` and `rw?`, work by creating a separate "Try this:" widget for each individual suggestion. This takes up a lot of space in the infoview—each suggestion includes its own "Tactic replacement" summary element and the words `Try this:`.

This was discussed briefly and requested as a std4 PR [on zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60tryThese.60.3F).

This PR also provides both widgets with configurable headers—the text `Try these:` and `Try this:` can now be modified at will. (This is motivated by upcoming tactics like `hint`, which may want different lists for different things, e.g. a list of tactics which close the goal, and a list of tactics which don't.)

This PR also modifies the form of the code actions. Currently, tryThis code actions all read simply `"Apply 'Try this'"`. We change this to the more informative `"Try this: {replacementText}"`. This is necessary in order to have distinguishable code actions now that we're handling a *list* of different possible replacements instead of just one.

Finally, this PR also allows for a bare `String` (instead of `TSyntax`) to be used as a suggested replacement via `SuggestionText`, and provides 6 basic styling options for each entry via `SuggestionStyle`: `.error`, `.warning`, `.success`; `.asHypothesis`, `.asInaccessible`; and `.value (t : Float)`, which draws from a red-yellow-green color gradient with red at `0.0` and green at `1.0`.

-----

I'm not 100% sure these are the "right" HTML classes to be using to style the text in a code-like fashion, even though they look alright so far.

### DIFF
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -172,16 +172,63 @@ def addSuggestion (ref : Syntax) {kind : Name} (suggestion : TSyntax kind)
     let json := Json.mkObj [("suggestion", text), ("range", toJson range), ("info", extraMsg)]
     Widget.saveWidgetInfo ``tryThisWidget json (.ofRange stxRange)
 
-/-- Add a `exact e` or `refine e` suggestion.
+/-- Holds a `suggestion` for replacement, along with an `info` string to be printed immediately
+after that suggestion and optional `MessageData` to represent the suggestion in logs. -/
+structure Suggestion (kind : Name) where
+  /-- Syntax to be used as a replacement via a code action. -/
+  suggestion : TSyntax kind
+  /-- Info to be printed immediately after replacement syntax. -/
+  info : String := ""
+  /-- How to represent the suggestion as `MessageData`. This is used only in the info diagnostic.
+  If `none`, we use `suggestion`. -/
+  messageData : Option MessageData := none
+
+instance {kind : Name} : Coe (TSyntax kind) (Suggestion kind) where
+  coe t := { suggestion := t }
+
+/-- Add a list of "try this" suggestions as a single "try these" suggestion. This has three effects:
+
+* An info diagnostic is displayed saying `Try these: <list of suggestions>`
+* A widget is registered, saying `Try these: <list of suggestions>` with a link on each
+  `<suggestion>` to apply the suggestion
+* A code action for each suggestion is added, which will apply the suggestion.
 
 The parameters are:
 * `ref`: the span of the info diagnostic
-* `e`: the replacement expression
+* `suggestions`: an array of `Suggestion`s, which each contain
+  * `suggestion`: the replacement syntax
+  * `info`: a string shown immediately after the replacement syntax in the infoview
+* `suggestionsForMessage?`: the messages to display in the info diagnostic (only).
+  The widget message uses only `suggestion`. If empty or too short, `suggestion` is used in both
+  places.
 * `origSpan?`: a syntax object whose span is the actual text to be replaced by `suggestion`.
   If not provided it defaults to `ref`.
+* `header`: a string that precedes the list. By default, it is `"Try these:"`.
 -/
-def addExactSuggestion (ref : Syntax) (e : Expr)
-    (origSpan? : Option Syntax := none) (addSubgoalsMsg := false) : TermElabM Unit := do
+def addSuggestions (ref : Syntax) {kind : Name} (suggestions : Array (Suggestion kind))
+    (origSpan? : Option Syntax := none)
+    (header : String := "Try these:") : MetaM Unit := do
+  let msgs := suggestions.map fun s => s.messageData.getD s.suggestion
+  let msgs := msgs.foldl (init := MessageData.nil) (fun msg m => msg ++ m!"\n• " ++ m)
+  logInfoAt ref m!"{header}{msgs}"
+  if let some range := (origSpan?.getD ref).getRange? then
+    let map ← getFileMap
+    let start := findLineStart map.source range.start
+    let body := map.source.findAux (· ≠ ' ') range.start start
+    let suggestions ← suggestions.mapM fun ⟨suggestion, info, _⟩ => do
+      let text ← PrettyPrinter.ppCategory kind suggestion
+      let text := Format.prettyExtra text
+        (indent := (body - start).1) (column := (range.start - start).1)
+      pure <| Json.mkObj [("suggestion", text), ("info", info)]
+    let stxRange := ref.getRange?.getD range
+    let stxRange :=
+    { start := map.lineStart (map.toPosition stxRange.start).line
+      stop := map.lineStart ((map.toPosition stxRange.stop).line + 1) }
+    let range := map.utf8RangeToLspRange range
+    let json := Json.mkObj [("suggestions", Json.arr suggestions), ("range", toJson range),
+      ("header", header)]
+    Widget.saveWidgetInfo ``tryTheseWidget json (.ofRange stxRange)
+
   let stx ← delabToRefinableSyntax e
   let mvars ← getMVars e
   let tac ← if mvars.isEmpty then `(tactic| exact $stx) else `(tactic| refine $stx)
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -152,7 +152,8 @@ apply the replacement.
       let .ok newText := s.getObjValAs? String "suggestion" | panic! "bad type"
       result.push {
         eager.title := "Try this: " ++ newText
-        eager.kind? := "refactor"
+        eager.kind? := "quickfix"
+        eager.isPreferred? := true
         eager.edit? := some <| .ofTextEdit params.textDocument.uri { range, newText }
       }
 
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -282,4 +282,17 @@ The parameters are:
 def addTermSuggestion (ref : Syntax) (e : Expr)
     (origSpan? : Option Syntax := none) : TermElabM Unit := do
   addSuggestion ref (← delabToRefinableSyntax e)
-    (suggestionForMessage? := e) (origSpan? := origSpan?)
+
+/-- Add term suggestions.
+
+The parameters are:
+* `ref`: the span of the info diagnostic
+* `es`: an array of the replacement expressions
+* `origSpan?`: a syntax object whose span is the actual text to be replaced by `suggestion`.
+  If not provided it defaults to `ref`.
+* `header`: a string which precedes the list of suggestions. By default, it's `"Try these:"`.
+-/
+def addTermSuggestions (ref : Syntax) (es : Array Expr)
+    (origSpan? : Option Syntax := none) (header : String := "Try these:") : TermElabM Unit := do
+  addSuggestions ref (← es.mapM (β := Suggestion `term) (delabToRefinableSyntax ·))
+    (origSpan? := origSpan?) (header := header)
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -115,7 +115,7 @@ export default function(props) {
   }
 
   // Choose between an inline 'Try this'-like display and a list-based 'Try these'-like display.
-  if (props.isSingular) {
+  if (props.isInline) {
     return e('div', {className: 'ml1'},
       e('pre', {className: 'font-code pre-wrap'},
         props.header,
@@ -345,11 +345,11 @@ def delabToRefinableSuggestion (e : Expr) : TermElabM Suggestion :=
 
 /-! # Widget hooks -/
 
-/-- Core of `addSuggestion` and `addSuggestions`. Whether we use a `Try This:` display or a
-`Try These:` display is controlled by `isSingular`. -/
+/-- Core of `addSuggestion` and `addSuggestions`. Whether we use an inline display for a single
+element or a list display is controlled by `isInline`. -/
 private def addSuggestionCore (ref : Syntax) (suggestions : Array Suggestion)
     (origSpan? : Option Syntax := none)
-    (header : String) (isSingular : Bool) : MetaM Unit := do
+    (header : String) (isInline : Bool) : MetaM Unit := do
   if let some range := (origSpan?.getD ref).getRange? then
     let map ← getFileMap
     let (indent, column) := getIndentAndColumn map range
@@ -363,7 +363,7 @@ private def addSuggestionCore (ref : Syntax) (suggestions : Array Suggestion)
       ("suggestions", Json.arr suggestions),
       ("range", toJson range),
       ("header", header),
-      ("isSingular", toJson isSingular)]
+      ("isInline", toJson isInline)]
     Widget.saveWidgetInfo ``tryThisWidget json (.ofRange stxRange)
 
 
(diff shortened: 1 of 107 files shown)


semorrison:
The [`LLMStep`](https://github.com/wellecks/llmstep) project has implemented something very similar, see https://github.com/wellecks/llmstep/blob/master/LLMstep/LLMstep.lean#L94 and also my PR refactoring it a bit https://github.com/wellecks/llmstep/pull/4.

Can we support their use case?

They are doing something a bit extra, which is actually running the tactics against the goal and putting in 🎉  emojis where appropriate.

thorimur:
@semorrison I think so! There are two pieces of functionality I see that we'd need to supply here:
* ability to customize the classname/highlight of the clickable link (to some extent; iirc there are specific highlight colors allowed and a type somewhere encoding them)
* ability to prepend text (e.g. 🎉)

If these both become fields of the `Suggestion` structure, then the running of tactics against the goal (and sorting, etc.) can all be done before feeding the array of `Suggestion`s to `addSuggestions`. I think that might be a good pattern in general: put all of the customization data in the `Suggestion`s, then feed them to `addSuggestion`(`s`).

There are two ways I can think of for handling the "bad suggestions".

1. I think there's a case to be made that we should only ever insert valid `Syntax`, and thus only `Syntax` should be clickable. (But maybe I'm not anticipating some cases...) This would mean that the bad suggestion strings should be relegated to the `info` field(s) when constructing the `Suggestion`s. We'd abstain from displaying syntax by either:
  a. making the `suggestion` field `Option` al
  b. asking the LLMStep project to take the `suggestion` field to be `⟨Syntax.missing⟩`, and special-casing missing syntax to simply not display (instead of displaying as `<missing>`)
2. Have the `suggestion` field be upgraded from`TSyntax kind` to `SuggestionText` or `SuggestionText kind`, which would more or less just be a descriptive version of `TSyntax kind ⊕ String` (with appropriate coercions from the two ponents, each of which would be handled appropriately, etc.). `kind` could either be a parameter to `SuggestionText` or an implicit argument to one of the constructors of `SuggestionText` (in which case it wouldn't need to be a parameter to `Suggestion` any longer, and suggestion arrays could be heterogeneously in their syntax kinds). (I tried the latter out just to make sure it's viable and it seems so.)

I'll think about this a bit, but if you have an opinion on which of these options would be the best way to handle the invalid suggestions, please let me know! :)

semorrison:
The reason there are non-parseable suggestions in the llmstep example is that the suggestions are coming from a language model, which sometimes has no idea what it is doing! :-) Nevertheless, the thinking is that sometimes there might be the germ of a good idea, which the human could still make use of. So for their application, they do want to be able to show plain text suggestions (whether or not they are clickable is probably less important: users will use these ones less often, and perhaps can be expected to copy and paste).

The wrapper for `TSyntax kind ⊕ String` seems the right way to go.

And yes, absolutely `addSuggestions` itself should not be doing any of the preparation of the `Suggestion`, just displaying it.

thorimur:
@semorrison Ok, I think that this should work! :) Note that while the LLMStep project were altering the `className` to e.g. make text red, this makes the links behaves strangely on mouseover (namely, they flash back to blue, then dim (while staying blue). Instead, I think providing a `style` is the way to go—so, there's now a `className? : Option String` and `style? : Option Json`, so that we can say something like `style? := Json.mkObj [("color","red")]`.

Std/Tactic/TryThis.lean
```diff
@@ -75,65 +263,140 @@ partial def replaceMVarsByUnderscores [Monad m] [MonadQuotation m]
 def delabToRefinableSyntax (e : Expr) : TermElabM Term :=
   return ⟨← replaceMVarsByUnderscores (← delab e)⟩
 
+/-- Delaborate `e` into a suggestion suitable for use in `refine`. -/
```
semorrison, line 266:
```suggestion
/-- Delaborate `e` into a suggestion suitable for use in `refine`. -/
```
"in `refine`" seems wrong here?

semorrison:
Could we have a test file? Even if CI can't check anything, I think we should have something committed that makes it easy for a human to verify reasonable behaviour.

semorrison:
Could `addSuggestion` be made to just call `addSuggestions` with a singleton array, to reduce duplication?

semorrison:
For some context here, implicit in this PR is the idea that we will also later have something like `analyzeSuggestions: Array String \to Array Suggestion`, which will:
* try to parse the Strings to Syntax
* run all suggestions against the goal, add put  🎉 in the `preInfo` of any suggestions that close the goal
* sort the suggestions so the ones that close come first, and the ones that don't even parse come last

This, and some slight variants, allow both a nice implementation of the `hint` tactic, and all the user interaction required for something like [`llmstep`](https://github.com/wellecks/llmstep).

digama0:
I am a bit concerned about feature creep and over-parameterizing this widget, which is mainly intended as a shortcut for people who don't want all the customization. LLMStep should probably just be using a custom widget if they want to set custom style options on text and such.

thorimur:
> For some context here, implicit in this PR is the idea that we will also later have something like `analyzeSuggestions: Array String \to Array Suggestion`

I was wondering if it would be worth upstreaming something like what I saw in LLMStep! :) I assumed it'd be in a separate PR to keep review easy, but give me a thumbs up and I'll start a dependent PR... :)

Std/Tactic/TryThis.lean
thorimur: You're right—this is vestigial copypasta from the docstring above it that I forgot to review. :) I'll figure out an appropriate rephrasing tomorrow.

semorrison:
@digama0, the thing that made me less concerned about the over-parametrization issue is that this PR was originally motivated (I think!) by the `hint` tactic --- which wants much the same thing: reporting a list of tactics, some of which close the goal and some of which don't.

The main potential over-parametrization here is accepting `String` suggestions as alternative to `Syntax`, which does seem pretty specific to an LLM. However at that point it's either a big duplication between the two widgets, or pushing the generality here.

semorrison:
> > For some context here, implicit in this PR is the idea that we will also later have something like `analyzeSuggestions: Array String \to Array Suggestion`
> 
> I was wondering if it would be worth upstreaming something like what I saw in LLMStep! :) I assumed it'd be in a separate PR to keep review easy; give me a thumbs up and I'll start a dependent PR... :)

Let's wait until the dust settles here. :-)

digama0:
> E.g., when suggesting tactics, we could color by how many subgoals remain, or by how long it takes to elaborate—or whatever's relevant for the particular application. Yes, you can write something in the info, but colors are immediate, informative, and, well, nice! :)

I think that's fine, *if* the coloring is mostly automatic and enabled by default. If the user is setting RGB values rather than just passing some kind of "emphasize this" or "highlight 20%" style hint, then it's too low level.

semorrison:
@thorimur, what is the status of this one? It is still marked WIP. I would like to see if we can merge this before LLMStep is forgotten. :-)

thorimur:
> @thorimur, what is the status of this one? It is still marked WIP. I would like to see if we can merge this before LLMStep is forgotten. :-)

@semorrison Still WIP! :) Sorry, it's been on my mind, but I just haven't had a whole lot of time. Now that I'm more or less finished with the `refine` bug, I've got two things in my personal queue—this and `fail_if_no_progress`. I'll prioritize this. :)

thorimur:
Ok, this PR now
* refactors everything into a single widget
* gives you a choice of 6 simple styles: `error`, `warning`, and `success`; `asHypothesis`, `asInaccessible` (in case you're suggesting hypotheses); and `value (t : Float)`, where `t` picks a color from a red-yellow-green gradient as it ranges from `0.0` to `1.0`.

thorimur:
awaiting-review

Std/Tactic/TryThis.lean
```diff
@@ -75,65 +339,135 @@ partial def replaceMVarsByUnderscores [Monad m] [MonadQuotation m]
...
+def delabToRefinableSuggestion (e : Expr) : TermElabM Suggestion :=
+  return { suggestion := ← delabToRefinableSyntax e, messageData? := e }
+
+/-! # Widget hooks -/
+
+/-- Core of `addSuggestion` and `addSuggestions`. Whether we use a `Try This:` display or a
+`Try These:` display is controlled by `isSingular`. -/
```
semorrison, line 349: Couldn't `isSingular` just be replaced by `suggestions.size = 1`?

thorimur: Not without changing the styling: that would change a list with one element to a non-list. `addSuggestions` with one suggestion would then produce
```
Try these: rfl
```
instead of

`Try these:`

* `rfl`

The first option looks weird—at least to me—but the second doesn't.

thorimur: (Renamed `isSingular` to `isInline` in 0023326, because however this resolves, that's a better name. :) )

semorrison: Okay, I understand now. That's good as is, and the rename is helpful!

test/tryThis.lean
```diff
@@ -0,0 +1,170 @@
...
+```
+Try these:
+• rfl -- red
+• rfl -- red-orange
+• rfl -- orange
+• rfl -- yellow
+• rfl -- yellow-green
+• rfl -- light green
+• rfl -- green
+```-/
```
digama0, line 92:
````suggestion
```
-/
````
Never put the triple-backtick on the same line as the comment end, it messes up syntax highlighting in github

File: test/tryThis.lean [hunk 4]
```diff
@@ -0,0 +1,170 @@
...
+-- `Try this: rfl` -- Lean green
+#demo1 {s with style? := some .success}
+
+-- `Try this: rfl` -- styled like a goal hypothesis
+#demo1 {s with style? := some .asHypothesis}
+
+-- `Try this: rfl` -- styled like an inaccessible goal hypothesis
+#demo1 {s with style? := some .asInaccessible}
+
+/--info: Try this: rfl-/
```
digama0, line 116: use the code action to generate the message, it uses better formatting

thorimur: Do you mean by moving the `logInfoAt` to inside the `tryThisProvider`? Or something else?

(There would be a couple issues with moving `logInfoAt`, but mainly, I don't think we have access to the `MessageData` at that point, just the string—and apparently we want to be able to specify and use custom `MessageData`. Not exactly sure what the use case is, though.)

Std/Tactic/TryThis.lean
thorimur: Oops, this one fell through the cracks. I changed the preposition to `by` in 830fc8c, but I've stared at it too long for me to tell if that sounds right...

test/tryThis.lean
digama0: I mean delete the docstring on the test and then use `ctrl-.` to make the code action put the docstring back

thorimur:
(Apologies for the force push, accidentally pushed some clutter.)

test/tryThis.lean
thorimur: Ok, done in efd52f6.

Std/Tactic/TryThis.lean
```diff
@@ -56,14 +144,189 @@ apply the replacement.
...
-      eager.edit? := some <| .ofTextEdit params.textDocument.uri { range, newText }
-    }
+    let .ok suggestions := props.getObjVal? "suggestions" | panic! "bad type"
+    let .ok suggestions := suggestions.getArr? | panic! "bad type"
+    suggestions.foldlM (init := result) fun result s => do
+      let .ok newText := s.getObjValAs? String "suggestion" | panic! "bad type"
+      result.push {
+        eager.title := "Try this: " ++ newText
+        eager.kind? := "quickfix"
+        eager.isPreferred? := true
```
thorimur, line 155: I wanted to check that it makes sense to have `eager.isPreferred? := true` for each suggestion; I tried changing it, and couldn't see any effect. (`Cmd+.``return` still worked fine.)

Std/Tactic/TryThis.lean
```diff
@@ -72,69 +335,139 @@ partial def replaceMVarsByUnderscores [Monad m] [MonadQuotation m]
...
 
+/-- Delaborate `e` into a suggestion suitable for use by `refine`. -/
+def delabToRefinableSuggestion (e : Expr) : TermElabM Suggestion :=
+  return { suggestion := ← delabToRefinableSyntax e, messageData? := e }
+
+/-! # Widget hooks -/
+
+/-- Core of `addSuggestion` and `addSuggestions`. Whether we use an inline display for a single
+element or a list display is controlled by `isInline`. -/
+private def addSuggestionCore (ref : Syntax) (suggestions : Array Suggestion)
```
thorimur, line 350: Should `addSuggestions` fail on an empty `suggestions` array, or should we leave that case to be handled by whoever's using it?

Std/Tactic/TryThis.lean
digama0: I believe if you set `isPreferred? := true`, the lightbulb turns big and blue, and you can trigger it with a single key command (`shift-alt-.` for me), which doesn't even pull up the menu and just applies it directly.

thorimur: Ah, got it—it's `Cmd+option+.` for me. Looks like if there are multiple preferred options, it pulls up the menu instead of inserting it immediately; I'd best change this.

thorimur: Ok, fixed in 4723c3b.

Std/Tactic/TryThis.lean
semorrison: I'm inclined to fail. Otherwise we'll be silently pretending to succeed in situations that should be failing.

thorimur: Ok, fixed in d9abcff. (I only touched `addSuggestions`, since we want to error before we log any info.)

semorrison:
@digama0, what do you think of this PR at this point? It would be nice to get this in so that both LeanInfer and LLMStep can adopt it, rather than having a version in each repository.

I'd also like it to restore the `hint` tactic.

Std/Tactic/TryThis.lean
```diff
@@ -20,28 +20,116 @@ namespace Std.Tactic.TryThis
...
 const e = React.createElement;
 export default function(props) {
   const editorConnection = React.useContext(EditorContext)
-  function onClick() {
-    editorConnection.api.applyEdit({
-      changes: { [props.pos.uri]: [{ range: props.range, newText: props.suggestion }] }
-    })
+
+  // Construct the style from the object corresponding to a given SuggestionStyle.
+  function getStyle(style) {
```
digama0, line 52: This is a lot of JS code. Can we upstream this to the vscode extension? (Or just not do this...)

digama0 (approved):
Apologies for not giving this a proper review for so long. I've made several changes, most notably to remove all of the `getStyle` logic and just let users pass in a json object with the style they want (and retaining the existing styles, as presets). Anyway it is now ready to merge IMO, but I will give @joehendrix @thorimur @semorrison time to speak up since I made some significant modifications.

semorrison:
For ease of reviewing, Mario's changes are:
* https://github.com/leanprover/std4/pull/215/commits/def3f72867537fc93f9b2836bba241acca8affbc and
* https://github.com/leanprover/std4/compare/f9d368d..0d39f21

thorimur:
@digama0 I quite like the refactor you've made to the style logic! I was also wary of all that JS and reluctant to constrain the user to my choices. This is much better. (And I don't think we had the `json%` elaborator in std when I started, so that's nice too!) The default style setting is a sensible addition as well.

I went through the code and only found an incorrect test description from myself, which I fixed. LGTM. :)

digama0:
> (And I don't think we had the json% elaborator in std when I started, so that's nice too!)

There's a reason it was merged just a few minutes ago...

PR #216, opened 2023-08-12
feat: some lemmas for get!

mo271:


mo271:
I'm not sure if
 1. it is at all desirable for those `get!` lemmas to be in std
 2. if this is the right place for it

digama0:
I think there should only be one simp lemma, `l.get! n = l.getD n default`

mo271:
> I think there should only be one simp lemma, `l.get! n = l.getD n default`

Not entirely sure if you meant "there should be only one lemma, and it should be simp" or
"of of all the lemma, only one should be simp".
In case it is the former, I'll delete everything but the `get!_eq_getD`

Also: right now there is still a simpNF error.

Std/Data/List/Lemmas.lean
```diff
@@ -524,10 +524,25 @@ theorem get?_eq_get : ∀ {l : List α} {n} h, l.get? n = some (get l ⟨n, h⟩
   | _ :: _, 0, _ => rfl
   | _ :: l, _+1, _ => get?_eq_get (l := l) _
 
+theorem get!_eq_get : ∀ {l : List α} {n} h, l.get? n = some (get l ⟨n, h⟩)
```
fgdorais, line 527: In this case, the explicit parameter `h` should have a type to guide usage. The type of `h` can be inferred by Lean but a human reading the code could use the hint that `h : n < l.length`.

fgdorais (changes requested):

fgdorais:
> Also: right now there is still a simpNF error.

That is a weird error involving a private lemma. Likely a bug. Can you have a look @digama0?

digama0:
I believe it is saying that `get!` already has a simp lemma, most likely the block at the top of `Std.Data.List.Init.Lemmas`

fgdorais:
@digama0 is recommending deleting `get!` from this line: https://github.com/leanprover/std4/blob/dbffa8cb31b0c51b151453c4ff8f00ede2a84ed8/Std/Data/List/Init/Lemmas.lean#L18C22-L18C26

I also think this is the right move but there might be additional problems with that change. Please try it and let us know if there are complications. I am happy to help resolve these.

Std/Data/List/Lemmas.lean
mo271: done, also in the lemma above for consistency

mo271:
That seems to have worked, @fgdorais

fgdorais (commented):
I think this looks great. Do you have time to add the companion `getD` lemmas?

mo271:
I'll give it a try

mo271:
Just to be on the same page, what companion `getD` lemmas were you thinking of exactly?
I noticed that there are already some `getD` lemmas in mathlib:
https://github.com/leanprover-community/mathlib4/blob/b5ece30947b3cd725c5ffa958481e36684395661/Mathlib/Data/List/Basic.lean#L4332-L4365

and `getD_eq_get?` is already in std.

mo271:
I'm happy to help, if I can. What is the plan? Move everything in `section getD` to std?
There might be a reason that it is in mathlib now.
In the spirit of atomic changes, it might be a good idea to do the changes to `getD` in a different pull request.

fgdorais:
All right! These are indeed the options I was considering. Let's consult Mario before going further. If the goal is to get it 99.99% right the first time rather than making incremental changes then we definitely need some input! Of course, like I emphasized earlier (but in different words) we can always fall back on the incremental route. In either case, I think Mario's input would be valuable.

@digama0 We would appreciate your opinion here. It makes sense to me to move all the `List.get{?,!,D}` API to Std. That would make simps easier to manage. In particular, we could make `get!_eq_getD` the only simp-outlet out of `get!`, as you recommended earlier. You may have other ideas that we haven't considered yet.

fgdorais:
I think we can move forward and fix it later on if necessary. Let's move the Mathlib `getD` section here. It's really the same as `get!`, with a potentially different choice of default value, so it makes perfect sense to handle them together. 

Can you move these lemmas here? I will then handle the Mathlib patch. I've done a bunch of these lately so it's easy for me to handle that part.

semorrison:
@mo271, I've added `awaiting-author`, as I think you're planning to do more work here. Please set back to `awaiting-review` (and if you don't have permission to do so, please ping on the zulip `Std` channel asking for `triage` permissions).

mo271:
Thanks! I thought we were still waiting for input from @digama0. 
In any case, I can prepare a patch moving the `getD` section. 
I guess it is best to keep the aligns in the mathlib file. 
Also I still see no real benefit (and the drawback of making things more complicated) of relating this change (which does not even require a std bump in mathlib) together with the `getD` refactor, but perhaps there is some simp-lint issue that I'm not seeing.

Std/Data/List/Lemmas.lean
```diff
@@ -705,6 +720,11 @@ theorem getD_eq_get? : ∀ l n (a : α), getD l n a = (get? l n).getD a
   | _a::_, 0, _ => rfl
   | _::l, _+1, _ => getD_eq_get? (l := l) ..
 
+@[simp] theorem get!_eq_getD [Inhabited α] : ∀ (l : List α) n, l.get! n = l.getD n default
+| [], _        => rfl
+| _a::_, 0     => rfl
+| _a::l, n+1   => get!_eq_getD l n
```
digama0, line 726:
```suggestion
  | [], _      => rfl
  | _a::_, 0   => rfl
  | _a::l, n+1 => get!_eq_getD l n
```

digama0:
Let's do the getD stuff in a separate PR. I do wonder about the usefulness of these theorems though if `get!` is unconditionally not a simp-normal form, but they seem harmless enough.

PR #218, opened 2023-08-15
feat: add `absurd` tactic

fgdorais:
Handy little tactic to "smartly" apply `absurd`.

### DIFF
--- a/Std/Tactic/Basic.lean
+++ b/Std/Tactic/Basic.lean
@@ -70,6 +70,13 @@ macro (name := byContra) tk:"by_contra" e?:(ppSpace colGt binderIdent)? : tactic
     | refine Classical.byContradiction fun $e => ?_)
 
 /--
+Given a proof `h` of `p`, `absurd h` changes the goal to `⊢ ¬ p`.
+If `p` is a negation `¬q` then the goal is changed to `⊢ q` instead.
+-/
+macro "absurd " h:term : tactic =>
+  `(tactic| first | apply absurd _ $h | apply absurd $h)
+
+/--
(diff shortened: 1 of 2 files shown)


test/absurd.lean
```diff
@@ -0,0 +1,48 @@
...
+
+-- Basic example
+/--
+error: unsolved goals
+p : Prop
+h : p
+⊢ ¬p
+-/
+#guard_msgs in
+example {p : Prop} (h : p) : False := by
```
digama0, line 14: you should use goals that are not just `False` in the tests

fgdorais: Any preference? Random terms? Or `sorry`?

digama0: Just anything else - you can't tell from this test whether the tactic requires that the goal is `False`

Std/Tactic/Basic.lean
```diff
@@ -69,6 +69,13 @@ macro (name := byContra) tk:"by_contra" e?:(ppSpace colGt binderIdent)? : tactic
     | refine Decidable.byContradiction fun $e => ?_
     | refine Classical.byContradiction fun $e => ?_)
 
+/--
+Given a proof `h` of `p`, `absurd h` changes the goal to `⊢ ¬ p`.
+If `p` is a negation `¬q` then the goal is changed to `⊢ q` instead.
+-/
+macro "absurd " h:term : tactic =>
+  `(tactic| first | apply absurd _ $h | apply absurd $h)
```
digama0, line 77:
```suggestion
  `(tactic| first | refine absurd ?_ $h | refine absurd $h ?_)
```
better not to use `apply` if you know the arity

PR #219, opened 2023-08-15
feat: add String to Substring coercion

fgdorais:
Split from #178

### DIFF
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -470,7 +470,7 @@ theorem split_of_valid (s p) : split s p = (List.splitOnP p s.1).map mk := by
 
 -- TODO: splitOn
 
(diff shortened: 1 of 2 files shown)


Std/Data/String/Lemmas.lean
```diff
@@ -470,7 +470,7 @@ theorem split_of_valid (s p) : split s p = (List.splitOnP p s.1).map mk := by
 
 -- TODO: splitOn
 
-@[simp] theorem toString_toSubstring (s : String) : s.toSubstring.toString = s :=
+@[simp] theorem toString_toSubstring (s : String) : Substring.toString s = s :=
```
digama0, line 473: this one should stay as is, otherwise it reads too confusingly

fgdorais:
I'll do a "quick" check to see if anything breaks in Mathlib later today.

fgdorais:
No serious issues at Mathlib, just one place where the coercion caused a type ambiguity.

PR #220, opened 2023-08-18
feat: upstream String.count

semorrison:


semorrison:
I have some files that only depend on Mathlib through this definition... Time to upstream!

fgdorais (approved):

PR #222, opened 2023-08-20
feat: upstream `compile_time_search_path%` term elaborator

semorrison:
We define `ToExpr SearchPath` by hand, as we have decided not to upstream the `ToExpr` derive handler for now.

### DIFF
--- a/Std/Lean/Util/Path.lean
+++ b/Std/Lean/Util/Path.lean
@@ -3,7 +3,7 @@ Copyright (c) 2022 Gabriel Ebner. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Gabriel Ebner
 -/
-import Lean
+import Lean.Elab.Term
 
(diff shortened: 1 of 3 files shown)


Std/Lean/Util/Path.lean
```diff
@@ -0,0 +1,19 @@
...
+
+/--
+Term elaborator that retrieves the current `SearchPath`.
+
+This must not be used in files that are potentially compiled on another machine and then
+imported.
+(That is, if used in an imported file it will embed the search path from whichever machine
+compiled the `.olean`.)
+-/
+elab "compile_time_searchPath%" : term =>
```
digama0, line 18:
```suggestion
elab "compile_time_search_path%" : term =>
```
This is an interesting test of the naming convention. This is a keyword (commands / attributes / tactic names are also in this category), not a theorem name, and AFAIK these are strictly snake case.

PR #225, opened 2023-08-21
fix: merge conflicts bot on forks

fgdorais:
Since forks have a read-only token, the bot fails every time it is run on a fork. Setting `continueOnMissingPermissions` makes that failure quiet.

### DIFF
--- a/.github/workflows/merge_conflicts.yml
+++ b/.github/workflows/merge_conflicts.yml
@@ -13,3 +13,4 @@ jobs:
         with:
           dirtyLabel: "merge-conflict"
           repoToken: "${{ secrets.MERGE_CONFLICTS_TOKEN }}"
+          continueOnMissingPermissions: true


semorrison:
Ah, that's unfortunate. I guess we could give the merge conflict bot maintainer permissions? (As most people will leave the box checked that gives maintainers the ability to write to PR branches even from forks.)

@digama0, what is your preference here?

digama0:
I don't really understand the alternatives being considered. It seems fine if the merge conflict bot doesn't run on forks, what does it need to do on a fork?

fgdorais:
It doesn't need to run on forks at all. Right now, it runs every 60 minutes and sends a failure report every time. That option is designed to let the job fail quietly on forks. (Basically the same as not running at all.) It doesn't affect behavior on the main repo since that token does have valid permissions.

digama0:
SGTM, I will let @semorrison confirm this since he set up the bot.

semorrison:
🤦‍♀️ I misunderstood! I thought this was affecting PRs coming from forks, rather than PRs to forks!

Yes, let's merge!

PR #228, opened 2023-08-24
feat: `alias` sets the deprecated target

digama0:
This makes `@[deprecated] alias bar := foo` set the deprecation target, as if you wrote `@[deprecated foo] alias bar := foo`. This way, `@[deprecated] alias bar := foo` can be used as the standard idiom for introducing a rename suggestion into downstream code.

### DIFF
--- a/test/alias.lean
+++ b/test/alias.lean
@@ -12,12 +12,13 @@ theorem foo : 1 + 1 = 2 := rfl
 /-- doc string for `alias foo` -/
 alias foo1 := foo
 @[deprecated] alias foo2 := foo
-alias _root_.B.foo3 := foo
+@[deprecated foo2] alias _root_.B.foo3 := foo
 
 example : 1 + 1 = 2 := foo1
 /-- warning: `A.foo2` has been deprecated, use `A.foo` instead -/
 #guard_msgs in example : 1 + 1 = 2 := foo2
-example : 1 + 1 = 2 := B.foo3
+/-- warning: `B.foo3` has been deprecated, use `A.foo2` instead -/
+#guard_msgs in example : 1 + 1 = 2 := B.foo3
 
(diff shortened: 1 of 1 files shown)


Std/Tactic/Alias.lean
```diff
@@ -70,10 +79,11 @@ elab (name := alias) mods:declModifiers "alias " alias:ident " := " name:ident :
...
     let const ← getConstInfo resolved
     let declMods ← elabModifiers mods
-    let declMods ← pure { declMods with
-        isNoncomputable := declMods.isNoncomputable || isNoncomputable (← getEnv) const.name
-        isUnsafe := declMods.isUnsafe || const.isUnsafe
-      }
+    let declMods := { declMods with
+      isNoncomputable := declMods.isNoncomputable || isNoncomputable (← getEnv) const.name
+      isUnsafe := declMods.isUnsafe || const.isUnsafe
+      attrs := setDeprecatedTarget resolved declMods.attrs
```
fgdorais, line 85: `resolved` is used here whereas `const.name` is used two lines before. They must be equal but it's better to use just one to avoid confusion. I have no particular preference which but I guess I did prefer `const.name` when I wrote the previous line :-)

fgdorais (changes requested):

test/alias.lean
```diff
@@ -11,11 +11,12 @@ theorem foo : 1 + 1 = 2 := rfl
 
 /-- doc string for `alias foo` -/
 alias foo1 := foo
-alias foo2 := foo
+@[deprecated] alias foo2 := foo
```
fgdorais, line 14: Test the iff case too?

fgdorais (approved):
Thanks for the extra cleanup! LGTM

PR #230, opened 2023-08-24
feat: ignore rhs of dischargers when linting for unreachable

alexjbest:
Before this change the linter would complain that `assumption` was unused
```lean
def t : ℕ → ℕ := sorry
@[simp]
lemma a : aa = 0 → t aa = 0 := sorry
lemma aaa (ha : aa = 0) : t aa = 0 := by
  simp (disch := assumption)
``` 

Also have `guardMsgs` run linters so that we can test them

### DIFF
--- a/Std/Linter/UnreachableTactic.lean
+++ b/Std/Linter/UnreachableTactic.lean
@@ -44,6 +44,7 @@ initialize ignoreTacticKindsRef : IO.Ref NameHashSet ←
     |>.insert ``Lean.Parser.Tactic.tacticStop_
     |>.insert ``Lean.Parser.Command.notation
     |>.insert ``Lean.Parser.Command.mixfix
+    |>.insert ``Lean.Parser.Tactic.discharger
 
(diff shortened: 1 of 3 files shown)


Std/Linter/UnreachableTactic.lean
```diff
@@ -58,6 +58,8 @@ This should be called from an `initialize` block.
 def addIgnoreTacticKind (kind : SyntaxNodeKind) : IO Unit :=
   ignoreTacticKindsRef.modify (·.insert kind)
 
+initialize Std.Linter.UnreachableTactic.addIgnoreTacticKind `Lean.Parser.Tactic.discharger
```
digama0, line 61: If it's in the same file, you may as well add it to the `HashSet` directly (line 46). Also always use double-backtick if possible.

alexjbest: Oh of course, thanks :), done

digama0:
could you add a test?

PR #231, opened 2023-08-25
feat: code action for alias + deprecated

digama0:
Now, when you declare a deprecated alias such as `@[deprecated] alias foo := bar`, occurrences of `foo` (which will be marked with a deprecation warning) will also have an auto-fix command to replace `foo` with `bar`.

### DIFF
--- a/Std/CodeAction/Deprecated.lean
+++ b/Std/CodeAction/Deprecated.lean
@@ -1,0 +1,69 @@
+/-
+Copyright (c) 2023 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro
+-/
+import Lean.Server.CodeActions
+import Std.CodeAction.Basic
+import Std.Lean.Position
+
+/-!
+# Code action for @[deprecated] replacements
+
+This is an opt-in mechanism for making machine-applicable `@[deprecated]` definitions. When enabled
+(by setting the `machineApplicableDeprecated` tag attribute), a code action will be triggered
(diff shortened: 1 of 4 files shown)


fgdorais:
Really cool! Is there a way to add tests for this? I'm really curious how it works for the iff case.

digama0:
> I'm really curious how it works for the iff case.

It doesn't. The code action is actually quite conservative: it does not fire on `@[deprecated] def` e.g. as might be used in core, only on `@[deprecated] alias foo := bar` where we can be sure that `bar` is a drop in replacement for `foo`. For anything more complicated we simply don't show the code action.

fgdorais:
Sounds good. I'll give it a test drive tomorrow.

Std/CodeAction/Deprecated.lean
```diff
@@ -0,0 +1,69 @@
...
+-/
+import Lean.Server.CodeActions
+import Std.CodeAction.Basic
+import Std.Lean.Position
+
+/-!
+# Code action for @[deprecated] replacements
+
+This is an opt-in mechanism for making machine-applicable `@[deprecated]` definitions. When enabled
+(by setting the `machineApplicableDeprecated` tag attribute), a code action will be triggered
```
semorrison, line 14: Could you say what command the user is actually meant to type?

digama0: As far as this file is concerned, it makes no user-visible syntax. You just call `machineApplicableDeprecated.tag` directly (and this is public API, available for other tactic writers). If one includes the `Alias.lean` file, the user syntax is `@[deprecated] alias foo := bar`.

PR #234, opened 2023-08-26
feat: change `extBinder` syntax to allow destructuring

kmill:
This changes the `extBinder` syntax to allow arbitrary terms as binders, to be able to do destructuring like in `fun` binders. To accomplish this, the `binderIdent` is switched to `term:max`. This creates an ambiguity between `extBinder` and `extBinderCollection` however since terms can start with `(`, so to break this ambiguity if a binder starts with `(` it is assumed to be the start of an `extBinderCollection`.

For the extensions to `∃` and `∀` as well as the `∃ᵉ` and `∀ᵉ` notation, we need to process `ident`/`hole` vs general terms ourselves, partly because `∃` and `∀` expect a `binderIdent`, and partly because we need to control elaboration order so that, for example, `∃ ⟨a⟩ ≤ n, a = 1` can successfully elaborate. The issue is that we need the expansion of the binder predicate to occur ahead of making sense of `⟨a⟩`, which requires an expected type.

This PR will support destructuring in notations defined by `notation3` in mathlib, which has been a requested feature.

### DIFF
--- a/Std/Util/ExtendedBinder.lean
+++ b/Std/Util/ExtendedBinder.lean
@@ -28,26 +28,28 @@ syntax "satisfies_binder_pred% " term:max binderPred : term
 
 /--
 The notation `∃ x < 2, p x` is shorthand for `∃ x, x < 2 ∧ p x`,
-and similarly for other binary operators.
+and similarly for other binary relations defined as being binder predicates.
 -/
-syntax "∃ " binderIdent binderPred ", " term : term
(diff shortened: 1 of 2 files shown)


kmill:
https://github.com/leanprover-community/mathlib4/pull/6795 is using these changes to implement destructuring in big sum & product notation.

kmill:
I think for the purpose of making `extBinders` being closer to `fun` binder syntax a different approach is needed, so I'm closing this.

PR #237, opened 2023-09-02
feat: module docs

fgdorais:
Add statement of purpose as requested in #203.


Std/Data/Nat/Lemmas.lean
```diff
@@ -344,7 +351,7 @@ protected theorem add_self_ne_one : ∀ (n : Nat), n + n ≠ 1
     have h1 : succ (succ (n + n)) = 1 := succ_add n n ▸ h
     Nat.noConfusion h1 fun.
 
-/-! ### sub -/
+/-! ## sub -/
```
YaelDillies, line 354: I think it makes sense to keep the headers with the same size as everywhere else even if we're not using the intermediate size.

YaelDillies (approved):
Thank you! I'm happy with this docstring, but I guess Mario should sign off.

Std/Data/Nat/Lemmas.lean
digama0: I don't think we are being very consistent about header size. It probably requires a global review.

YaelDillies: At least, we're consistent within mathlib.

fgdorais: Should I change from `##` to `###`? The file was inconsistent so I picked one. I have no personal preference for 2 or 3.

YaelDillies: I'd personally go for `###` but I guess Mario might have a different opinion.

Std/Data/Nat/Lemmas.lean
```diff
@@ -8,6 +8,13 @@ import Std.Tactic.Basic
 import Std.Data.Nat.Init.Lemmas
 import Std.Data.Nat.Basic
 
+/-! # Basic lemmas about natural numbers
+
+The primary purpose of the lemmas in this file is to assist with reasonning
```
semorrison, line 13:
```suggestion
The primary purpose of the lemmas in this file is to assist with reasoning
```

semorrison (approved):

PR #238, opened 2023-09-04
chore: upstream left and right tactics

semorrison:


Std/Tactic/LeftRight.lean
```diff
@@ -0,0 +1,45 @@
...
+checking that it is an inductive type,
+and that there are the expected number of constructors.
+-/
+def nthConstructor (name : Name) (idx : Nat) (max : Option Nat) (goal : MVarId) :
+    MetaM (List MVarId) := do
+  goal.withContext do
+    goal.checkNotAssigned name
+    matchConstInduct (← goal.getType').getAppFn
+      (fun _ ↦ throwTacticEx `constructor goal "target is not an inductive datatype")
+      fun ival us ↦ do
```
digama0, line 23:
```suggestion
      (fun _ => throwTacticEx `constructor goal "target is not an inductive datatype")
      fun ival us => do
```

digama0:
needs a test

semorrison:
Added a test.

digama0:
also examples for inductives with one constructor, three constructors, definitions that unfold to inductives, and non-inductives

semorrison:
Completed tests.

PR #241, opened 2023-09-06
chore: upstream on_goal / pick_goal / swap

semorrison:


fgdorais:
I think this needs more justification. I can see this is of some use but I'm wondering why structured proofs don't solve most of the use cases. Some parts would be handy for writing basic tactics that need to reorder goals but there are other ways to do that. Bottom line, this looks like legacy Lean 3 code for the most part.

I would like to see a less complicated API and more convincing test cases.

digama0:
I believe the motivation is at least in part so that they can be generated by tools like `aesop` that generate proof scripts replicating what they did on the goal, which are still largely readable and suitable for inclusion in the source text.

fgdorais:
I understand now. Also, the tests for #242 are exactly the kind of use cases I was looking for. I approve.

fgdorais (approved):

joehendrix (approved):

PR #242, opened 2023-09-06
chore: upstream classical tactic

semorrison:


fgdorais:
Does this really depend on #241? I think it could be made to do without.

semorrison:
> Does this really depend on #241? I think it could be made to do without.

I was not anticipating this being nearly as slow; I thought when I started that both of these PRs would be merged 30 minutes-hours later, but here we are 60 days later.

Yes, the only "dependence" is through a test.

semorrison:
This PR is not essential for anything I'm doing in the short term; if you'd like to disentangle them please do!

Std/Tactic/Classical.lean
```diff
@@ -0,0 +1,51 @@
...
+def bar : Bool := by
+  classical
+  have := ∀ p, decide p -- uses the classical instance
+  exact decide (0 < 1) -- uses the decidable instance
+```
+Note that (unlike lean 3) `classical` is a scoping tactic - it adds the instance only within the
+scope of the tactic.
+-/
+-- FIXME: using ppDedent looks good in the common case, but produces the incorrect result when
+-- the `classical` does not scope over the rest of the block.
```
fgdorais, line 46: Can this be fixed in this PR?

fgdorais (approved):
Looks good to me modulo the question about the FIXME.

joehendrix (approved):

PR #243, opened 2023-09-06
chore: upstream split_ifs

semorrison:


digama0:
This one seems like big sadness to me, why can't we use `split` instead? Have we given up on progressing https://github.com/leanprover/lean4/issues/2414 or is the issue something else?

semorrison:
@digama0 my main complaint is lack of `split with h`.

digama0:
Is that a feature request I hear?

semorrison:
I'm going to close this now, let's see if we can work on `split` instead.

PR #244, opened 2023-09-06
chore: upstream rfl and symm tactics

semorrison:
The `trans` tactic in Mathlib needs more polish than these, so I'll do that in a separate PR to ease reviewing.

### DIFF
--- a/Std/Lean/Elab/Tactic.lean
+++ b/Std/Lean/Elab/Tactic.lean
@@ -3,7 +3,7 @@ Copyright (c) 2022 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
-import Lean
+import Lean.Elab.Tactic.Basic
 
 /-!
 # Tactic combinators in `TacticM`.
--- a/Std/Lean/Elab/Tactic.lean
+++ b/Std/Lean/Elab/Tactic.lean
@@ -1,0 +1,18 @@
+/-
+Copyright (c) 2022 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro
+-/
+import Lean
+
+/-!
+# Tactic combinators in `TacticM`.
+-/
+
+namespace Lean.Elab.Tactic
+
+/-- Analogue of `liftMetaTactic` for tactics that do not return any goals. -/
+def liftMetaFinishingTactic (tac : MVarId → MetaM Unit) : TacticM Unit :=
+  liftMetaTactic fun g => do tac g; pure []
+
+end Lean.Elab.Tactic
--- a/Std/Tactic/Relation/Rfl.lean
+++ b/Std/Tactic/Relation/Rfl.lean
@@ -13,7 +13,7 @@ This extends the `rfl` tactic so that it works on any reflexive relation,
 provided the reflexivity lemma has been marked as `@[refl]`.
 -/
 
-namespace Mathlib.Tactic
+namespace Std.Tactic
 
 open Lean Meta
 
--- a/Std/Tactic/Relation/Rfl.lean
+++ b/Std/Tactic/Relation/Rfl.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Newell Jensen, Thomas Murrills
 -/
 import Std.Lean.Elab.Tactic
+import Lean.Meta.Tactic.Apply
 
 /-!
 # `rfl` tactic extension for reflexive relations
--- a/Std/Tactic/Relation/Rfl.lean
+++ b/Std/Tactic/Relation/Rfl.lean
@@ -21,14 +21,14 @@ open Lean Meta
 initialize reflExt :
     SimpleScopedEnvExtension (Name × Array (DiscrTree.Key true)) (DiscrTree Name true) ←
   registerSimpleScopedEnvExtension {
-    addEntry := fun dt (n, ks) ↦ dt.insertCore ks n
+    addEntry := fun dt (n, ks) => dt.insertCore ks n
     initial := {}
   }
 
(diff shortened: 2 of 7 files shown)


Std/Lean/Elab/Tactic.lean
```diff
@@ -0,0 +1,18 @@
+/-
+Copyright (c) 2022 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro
+-/
+import Lean
```
digama0, line 6: minimize import

semorrison: Done, and linter at #259.

joehendrix:
These seem like instances of a potential more general "resolution" tactic that would try to resolve a collection of implications  `A1 -> ... AN -> B` against the goal or a hypothesis.

Is that correct?

digama0:
Yes (you are describing `solve_by_elim`), but that's an incredibly heavy hammer and `rfl` and `symm` need to have more controlled results because they are used as structural tactics (nonterminally)

Std/Tactic/Relation/Symm.lean
```diff
@@ -0,0 +1,128 @@
...
+    let .app (.app rel _) _ := targetTy | fail
+    let key ← withReducible <| DiscrTree.mkPath rel
+    symmExt.add (decl, key) kind
+}
+
+end Std.Tactic
+
+open Std.Tactic
+
+namespace Lean.Expr
```
digama0, line 48: All of these functions should be in the `Std.Tactic` namespace, I don't think it's a good idea to create `Expr.symm` here, this is likely to conflict with uses outside std. Ditto for `MVarId.symm`, we don't have a precedent for doing tactic functions with `MVarId` dot notation in std, only lean core does that and I am concerned about name collisions.

semorrison:
> Yes (you are describing `solve_by_elim`), but that's an incredibly heavy hammer and `rfl` and `symm` need to have more controlled results because they are used as structural tactics (nonterminally)

I'll be upstreaming `solve_by_elim` soon.

Std/Tactic/Relation/Symm.lean
semorrison: I guess so. It makes me a little sad to lose the dot notation for both `Expr.symm` and `MVarId.symm`.

digama0: Is it used ever (outside this file)?

semorrison: `MVarId.symm` is used in one tactic in Mathlib, and `MVarId.rfl` is used in two (`rw?` and `gcongr`). So it's no great loss. I just hadn't realised you wanted to be as cautious about the namespaces for Std; given Std is always going to be near the top of any import hierarchy I'd thought we'd be safe to insert things into core namespaces.

digama0: I think this is mostly fine, but I would try to limit this to things that are both public API and also not likely to be desired for other purposes either upstream or downstream.

semorrison: Okay, I have changed the incorrect `Mathlib` namespaces to `Std`, and made the `symmAux` functions private.

joehendrix: I also slightly prefer a longer name that indicates it's applying a tactic for these (`applySymm`?), but I'll defer to the author on this.

Is there any remaining changes desired before merging?

semorrison: Indeed, the longer name is better. Done now. I'm happy for these to go in.

Std/Tactic/Relation/Rfl.lean
```diff
@@ -0,0 +1,75 @@
...
+
+/-- `MetaM` version of the `rfl` tactic.
+
+This tactic applies to a goal whose target has the form `x ~ x`, where `~` is a reflexive
+relation, that is, a relation which has a reflexive lemma tagged with the attribute [refl].
+-/
+def _root_.Lean.MVarId.applyRfl (goal : MVarId) : MetaM Unit := do
+  let .app (.app rel _) _ ← whnfR <|← instantiateMVars <|← goal.getType
+    | throwError "reflexivity lemmas only apply to binary relations, not
+      {indentExpr (← goal.getType)}"
```
digama0, line 52:
```suggestion
    | throwError "reflexivity lemmas only apply to binary relations, not{
        indentExpr (← goal.getType)}"
```

Std/Tactic/Relation/Rfl.lean
```diff
@@ -0,0 +1,75 @@
...
+  let s ← saveState
+  let mut ex? := none
+  for lem in ← (reflExt.getState (← getEnv)).getMatch rel do
+    try
+      let gs ← goal.apply (← mkConstWithFreshMVarLevels lem)
+      if gs.isEmpty then return () else
+        logError <| MessageData.tagged `Tactic.unsolvedGoals <| m!"unsolved goals\n
+          {goalsToMessageData gs}"
```
digama0, line 60:
```suggestion
        logError <| MessageData.tagged `Tactic.unsolvedGoals <| m!"unsolved goals\n{
          goalsToMessageData gs}"
```

Std/Tactic/Relation/Rfl.lean
```diff
@@ -0,0 +1,75 @@
...
+    throw e
+  else
+    throwError "rfl failed, no lemma with @[refl] applies"
+
+/--
+This tactic applies to a goal whose target has the form `x ~ x`, where `~` is a reflexive
+relation, that is, a relation which has a reflexive lemma tagged with the attribute [refl].
+-/
+elab_rules : tactic
+| `(tactic| rfl) => withMainContext do liftMetaFinishingTactic (·.applyRfl)
```
digama0, line 75:
```suggestion
  | `(tactic| rfl) => withMainContext do liftMetaFinishingTactic (·.applyRfl)
```

PR #245, opened 2023-09-06
chore: upstream @[eqns]

semorrison:


Std/Tactic/Eqns.lean
```diff
@@ -0,0 +1,55 @@
...
+This file provides the `eqns` attribute as a way of overriding the default equation lemmas.
+-/
+open Lean Elab
+
+namespace Std.Tactic
+
+/--
+Overrides the default equation lemmas. For example
+
+```lean4
```
alexjbest, line 22: seems a bit unnecessary to say lean4 here and not in every other file
````suggestion
```
````

File: test/eqns.lean [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+  transpose A i j = A j i := rfl
+
+attribute [eqns transpose_apply] transpose
+
+theorem transpose_const {m n} (c : Nat) :
+    transpose (fun (_i : m) (_j : n) => c) = fun _j _i => c := by
+  fail_if_success rw [transpose]
+  fail_if_success simp [transpose]
+  funext i j
+  rw [transpose]
```
joehendrix, line 19: At least with lean `v.4.5.0-rc1`, the example works even without the `eqns` attribute above.

Perhaps the heuristics have changed?  Is this extension still useful?

semorrison: I just checked that there are `attribute [eqns]` in Mathlib that have effect / are needed (although some are not needed!).

I don't think there is any particular clamour for this PR: it's just trying to remove tactics from Mathlib that are "needed" before the basic algebraic hierarchy.

I will see if I can move a better test case up from Mathlib, but low priority.

semorrison: @eric-wieser, this is actually the example in the doc-string in Mathlib, but it seems it is no longer relevant there. Would you be willing to rewrite the doc-string?

semorrison: The main use case of `eqns` is actually the `irreducible_def X := ...` command in Mathlib, which makes use of `opaque` to give a completely irreducible definition `X` and a `X_def` theorem, and then uses `attribute [eqns]` to allow `rw [X]` to achieve the same as `rw [X_def]`.

fgdorais:
This feature has new utility now that definitions by wellfounded recursion are irreducible.

PR #246, opened 2023-09-06
chore: upstream spread syntax `__ :=`

kim-em:


digama0:
I don't really like this being an extension to all structure instances, it's a potential performance issue. Is there a reason this can't or shouldn't be upstreamed to core?

semorrison:
No, I think that would be reasonable. I'll mark this as `help-wanted` in the hope that someone is interested in taking this on!

fgdorais:

PR #248, opened 2023-09-07
feat: some string length lemmas

alexjbest:
not sure if they belong here or in mathlib but they seemed widely useful

### DIFF
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -23,6 +23,16 @@ theorem ext_iff {s₁ s₂ : String} : s₁ = s₂ ↔ s₁.data = s₂.data := 
 
 @[simp] theorem mk_length (s : List Char) : (String.mk s).length = s.length := rfl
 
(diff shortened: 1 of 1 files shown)


Std/Data/String/Lemmas.lean
```diff
@@ -23,6 +23,16 @@ theorem ext_iff {s₁ s₂ : String} : s₁ = s₂ ↔ s₁.data = s₂.data :=
 
 @[simp] theorem mk_length (s : List Char) : (String.mk s).length = s.length := rfl
 
+@[simp] theorem length_toString (c : Char) : c.toString.length = 1 := rfl
```
digama0, line 26: I think this would make more sense in the `Char` namespace, otherwise the name is too general

semorrison:
awaiting-review

alexjbest:
Damn I had no idea that I typo'd. This bot labelling thing is kinda annoying.

semorrison:
Ask @digama0 for triage access. :-)

PR #251, opened 2023-09-12
chore: upstream from Mathlib.Data.Sum.Basic

semorrison:


Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,220 @@
...
+
+## Further material
+
+See `Std.Data.Sum.Basic` for theorems about these definitions.
+
+## Notes
+
+The definition of `Sum` takes values in `Type _`. This effectively forbids `Prop`- valued sum types.
+To this effect, we have `PSum`, which takes value in `Sort _` and carries a more complicated
+universe signature in consequence. The `Prop` version is `or`.
```
digama0, line 39:
```suggestion
universe signature in consequence. The `Prop` version is `Or`.
```

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,220 @@
...
+-/
+
+
+universe u v w x y z
+
+variable {α : Type u} {α' : Type w} {β : Type v} {β' : Type x} {γ : Type y} {δ : Type z}
```
digama0, line 45:
```suggestion
```
Use autoimplicits where possible, otherwise use implicit args on the definitions

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,220 @@
...
+def getLeft? : Sum α β → Option α
+  | inl a => some a
+  | inr _ => none
+
+/-- Check if a sum is `inr` and if so, retrieve its contents. -/
+def getRight? : Sum α β → Option β
+  | inr b => some b
+  | inl _ => none
+
+variable {x y : Sum α β}
```
digama0, line 82:
```suggestion
```

Std/Data/Sum/Basic.lean
semorrison: Done! Just a matter of deleting them. If only Mathlib were so enlightened. :-)

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,220 @@
...
+
+end get
+
+/-- Define a function on `α ⊕ β` by giving separate definitions on `α` and `β`. -/
+protected def elim {α β γ : Sort _} (f : α → γ) (g : β → γ) : Sum α β → γ :=
+  fun x ↦ Sum.casesOn x f g
+
+@[simp]
+theorem elim_inl {α β γ : Sort _} (f : α → γ) (g : β → γ) (x : α) : Sum.elim f g (inl x) = f x :=
+  rfl
```
digama0, line 105:
```suggestion
@[simp] theorem elim_inl {α β γ : Sort _} (f : α → γ) (g : β → γ) (x : α) : Sum.elim f g (inl x) = f x := rfl
```
Both `@[simp]` and `rfl` go on the same line if possible, split before the return type or after the `:=` if it doesn't fit on the line

Std/Data/Sum/Lemmas.lean
```diff
@@ -42,7 +37,7 @@ theorem forall_sum_pi {γ : α ⊕ β → Sort _} (p : (∀ ab, γ ab) → Prop)
 
 section get
 
-variable {x y : Sum α β}
+variable {x : Sum α β}
```
digama0, line 40: can we get rid of this one too?

semorrison: Sure. Done!

Std/Data/Sum/Basic.lean
semorrison: Done.

I put `fun.` on the same line too?

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,189 @@
...
+  ⟨fun h ↦ by
+    cases h
+    assumption, LiftRel.inl⟩
+
+@[simp] theorem not_liftRel_inl_inr : ¬LiftRel r s (inl a) (inr d) := fun.
+
+@[simp] theorem not_liftRel_inr_inl : ¬LiftRel r s (inr b) (inl c) := fun.
+
+@[simp] theorem liftRel_inr_inr : LiftRel r s (inr b) (inr d) ↔ s b d :=
+  ⟨fun h ↦ by
```
digama0, line 138:
```suggestion
  ⟨fun h => by
```
Put this on your list of find/replaces when migrating things from mathlib to std

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,189 @@
...
+inductive LiftRel (r : α → γ → Prop) (s : β → δ → Prop) : Sum α β → Sum γ δ → Prop
+  /-- `inl a` and `inl c` are related via `LiftRel r s` if `a` and `c` are related via `r`. -/
+  | protected inl {a c} : r a c → LiftRel r s (inl a) (inl c)
+  /-- `inr b` and `inr d` are related via `LiftRel r s` if `b` and `d` are related via `s`. -/
+  | protected inr {b d} : s b d → LiftRel r s (inr b) (inr d)
+
+@[simp] theorem liftRel_inl_inl : LiftRel r s (inl a) (inl c) ↔ r a c :=
+  ⟨fun h ↦ by
+    cases h
+    assumption, LiftRel.inl⟩
```
digama0, line 131:
```suggestion
  ⟨fun h => by cases h; assumption, LiftRel.inl⟩
```
I think this syntax was messed up from an early version of mathport

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,189 @@
...
+  | inl _, inl _ => decidable_of_iff' _ lex_inl_inl
+  | inl _, inr _ => Decidable.isTrue (Lex.sep _ _)
+  | inr _, inl _ => Decidable.isFalse lex_inr_inl
+  | inr _, inr _ => decidable_of_iff' _ lex_inr_inr
+
+end Lex
+
+end Sum
+
+open Sum
```
digama0, line 189:
```suggestion
```

Std/Data/Sum/Basic.lean
semorrison: I'll add a linter, I think.

Std/Data/Sum/Basic.lean
semorrison: Done (x4)

Std/Data/Sum/Basic.lean
digama0: yeah, if the body fits on one line you should, even if it's something more complicated

Std/Data/Sum/Basic.lean
semorrison: Of course the `open Sum` should go, but you really want to delete perfectly valid `end` statements? Isn't it just making life harder for whoever comes next?

Std/Data/Sum/Basic.lean
semorrison: Okay, removed a few more newlines.

Std/Data/Sum/Basic.lean
digama0: I think the policy should be something like, there should not be an end statement for the "main namespace" of the file, but if it is a "local section" then it should have an end statement. Hence `end Sum` should be deleted but not `end Lex`

semorrison: Great.

semorrison: Done.

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,252 @@
...
+Authors: Mario Carneiro, Yury G. Kudryashov
+-/
+
+import Std.Data.Sum.Basic
+import Std.Tactic.Ext
+
+/-!
+# Disjoint union of types
+
+Theorems about the definitions introduced in `Std.Data.Sum.Defs`.
```
digama0, line 13:
```suggestion
Theorems about the definitions introduced in `Std.Data.Sum.Basic`.
```

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,252 @@
...
+-/
+
+namespace Sum
+
+@[simp] theorem «forall» {p : Sum α β → Prop} : (∀ x, p x) ↔ (∀ a, p (inl a)) ∧ ∀ b, p (inr b) :=
+  ⟨fun h => ⟨fun _ => h _, fun _ => h _⟩, fun ⟨h₁, h₂⟩ => Sum.rec h₁ h₂⟩
+
+@[simp] theorem «exists» {p : Sum α β → Prop} : (∃ x, p x) ↔ (∃ a, p (inl a)) ∨ ∃ b, p (inr b) :=
```
digama0, line 21:
```suggestion
@[simp] protected theorem «forall» {p : Sum α β → Prop} : (∀ x, p x) ↔ (∀ a, p (inl a)) ∧ ∀ b, p (inr b) :=
  ⟨fun h => ⟨fun _ => h _, fun _ => h _⟩, fun ⟨h₁, h₂⟩ => Sum.rec h₁ h₂⟩

@[simp] protected theorem «exists» {p : Sum α β → Prop} : (∃ x, p x) ↔ (∃ a, p (inl a)) ∨ ∃ b, p (inr b) :=
```
seems bad to have forall and exists get shadowed, although maybe this isn't an issue since they are keywords anyway?

semorrison: Yeah, probably nicer to have these protected.

Done.

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+
+open Function
+
+theorem elim_const_const (c : γ) :
+    Sum.elim (const _ c : α → γ) (const _ c : β → γ) = const _ c := by
+  ext x
+  cases x <;> rfl
+
+@[simp] theorem elim_lam_const_lam_const (c : γ) :
+    (Sum.elim (fun _ : α => c) fun _ : β => c) = fun _ => c :=
```
digama0, line 251:
```suggestion
    Sum.elim (fun _ : α => c) (fun _ : β => c) = fun _ => c :=
```

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,176 @@
...
+  if it's coming from the left.
+* `Sum.map`: Maps `α ⊕ β` to `γ ⊕ δ` component-wise.
+* `Sum.elim`: Nondependent eliminator/induction principle for `α ⊕ β`.
+* `Sum.swap`: Maps `α ⊕ β` to `β ⊕ α` by swapping components.
+* `Sum.LiftRel`: The disjoint union of two relations.
+* `Sum.Lex`: Lexicographic order on `α ⊕ β` induced by a relation on `α` and a relation on `β`.
+
+## Further material
+
+See `Std.Data.Sum.Basic` for theorems about these definitions.
```
digama0, line 34:
```suggestion
See `Std.Data.Sum.Lemmas` for theorems about these definitions.
```

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,176 @@
...
+
+@[simp] theorem getLeft?_inl (x : α) : (inl x : α ⊕ β).getLeft? = some x := rfl
+@[simp] theorem getLeft?_inr (x : β) : (inr x : α ⊕ β).getLeft? = none := rfl
+@[simp] theorem getRight?_inl (x : α) : (inl x : α ⊕ β).getRight? = none := rfl
+@[simp] theorem getRight?_inr (x : β) : (inr x : α ⊕ β).getRight? = some x := rfl
+
+end get
+
+/-- Define a function on `α ⊕ β` by giving separate definitions on `α` and `β`. -/
+protected def elim {α β γ : Sort _} (f : α → γ) (g : β → γ) : Sum α β → γ :=
```
digama0, line 94:
```suggestion
protected def elim {α β γ : Sort _} (f : α → γ) (g : β → γ) : α ⊕ β → γ :=
```
Usage of the notation seems to be inconsistent in this file. I suggest using it everywhere.

semorrison: Done!

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+@[simp] protected theorem «exists» {p : α ⊕ β → Prop} :
+    (∃ x, p x) ↔ (∃ a, p (inl a)) ∨ ∃ b, p (inr b) :=
+  ⟨ fun
+    | ⟨inl a, h⟩ => Or.inl ⟨a, h⟩
+    | ⟨inr b, h⟩ => Or.inr ⟨b, h⟩,
+    fun
+    | Or.inl ⟨a, h⟩ => ⟨inl a, h⟩
+    | Or.inr ⟨b, h⟩ => ⟨inr b, h⟩⟩
+
+theorem forall_sum_pi {γ : α ⊕ β → Sort _} (p : (∀ ab, γ ab) → Prop) :
```
digama0, line 31:
```suggestion
theorem forall_sum {γ : α ⊕ β → Sort _} (p : (∀ ab, γ ab) → Prop) :
```
I'm not sure what the `pi` is referring to here, and in any case we don't use `Pi` to refer to the pi-type anymore (at least in core, `PiInstances.lean` may beg to differ).

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+  cases x <;> simp only [getLeft?, Option.some.injEq, inl.injEq]
+
+@[simp] theorem getRight?_eq_some_iff {b : β} : x.getRight? = some b ↔ x = inr b := by
+  cases x <;> simp only [getRight?, Option.some.injEq, inr.injEq]
+
+@[simp] theorem not_isLeft (x : α ⊕ β) : not x.isLeft = x.isRight := by cases x <;> rfl
+
+@[simp] theorem isLeft_eq_false {x : α ⊕ β} : x.isLeft = false ↔ x.isRight := by cases x <;> simp
+
+theorem Not_isLeft {x : α ⊕ β} : ¬x.isLeft ↔ x.isRight := by simp
```
digama0, line 73: This is an annoying name clash. I think I would rather we call the other one `bnot_isLeft` and prefer `not_isLeft` here, although it may require a larger effort and/or bringing it up on zulip...

semorrison: Let's just rename here and hope for the best downstream. :-)

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+@[simp] theorem getRight_eq_iff {b : β} (h : x.isRight) : x.getRight h = b ↔ x = inr b := by
+  cases x <;> simp at h ⊢
+
+@[simp] theorem getLeft?_eq_some_iff {a : α} : x.getLeft? = some a ↔ x = inl a := by
+  cases x <;> simp only [getLeft?, Option.some.injEq, inl.injEq]
+
+@[simp] theorem getRight?_eq_some_iff {b : β} : x.getRight? = some b ↔ x = inr b := by
+  cases x <;> simp only [getRight?, Option.some.injEq, inr.injEq]
+
+@[simp] theorem bnot_isLeft (x : α ⊕ β) : not x.isLeft = x.isRight := by cases x <;> rfl
```
digama0, line 69:
```suggestion
@[simp] theorem bnot_isLeft (x : α ⊕ β) : !x.isLeft = x.isRight := by cases x <;> rfl
```

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+
+theorem inr.inj_iff {a b} : (inr a : α ⊕ β) = inr b ↔ a = b := ⟨inr.inj, congrArg _⟩
+
+theorem inl_ne_inr {a : α} {b : β} : inl a ≠ inr b := fun.
+
+theorem inr_ne_inl {a : α} {b : β} : inr b ≠ inl a := fun.
+
+/-! ### `Sum.elim` -/
+
+@[simp] theorem elim_comp_inl {α β γ : Sort _} (f : α → γ) (g : β → γ) : Sum.elim f g ∘ inl = f :=
```
digama0, line 97: haven't tested but I guess most of these `: Sort _` variables are unnecessary

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+  ⟨fun x => Sum.recOn x aca fun b => lex_acc_inr aca (hb.apply b)⟩
+
+end Lex
+
+end Sum
+
+open Sum
+
+namespace Sum
+
```
digama0, line 242:
```suggestion
```

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+theorem elim_const_const (c : γ) :
+    Sum.elim (const _ c : α → γ) (const _ c : β → γ) = const _ c := by
+  ext x
+  cases x <;> rfl
+
+@[simp] theorem elim_lam_const_lam_const (c : γ) :
+    Sum.elim (fun _ : α => c) (fun _ : β => c) = fun _ => c :=
+  Sum.elim_const_const c
+
+end Sum
```
digama0, line 254:
```suggestion
```

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,254 @@
...
+
+end Lex
+
+end Sum
+
+open Sum
+
+namespace Sum
+
+open Function
```
digama0, line 243: Without the namespaces I think this `open` is a bit discouraged, coming in the middle of a section like this, but it should be fine to move it to the top of the file.

Std/Data/Sum/Basic.lean
```diff
@@ -0,0 +1,176 @@
...
+
+@[simp] theorem getLeft?_inl : (inl x : α ⊕ β).getLeft? = some x := rfl
+@[simp] theorem getLeft?_inr : (inr x : α ⊕ β).getLeft? = none := rfl
+@[simp] theorem getRight?_inl : (inl x : α ⊕ β).getRight? = none := rfl
+@[simp] theorem getRight?_inr : (inr x : α ⊕ β).getRight? = some x := rfl
+
+end get
+
+/-- Define a function on `α ⊕ β` by giving separate definitions on `α` and `β`. -/
+protected def elim {α β γ : Sort _} (f : α → γ) (g : β → γ) : α ⊕ β → γ :=
```
digama0, line 94:
```suggestion
protected def elim {α β γ} (f : α → γ) (g : β → γ) : α ⊕ β → γ :=
```
Why is this needed? (In any case using `Sort _` is misleading since the first two are actually `Type _`)

semorrison: Without this is puts these arguments in the `α γ β` order, which isn't very helpful.

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,246 @@
...
+
+protected theorem LiftRel.swap (h : LiftRel r s x y) : LiftRel s r x.swap y.swap := by
+  cases h
+  · exact LiftRel.inr ‹_›
+  · exact LiftRel.inl ‹_›
+
+@[simp] theorem liftRel_swap_iff : LiftRel s r x.swap y.swap ↔ LiftRel r s x y :=
+  ⟨fun h => by
+    rw [← swap_swap x, ← swap_swap y]
+    exact h.swap, LiftRel.swap⟩
```
digama0, line 189:
```suggestion
  ⟨fun h => by rw [← swap_swap x, ← swap_swap y]; exact h.swap, LiftRel.swap⟩
```

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,246 @@
...
+    fun
+    | Or.inl ⟨a, h⟩ => ⟨inl a, h⟩
+    | Or.inr ⟨b, h⟩ => ⟨inr b, h⟩⟩
+
+theorem forall_sum {γ : α ⊕ β → Sort _} (p : (∀ ab, γ ab) → Prop) :
+    (∀ fab, p fab) ↔ (∀ fa fb, p (Sum.rec fa fb)) :=
+  ⟨fun h fa fb => h _, fun h fab => by
+    have h1 : fab = Sum.rec (fun a => fab (Sum.inl a)) (fun b => fab (Sum.inr b)) := by
+      ext ab; cases ab <;> rfl
+    rw [h1]; exact h _ _⟩
```
digama0, line 38:
```suggestion
  ⟨ fun h fa fb => h _,
    fun h fab => by
      have h1 : fab = Sum.rec (fun a => fab (Sum.inl a)) (fun b => fab (Sum.inr b)) := by
        ext ab; cases ab <;> rfl
      rw [h1]; exact h _ _⟩
```
option 1

Std/Data/Sum/Lemmas.lean
```diff
@@ -0,0 +1,244 @@
...
+@[simp] theorem map_id_id : Sum.map (@id α) (@id β) = id :=
+  funext fun x => Sum.recOn x (fun _ => rfl) fun _ => rfl
+
+theorem elim_map {f₁ : α → β} {f₂ : β → ε} {g₁ : γ → δ} {g₂ : δ → ε} {x} :
+    Sum.elim f₂ g₂ (Sum.map f₁ g₁ x) = Sum.elim (f₂ ∘ f₁) (g₂ ∘ g₁) x := by
+  cases x <;> rfl
+
+theorem elim_comp_map {f₁ : α → β} {f₂ : β → ε} {g₁ : γ → δ} {g₂ : δ → ε} :
+    Sum.elim f₂ g₂ ∘ Sum.map f₁ g₁ = Sum.elim (f₂ ∘ f₁) (g₂ ∘ g₁) :=
+  funext $ fun _ => elim_map
```
digama0, line 136:
```suggestion
  funext fun _ => elim_map
```
was going to suggest `<|` as another item for the find/replace, but neither is needed here.

PR #252, opened 2023-09-12
chore: replace ↦ with =>

semorrison:
Additional proposal:

* We still use `↦` in documentation. Let's change to some different symbol, so one can simply search for `↦` and know that every appearance is "wrong".
* Lint for this.

### DIFF
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -280,7 +280,7 @@ def insert (self : HashMap α β) (a : α) (b : β) : HashMap α β := ⟨self.1
 
 /--
 Similar to `insert`, but also returns a boolean flag indicating whether an existing entry has been
-replaced with `a ↦ b`.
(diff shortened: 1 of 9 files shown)


digama0:
I'm unclear on your first point. Did you replace all occurrences of `↦` in docs or not? I see one replacement in docs here but I'm not sure if this is all of them.

semorrison:
There are lots of docs regarding HashMap that use this symbol, e.g.

```
/--
Similar to `insert`, but also returns a boolean flag indicating whether an existing entry has been
replaced with `a ↦ b`.
-/
```

The one use in documentation that I replaced here was the "standard" (Mathlib) meaning.

digama0:
Hm, I see. I suppose that it wouldn't be that bad to just use `=>` in those comments as well.

semorrison:
Okay, in that case I will update this PR to use `=>` everywhere, and then think about adding a linter.

PR #253, opened 2023-09-12
chore upstream List.partitionM

semorrison:
* I need this for `exact?`
* Happy to add the supremum of `List.partitionMap` and `List.partitionM` as well if that is desirable.

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -685,6 +685,20 @@ partitionMap (id : Nat ⊕ Nat → Nat ⊕ Nat) [inl 0, inr 1, inl 2] = ([0, 2],
     | .inl a => go xs (acc₁.push a) acc₂
     | .inr b => go xs acc₁ (acc₂.push b)
 
+/-- Monadic generalization of `List.partition`. -/
+@[inline] def partitionM [Monad m] (p : α → m Bool) (l : List α) : m (List α × List α) :=
+  go l #[] #[] where
+  /-- Auxiliary for `partitionM`:
+  `partitionM.go p l acc₁ acc₂` returns `(acc₁.toList ++ left, acc₂.toList ++ right)`
+  if `partitionM p l` returns `(left, right)`. -/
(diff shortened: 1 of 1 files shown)


Std/Data/List/Basic.lean
```diff
@@ -685,6 +685,20 @@ partitionMap (id : Nat ⊕ Nat → Nat ⊕ Nat) [inl 0, inr 1, inl 2] = ([0, 2],
     | .inl a => go xs (acc₁.push a) acc₂
     | .inr b => go xs acc₁ (acc₂.push b)
 
+/-- Monadic generalization of `List.partition`. -/
+@[inline] def partitionM [Monad m] (p : α → m Bool) (l : List α) : m (List α × List α) :=
+  go l #[] #[] where
```
digama0, line 690:
```suggestion
  go l #[] #[]
where
```
The convention here is to put `where` on its own line, unless it can fit on the same line as the body and the `:=`

Std/Data/List/Basic.lean
```diff
@@ -685,6 +685,20 @@ partitionMap (id : Nat ⊕ Nat → Nat ⊕ Nat) [inl 0, inr 1, inl 2] = ([0, 2],
...
+  /-- Auxiliary for `partitionM`:
+  `partitionM.go p l acc₁ acc₂` returns `(acc₁.toList ++ left, acc₂.toList ++ right)`
+  if `partitionM p l` returns `(left, right)`. -/
+  @[specialize] go : List α → Array α → Array α → m (List α × List α)
+  | [], acc₁, acc₂ => pure (acc₁.toList, acc₂.toList)
+  | x :: xs, acc₁, acc₂ => do
+      if ← p x then
+        go xs (acc₁.push x) acc₂
+      else
+        go xs acc₁ (acc₂.push x)
```
digama0, line 700:
```suggestion
    if ← p x then
      go xs (acc₁.push x) acc₂
    else
      go xs acc₁ (acc₂.push x)
```

PR #254, opened 2023-09-12
chore: MVarId.isIndependentOf

semorrison:
Another one needed for `exact?`.


Std/Lean/Meta/Basic.lean
```diff
@@ -274,6 +274,57 @@ where
...
+  by proof irrelevance.
+* Similarly given `f : Fin 0 → Type`, `?m_1 : Fin 0` is independent of `?m_2 : f ?m_1`,
+  because `Fin 0` is a subsingleton.
+* Finally `?m_1 : Nat` is independent of `?m_2 : α`,
+  as long as `?m_1` does not appear in `Meta.getMVars α`
+  (note that `Meta.getMVars` follows delayed assignments).
+
+This function only calculates a conservative approximation of this condition.
+(That is, it may return `false` when it should return `true`.
+In particular it returns false whenever the type of `g` contains a metavariable,
```
digama0, line 308:
```suggestion
That is, it may return `false` when it should return `true`.
(In particular it returns false whenever the type of `g` contains a metavariable,
```

Std/Lean/Meta/Basic.lean
```diff
@@ -274,6 +274,57 @@ where
...
+  let t ← instantiateMVars (← g.getType)
+  if t.hasExprMVar then
+    -- If the goal's type contains other meta-variables,
+    -- we conservatively say that `g` is not independent.
+    -- It would be possible to check if `L` depends on these meta-variables.
+    return false
+  if t.isProp then
+    -- If the goal is propositional,
+    -- proof-irrelevance ensures it is independent of any other goals.
+    return true
```
digama0, line 321: Are you sure this is testing what you want? This will only return true when the goal is `|- Prop`

semorrison: That was definitely wrong, but always being caught by the next step `\l g.isSubsingleton`. Fixed.

PR #255, opened 2023-09-12
chore: add grep-based linter for ↦

semorrison:
Simplest possible implementation of a linter forbidding `↦`. I propose not building more general infrastructure until we want it.


semorrison:
Expecting this to fail at first, then I'll rebase onto #252.

digama0:
Another linter in the same category is the line length check.

PR #256, opened 2023-09-12
chore: CI linter for long lines

semorrison:


semorrison:
I haven't quite fixed all the long lines here. #guard_msgs won't let you insert newlines.

Options:
* Don't run this linter in `test`
* Teach #guard_msgs to normalize whitespace
* ...?

digama0:
(Arguably we should just fix these error messages in core to not use such long lines.) We could do something to detect lines that precede `#guard_msgs`, but probably this would exceed what you could do in a bash one-liner. I guess a really simple whitespace normalization which is sufficient for line breaking without losing too much detail is `expected.replace '\n' ' ' == doc.replace '\n' ' '`, maybe that's the easiest quick fix for this PR.

PR #257, opened 2023-09-12
chore: fix argument order in Sum/Lemmas

semorrison:
I guess we really do need to make the matching Mathlib PRs before merging upstreaming PRs...


digama0:
I kind of just want to say that we shouldn't be bothered about this in most cases... If you are giving the arguments you should use named argument style `(α := Nat)` anyway, so you can write them in whatever order you like. More precisely, I'm not sure we want to make implicit argument order a stable guarantee in the majority of lemmas and a decent chunk of definitions.

semorrison:
Okay! I will use `\a := Nat` in mathlib4#7126, and you can close or merge here as you please!

PR #258, opened 2023-09-12
chore: upstream LabelAttr

semorrison:


semorrison:
The file `Std.Tactic.DummyLabelAttr` is obviously annoying, but apparently necessary for testing. Suggestions (even if only a rename) welcome.

Std/Tactic/LabelAttr.lean
```diff
@@ -0,0 +1,95 @@
...
+def registerLabelAttr (attrName : Name) (attrDescr : String)
+    (ref : Name := by exact decl_name%) : IO LabelExtension := do
+  let ext ← mkLabelExt ref
+  mkLabelAttr attrName attrDescr ext ref
+  labelExtensionMapRef.modify fun map => map.insert attrName ext
+  return ext
+
+/--
+Initialize a new "label" attribute.
+Declarations tagged with the attribute can be retrieved using `Mathlib.Tactic.LabelAttr.labelled`.
```
tydeu, line 78:
```suggestion
Declarations tagged with the attribute can be retrieved using `Std.Tactic.LabelAttr.labelled`.
```

Std/Tactic/DummyLabelAttr.lean
digama0, line 1: Let's move this to `Std.Util.DummyLabelAttr`

tydeu: I would usually suggest `Std.Test`. Though that may end up being used for whatever public testing add-ons `Std`  provides. So perhaps `Std.Test.Internal` would be best?

digama0: `Std.Test.Internal` works for me. The point of the `Internal` is to indicate that this may not stick around, for example if we get some other real label attribute we might just use that for testing instead.

PR #259, opened 2023-09-13
chore: add linter for 'import Lean'

semorrison:


semorrison:
I guess we should upstream `#minimize_imports` too, so people don't have to do this by hand. :-)

.github/workflows/build.yml
```diff
@@ -28,24 +28,35 @@ jobs:
...
             echo "Error: Found forbidden character ↦"
             exit 1
           fi
 
       - name: Check for long lines
+        if: always()
         run: |
           ! (find . -name "*.lean" -type f -exec grep -E -H -n '^.{101,}$' {} \; | grep -v -E 'https?://')
+
+      - name: Check for forbidden 'import Lean'
```
digama0, line 59:
```suggestion
      - name: Don't 'import Lean', use precise imports
```
Hopefully this should help clarify why this lint exists

PR #260, opened 2023-09-13
chore: remove simp attributes, add simp lemmas

semorrison:
Companion RFC PR to discussion on [zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/simp.20attributes.20in.20Std/near/390695909).


scripts/nolints.json
```diff
@@ -49,4 +49,5 @@
  ["docBlame", "String.foldrAuxWF.loop"],
  ["docBlame", "String.replaceWF.loop"],
  ["docBlame", "String.substrEqWF.loop"],
- ["docBlame", "Substring.splitOnWF.loop"]]
\ No newline at end of file
+ ["docBlame", "Substring.splitOnWF.loop"],
+ ["simpNF", "List.get_cons_zero"]]
```
digama0, line 53: this should be a `nolint` on the relevant declaration

semorrison: Oops. `nolint` is not available in that file. However this is a leftover from an earlier draft in which `List.get_cons_zero` said something different. Now the linter doesn't complain anyway.

The remaining problem stemming from this lemma is `Fin.zero_eta` which I've currently left stranded at the top of `Std.Data.List.Init.Lemmas`.

Std/Data/Fin/Lemmas.lean
```diff
@@ -29,7 +29,7 @@ theorem pos_iff_nonempty {n : Nat} : 0 < n ↔ Nonempty (Fin n) :=
 
 /-! ### coercions and constructions -/
 
-@[simp] protected theorem eta (a : Fin n) (h : a < n) : (⟨a, h⟩ : Fin n) = a := by cases a; rfl
+@[simp] protected theorem eta (a : Fin n) (h : a < n) : (⟨a, h⟩ : Fin n) = a := rfl
```
semorrison, line 32: It would be great to have a linter for this (`@[simp]` lemmas that could be proved by `rfl` but aren't). It is a headache for users when `simp` refuses to go into `dsimp` mode with lemmas like this.

Std/Data/AssocList.lean
```diff
@@ -3,7 +3,7 @@ Copyright (c) 2019 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
-import Std.Data.List.Basic
```
digama0, line 6: Can you revert this and move the relevant lemmas to `Std.Data.List.Init.Lemmas`?

Std/Data/List/Lemmas.lean
```diff
@@ -1558,6 +1596,18 @@ theorem forIn_eq_bindList [Monad m] [LawfulMonad m]
   induction l generalizing init <;> simp [*, map_eq_pure_bind]
   congr; ext (b | b) <;> simp
 
+-- TODO: Is `List.forM` or `ForM.forM` the simp normal form?
```
digama0, line 1599: I think `List.forM` is the simp-normal form. `ForM.forM` is unused AFAIK

Std/Data/AssocList.lean
semorrison: Done.

Std/Data/Fin/Lemmas.lean
semorrison: I've punted this to a [request](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/linter.20for.20.60rfl.60.20lemmas.3F/near/390817700) on zulip, it's too far a distraction for today!

Std/Data/List/Init/Lemmas.lean
```diff
@@ -15,15 +19,12 @@ These are theorems used in the definitions of `Std.Data.List.Basic`.
 New theorems should be added to `Std.Data.List.Lemmas` if they are not needed by the bootstrap.
 -/
 
-attribute [simp] get reverseAux eraseIdx map join find? findSome?
-  replace elem lookup drop take foldl foldr zipWith unzip range.loop enumFrom
-  intersperse isPrefixOf isEqv dropLast iota mapM.loop mapA List.forM forA filterAuxM
-  filterMapM.loop List.foldlM firstM anyM allM findM? findSomeM? forIn.loop forIn'.loop
-  concat_eq_append append_assoc
```
digama0, line 22: These have not been replaced with the appropriate simp lemmas. For example, instead of `attribute [simp] reverseAux`, there should be:
```lean
@[simp] theorem reverseAux_nil : reverseAux [] r = r := rfl
@[simp] theorem reverseAux_cons : reverseAux (a::l) r = reverseAux l (a::r) := rfl
```
These should also be in `Data.List.Init.Lemmas`, not `Data.List.Lemmas` because the equation lemmas are considered part of the definition itself.

Std/Data/List/Init/Lemmas.lean
```diff
@@ -4,6 +4,10 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Parikshit Khanna, Jeremy Avigad, Leonardo de Moura, Floris van Doorn, Mario Carneiro
 -/
 
+-- TODO: I'm not sure where this should live.
```
jcommelin, line 7: Is `Mathlib/Init/Data/Fin/Basic.lean` a good place?

jcommelin (approved):
bors d+

Std/Data/List/Init/Lemmas.lean
semorrison: @jcommelin, I think you missed this is a PR at Std. :-)

semorrison: Nevertheless the same idea but `Std/Data/Fin/Init.lean` is good.

digama0: the naming convention is `Std.Data.Fin.Init.Basic`

Std/Data/Fin/Init.lean
digama0, line 1: `Std.Data.Fin.Init.Lemmas`

semorrison: Done!

Std/Data/List/Init/Lemmas.lean
```diff
@@ -202,26 +229,172 @@ theorem foldr_eq_foldrM (f : α → β → β) (b) (l : List α) :
...
 @[simp] theorem foldr_self_append (l : List α) : l.foldr cons l' = l ++ l' := by
   induction l <;> simp [*]
 
 theorem foldr_self (l : List α) : l.foldr cons [] = l := by simp
 
+/-! ### mapM -/
+
 /-- Alternate (non-tail-recursive) form of mapM for proofs. -/
-@[simp] def mapM' [Monad m] (f : α → m β) : List α → m (List β)
+def mapM' [Monad m] (f : α → m β) : List α → m (List β)
```
digama0, line 248: this needs equation lemmas

Std/Data/List/Init/Lemmas.lean
```diff
@@ -202,26 +229,172 @@ theorem foldr_eq_foldrM (f : α → β → β) (b) (l : List α) :
...
+/-! ### dropLast -/
+
+@[simp] theorem dropLast_nil : ([] : List α).dropLast = [] := rfl
+@[simp] theorem dropLast_single : [x].dropLast = [] := rfl
+@[simp] theorem dropLast_cons₂ :
+    (x::y::zs).dropLast = x :: (y::zs).dropLast := rfl
+
+-- We may want to replace these `simp` attributes with explicit equational lemmas,
+-- as we already have for all the non-monadic functions.
+attribute [simp] mapA forA List.forM filterAuxM firstM anyM allM findM? findSomeM?
```
digama0, line 395:
```suggestion
attribute [simp] mapA forA filterAuxM firstM anyM allM findM? findSomeM?
```

digama0 (approved):

Std/Data/List/Init/Lemmas.lean
semorrison: Done.

PR #267, opened 2023-09-19
feat: add the json% macro from ProofWidgets4

PatrickMassot:
---

See Zulip discussion at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/json.20elaborator

### DIFF
--- a/Std/Data/Json.lean
+++ b/Std/Data/Json.lean
@@ -11,7 +11,7 @@ import Std.Lean.Json
 Now you can write
 
 ```lean
-open scoped ProofWidgets.Json
+open scoped Std.Json
 
 #eval json% {
   hello : "world",
--- a/Std/Data/Json.lean
+++ b/Std/Data/Json.lean
@@ -1,9 +1,10 @@
 /-
  Copyright (c) 2022 E.W.Ayers. All rights reserved.
  Released under Apache 2.0 license as described in the file LICENSE.
- Authors: E.W.Ayers
+ Authors: E.W.Ayers, Wojciech Nawrocki
 -/
 import Std.Lean.Json
+import Lean.Syntax
 
 /-!
 # JSON-like syntax for Lean.
@@ -11,7 +12,7 @@ import Std.Lean.Json
 Now you can write
 
 ```lean
-open scoped Std.Json
+open Std.Json
 
 #eval json% {
   hello : "world",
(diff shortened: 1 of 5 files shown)


digama0 (changes requested):
Comments/concerns:
* It should be in `Std` not `ProofWidgets`
* The instances at the start are misplaced, and should go in `Std.Lean` or in lean itself
* The syntax should not be scoped, and if this causes problems we need to fix those before merge (mathlib bump test needed)
* The `$(` token conflicts with lean antiquotations

PatrickMassot:
Do you have a suggestion to replace `$(`?

digama0:
ideally it would be `$(`, but some research is needed to determine how to do this without messing up other uses of antiquotations. Perhaps the implementation in Qq could be helpful...

PatrickMassot:
I made the easy changes but the quotation thing may be above my pay grade.

PatrickMassot:
@gebner do you have any insight about reusing the antiquotation syntax coming from Qq?

Std/Data/Json.lean
```diff
@@ -0,0 +1,86 @@
...
+-/
+import Std.Lean.Json
+
+/-!
+# JSON-like syntax for Lean.
+
+Now you can write
+
+```lean
+open scoped ProofWidgets.Json
```
sgouezel, line 14:
```suggestion
open scoped Std.Json
```

Vtec234:
Hello, I improved the antiquotation story [here](https://github.com/EdAyers/ProofWidgets4/commit/9f68f14df384f43b74aeb2908b97ae54a0ad9d95). Feel free to update the PR with those changes. I kept the PW4 variant `scoped` because even with the changes I am not confident that it won't break global syntax. Probably if you remove `scoped` here and both std4/mathlib4 still compile, then it's ok.

gebner:
> do you have any insight about reusing the antiquotation syntax coming from Qq?

Qq uses the same quotation syntax as Lean core.  Wojciech has now changed the json macro in proofwidgets to do the same. ^^

PatrickMassot:
I tried to incorporate changes from ProofWidgets.

Std/Data/Json.lean
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Std.Json
+open Lean
+
+/-- Json data syntax. -/
+declare_syntax_cat jso (behavior := symbol)
+/-- Json data field syntax. -/
+declare_syntax_cat jso_field
+/-- Json identifier syntax. -/
+declare_syntax_cat jso_ident
```
Vtec234, line 36:
```suggestion
```

Std/Data/Json.lean
```diff
@@ -0,0 +1,84 @@
...
+syntax "-"? num : jso
+/-- Json number negation syntax for scientific numbers. -/
+syntax "-"? scientific : jso
+/-- Json array syntax. -/
+syntax "[" jso,* "]" : jso
+/-- Json identifier syntax. -/
+syntax jsoIdent := ident <|> str
+/-- Json key/value syntax. -/
+syntax jsoField := jsoIdent ": " jso
+/-- Json dictionary syntax. -/
```
Vtec234, line 55:
```suggestion
/-- Json object syntax. -/
```

Std/Data/Json.lean
```diff
@@ -0,0 +1,84 @@
...
+  isBug : null,
+  lookACalc: $(23 + 54 * 2)
+}
+```
+-/
+
+namespace Std.Json
+open Lean
+
+/-- Json data syntax. -/
```
Vtec234, line 31:
```suggestion
/-- Json syntactic category -/
```

Vtec234 (approved):

Vtec234:
@digama0 this is good to go I think.

Std/Lean/Json.lean
```diff
@@ -0,0 +1,35 @@
...
+    | none =>
+      match s with
+      | "inf" => "inf" -- [todo] emit a warning
+      | "-inf" => "-inf"
+      | "nan" => "nan"
+      | _ => panic! s!"unhandled float string {s}"
+    | some (m, e, de) =>
+      let j : JsonNumber := OfScientific.ofScientific m e de
+      let j := if x < 0.0 then -j else j
+      Json.num j
```
digama0, line 35: This instance looks super hackish, translating stuff to strings and then back, and constructing invalid JSON literals like `nan`. What is going on here? (If you want to translate invalid floats I think the usual method is to use `null` instead.)

digama0 (approved):

PR #272, opened 2023-09-25
De-mathlib `List.take` and `List.drop` lemmas

Vtec234:


Vtec234:
I could use help with the `simpNF` lint. It reports
```
-- Std.Data.List.Lemmas
./Std/Data/List/Lemmas.lean:526:9: error: List.dropLast_append_cons.{u_1} simp can prove this:
  by simp only [@ne_eq, not_false_eq_true, @List.dropLast_append_of_ne_nil]
```
The lemma `dropLast_append_of_ne_nil` appearing in the simp-set is declared *after* `dropLast_append_cons`, so surely it couldn't be used to prove `dropLast_append_cons`!

semorrison:
The `simpNF` linter is global, and does not care about the order of declarations. The messages could be clearer but it is trying to indicate a redundancy in the simp set --- that `List.dropLast_append_cons` does not need to be marked as @[simp] because the simplifier will already achieve the same result.

Vtec234:
@semorrison ok, that makes sense. I am not sure how to reconcile it with Mario's remark [here](https://github.com/leanprover/std4/pull/89#discussion_r1088422480). Has the feeling about when to mark things `@[simp]` changed since then, or is it that somehow the `simpNF` linter knows that `simp` will be able to derive this when processing an arbitrary subterm?

semorrison:
I *think* the reconciliation is that if you find that `simp` can't actually do this in context, then you add a `@[nolint simpNF]` attribute and a comment explaining it.

But the simpNF linter still scares me, so happy to hear other takes on this!

digama0:
The `simpNF` linter tries to avoid "shortcut" proofs, but they are generally harmless as long as they result in a confluent system of rewrites. (But they may not be applied when you would like them to if you don't fiddle with priorities.) You can use `@[nolint simpNF]` if you think a shortcut rewrite is important.

Std/Data/List/Lemmas.lean
```diff
@@ -760,23 +774,138 @@ theorem getD_eq_get? : ∀ l n (a : α), getD l n a = (get? l n).getD a
...
+theorem take_append_eq_append_take {l₁ l₂ : List α} {n : Nat} :
+    take n (l₁ ++ l₂) = take n l₁ ++ take (n - l₁.length) l₂ := by
+  induction l₁ generalizing n; {simp}
+  cases n <;> simp [*]
 
-theorem take_eq_nil_of_eq_nil : ∀ {as : List α} {i}, as = [] → as.take i = []
-  | _, _, rfl => take_nil
+theorem take_append_of_le_length {l₁ l₂ : List α} {n : Nat} (h : n ≤ l₁.length) :
+    (l₁ ++ l₂).take n = l₁.take n :=
+  by simp [take_append_eq_append_take, Nat.sub_eq_zero_of_le h]
```
digama0, line 825:
```suggestion
    (l₁ ++ l₂).take n = l₁.take n := by
  simp [take_append_eq_append_take, Nat.sub_eq_zero_of_le h]
```

Std/Data/List/Lemmas.lean
```diff
@@ -788,17 +917,162 @@ theorem map_eq_append_split {f : α → β} {l : List α} {s₁ s₂ : List β}
...
+      exact Trans.trans (lih _) (Nat.le_add_left _ _)
+
+/-- The `i + j`-th element of a list coincides with the `j`-th element of the list obtained by
+dropping the first `i` elements. Version designed to rewrite from the big list to the small list. -/
+theorem get_drop (L : List α) {i j : Nat} (h : i + j < L.length) :
+    get L ⟨i + j, h⟩ = get (L.drop i) ⟨j, by
+      have A : i < L.length := Nat.lt_of_le_of_lt (Nat.le.intro rfl) h
+      rw [(take_append_drop i L).symm] at h
+      simpa only [Nat.le_of_lt A, Nat.min_eq_left, Nat.add_lt_add_iff_left, length_take,
+        length_append] using h⟩ := by
```
digama0, line 1008: this should be extracted into a named lemma

Std/Data/List/Pairwise.lean
```diff
@@ -213,7 +213,7 @@ theorem map_get_sublist {l : List α} {is : List (Fin l.length)} (h : is.Pairwis
     simp; cases hl'
     have := IH h.of_cons (hd+1) _ rfl (pairwise_cons.mp h).1
     specialize his hd (.head _)
-    have := get_cons_drop .. ▸ this.cons₂ (get l hd)
+    have := (drop_eq_get_cons ..).symm ▸ this.cons₂ (get l hd)
```
digama0, line 216: is the symm here necessary? `▸` rewrites in both directions so `symm` is usually inferrable

Vtec234: It fails without `symm`.

Vtec234:
Thanks for the review @digama0, it is addressed now.

Vtec234:
Gentle reminder that this is good to go on my side, @digama0

Std/Data/List/Lemmas.lean
```diff
@@ -749,7 +759,11 @@ theorem getD_eq_get? : ∀ l n (a : α), getD l n a = (get? l n).getD a
   | _a::_, 0   => rfl
   | _a::l, n+1 => get!_eq_getD l n
 
-/-! ### take and drop -/
+/-! ### take -/
+
+@[simp]
+theorem take_cons (n) (a : α) (l : List α) : take (succ n) (a :: l) = a :: take n l :=
```
fgdorais, line 765: This is essentially the same as the next theorem. There is a difference but the names should be swapped for when that distinction matters.

Std/Data/List/Lemmas.lean
```diff
@@ -760,23 +774,138 @@ theorem getD_eq_get? : ∀ l n (a : α), getD l n a = (get? l n).getD a
...
+/-- Taking the first `n` elements in `l₁ ++ l₂` is the same as appending the first `n` elements
+of `l₁` to the first `n - l₁.length` elements of `l₂`. -/
+theorem take_append_eq_append_take {l₁ l₂ : List α} {n : Nat} :
+    take n (l₁ ++ l₂) = take n l₁ ++ take (n - l₁.length) l₂ := by
+  induction l₁ generalizing n; {simp}
+  cases n <;> simp [*]
 
-theorem take_eq_nil_of_eq_nil : ∀ {as : List α} {i}, as = [] → as.take i = []
-  | _, _, rfl => take_nil
+theorem take_append_of_le_length {l₁ l₂ : List α} {n : Nat} (h : n ≤ l₁.length) :
```
fgdorais, line 823: Shorter and less confusing?
```suggestion
theorem take_append_of_le {l₁ l₂ : List α} {n : Nat} (h : n ≤ l₁.length) :
```

fgdorais: Maybe add 
```
take_append_of_ge {l₁ l₂ : List α} {n : Nat} (h : n ≥ l₁.length) : take n (l₁ ++ l₂) = l₁ ++ take (n-l₁.length) l₂ := sorry
```

Std/Data/List/Lemmas.lean
```diff
@@ -788,17 +917,164 @@ theorem map_eq_append_split {f : α → β} {l : List α} {s₁ s₂ : List β}
...
 /-- Dropping the elements up to `n` in `l₁ ++ l₂` is the same as dropping the elements up to `n`
 in `l₁`, dropping the elements up to `n - l₁.length` in `l₂`, and appending them. -/
-theorem drop_append_eq_append_drop {l₁ l₂ : List α} :
+theorem drop_append_eq_append_drop {l₁ l₂ : List α} {n : Nat} :
     drop n (l₁ ++ l₂) = drop n l₁ ++ drop (n - l₁.length) l₂ := by
   induction l₁ generalizing n; · simp
   cases n <;> simp [*]
 
-theorem drop_append_of_le_length {l₁ l₂ : List α} (h : n ≤ l₁.length) :
+theorem drop_append_of_le_length {l₁ l₂ : List α} {n : Nat} (h : n ≤ l₁.length) :
```
fgdorais, line 982: Same as `take` comments above.

Std/Data/List/Lemmas.lean
Vtec234: I removed the `succ` variant.

Std/Data/List/Lemmas.lean
Vtec234: I would find the shorter version more confusing since it doesn't say what the `le` relation is on. The latter lemma is already `take_append_eq_append_take`.

Std/Data/List/Lemmas.lean
Vtec234: Similarly here, I would say specifying what is less-equal is helpful.

fgdorais (approved):

PR #274, opened 2023-09-26
feat: add non-monotone map function to RBSet

lambda-fairy:
The rename makes Lean more consistent with OCaml and Haskell, where the default `Set.map` is not monotone.

- https://v2.ocaml.org/api/Set.S.html#VALmap
- https://hackage.haskell.org/package/containers-0.6.7/docs/Data-Set.html#g:10

Note that mathlib's `Finset` also has separate `map` and `image` functions, with `map` requiring extra conditions (injectivity). I think that reasoning doesn't apply for `RBSet` as the non-monotone case is much more common, and in this case consistency with Haskell/OCaml is more valuable than with mathlib.

### DIFF
--- a/Std/Data/RBMap/Basic.lean
+++ b/Std/Data/RBMap/Basic.lean
@@ -835,6 +835,13 @@ def filter (t : RBSet α cmp) (p : α → Bool) : RBSet α cmp :=
   t.foldl (init := ∅) fun acc a => bif p a then acc.insert a else acc
 
 /--
+`O(n * log n)`. Map a function on every value in the set.
+If the function is monotone, consider using the more efficient `RBSet.mapMonotone` instead.
+-/
+def map (t : RBSet α cmpα) (f : α → β) : RBSet β cmpβ :=
+  t.foldl (fun acc a => acc.insert <| f a) (mkRBSet β cmpβ)
+
+/--
 `O(n₁ * (log n₁ + log n₂))`. Constructs the set of all elements of `t₁` that are not in `t₂`.
 -/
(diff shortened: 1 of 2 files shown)


semorrison:
Why would `RBSet.image` be a strange name? I think consistency in the API is valuable here.

lambda-fairy:
I guess that while I've seen "image" a lot in maths, I haven't seen it used in a programming context. So it feels strange from that perspective.

Then again, I'm not sure where Lean stands on this point, so if we feel strongly about `RBSet.image` then I can revert the rename.

lambda-fairy:
I reverted the rename.

lambda-fairy:
awaiting-review

joehendrix:
Coming from Haskell, I liked the original `mapMonotone`/`map`, but this may be bike shedding.  Otherwise, the function itself seems like it'd be a useful addition.

joehendrix:
@semorrison What are your thoughts on naming consistency here?  Coming from a CS background `mapMonotone/map` seem more conventional to me.

semorrison:
Looking at it again, I agree `map` and `mapMonotone` are better, and the argument for consistency with Mathlib is quite weak.

lambda-fairy:
... just when I started coming around to `image`. It seems like my opinion is phase-shifted from Scott's 😆

@joehendrix @semorrison I've reverted it back. Would you be happy to merge? I don't feel strongly either way, so feel free to chop off the final commit yourself if you change your mind again.

joehendrix (approved):

lambda-fairy:
Thank you @joehendrix!

PR #275, opened 2023-09-28
feat: some lemmas about variations List.get and List.getLast, fix duplicate lemma

javra:


semorrison:

semorrison:
awaiting-review

fgdorais:
These changes are now all in Lean core.

PR #276, opened 2023-10-01
fix: make Fin.val_succ `dsimp`

timotree3:


semorrison:

semorrison:
awaiting-review

PR #277, opened 2023-10-01
feat: add varous lemmas for (Fin.rev _)

timotree3:
These came up in [Zulip](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/cons.20is.20to.20snoc.20as.20Fin.2Eeq_zero_or_eq_succ.20is.20to). Let me know if there's a better place to put these lemmas.

Also, I'd be happy to mark as `@[simp]`, but I wasn't sure if we wanted to commit to pushing `rev`s in or out as simp-normal form. Pushing them in (as these lemmas do) seems natural as it's what we do with negation, but `k.rev` is `-1 - k`, not `-k`, so I'm not sure what to think.


semorrison:

semorrison:
awaiting-review

joehendrix (approved):

semorrison (approved):

PR #279, opened 2023-10-05
Add HashMap model

Vtec234:
This adds a function `Buckets.toListModel`. It models an array of buckets as a list of (key, value) pairs. We prove a number of lemmas about the behaviour of this model w.r.t. various map operations. No further propositions are proven in this PR, in particular it doesn't add any facts about the relations between user-facing HashMap operations just yet.

Depends on #89 and #272.

### DIFF
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -14,19 +14,22 @@ class LawfulHashable (α : Type _) [BEq α] [Hashable α] : Prop where
   /-- Two elements which compare equal under the `BEq` instance have equal hash. -/
   hash_eq {a b : α} : a == b → hash a = hash b
 
+instance [BEq α] [LawfulBEq α] [Hashable α] : LawfulHashable α where
+  hash_eq h := by rw [LawfulBEq.eq_of_beq h]
+
 namespace Imp
 
 /--
 The bucket array of a `HashMap` is a nonempty array of `AssocList`s.
 (This type is an internal implementation detail of `HashMap`.)
 -/
-def Buckets (α : Type u) (β : Type v) := {b : Array (AssocList α β) // 0 < b.size}
+def Buckets (α : Type u) (β : Type v) := {b : Array (AssocList α β) // b.size.isPowerOfTwo}
 
 namespace Buckets
 
-/-- Construct a new empty bucket array with the specified capacity. -/
-def mk (buckets := 8) (h : 0 < buckets := by decide) : Buckets α β :=
-  ⟨mkArray buckets .nil, by simp [h]⟩
+/-- Construct a new empty bucket array with the specified number of buckets. -/
+def mk (nBuckets : Nat) (h : nBuckets.isPowerOfTwo) : Buckets α β :=
+  ⟨mkArray nBuckets .nil, by simp [h]⟩
 
 /-- Update one bucket in the bucket array with a new value. -/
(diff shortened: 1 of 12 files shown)


Std/Data/HashMap/Basic.lean
```diff
@@ -80,28 +83,24 @@ A "load factor" of 0.75 is the usual standard for hash maps, so we return `capac
...
 
 /-- Constructs an empty hash map with the specified target capacity. -/
-def empty (capacity := 0) : Imp α β :=
-  let nbuckets := numBucketsForCapacity capacity
-  let n : {n : Nat // 0 < n} :=
-    if h : nbuckets = 0 then ⟨8, by decide⟩
-    else ⟨nbuckets, Nat.zero_lt_of_ne_zero h⟩
-  empty' n n.2
+def empty (capacity := 8) : Imp α β :=
+  let nBuckets := numBucketsForCapacity capacity |>.nextPowerOfTwo
```
Vtec234, line 91: Note that this changes the behaviour when `empty` is called with `capacity := 0`, but I restore it in the default case by setting `capacity := 8`.

Std/Logic.lean
```diff
@@ -145,5 +158,8 @@ theorem subsingleton_of_forall_eq (x : α) (h : ∀ y, y = x) : Subsingleton α
 theorem subsingleton_iff_forall_eq (x : α) : Subsingleton α ↔ ∀ y, y = x :=
   ⟨fun _ y => Subsingleton.elim y x, subsingleton_of_forall_eq x⟩
 
+example [Subsingleton α] (p : α → Prop) : Subsingleton (Subtype p) :=
+  ⟨fun ⟨x, _⟩ ⟨y, _⟩ => by congr; exact Subsingleton.elim x y⟩
```
semorrison, line 162: Why is this an example?

Std/Logic.lean
```diff
@@ -125,6 +125,19 @@ theorem ne_of_mem_of_not_mem' (h : a ∈ s) : a ∉ t → s ≠ t := mt fun e =>
 
 end Mem
 
+/-! ## Bool -/
+
+theorem Bool.beq_eq_false_iff [BEq α] {a a' : α} :
+    (a == a') = false ↔ a != a' := by
+  simp only [bne, not_eq_true']
+
+@[simp] theorem Bool.bne_eq_false_iff [BEq α] {a a' : α} :
```
semorrison, line 134: This will compete with
```
@[simp] theorem bne_eq_false_iff_eq [BEq α] [LawfulBEq α] (a b : α) :
```
in Lean.

Std/Data/List/Lemmas.lean
```diff
@@ -1123,6 +1123,13 @@ theorem map_drop {α β : Type u} (f : α → β) :
     dsimp
     rw [map_drop f t]
 
+theorem drop_ext (l₁ l₂ : List α) (j : Nat)
+    : (∀ i ≥ j, l₁.get? i = l₂.get? i) → l₁.drop j = l₂.drop j := by
```
semorrison, line 1127:
```suggestion
theorem drop_ext (l₁ l₂ : List α) (j : Nat) : 
    (∀ i ≥ j, l₁.get? i = l₂.get? i) → l₁.drop j = l₂.drop j := by
```

semorrison:
Could we split this up? I'd happily merge the Std/Data/List part immediately, and that then makes the rest easier.

PR #281, opened 2023-10-09
fix: spell `List.Is{Pre,Suf,In}fix` with a capital `I`

eric-wieser:


semorrison:
There wasn't a matching Mathlib PR, but obviously this is going to break things. There should have been, and ideally it would have been linked from here, and merged immediately after this was merged.

I think I've made most of the changes in #7602 now. (Which I needed in order to test a Lean 4 against Mathlib.)

eric-wieser:
Ah, I wasnt fully aware of that process. Could you add some documentation to that effect to `CONTRIBUTING.md`?

PR #283, opened 2023-10-10
chore: adapt to leanprover/lean4#2644

semorrison:
This should wait until leanprover/lean4#2644 hits a release.


semorrison:
No longer needed after changes to the upstream PR.

PR #285, opened 2023-10-13
feat: Upstream more `DiscrTree` APIs

nomeata:
while working on https://github.com/leanprover/lean4/pull/2577 I noticed
that functions on `DiscrTree`s are scattered throughout std4, aesop and
mathlib4. This PR upstreams those functions look at the internals of
`DiscrTree`, in particular

* `mapArrays` from mathlib
  <https://github.com/leanprover-community/mathlib4/blob/c00742dc87c9656c6a5502317d75344c0284751b/Mathlib/Lean/Meta/DiscrTree.lean#L79-L81>
* `filterFoldM` from aesop
  <https://github.com/JLimperg/aesop/blob/831865d91f9dcc8ef9295d6d6053606766cdb6a3/Aesop/Util/Basic.lean#L123-L138>
  (renamed to put the `Trie`/`DiscrTree` into the namespace)

The other function in mathlib that operate on `DiscrTree`, is a reimplementation of `.values`.

Also drop the reimplementation `.elements` in std4.

Collecting all operations in one place not only avoids such reimplementations (hopefully), but
also makes it easier to refactor the internals of the data structure.

### DIFF
--- a/Std/Lean/Meta/DiscrTree.lean
+++ b/Std/Lean/Meta/DiscrTree.lean
@@ -88,6 +88,12 @@ partial def size : Trie α s → Nat
     children.foldl (init := vs.size) fun n (_, c) => n + size c
 
 /--
+Whether the `Trie` is empty.
+-/
+partial def isEmpty : Trie α s → Bool
+  | .node vs children => vs.isEmpty && children.all fun (_, c) => c.isEmpty
+
+/--
 Merge two `Trie`s. Duplicate values are preserved.
 -/
 partial def mergePreservingDuplicates : Trie α s → Trie α s → Trie α s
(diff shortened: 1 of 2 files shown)


Std/Lean/Meta/DiscrTree.lean
```diff
@@ -101,6 +107,57 @@ where
...
+  | empty : Ctxt
+  | ctxt : Array (Key s × Trie β s) → Array β → Array (Key s × Trie α s) → Key s → Ctxt → Ctxt
+
+/-- Apply a function to the array of values at each node in a `DiscrTree`. -/
+partial def mapArrays (t : Trie α s) (f : Array α → Array β) : Trie β s :=
+  let .node vs0 cs0 := t
+  go (.mkEmpty cs0.size) (f vs0) cs0.reverse Ctxt.empty
+where
+  /-- This implementation as a single tail-recursive function is chosen to not blow the
+      interpreter stack when the `Trie` is very deep -/
```
digama0, line 121: This is not a doc comment. This should be a plain comment and the doc comment should say what the function does. In this case I would try to write a theorem which defines `go` non-recursively in terms of `mapArrays` as a means of specification (see the doc comments on `List.setTR.go` and other worker-wrapper functions in Std.Data.List.Basic for examples).

Std/Lean/Meta/DiscrTree.lean
```diff
@@ -87,6 +87,12 @@ partial def size : Trie α s → Nat
   | Trie.node vs children =>
     children.foldl (init := vs.size) fun n (_, c) => n + size c
 
+/--
+Whether the `Trie` is empty.
+-/
```
digama0, line 92:
```suggestion
/-- Whether the `Trie` is empty. -/
-- TODO: remove partial
```

Std/Lean/Meta/DiscrTree.lean
```diff
@@ -101,6 +107,57 @@ where
...
+private partial def filterFoldM [Monad m] [Inhabited σ] (f : σ → α → m σ)
+    (p : α → m (ULift Bool)) (init : σ) : Trie α s → m (Trie α s × σ)
+  | .node vs children => do
+    let (vs, acc) ← vs.foldlM (init := (#[], init)) λ (vs, acc) v => do
+      if (← p v).down then
+        return (vs.push v, acc)
+      else
+        return (vs, ← f acc v)
+    let (children, acc) ← go acc 0 children
+    let children := children.filter λ (_, c) => ! c.isEmpty
```
digama0, line 148:
```suggestion
    let children := children.filter fun (_, c) => !c.isEmpty
```
ditto elsewhere: use `fun` in place of `λ`, no space after `!`

Std/Lean/Meta/DiscrTree.lean
```diff
@@ -155,10 +212,37 @@ Get the number of values stored in a `DiscrTree`. O(n) in the size of the tree.
...
 def mergePreservingDuplicates (t u : DiscrTree α s) : DiscrTree α s :=
   ⟨t.root.mergeWith u.root fun _ trie₁ trie₂ =>
     trie₁.mergePreservingDuplicates trie₂⟩
+
+/-- Apply a function to the array of values at each node in a `DiscrTree`. -/
+def mapArrays (d : DiscrTree α s) (f : Array α → Array β) : DiscrTree β s :=
+  { root := d.root.map (fun t => t.mapArrays f) }
+
+/--
+Remove elements for which `p` returns `false` from the given `DiscrTrie`.
```
digama0, line 234:
```suggestion
Remove elements for which `p` returns `false` from the given `DiscrTree`.
```

Std/Lean/Meta/DiscrTree.lean
nomeata: It was a plain comment before :-)
I only changed this because the doc linter insisted one one, despite this being a local definition. Is it a bug in the linter that it insists on docstrings on `where`'d definitions?

nomeata: I’d rather not invest much into this code (e.g. theorems relating to other specifications) if I get to dump it anyways soon if https://github.com/leanprover/lean4/pull/2577 can land. But I addressed the docstring issue.

Std/Lean/Meta/DiscrTree.lean
nomeata: It’s as `partial` as the code around it, should I add `TODO`s to the existing functions (`size` for example)?

semorrison: Maybe a TODO at the top of the file will suffice, given how many there are.

semorrison:

semorrison:
awaiting-review

Std/Lean/Meta/DiscrTree.lean
nomeata: Added a TODO.

Incidentially, how would I go about unpartializing these functions? They are “morally” structurally recursive, but lean doesn’t see that. So does this need to go via well-founded recursion? If so, do we have to define `Array.attach` (analogous to `List.attach` here, and use it in all these function definitions?

Actually, let me take this to Zulip, I’ll summarize the answer where.

nomeata: In https://github.com/leanprover/std4/pull/332 I am exploring how to make these definitions not `unsafe`; the PRs are independent.

joehendrix:
I updated tags to reflect discussion mentioning #332.

PR #286, opened 2023-10-14
feat: add basic `BitVec` library

leodemoura:
This library was started by Joe Hendrix and Wojciech Nawrocki. It is crucial for software and hardware verification projects. We need it in `Std`. Moreover `lean-auto` project needs it.

### DIFF
--- a/Std/Data/BitVec.lean
+++ b/Std/Data/BitVec.lean
@@ -1,7 +1,1 @@
-/-
-Copyright (c) 2022 by the authors listed in the file AUTHORS and their
-institutional affiliations. All rights reserved.
-Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura
--/
 import Std.Data.BitVec.Basic
--- a/Std/Data/BitVec.lean
+++ b/Std/Data/BitVec.lean
@@ -1,0 +1,7 @@
+/-
+Copyright (c) 2022 by the authors listed in the file AUTHORS and their
+institutional affiliations. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura
+-/
+import Std.Data.BitVec.Basic
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -56,7 +56,6 @@ protected def toNat (a : BitVec n) : Nat := a.toFin.val
 /-- Return the `i`-th least significant bit. -/
 @[inline] def getMsb (x : BitVec w) (i : Nat) : Bool := getLsb x (w-1-i)
 
-
 /-- Return most-significant bit in bitvector. -/
 @[inline] protected def msb (a : BitVec n) : Bool := getMsb a 0
 
@@ -66,7 +65,7 @@ protected def ofInt (n : Nat) (i : Int) : BitVec n :=
   | Int.ofNat a => .ofNat n a
   | Int.negSucc a => .ofNat n (2^n - 1 - a % 2^n)
 
-/-- Given a bitvector -/
+/-- Interpret the bitvector as an integer stored in two's complement form. -/
 protected def toInt (a : BitVec n) : Int :=
   if a.msb then Int.ofNat a.toNat - Int.ofNat (2^n) else a.toNat
 
@@ -78,13 +77,12 @@ instance : Inhabited (BitVec n) where default := .zero n
 instance : OfNat (BitVec n) i where ofNat := .ofNat n i
 
 /-- Notation for bit vector literals. `i#n` is a shorthand for `BitVec.ofNat n i`. -/
--- TODO: is this a good notation? Do we need it? We can also write `(i : BitVec n)`
-scoped syntax:max num noWs "#" noWs num : term
-macro_rules | `($i:num#$n:num) => `(BitVec.ofNat $n $i)
+scoped syntax:max term:max noWs "#" noWs term:max : term
+macro_rules | `($i#$n) => `(BitVec.ofNat $n $i)
 
 /-- Unexpander for bit vector literals. -/
 @[app_unexpander BitVec.ofNat] def unexpandBitVecOfNat : Lean.PrettyPrinter.Unexpander
-  | `($(_) $n:num $i:num) => `($i:num#$n:num)
+  | `($(_) $n $i) => `($i#$n)
   | _ => throw ()
 
 /-- Convert bitvector into a fixed-width hex number. -/
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -429,6 +429,3 @@ SMT-Lib name: `sign_extend`.
 -/
 def signExtend (i : Nat) (x : BitVec w) : BitVec (w+i) :=
   Nat.add_comm .. ▸ (fill i x.msb ++ x)
-
-#print Fin.cast
-def cast {n : Nat} : u = v -> BitVec u -> BitVec v
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -2,7 +2,7 @@
 Copyright (c) 2022 by the authors listed in the file AUTHORS and their
 institutional affiliations. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura
+Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura, Mario Carneiro
 -/
 import Std.Data.Nat.Init.Lemmas
 
@@ -24,30 +24,34 @@ A bitvector of the specified width. This is represented as the underlying `Nat` 
 in both the runtime and the kernel, inheriting all the special support for `Nat`.
 -/
 structure BitVec (w : Nat) where
-  /-- Bitvector is represented as a `Fin`. -/
+  /-- Construct a `BitVec w` from a number less than `2^w`.
+  O(1), because we use `Fin` as the internal representation of a bitvector. -/
+  mk ::
+  /-- Interpret a bitvector as a number less than `2^w`.
+  O(1), because we use `Fin` as the internal representation of a bitvector. -/
   val : Fin (2^w)
   deriving DecidableEq
 
 namespace BitVec
 
-/-- The BitVector `i mod 2^n`. -/
+/-- The `BitVector` with value `i mod 2^n`. Treated as an operation on bitvectors,
+this is truncation of the high bits when downcasting and zero-extension when upcasting. -/
 protected def ofNat (n : Nat) (i : Nat) : BitVec n where
   val := Fin.ofNat' i (Nat.pow_two_gt_zero _)
 
-/-- Given a bitvector `a`, return the underlying `Nat`. -/
+/-- Given a bitvector `a`, return the underlying `Nat`. This is O(1) because `BitVec` is a
+(zero-cost) wrapper around a `Nat`. -/
 protected def toNat (a : BitVec n) : Nat := a.val.val
 
-/-- Return a bitvector `0` of size `n`.  -/
+/-- Return a bitvector `0` of size `n`. This is the bitvector with all zero bits. -/
 protected def zero (n : Nat) : BitVec n := .ofNat n 0
 
-instance : Inhabited (BitVec n) where default := BitVec.zero n
+instance : Inhabited (BitVec n) where default := .zero n
 
-instance : OfNat (BitVec n) i where ofNat := BitVec.ofNat n i
+instance : OfNat (BitVec n) i where ofNat := .ofNat n i
 
-/--
-Notation for bit vector literals. `i#n` is a shorthand for `BitVec.ofNat n i`.
-TODO: is this a good notation? Do we need it? We can also write `(i : BitVec n)`
--/
+/-- Notation for bit vector literals. `i#n` is a shorthand for `BitVec.ofNat n i`. -/
+-- TODO: is this a good notation? Do we need it? We can also write `(i : BitVec n)`
 scoped syntax:max num noWs "#" noWs num : term
 macro_rules | `($i:num#$n:num) => `(BitVec.ofNat $n $i)
 
@@ -56,145 +60,154 @@ macro_rules | `($i:num#$n:num) => `(BitVec.ofNat $n $i)
   | `($(_) $n:num $i:num) => `($i:num#$n:num)
   | _ => throw ()
 
-instance : Repr (BitVec n) where
-  reprPrec a _ := repr a.val ++ "#" ++ repr n
+instance : Repr (BitVec n) where reprPrec a _ := repr a.val ++ "#" ++ repr n
 
-instance : ToString (BitVec n) where
-  toString a := toString (repr a)
+instance : ToString (BitVec n) where toString a := toString (repr a)
 
 /-- Theorem for normalizing the bit vector literal representation. -/
+-- TODO: This needs more usage data to assess which direction the simp should go.
 @[simp] theorem ofNat_eq_ofNat : @OfNat.ofNat (BitVec n) i _ = BitVec.ofNat n i := rfl
 
 /--
-Addition for bit vectors.
+Addition for bit vectors. This can be interpreted as either signed or unsigned addition
+modulo `2^n`.
 
 SMT-Lib name: `bvadd`.
 -/
 protected def add (x y : BitVec n) : BitVec n where val := x.val + y.val
-
-/-- Subtraction for bit vectors. -/
-protected def sub (x y : BitVec n) : BitVec n where val := x.val - y.val
+instance : Add (BitVec n) := ⟨BitVec.add⟩
 
 /--
-Negation for bit vectors.
+Subtraction for bit vectors. This can be interpreted as either signed or unsigned subtraction
+modulo `2^n`.
+-/
+protected def sub (x y : BitVec n) : BitVec n where val := x.val - y.val
+instance : Sub (BitVec n) := ⟨BitVec.sub⟩
+
+/--
+Negation for bit vectors. This can be interpreted as either signed or unsigned negation
+modulo `2^n`.
 
 SMT-Lib name: `bvneg`.
 -/
 protected def neg (x : BitVec n) : BitVec n := .sub 0 x
+instance : Neg (BitVec n) := ⟨.neg⟩
 
 /-- Bit vector of size `n` where all bits are `1`s -/
-protected def allOnes (n : Nat) : BitVec n := .neg 1
+def allOnes (n : Nat) : BitVec n := -1
 
 /--
-Multiplication for bit vectors.
+Multiplication for bit vectors. This can be interpreted as either signed or unsigned negation
+modulo `2^n`.
 
 SMT-Lib name: `bvmul`.
 -/
 protected def mul (x y : BitVec n) : BitVec n where val := x.val * y.val
+instance : Mul (BitVec n) := ⟨BitVec.mul⟩
 
 /--
-Modulo for bit vectors.
+Unsigned modulo for bit vectors.
 
 SMT-Lib name: `bvurem`.
 -/
 protected def mod (x y : BitVec n) : BitVec n where val := x.val % y.val
+instance : Mod (BitVec n) := ⟨BitVec.mod⟩
 
 /--
-Division for bit vectors.
+Unsigned division for bit vectors using the Lean convention where division by zero returns zero.
+-/
+protected def div' (x y : BitVec n) : BitVec n where val := x.val / y.val
+
+/--
+Unsigned division for bit vectors using the
+[SMT-Lib convention](http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml)
+where division by zero returns the `allOnes` bitvector.
 
 SMT-Lib name: `bvudiv`.
-
-Note that `div x 0 = allOnes n` as specified by the
-[SMT-Lib standard](http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml).
-This is also consistent with a simple division circuit.
 -/
-protected def div (x y : BitVec n) : BitVec n :=
-  if y = BitVec.zero n then
-    BitVec.allOnes n
-  else
-    { val := x.val / y.val }
-/-- Division for bit vectors using the Lean convention where division by zero returns zero. -/
-protected def div' (x y : BitVec n) : BitVec n := { val := x.val / y.val }
+protected def div (x y : BitVec n) : BitVec n := if y = 0 then -1 else .div' x y
+instance : Div (BitVec n) := ⟨.div⟩
 
 /--
-Less than for bit vectors.
+Unsigned less-than for bit vectors.
 
 SMT-Lib name: `bvult`.
 -/
 protected def lt (x y : BitVec n) : Bool := x.val < y.val
+instance : LT (BitVec n) where lt x y := x.val < y.val
 
-/-- Less than or equal to for bit vectors. -/
+/-- Unsigned less-than-or-equal-to for bit vectors. -/
 protected def le (x y : BitVec n) : Bool := x.val ≤ y.val
-
-instance : Add (BitVec n) := ⟨BitVec.add⟩
-instance : Sub (BitVec n) := ⟨BitVec.sub⟩
-instance : Mul (BitVec n) := ⟨BitVec.mul⟩
-instance : Mod (BitVec n) := ⟨BitVec.mod⟩
-instance : Div (BitVec n) := ⟨BitVec.div⟩
-instance : Neg (BitVec n) := ⟨BitVec.neg⟩
-instance : LT (BitVec n)  := ⟨fun x y => BitVec.lt x y⟩
-instance : LE (BitVec n)  := ⟨fun x y => BitVec.le x y⟩
+instance : LE (BitVec n) where le x y := x.val ≤ y.val
 
 /--
-Bitwise and for bit vectors.
+Bitwise AND for bit vectors.
 
 SMT-Lib name: `bvand`.
 -/
-protected def and (x y : BitVec n) : BitVec n := { val := x.val &&& y.val }
+protected def and (x y : BitVec n) : BitVec n where val := x.val &&& y.val
+instance : AndOp (BitVec w) := ⟨.and⟩
 
 /--
-Bitwise or for bit vectors.
+Bitwise OR for bit vectors.
 
 SMT-Lib name: `bvor`.
 -/
-protected def or (x y : BitVec n) : BitVec n := { val := x.val ||| y.val }
+protected def or (x y : BitVec n) : BitVec n where val := x.val ||| y.val
+instance : OrOp (BitVec w) := ⟨.or⟩
 
-/-- Bitwise xor for bit vectors. -/
-protected def xor (x y : BitVec n) : BitVec n := { val := x.val ^^^ y.val }
+/-- Bitwise XOR for bit vectors. -/
+protected def xor (x y : BitVec n) : BitVec n where val := x.val ^^^ y.val
+instance : Xor (BitVec w) := ⟨.xor⟩
 
 /--
-Complement for bit vectors.
+Bitwise NOT for bit vectors.
 
 SMT-Lib name: `bvnot`.
 -/
 protected def not (x : BitVec n) : BitVec n := -(x + .ofNat n 1)
+instance : Complement (BitVec w) := ⟨.not⟩
 
 /--
-Shift left for bit vectors.
+Left shift for bit vectors. The low bits are filled with zeros. As a numeric operation, this is
+equivalent to `a * 2^s`, modulo `2^n`.
 
 SMT-Lib name: `bvshl`.
 -/
 protected def shiftLeft (a : BitVec n) (s : Nat) : BitVec n := .ofNat n (a.toNat <<< s)
+instance : HShiftLeft (BitVec w) Nat (BitVec w) := ⟨.shiftLeft⟩
 
 /--
-Shift right for bit vectors.
+(Logical) right shift for bit vectors. The high bits are filled with zeros.
+As a numeric operation, this is equivalent to `a / 2^s`, rounding down.
 
 SMT-Lib name: `bvlshr`.
 -/
 protected def shiftRight (a : BitVec n) (s : Nat) : BitVec n := .ofNat n (a.toNat >>> s)
-
-instance : Complement (BitVec w) := ⟨.not⟩
-instance : AndOp (BitVec w) := ⟨.and⟩
-instance : OrOp (BitVec w) := ⟨.or⟩
-instance : Xor (BitVec w) := ⟨.xor⟩
-instance : HShiftLeft (BitVec w) Nat (BitVec w) := ⟨.shiftLeft⟩
 instance : HShiftRight (BitVec w) Nat (BitVec w) := ⟨.shiftRight⟩
 
-/-- Rotate left. -/
+/--
+Rotate left for bit vectors. All the bits of `x` are shifted to higher positions, with the top `n`
+bits wrapping around to fill the low bits.
+-/
 def rotateLeft (x : BitVec w) (n : Nat) : BitVec w := x <<< n ||| x >>> (w - n)
 
-/-- Rotate right. -/
+/--
+Rotate right for bit vectors. All the bits of `x` are shifted to lower positions, with the
+bottom `n` bits wrapping around to fill the high bits.
+-/
 def rotateRight (x : BitVec w) (n : Nat) : BitVec w := x >>> n ||| x <<< (w - n)
 
 /--
-Append.
+Concatenation of bitvectors. This uses the "big endian" convention that the more significant
+input is on the left, so `0xAB#8 ++ 0xCD#8 = 0xABCD#16`.
 
 SMT-Lib name: `concat`.
 -/
 def append (msbs : BitVec n) (lsbs : BitVec m) : BitVec (n+m) :=
-  .ofNat (n+m) (msbs.toNat <<< m + lsbs.toNat)
+  .ofNat (n + m) (msbs.toNat <<< m + lsbs.toNat)
 
-instance : HAppend (BitVec w) (BitVec v) (BitVec (w+v)) := ⟨.append⟩
+instance : HAppend (BitVec w) (BitVec v) (BitVec (w + v)) := ⟨.append⟩
 
 /--
 Extraction of bits `start` to `start + len - 1` from a bit vector of size `n` to yield a
@@ -211,7 +224,7 @@ SMT-Lib name: `extract`.
 -/
 def extract (hi lo : Nat) (a : BitVec n) : BitVec (hi - lo + 1) := extract' lo _ a
 
-/-- `replicate i x` means concatenate `i` copies of `x`. -/
+/-- `replicate i x` concatenates `i` copies of `x` into a new vector of length `w*i`. -/
 def replicate : (i : Nat) → BitVec w → BitVec (w*i)
   | 0,   _ => 0
   | n+1, x =>
@@ -219,16 +232,22 @@ def replicate : (i : Nat) → BitVec w → BitVec (w*i)
       rw [Nat.mul_add, Nat.add_comm, Nat.mul_one]
     hEq ▸ (x ++ replicate n x)
 
-/-- Zero extension. -/
+/-- Zero extend a vector of length `w` with `i` additional zeros in the high bits. -/
 def zeroExtend (i : Nat) (x : BitVec w) : BitVec (w+i) := Nat.add_comm .. ▸ ((0 : BitVec i) ++ x)
 
-/-- Signed extension. Left-pads with copies of the left-most bit. -/
+/--
+Sign extend a vector of length `w`, extending with `i` additional copies of the most significant
+bit in `x`. If `x` is an empty vector, then the sign is treated as zero.
+-/
 def signExtend (i : Nat) (x : BitVec w) : BitVec (w+i) :=
   have hEq : ((w-1) - (w-1) + 1)*i + w = w+i := by
     rw [Nat.sub_self, Nat.zero_add, Nat.one_mul, Nat.add_comm]
   hEq ▸ ((replicate i (extract (w-1) (w-1) x)) ++ x)
 
-/-- Return a prefix of size `v` of bit vector `x`. -/
+/--
+Truncate the high bits of bitvector `x` of length `w`, resulting in a vector of length `v`.
+If `v > w` then it zero-extends the vector instead.
+-/
 def shrink (v : Nat) (x : BitVec w) : BitVec v :=
   if hZero : 0 < v then
     have hEq : v - 1 + 0 + 1 = v := by
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -9,7 +9,7 @@ namespace Std
 We define bitvectors. We choose the `Fin` representation over others for its relative efficiency
 (Lean has special support for `Nat`), alignment with `UIntXY` types which are also represented
 with `Fin`, and the fact that bitwise operations on `Fin` are already defined. Some other possible
-representations are `List Bool`, `{ l : List Bool // l.length = w}`, `Fin w → Bool`.
+representations are `List Bool`, `{ l : List Bool // l.length = w }`, `Fin w → Bool`.
 
 We also define many bitvector operations from the
 [`QF_BV` logic](https://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV).
@@ -54,7 +54,7 @@ instance : OfNat (BitVec n) i where
 Notation for bit vector literals. `i#n` is a shorthand for `BitVec.ofNat n i`.
 TODO: is this a good notation? Do we need it? We can also write `(i : BitVec n)`
 -/
-syntax:max num noWs "#" noWs num : term
+scoped syntax:max num noWs "#" noWs num : term
 macro_rules | `($i:num#$n:num) => `(BitVec.ofNat $n $i)
 
 /-- Unexpander for bit vector literals. -/
@@ -140,6 +140,7 @@ def append (a : BitVec n) (b : BitVec m) : BitVec (n+m) :=
   .ofNat (n+m) (a.toNat <<< m + b.toNat)
 
 instance : HAppend (BitVec w) (BitVec v) (BitVec (w+v)) := ⟨BitVec.append⟩
+
 /--
 Extraction of bits `i` down to `j` from a bit vector of size `n` to yield a
 new bitvector of size `i - j + 1`
@@ -163,7 +164,7 @@ def zeroExtend (i : Nat) (x : BitVec w) : BitVec (w+i) :=
   have hEq : w+i = i+w := Nat.add_comm _ _
   hEq ▸ ((0 : BitVec i) ++ x)
 
-/-- Signed extension. -/
+/-- Signed extension. Left-pads with copies of the left-most bit. -/
 def signExtend (i : Nat) (x : BitVec w) : BitVec (w+i) :=
   have hEq : ((w-1) - (w-1) + 1)*i + w = w+i := by
     rw [Nat.sub_self, Nat.zero_add, Nat.one_mul, Nat.add_comm]
@@ -182,4 +183,3 @@ def shrink (v : Nat) (x : BitVec w) : BitVec v :=
 def lsbGet (x : BitVec w) (i : Nat) : Bool :=
   x.extract i i != 0
 
-end Std.BitVec
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -38,7 +38,8 @@ structure BitVec (w : Nat) where
 namespace BitVec
 
 /-- `cast eq i` embeds `i` into an equal `BitVec` type. -/
-@[inline] def cast (eq : n = m) (i : BitVec n) : BitVec m := .ofFin (Fin.cast (congrArg _ eq) i.toFin)
+@[inline] def cast (eq : n = m) (i : BitVec n) : BitVec m :=
+  .ofFin (Fin.cast (congrArg _ eq) i.toFin)
 
 /-- The `BitVec` with value `i mod 2^n`. Treated as an operation on bitvectors,
 this is truncation of the high bits when downcasting and zero-extension when upcasting. -/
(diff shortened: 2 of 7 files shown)


leodemoura:
https://github.com/avigad/lean-auto

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+institutional affiliations. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura
+-/
+namespace Std
+/-!
+We define bitvectors. We choose the `Fin` representation over others for its relative efficiency
+(Lean has special support for `Nat`), alignment with `UIntXY` types which are also represented
+with `Fin`, and the fact that bitwise operations on `Fin` are already defined. Some other possible
+representations are `List Bool`, `{ l : List Bool // l.length = w}`, `Fin w → Bool`.
```
semorrison, line 12:
```suggestion
representations are `List Bool`, `{ l : List Bool // l.length = w }`, `Fin w → Bool`.
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+  default := BitVec.zero n
+
+instance : OfNat (BitVec n) i where
+  ofNat := BitVec.ofNat n i
+
+/--
+Notation for bit vector literals. `i#n` is a shorthand for `BitVec.ofNat n i`.
+TODO: is this a good notation? Do we need it? We can also write `(i : BitVec n)`
+-/
+syntax:max num noWs "#" noWs num : term
```
semorrison, line 57:
```suggestion
scoped syntax:max num noWs "#" noWs num : term
```

semorrison: If we keep the notation, probably best to use it below, to make it more familiar to API users.

(Thumbs up and I'll make this change.)

semorrison: It seems narrowly useful if it only works for literals. I'd say we don't want it.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+/-- Rotate right. -/
+def rotateRight (x : BitVec w) (n : Nat) : BitVec w :=
+  x >>> n ||| x <<< (w - n)
+
+/-- Append. -/
+def append (a : BitVec n) (b : BitVec m) : BitVec (n+m) :=
+  .ofNat (n+m) (a.toNat <<< m + b.toNat)
+
+instance : HAppend (BitVec w) (BitVec v) (BitVec (w+v)) := ⟨BitVec.append⟩
+/--
```
semorrison, line 143:
```suggestion

/--
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+    have hEq : w + w*n = w*(n + 1) := by
+      rw [Nat.mul_add, Nat.add_comm, Nat.mul_one]
+    hEq ▸ (x ++ repeat_ n x)
+
+/-- Zero extension. -/
+def zeroExtend (i : Nat) (x : BitVec w) : BitVec (w+i) :=
+  have hEq : w+i = i+w := Nat.add_comm _ _
+  hEq ▸ ((0 : BitVec i) ++ x)
+
+/-- Signed extension. -/
```
semorrison, line 166:
```suggestion
/-- Signed extension. Left-pads with copies of the left-most bit. -/
```
(Maybe this is obvious to users of bitvectors, and can be ignored.)

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+def signExtend (i : Nat) (x : BitVec w) : BitVec (w+i) :=
+  have hEq : ((w-1) - (w-1) + 1)*i + w = w+i := by
+    rw [Nat.sub_self, Nat.zero_add, Nat.one_mul, Nat.add_comm]
+  hEq ▸ ((repeat_ i (extract (w-1) (w-1) x)) ++ x)
+
+/-- Return a prefix of size `v` of bit vector `x`. -/
```
semorrison, line 172: There is some confusion here about endianness. Relative to the `++` operation, this is returning a suffix, not a prefix.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+      rw [Nat.add_zero]
+      exact Nat.sub_add_cancel hZero
+    hEq ▸ x.extract (v - 1) 0
+  else 0
+
+/-- Return the `i`-th least significant bit. -/
+def lsbGet (x : BitVec w) (i : Nat) : Bool :=
+  x.extract i i != 0
+
+end Std.BitVec
```
semorrison, line 185:
```suggestion
```

for better or worse, the current Std convention is not close namespaces at the end of the file.

alexkeizer:
It might be worth mentioning that there is an open [Mathlib PR](https://github.com/leanprover-community/mathlib4/pull/5920/) which I believe was based on the same `BitVec` library as this PR, and has undergone several rounds of review.

digama0:
It might be better to get this into mathlib first, rather than having to copy things around if it is already under review there.

Std/Data/BitVec/Basic.lean
leodemoura: It makes proof goals way more readable. Specially the ones coming from software/hardware verification.

leodemoura:
@alexkeizer and @digama0 Bit vectors must be in the Std. They are essential for software and hardware verification, and support for bit vectors in `lean-auto` is blocked because we don't have bit vectors in Std. `lean-auto` is a package for invoking SMT solvers from Lean. I need `lean-auto` for applications.
I am happy to move the Mathlib PR here myself or tolerate the duplication for now. Happy with any of these two options, tell me which one you prefer and I will get it done by tomorrow.

Std/Data/BitVec/Basic.lean
leodemoura: It looks correct to me. 
```
example : signExtend 2 2#4  = 2#6  := rfl
example : signExtend 1 10#4 = 26#5 := rfl
example : signExtend 2 10#4 = 58#6 := rfl
```

digama0:
> @alexkeizer and @digama0 Bit vectors must be in the Std. They are essential for software and hardware verification, and support for bit vectors in `lean-auto` is blocked because we don't have bit vectors in Std.

To be clear, I don't contest this and would like us to have a bitvectors API in Std. My concern is about whether the API here is the right one, or whether we need to expose things in a different way or with different definitions (e.g. little endian vs big endian). My main tool for evaluating this is to look at usage of the API in a downstream project. It is relatively easier to merge PRs migrating from mathlib because they have already undergone some API battle testing in order to incorporate them in the library.

How exactly is lean-auto blocked? (Not contesting the claim, asking for details.) Should it not be possible to simply have this module in lean-auto since dependencies of it will be able to use this? This is what aesop and mathlib have done in the past when it comes to merging things upstream: they keep a copy of the module and use that instead pending the merge. This practice also helps to get more data and refine the API.

Std/Data/BitVec/Basic.lean
digama0: I agree with Scott, the only thing this does is add a type ascription so it seems more suited as either `#n` as a notation for `Bitvec n` (so `(5 : #3)` works), or letting downstream projects define the notation themselves as a local notation (it's a one-liner). I don't find the particular syntax very descriptive and would prefer the definition be moved closer to the use for readability.

PratherConid:
Lean-auto is not necessarily "blocked", but importing the current implementation of BitVec from Mathlib makes compilation and initialization unacceptably slow. @digama0 It seemed to me that there would be no problem to have Bitvec in Std immediately. But since it contradicts the method used to evaluate APIs, I'd be happy to just have this module in Lean-auto.

digama0:
I'm not trying to block this PR, I'm soliciting information on how it is used. Maybe there could be a test file or something, to demonstrate intended usage?

PratherConid:
Lean-auto just imports those Bitvec definitions, and matches their name against constant names in the input so that it can recognize them as Bitvec operations. The actual implementation of Bitvec has no impact on the performance of Lean-auto if we're only using the smt translation part of it.

@digama0 I think the slow compilation might not be caused by the Bitvec implementation. Generally, whenever I import files from mathlib, the compilation of lean-auto becomes significantly slower, so I've been avoiding importing anything from mathlib, and I have several modules copied from mathlib in Lean-auto (I didn't do the same to Bitvec because I thought it makes sense to have it in std). I'll do some more profiling to determine by how many factors the compilation is slowed down by importing mathlib, and whether it's related to the Bitvec implementation.

leodemoura:
@PratherConid Could you provide a list of all definitions you need in the `BitVec` library? I am assuming you need all the functions that are available on SMT-Lib, correct? The actual names will not be exactly the same, but we can put the corresponding SMT-Lib names in the doc strings. 

@digama0 I am going to be using the `BitVec` library in actual applications, and we will polish the API and add theorems as we go. @joehendrix will help making sure the library is good for software and hardware verification applications. AFACIT, the `Bitvec` module in Mathlib has no users. So, this PR and future ones will have no impact on Mathlib.
BTW, SMT-Lib is a great reference for bit vectors since SMT solvers are already used in many hardware and software verification applications.

PratherConid:
@leodemoura Yes, I only need the functions available on smt-lib 2.6 https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+
+/-- Append. -/
+def append (a : BitVec n) (b : BitVec m) : BitVec (n+m) :=
+  .ofNat (n+m) (a.toNat <<< m + b.toNat)
+
+instance : HAppend (BitVec w) (BitVec v) (BitVec (w+v)) := ⟨BitVec.append⟩
+
+/--
+Extraction of bits `i` down to `j` from a bit vector of size `n` to yield a
+new bitvector of size `i - j + 1`
```
eric-wieser, line 146: In mathlib4 we were leaning towards an API that satisfied
```suggestion
new bitvector of size `i - j`
```
As this saves some branching in `shrink`

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+  { val := x.val - y.val }
+/-- Multiplication for bit vectors. -/
+protected def mul (x y : BitVec n) : BitVec n :=
+  { val := x.val * y.val }
+/-- Modulo for bit vectors. -/
+protected def mod (x y : BitVec n) : BitVec n :=
+  { val := ⟨x.val % y.val, Nat.lt_of_le_of_lt (Nat.mod_le _ _) x.val.isLt⟩ }
+/-- Division for bit vectors. -/
+protected def div (x y : BitVec n) : BitVec n :=
+  { val := ⟨x.val / y.val, Nat.lt_of_le_of_lt (Nat.div_le_self _ _) x.val.isLt⟩ }
```
eric-wieser, line 88: Can we reuse fin division here?

Std/Data/BitVec/Basic.lean
eric-wieser: I imagine this does not align with SMTlib (and am aware it does not align with the convention in VHDL and Verilog), but I imagine it will be much more pleasant to prove things avout

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,185 @@
...
+/-- Rotate left. -/
+def rotateLeft (x : BitVec w) (n : Nat) : BitVec w :=
+  x <<< n ||| x >>> (w - n)
+
+/-- Rotate right. -/
+def rotateRight (x : BitVec w) (n : Nat) : BitVec w :=
+  x >>> n ||| x <<< (w - n)
+
+/-- Append. -/
+def append (a : BitVec n) (b : BitVec m) : BitVec (n+m) :=
```
eric-wieser, line 139: As I commented on the mathlib PR, this would be clearer with the arguments renamed as `append msb lsb`.

alexkeizer:
We are also looking to use bitvectors in our project. We started using the current bitvector API, but when the discussed refactor was proposed we stopped until it became clear which representation mathlib would use.

I did not necessarily mean to suggest postponing this PR, I would be happy to have bitvectors in STD ASAP, I merely wanted to point out the existing efforts.

The refactor also got discussed on Zulip (https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Proposed.20refactor.20of.20Bitvec).

Std/Data/BitVec/Basic.lean
eric-wieser: I think @semorrison is commenting on `shrink` (for which this line is the docstring) not `signExtend`

Std/Data/BitVec/Basic.lean
leodemoura: @eric-wieser Thanks for the feedback. Yes, the goal is to align with SMTlib which is a standard in software and hardware verification. SMT solvers are widely used in industry, and they all use the SMTlib standard.

Std/Data/BitVec/Basic.lean
leodemoura: Joe and Wojciech wrote their own version because it more efficient than the one in `Fin`. I will upstream this change to `Fin`, and use the `Fin` version here. Thanks!

Std/Data/BitVec/Basic.lean
leodemoura: I will do that. Thanks.

leodemoura:
@alexkeizer Thanks for sharing. The good news is that things will start moving really fast here in `Std` :) Our plans are

- `Std` should be great for software/hardware verification projects.
- `Std` should be great for software development, and should provide a library with the same capabilities of the Rust std.
- Better automation for software/hardware verification projects using `Std`.

We're lucky to have Joe Hendrix (@joehendrix) on board with us at Lean FRO. He's a pro when it comes to software verification and development. Joe's got our back in hitting our goals and will help us get those PRs merged super fast.

joehendrix:
Thanks for sharing the link to the Zulip discussion.  I hadn't seen that.  I thought I'd reply here since we have a concrete proposal that we want in quickly.  Sorry for the long thread.

There are a few contentious choices in designing a bitvector theory:
1. What representation to use?
2. What is the bit order that order-dependent operations like contact, extract, get/set use?  Is it most-significant or least-significant first?
3. Should comparison have a signed or unsigned default?  e.g., What does`x < y` mean?  Is it defined?  If so, is it signed or unsigned?

I'll address each of these:

**Representation**. I think the main candidates are `Nat` subtype (e.g., `Fin n`), a C byte array subtype, a vector of Booleans, `Fin n -> Bool`.

Computational predictability and efficiency in the interpreter/compiler matters, and I think this rules out Vector of Booleans and `Fin n -> Bool`.

Between `Fin n` and byte arrays, I like the `Fin n` option currently because it already exists with efficient large-word arithmetic operations.  In the long term, I think the optimal choice would be a byte array subtype eventually because then you have predictable non-data dependent memory usage.  However, that's a lot of work to implement (and would be fastest in C rather than Lean with the current Lean compiler).

**Bit Order**.  I think we should have index operations with both bit orders eventually, but `msb = bit 0` should be the default as it matches the pretty printed representation.  e.g., `concat 0xAB 0xCD = 0xABCD`.  [Cryptol](https://cryptol.net/) used to be least-significant first and got ported to most-significant first because of this. 

The alternative order of `lsb = bit 0` has the advantages that the `i`-th bit matched the bit set by `2^^i` and is more intuitive when implementing adder circuits and other arithmetic operations.  However, I think for library users, the visual symmetry between the value printed as a number like `0xAB` and the list of bits is more important `[1, 0, 1, 0, 1, 0, 1, 1]`

**Comparison**. I think the ordering from `x < y` should be defined and consistent with how bitvectors are rendered.   I think using a unsigned rendering of bitvectors to hex constants makes a lot of sense, and so defaulting to an unsigned comparison would be consistent with that.

I think compatibility with SMT-LIB is a very important requirement, but consistency with it drive all the design decisions.

digama0:
> Bit Order. I think we should have index operations with both bit orders eventually, but msb = bit 0 should be the default as it matches the pretty printed representation. e.g., concat 0xAB 0xCD = 0xABCD. [Cryptol](https://cryptol.net/) used to be least-significant first and got ported to most-significant first because of this.
>
> The alternative order of lsb = bit 0 has the advantages that the i-th bit matched the bit set by 2^^i and is more intuitive when implementing adder circuits and other arithmetic operations. However, I think for library users, the visual symmetry between the value printed as a number like 0xAB and the list of bits is more important [1, 0, 1, 0, 1, 0, 1, 1].

I feel pretty strongly that lsb = bit 0 (aka little-endian) should be the preferred form. Changing printing order is relatively easy, just use a snoc-like function instead of a cons-like function. For showing lists of bits we can customize delaborators all we like. But indexing should not depend on the bit width, there are so many things that get easier if you do this. Note that bit-indexing on Nat is also little-endian, and indeed big-endian isn't even coherent in that setting.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,184 @@
...
+  { val := x.val - y.val }
+/-- Multiplication for bit vectors. -/
+protected def mul (x y : BitVec n) : BitVec n :=
+  { val := x.val * y.val }
+/-- Modulo for bit vectors. -/
+protected def mod (x y : BitVec n) : BitVec n :=
+  { val := x.val % y.val }
+/-- Division for bit vectors. -/
+protected def div (x y : BitVec n) : BitVec n :=
+  { val := x.val / y.val }
```
joehendrix, line 88: Should we be compatible with SMT-LIB for division by zero?  Specifically, SMT-LIB defines `x / 0` on bitvectors to be all ones (see the [spec](http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml)).  This is consistent with a simple division circuit.

Here I am getting `0xABCD#16 / 0#16` to be `0`.

leodemoura: Yes, we should. Good catch.

digama0: We certainly at least want an `smtDiv` in that case. I think we should bias the notation to whatever leads to shorter theorems, but I don't think we have enough data on what that is yet. I have a slight preference for consistency with other lean definitions since then theorem transfer will be easier.

Std/Data/BitVec/Basic.lean
```diff
@@ -77,15 +77,26 @@ protected def add (x y : BitVec n) : BitVec n :=
 /-- Subtraction for bit vectors. -/
 protected def sub (x y : BitVec n) : BitVec n :=
   { val := x.val - y.val }
+/-- Bit vector of size `n` where all bits are `1`s -/
+protected def allOnes (n : Nat) : BitVec n :=
+  BitVec.sub (BitVec.ofNat n 0) (BitVec.ofNat n 1)
```
digama0, line 82: Isn't this just `-1`?

Std/Data/BitVec/Basic.lean
leodemoura: @digama0 Thanks for sharing your position. Nothing is final here and all decisions can be reverted later. We will move forward quickly and get feedback for the `BitVec` library from people that will be using Lean in software/hardware verification projects. Having industrial applications and users is super important for the Lean project future. 
The good news is that the work on `BitVec` does not affect Mathlib.

Std/Data/BitVec/Basic.lean
```diff
@@ -77,15 +77,26 @@ protected def add (x y : BitVec n) : BitVec n :=
...
 protected def mod (x y : BitVec n) : BitVec n :=
   { val := x.val % y.val }
-/-- Division for bit vectors. -/
+/--
+Division for bit vectors.
+Note that `div x 0 = allOnes n` as specified by the SMT-Lib standard.
+See the spec at http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml.
+This is also consistent with a simple division circuit.
+-/
 protected def div (x y : BitVec n) : BitVec n :=
```
digama0, line 95: I will repeat my suggestion from before: If there are two conventions then we need definitions for both of them. If you really want to make the SMT-lib definition be default, then there should be `div'` with the lean convention.

Std/Data/BitVec/Basic.lean
leodemoura: Yes.

Std/Data/BitVec/Basic.lean
leodemoura: Happy to add `div'`.

Std/Data/BitVec/Basic.lean
digama0: @leodemoura I will have you know that I don't need to be reminded how "we will move forward quickly". As a maintainer, I'm feeling pretty bulldozed right now. Please allow me to do due diligence in keeping the high quality standards of this library. I assure you it won't take that long; I just want to make sure that the rationales are all written down (either here or in the code itself) so that we can examine them later if we need to make changes in the future.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,200 @@
...
+in both the runtime and the kernel, inheriting all the special support for `Nat`.
+-/
+structure BitVec (w : Nat) where
+  /-- Bitvector is represented as a `Fin`. -/
+  val : Fin (2^w)
+  deriving DecidableEq
+
+namespace BitVec
+
+theorem pow_two_gt_zero (w : Nat) : 2^w > 0 := by
```
digama0, line 32: This should be in `Std.Data.Nat.Init.Lemmas` in the `Nat` namespace:
```lean
theorem pow_two_pos (n : Nat) : 0 < 2 ^ n := Nat.pos_pow_of_pos (by decide)
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,200 @@
...
+  | _                => throw ()
+
+instance : Repr (BitVec n) where
+  reprPrec a _ := repr a.val ++ "#" ++ repr n
+
+instance : ToString (BitVec n) where
+  toString a := toString (repr a)
+
+/-- Theorem for normalizing the bit vector literal representation. -/
+@[simp] theorem ofNat_eq_ofNat : @OfNat.ofNat (BitVec n) i _ = BitVec.ofNat n i := rfl
```
digama0, line 72: Why is `ofNat` being used as the simp-normal form, instead of plain numerals?

Std/Data/BitVec/Basic.lean
digama0: ...so why not write it that way? (I see we also need a `Neg` instance to make it work.)

Std/Data/BitVec/Basic.lean
digama0: How about defining `extract` in terms of `extract'` using Eric's convention?

Std/Data/BitVec/Basic.lean
leodemoura: @digama0 I want to start by expressing my deep appreciation for everything you've done for Std4. Your commitment and efforts since stepping up as its maintainer last year, when I ask you for help, have been nothing short of exemplary. I recognize the weight of responsibilities that came with it, and I genuinely regret if my recent comments added any undue pressure.

Please know that your concerns about maintaining the high quality of this library resonate with me. I've always valued your dedication to thoroughness and meticulousness. The intention behind mentioning moving forward quickly was never to undermine your role or the due diligence you bring.

As I had mentioned [here](https://github.com/leanprover/lean4/pull/2542#issuecomment-1723352703), help was coming. I did not forget the burden you were carrying :) Joe has now taken on the responsibility of being the full-time maintainer for Std4, aiming to progress the project further. This, however, doesn’t diminish the value of your insights or your role. You'll continue to have maintainer privileges, and your feedback will always be invaluable to us.

If you ever feel the need to lighten your workload, please know that the title of 'honorary maintainer' is open to you. This isn't about reducing your role but rather an acknowledgment of the immense work you've done and a gesture of gratitude. The priority here is your well-being and ensuring that you feel respected and acknowledged.

Thank you for everything, Mario.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,200 @@
...
+new bitvector of size `i - j + 1`
+-/
+def extract (i j : Nat) (a : BitVec n) : BitVec (i - j + 1) :=
+  BitVec.ofNat _ (a.toNat >>> j)
+
+/--
+`repeat_ i x` means concatenate `i` copies of `x`.
+Recall that `repeat` is a keyword in Lean.
+-/
+def repeat_ : (i : Nat) → BitVec w → BitVec (w*i)
```
digama0, line 171:
```suggestion
def replicate : (i : Nat) → BitVec w → BitVec (w*i)
```
List and String also use `replicate`, although in both cases it means to repeat a single character. The other standard naming convention for avoiding existing names is `repeat'`.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,200 @@
...
+We define bitvectors. We choose the `Fin` representation over others for its relative efficiency
+(Lean has special support for `Nat`), alignment with `UIntXY` types which are also represented
+with `Fin`, and the fact that bitwise operations on `Fin` are already defined. Some other possible
+representations are `List Bool`, `{ l : List Bool // l.length = w }`, `Fin w → Bool`.
+
+We also define many bitvector operations from the
+[`QF_BV` logic](https://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV).
+of SMT-LIBv2.
+
+TODO(Leo): match the interface to what SMT-LIB provides
```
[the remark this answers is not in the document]
leodemoura: Good question. `bvnot` is missing. The whole list is here https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml
@joehendrix SMT-Lib used to have operators such as `bvsdiv` (signed division) and `bvsle`. Were they removed from the standard?

Std/Data/BitVec/Basic.lean
leodemoura: The proof goals are easier to understand when we make the dimensions explicit.

Std/Data/BitVec/Basic.lean
leodemoura: Nice.

Std/Data/BitVec/Basic.lean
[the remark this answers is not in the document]
digama0: We have `complement`, isn't that `bvnot`? Maybe it should be called `not` like the bool function.

Std/Data/BitVec/Basic.lean
leodemoura: Yes, we would need the `Neg`. Happy with this change.

Std/Data/BitVec/Basic.lean
leodemoura: Good catch.

Std/Data/BitVec/Basic.lean
digama0: Is this something that we could do with a delaborator or delaborator option? (I don't mind if we mark this as an issue to be revisited later.)

Std/Data/BitVec/Basic.lean
digama0: I appreciate that. I would rather not step down from the maintainer role though, I'd rather be an "honorary FRO member" and retain code ownership over my work. I'm glad Joe is coming on board, two maintainers are better than one.

Std/Data/BitVec/Basic.lean
leodemoura: We will keep an eye on it, and revise in the future if needed. The current solution uses only an `app_unexpander` which is super light weight.

Std/Data/BitVec/Basic.lean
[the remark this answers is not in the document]
leodemoura: Yes, let's rename `complement` to `not`.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,200 @@
...
+new bitvector of size `i - j + 1`
+-/
+def extract (i j : Nat) (a : BitVec n) : BitVec (i - j + 1) :=
+  BitVec.ofNat _ (a.toNat >>> j)
+
+/--
+`repeat_ i x` means concatenate `i` copies of `x`.
+Recall that `repeat` is a keyword in Lean.
+-/
+def replicate : (i : Nat) → BitVec w → BitVec (w*i)
```
digama0, line 170:
```suggestion
-- TODO: write this using multiplication
def replicate : (i : Nat) → BitVec w → BitVec (w*i)
```
This should be equivalent to `x * (2^(w*i) - 1) / (2^w - 1)`, although there is some nontrivial proof to show it is in range and some even more nontrivial proof to show it replicates the bits in the same way as this linear time solution.

PS, while I have you here: `1 <<< n` on Nat does not have a builtin kernel implementation, which throws off the asymptotics of some algorithms. (I think you just get `2^n` implemented as `n` multiplications.) What do you think about adding bitwise ops on Nat to the list of accelerated kernel ops?

leodemoura: I love these suggestions. It will help for sure people working on software and hardware verification.

Std/Data/BitVec/Basic.lean
leodemoura: I love that. How are your expectations regarding the "Honorary FRO member" role? Happy to sync with you on a call this week.

Std/Data/BitVec/Basic.lean
leodemoura: Happy to change the notation, but it would be great to have something very compact and working "out of the box" for people working on software/hardware verification.

digama0: My gut feeling is that writing the arity all the time will get old quick, especially in cases where the arity is mostly fixed (e.g. everything is on 32 bit words). For sure frequently varied arity definitely comes up too, especially in hardware verification, but I don't think it is the majority case, and inventing a new domain specific notation for type ascription seems like it would lead to more confusion and not less, unless this is a very standard notation in the trade (and my impression is that it is not). It might be good to do a survey of what VHDLs and such do for this situation.

Std/Data/BitVec/Basic.lean
leodemoura: `extract'` is a good idea. Please feel free to add a commit with its implementation.

joehendrix:
> I feel pretty strongly that lsb = bit 0 (aka little-endian) should be the preferred form.

I used to also strong prefer lsb = bit 0, but changed my mind after being convinced by John Launchbury at Galois.  For implementing bitvector operations, least-significant bit first is often more natural for traversing bits in a bitvec, but I think for users we should priortize be consistent with the printing order of a bitvector as a number.  It makes documentation and demos a lot more clear.

In cryptography, one often needs to split a bitvector into a vector of bitvectors (e.g, split a `BitVec 256` into a vector of 32 bytes), do some manipulation and flatten back into a single result.  It's really much more readable if one can print the splitting operations and preserve the order of individual elements. e.g., 0x12345678#64 gets split into `[0x12 0x34 0x56 0x78]`).

To minimize index arithmetic, we should offer operations for getting/setting bits (and ranges) using a least-significant bit first (see [here](https://github.com/joehendrix/lean-crypto/blob/main/lib/Crypto/BitVec.lean#L9-L10) for some operations we could add), but I think the default should be biased towards being consistent with how the elements are printed - which is most-signicant bits/hex digit first.

Std/Data/BitVec/Basic.lean
joehendrix: I think I am slightly in favor of the notation.  This is probably orthogonal, but it'd be neat if there was a way to have access to the number of digits in a hex or binary literal (e.g., if we knew `0b01010` was a BitVec, then it would default to having 5 bits).  Similar with hex digits having a multiple of the input `0x0010` is a `BitVec 16`.  Cryptol does this, but also is predominantly about bitvector programming.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,237 @@
...
+scoped syntax:max num noWs "#" noWs num : term
+macro_rules | `($i:num#$n:num) => `(BitVec.ofNat $n $i)
+
+/-- Unexpander for bit vector literals. -/
+@[app_unexpander BitVec.ofNat] def unexpandBitVecOfNat : Lean.PrettyPrinter.Unexpander
+  | `($(_) $n:num $i:num) => `($i:num#$n:num)
+  | _                => throw ()
+
+instance : Repr (BitVec n) where
+  reprPrec a _ := repr a.val ++ "#" ++ repr n
```
joehendrix, line 63: How about we print in hex with a fixed width of `(n+3)/4` digits shown?  I have some similar code for doing that with strings [here](https://github.com/joehendrix/lean-crypto/blob/main/lib/Crypto/BitVec.lean#L55-L57).

eric-wieser:
There seems to be a tradeoff here between the debug output via printing being readable, and the indexing code being readable.

The two aren't mutually exclusive; we could have `split (b : Bitvector ℕ) : VectorParts _` so that the default `Repr` is `parts%[0x12, 0x34, 0x56, 0x78]` or similar, while `b.split.toList` contains the underlying list in LSB order

Std/Data/BitVec/Basic.lean
leodemoura: @joehendrix  Great idea. Let's get it done this week.

Std/Data/BitVec/Basic.lean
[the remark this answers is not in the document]
Vtec234: > @joehendrix SMT-Lib used to have operators such as bvsdiv (signed division) and bvsle. Were they removed from the standard?

I was once caught off-guard by this :) They are not in the page you linked but here: https://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV

joehendrix:
> There seems to be a tradeoff here between the debug output via printing being readable, and the indexing code being readable.
> 
> The two aren't mutually exclusive; we could have `split (b : Bitvector ℕ) : VectorParts _` so that the default `Repr` is `parts%[0x12, 0x34, 0x56, 0x78]` or similar, while `b.split.toList` contains the underlying list in LSB order

What is `VectorParts`?  Is that just staging for a split operation?  I think my main concern is keeping things easy to understand for new users of the library.  If readability of LSB indexing is problematic, we could also introduce more syntax. 
 Cryptol uses `x @ 0` to denote the most significant bit and `x ! 0` to denote the least.

I think there should be indexing and likely other operators for both orders.  The question is what is the default for split, join, and conversion to lists/arrays.

digama0:
I think it is also possible to just have *no* default here. For get/set you just have `lsb` and `msb` variants in the function names, same for list conversion, and for join you have some symmetric pair of operators like `<+` / `+>`. This is what [Rust](https://doc.rust-lang.org/std/primitive.u32.html#method.from_le_bytes) does, and I think it makes good documentation to be explicit about the ordering if we can't banish big-endian properly.

digama0:
I pushed some changes. Summary:

* Added more documentation to all of the operations
* General style guide things (indentation etc)
* Simplified some of the definitions
* Added `fill` as a version of `replicate` that just splats a bool across the vector
* Changed `LT` and `LE` to delegate to the `Fin` LT/LE (which delegates to the Nat one), instead of using the Bool function which will cause spurious `decide (x < y) = true` things in proofs using these ops
* Clarified things like what zero extension, left/right shift, and other bitwise ops mean in terms of bit positions and also numbers where applicable. I tried to avoid saying things like "the left side" in favor of "the low bits" to avoid ambiguity depending on how the user's mental picture is laid out. (The [Rust](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_left) stdlib also uses this guideline, because RTL languages exist too.)
* Renamed `shrink` to `truncate`, because `shrink` is not particularly explanatory regarding how the shrinking is supposed to happen (are we losing high bits or low bits?), while "truncate" is more consistently used for dropping high bits in integer conversion.
* `shrink` and `zeroExtend` are actually aliases, but they have documentation which focuses on the case in the name.

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,259 @@
...
+  O(1), because we use `Fin` as the internal representation of a bitvector. -/
+  mk ::
+  /-- Interpret a bitvector as a number less than `2^w`.
+  O(1), because we use `Fin` as the internal representation of a bitvector. -/
+  val : Fin (2^w)
+  deriving DecidableEq
+
+namespace BitVec
+
+/-- The `BitVector` with value `i mod 2^n`. Treated as an operation on bitvectors,
```
eric-wieser, line 38: This docstring says `BitVector` (in backticks) while the type is `BitVec`. I don't intend to do any name bikeshedding here, but presumably one way or the other this is a remnant of an incomplete rename that should be concluded?

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,259 @@
...
+A bitvector of the specified width. This is represented as the underlying `Nat` number
+in both the runtime and the kernel, inheriting all the special support for `Nat`.
+-/
+structure BitVec (w : Nat) where
+  /-- Construct a `BitVec w` from a number less than `2^w`.
+  O(1), because we use `Fin` as the internal representation of a bitvector. -/
+  mk ::
+  /-- Interpret a bitvector as a number less than `2^w`.
+  O(1), because we use `Fin` as the internal representation of a bitvector. -/
+  val : Fin (2^w)
```
eric-wieser, line 33: Given the multitude of possible representations, would calling `mk`/`val` as `ofFin` / `toFin` instead  be reasonable? That would make them feel on equal footing with `ofBits` / `toBits` or similar.

Std/Data/BitVec/Basic.lean
digama0: That's my fault, it was written without backticks but with camel case and I added the backticks.
```suggestion
/-- The `BitVec` with value `i mod 2^n`. Treated as an operation on bitvectors,
```

Std/Data/BitVec/Basic.lean
joehendrix: I like that idea.

joehendrix:
@digama0 I personally think the comment changes and cleanups improve readability so thanks.

One interesting thing is that most-significant being on the left is part of our [Hindu Arabic numeral system](https://en.wikipedia.org/wiki/Hindu%E2%80%93Arabic_numeral_system) -- even traditionally right-to-left languages have most-significant digits on the left.

I don't want the bit order discussion to block this PR from being merged, so if you haven't been persuaded I am ok with avoiding committing to a bit order in this PR.

digama0:
As things stand currently, the only place where big endian has to be explicitly acknowledged is in the argument order for `append`, and actually we might also want the reversed argument order not simply because of endianness but also because it would have the addition in the opposite order, which would make it a bit easier re: dependent typing in some cases (note that there are some uses of `Nat.add_comm .. ▸ (v1 ++ v2)` in this PR, and probably even more in usage code). 

In general, dependent typing hell will definitely arise when working with this API when the bitwidths are not numeric literals, so we probably need to do something to help with that (although as usual it is difficult to predict in advance what that might be). One thing is that we probably want a `ofEq` function for coercing between bitvector types with equal but not defeq bitwidth; mathlib makes a point of defining these for all dependent type families because they compute better than `Eq.rec` terms.

joehendrix:
I've pushed `ofFin`/`toFin` changes and a change to use hex for rendering bitvectors.  I don't know if people will always like that, but is is easy to convert to a Nat if one wants a decimal rendering.

`digama0` I can add a `ofEq`.  Is the conventional signature something like `ofEq : u = v -> BitVec u -> BitVec v`?

digama0:
> I've pushed ofFin/toFin changes and a change to use hex for rendering bitvectors. I don't know if people will always like that, but is is easy to convert to a Nat if one wants a decimal rendering.

Actually I was thinking you might want binary rendering. I can't imagine you would ever want decimal, that seems exceedingly unhelpful for understanding a bitvector value. But binary is a bit too verbose and is probably not suitable for more than 4-8 bits.

alexkeizer:
>  I've pushed `ofFin`/`toFin` changes and a change to use hex for rendering bitvectors.  I don't know if people will always like that, but is is easy to convert to a Nat if one wants a decimal rendering.
> 
> `digama0` I can add a `ofEq`.  Is the conventional signature something like `ofEq : u = v -> BitVec u -> BitVec v`?

The signature is right, but I think these kind of functions are usually called `cast`, as in https://leanprover-community.github.io/mathlib4_docs/Std/Data/Fin/Basic.html#Fin.cast

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,427 @@
...
+this is truncation of the high bits when downcasting and zero-extension when upcasting. -/
+protected def ofNat (n : Nat) (i : Nat) : BitVec n where
+  toFin := Fin.ofNat' i (Nat.pow_two_pos _)
+
+/-- Given a bitvector `a`, return the underlying `Nat`. This is O(1) because `BitVec` is a
+(zero-cost) wrapper around a `Nat`. -/
+protected def toNat (a : BitVec n) : Nat := a.toFin.val
+
+/-- Return most-significant bit in bitvector. -/
+protected def msb (a : BitVec n) : Bool := (a.toNat &&& (2^ (n-1))) ≠ 0
```
digama0, line 53:
```suggestion
protected def msb (a : BitVec n) : Bool := getLsb a (n-1)
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,427 @@
...
+
+/-- The `BitVec` with value `(2^n + (i mod 2^n)) mod 2^n`.  -/
+protected def ofInt (n : Nat) (i : Int) : BitVec n :=
+  match i with
+  | Int.ofNat a => .ofNat n a
+  | Int.negSucc a => .ofNat n (2^n - 1 - (a % 2^n))
```
digama0, line 59:
```suggestion
  | Int.negSucc a => .ofNat n (2^n - 1 - a % 2^n)
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,427 @@
...
+protected def neg (x : BitVec n) : BitVec n := .sub 0 x
+instance : Neg (BitVec n) := ⟨.neg⟩
+
+/-- Bit vector of size `n` where all bits are `1`s -/
+def allOnes (n : Nat) : BitVec n := -1
+
+/--
+Return the absolute value of a signed bitvector.
+-/
+protected def abs (s : BitVec n) : BitVec n := if s.msb then (.neg s) else s
```
digama0, line 127:
```suggestion
protected def abs (s : BitVec n) : BitVec n := if s.msb then .neg s else s
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,427 @@
...
+instance : Mod (BitVec n) := ⟨.umod⟩
+
+/--
+Unsigned division for bit vectors using the
+[SMT-Lib convention](http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml)
+where division by zero returns the `allOnes` bitvector.
+
+SMT-Lib name: `bvudiv`.
+-/
+def smt_udiv (x y : BitVec n) : BitVec n := if y = 0 then -1 else .udiv x y
```
digama0, line 159:
```suggestion
def smtUDiv (x y : BitVec n) : BitVec n := if y = 0 then -1 else .udiv x y
```
Ditto for all other snake case names here

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,427 @@
...
+def append (msbs : BitVec n) (lsbs : BitVec m) : BitVec (n+m) :=
+  .ofNat (n + m) (msbs.toNat <<< m ||| lsbs.toNat)
+
+instance : HAppend (BitVec w) (BitVec v) (BitVec (w + v)) := ⟨.append⟩
+
+/-- Return the `i`-th least significant bit. -/
+def get_lsb (x : BitVec w) (i : Nat) : Bool := x.toNat &&& (1 <<< i) != 0
+
+/-- Return the `i`-th least significant bit. -/
+def get_msb (x : BitVec w) (i : Nat) : Bool := x.toNat &&& (1 <<< (w-1-i)) != 0
```
digama0, line 382:
```suggestion
def getMsb (x : BitVec w) (i : Nat) : Bool := getLsb x (w-1-i)
```
probably both of these should be `@[inline]` though

Std/Data/Int/Basic.lean
```diff
@@ -126,3 +126,41 @@ Divisibility of integers. `a ∣ b` (typed as `\|`) says that
...
+
+Interprets the integer as an infinite sequence of bits in two's complement
+and complements each bit.
+```
+~~~(0:Int) = -1
+~~~(1:Int) = -2
+~~~(-1:Int) = 0
+```
+-/
+protected def complement : Int -> Int
```
digama0, line 143:
```suggestion
protected def not : Int -> Int
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,432 @@
...
+  if t.msb then
+    if s.msb then
+      .udiv (.neg s) (.neg t)
+    else
+      .neg (.udiv s (.neg t))
+  else
+    if s.msb then
+      .neg (.udiv (.neg s) t)
+    else
+      .udiv s t
```
digama0, line 183:
```suggestion
  match s.msb, t.msb with
  | true, true => .udiv (.neg s) (.neg t)
  | false, true => .neg (.udiv s (.neg t))
  | true, false => .neg (.udiv (.neg s) t)
  | false, false => .udiv s t
```
ditto for `smt_sdiv`, `srem`, `smod`

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,424 @@
...
+protected def zero (n : Nat) : BitVec n := .ofNat n 0
+
+instance : Inhabited (BitVec n) where default := .zero n
+
+instance : OfNat (BitVec n) i where ofNat := .ofNat n i
+
+/-- Notation for bit vector literals. `i#n` is a shorthand for `BitVec.ofNat n i`. -/
+-- TODO: is this a good notation? Do we need it? We can also write `(i : BitVec n)`
+scoped syntax:max num noWs "#" noWs num : term
+macro_rules | `($i:num#$n:num) => `(BitVec.ofNat $n $i)
```
digama0, line 83:
```suggestion
scoped syntax:max term:max noWs "#" noWs term:max : term
macro_rules | `($i#$n) => `(BitVec.ofNat $n $i)
```

Std/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,424 @@
...
+
+/-- Unexpander for bit vector literals. -/
+@[app_unexpander BitVec.ofNat] def unexpandBitVecOfNat : Lean.PrettyPrinter.Unexpander
+  | `($(_) $n:num $i:num) => `($i:num#$n:num)
```
digama0, line 87:
```suggestion
  | `($(_) $n $i) => `($i#$n)
```

Std/Data/BitVec.lean
```diff
@@ -0,0 +1,7 @@
+/-
+Copyright (c) 2022 by the authors listed in the file AUTHORS and their
+institutional affiliations. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura
+-/
```
digama0, line 6:
```suggestion
```
Files with only imports don't get a copyright header

digama0:
Thanks @alexkeizer , @leodemoura , @joehendrix and everyone else for working on this. I think it looks really good now, and although there is still more to do (`toList`, `split`, lemmas, automation) this looks like a good starting point from which to start getting user feedback and iterate further.

PR #291, opened 2023-10-15
chore: add @[simp] to useful lemmas about Int

semorrison:


semorrison:
This can wait for CI to complete on https://github.com/leanprover-community/mathlib4/pull/7688, to make sure we don't mess up the linter there.

digama0:
In case you aren't aware, the reason these are not marked as global simp lemmas is because they are subsumed by more general theorems in mathlib. That said I'm not particularly happy about handling the situation in this way because it's less good to users of std without mathlib.

semorrison:
Yes, I assumed as much, but I think that is increasingly less relevant as other consumers of Std appear. Mathlib will just have to cope with the duplication.

Std/Data/Int/DivMod.lean
```diff
@@ -43,7 +43,7 @@ theorem negSucc_ediv (m : Nat) {b : Int} (H : 0 < b) : -[m+1] / b = -(div m b +
   | -[_+1] => rfl
 
 -- Will be generalized to Euclidean domains.
```
fgdorais, line 45: Delete comment.

fgdorais (approved):
Looks great to me!

Std/Data/Int/DivMod.lean
semorrison, line 45:
```suggestion
```

PR #298, opened 2023-10-16
chore: upstream some logic

semorrison:


Std/Logic.lean
```diff
@@ -493,6 +496,16 @@ theorem forall_comm {p : α → β → Prop} : (∀ a b, p a b) ↔ (∀ b a, p
 theorem exists_comm {p : α → β → Prop} : (∃ a b, p a b) ↔ (∃ b a, p a b) :=
   ⟨fun ⟨a, b, h⟩ => ⟨b, a, h⟩, fun ⟨b, a, h⟩ => ⟨a, b, h⟩⟩
 
+@[simp] theorem forall_apply_eq_imp_iff {f : α → β} {p : β → Prop} :
+    (∀ b a, f a = b → p b) ↔ ∀ a, p (f a) := by simp [forall_comm]
+
+@[simp] theorem forall_eq_apply_imp_iff {f : α → β} {p : β → Prop} :
```
eric-wieser, line 502: I think it's a bit disingenuous to use an "upstream" PR to silently rename a lemma, and then make a PR to mathlib that uses the upstreaming (in which the rename was not reviewed) to justify the renaming!

Of course the lemma here is almost never referred to by name by a human so in this case it doesn't matter much; but it's not a great precedent to set.

digama0: Agreed that this is something we should avoid, although at the same time I don't want us to ban renames during Mathlib -> Std ports. Perhaps PRs need to come with a list of renames performed. There is also the issue that PRs here are not really reviewed by mathlib maintainers (other than me), yet they can be used to strong-arm changes to mathlib during the std bump.

I don't really have a suggestion for this, other than to suggest that interested mathlib maintainers should also review mathlib material contained in std upstreaming PRs, and PR authors should ping maintainers that they think want to review changes. For extreme cases, we (mathlib maintainers) should also reserve the right to reject std bump PRs and/or open a revert PR to std when mathlib can't handle some change (but this should be a backup plan mainly because it's a lot more paperwork).

semorrison: We may have to rethink how the relationship between Mathlib and Std works. Std needs to serve many users besides Mathlib, and sometimes will not be able to wait for the glacial process of Mathlib PRs.

digama0: I wouldn't describe the Mathlib PR process as glacial. It moves quite fast in aggregate, but we are still bandwidth-limited by review. I don't really like the idea of Std being used as a test bed for rapid iteration. Can this testing not be done in a staging library? Clients of Std value stability.

PR #299, opened 2023-10-16
feat: make try this suggestions fit in the suggested width

alexjbest:


alexjbest:
awaiting-review

test/add_suggestion.lean
```diff
@@ -0,0 +1,28 @@
...
+
+set_option hygiene false in
+elab "test" : tactic => do
+  addSuggestion (← getRef) (←
+  `(tactic| exact longdef a b h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h h))
+
+end Std.Tactic.TryThis
+
+-- ideally we would have a #guard_widgets or #guard_infos too, but instead we can simply check by
+-- hand that  the widget suggestion (not the message fits into 100 chars)
```
alexjbest, line 21:
```suggestion
-- hand that the widget suggestion (not the message) fits into 100 columns
```

alexjbest:
awaiting-review

semorrison (approved):

Std/Tactic/TryThis.lean
```diff
@@ -76,6 +76,9 @@ partial def replaceMVarsByUnderscores [Monad m] [MonadQuotation m]
 def delabToRefinableSyntax (e : Expr) : TermElabM Term :=
   return ⟨← replaceMVarsByUnderscores (← delab e)⟩
 
+/-- The maximum width of an ideal line in source code, 100 is the current convention. -/
+def lineWidth : Nat := 100
```
digama0, line 80: It's a bit annoying that we are replacing one [magic number](https://github.com/leanprover/lean4/blob/fb0d0245db9bcb9ff649794f57522c52cb80c993/src/Init/Data/Format/Basic.lean#L280) with another. Why not just use the [formatting option](https://github.com/leanprover/lean4/blob/fb0d0245db9bcb9ff649794f57522c52cb80c993/src/Lean/Data/Format.lean#L17-L20) that exists for the purpose?

alexjbest: Well the default there is too wide for most suggestions, it's exactly `defWidth`. So I want to override that with 100 somehow and figured I may as well give it a name. Are you suggesting to use `w` equal to 100 if the option is unset and the option otherwise?
I don't want to change the default value completely as that seems way too invasive, and setting the option before calling format seems functionally the same as setting `w`

digama0: I think we shouldn't be adding more magic numbers. If the default should be 100 for everything, that should be upstreamed to lean core. If the default is only 100 for mathlib, then they should set this as an option in the lakefile and the code here should honor that.

alexjbest: I take your point about non-configurable magic numbers, I have pushed a commit that turns this into a user option at least.

I do however think that there are two numbers that we care about that have been conflated up till now.
The width of what we see in the tactic state, and separately the width of what we type into the editor.
Of course people can do what they want but the current status quo is that every time someone uses `simp?` in mathlib (a lot) they have to reflow the lines manually which is just annoying.
On the flipside I (and many others I'm sure) do find that having the tactic state be a bit wider than the "ruler" marking the maximum input width is nice, as most likely the ruler is not the rightmost part of people's editors (as we often go over 100 characters when writing so having the ruler at the edge of the screen is not ideal). So there would be a chunk of dead space in the tactic state if we limit that to exactly the desired input width.

As with web development there is not really a way to optimize these numbers without a survey of peoples set-ups and preferences, but I do think we shouldn't treat the two values as the same, and having them both (now) configurable options with sensible defaults is a good start

digama0: > On the flipside I (and many others I'm sure) do find that having the tactic state be a bit wider than the "ruler" marking the maximum input width is nice, as most likely the ruler is not the rightmost part of people's editors (as we often go over 100 characters when writing so having the ruler at the edge of the screen is not ideal).

FWIW on my setup the 100 character line takes up about 55% of the screen, so with the sidebar also showing 100 characters I'm a bit over budget and have to frequently move things back and forth to see. So I don't have any "dead space" to speak of, and 100 + 120 is even more over budget.

alexjbest: Well thats precisely my point about this being like web development! Without knowing peoples devices you cant really please everyone, hence the options!

FWIW i put the goal view under my code, it works great and i have loads of space ;)

Std/Tactic/TryThis.lean
```diff
@@ -76,6 +76,19 @@ partial def replaceMVarsByUnderscores [Monad m] [MonadQuotation m]
...
+
+/-- an option allowing the user to customize the ideal input width, this controls output format when
+the output is intended to be copied back into a lean file -/
+register_option format.inputWidth : Nat := {
+  defValue := inputWidth
+  descr := "ideal input width"
+}
+
+/-- get the input width specified in the options -/
+def getInputWidth (o : Options) : Nat := o.get `format.inputWidth inputWidth
```
digama0, line 90:
```suggestion
def getInputWidth (o : Options) : Nat := format.inputWidth.get o
```

PR #301, opened 2023-10-18
perf: improve kernel reduction of `List.sublists`

eric-wieser:
This replaces leanprover-community/mathlib4#7388, where @ChrisHughes notes

> I suspect the cause of the slowdown was the use of `Array` in `List.sublists`. An `Array` in the Kernel is just a `List`, but compiled to something more efficient. `List.sublists` uses `Array.push`, which is slow in the kernel becuase it's `l ++ [a]` on the underlying list, so it's linear time instead of constant time. So this new implementation is a better complexity in the kernel.

The difference can be seen as follows; this is fast:
```lean
set_option maxRecDepth 561 in
example : [1, 2, 3].sublists.sublists.length = 256 := rfl
```
but this hits a deterministic timeout
```lean
example : [1, 2, 3].sublistsFast.sublistsFast.length = 256 := rfl
```

---

This should not be merged until leanprover-community/mathlib4#7746 is delegated.

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -897,8 +897,10 @@ def sublistsFast (l : List α) : List (List α) :=
     fun r l => (r.push l).push (a :: l)
   (l.foldr f #[[]]).toList
 
--- The fact that this transformation is safe is proved in mathlib. Using a `csimp` lemma here is
--- impractical as we are missing a lot of lemmas about lists.
+-- The fact that this transformation is safe is proved in mathlib4 as `sublists_eq_sublistsFast`.
+-- Using a `csimp` lemma here is impractical as we are missing a lot of lemmas about lists.
+-- TODO(std4#307): upstream the necessary results about `sublists` and put the `csimp` lemma in
+-- `Std/Data/List/Lemmas.lean`.
 attribute [implemented_by sublistsFast] sublists
 
 section Forall₂
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -889,9 +889,17 @@ sublists [1, 2, 3] = [[], [1], [2], [1, 2], [3], [1, 3], [2, 3], [1, 2, 3]]
 ```
 -/
 def sublists (l : List α) : List (List α) :=
+  l.foldr (fun a acc => join (acc.map fun x => [x, a :: x])) [[]]
+
+/-- A version of `List.sublists` that has faster runtime performance but worse kernel performance -/
+def sublistsFast (l : List α) : List (List α) :=
   let f a arr := arr.foldl (init := Array.mkEmpty (arr.size * 2))
     fun r l => (r.push l).push (a :: l)
   (l.foldr f #[[]]).toList
+
+-- The fact that this transformation is safe is proved in mathlib. Using a `csimp` lemma here is
+-- impractical as we are missing a lot of lemmas about lists.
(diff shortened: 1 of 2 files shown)


Std/Data/List/Basic.lean
```diff
@@ -889,10 +889,18 @@ sublists [1, 2, 3] = [[], [1], [2], [1, 2], [3], [1, 3], [2, 3], [1, 2, 3]]
...
+
+/-- A version of `List.sublists` that has faster runtime performance but worse kernel performance -/
+def sublistsFast (l : List α) : List (List α) :=
   let f a arr := arr.foldl (init := Array.mkEmpty (arr.size * 2))
     fun r l => (r.push l).push (a :: l)
   (l.foldr f #[[]]).toList
 
+-- The fact that this transformation is safe is proved in mathlib. Using a `csimp` lemma here is
+-- impractical as we are missing a lot of lemmas about lists.
+attribute [implemented_by sublistsFast] sublists
```
digama0, line 902: These lemmas should be upstreamed, then.

eric-wieser: I think this is a reasonable number of lemmas across multiple files; even if we did upstream them, they would probably need to go in a different file so as not to make this one grow too much, and then anyone importing only this file would still not get the csimp lemma.

Is this you drawing a hard line saying "we don't ever want to use `implemented_by` in this way", or the weaker "we should put upstreaming these results on our to-do list"? Ideally we could merge this before doing any upstreaming, as everyone downstream in mathlib (multiset and Finset power sets) inherits the speed boost.

An alternative that wouldn't have to wait for upstreaming is to define `sublistsFastKernel` in mathlib, and then change all the downstream lemmas and definitions to be about that instead.

eric-wieser:
Ah sorry, I thought the draft status was something I added by accident; I now see you put it there deliberately!

digama0:
I put it there because you said it is waiting for something downstream to happen, mark it as ready when that is done

Std/Data/List/Basic.lean
digama0: I think if `sublists` has too many requirements we should just take it out of `Data.List.Basic` and move it to `Data.List.Sublists` with appropriate imports.

eric-wieser: Ok, so this PR should be considered blocked by upstreaming the required results, and having the `csimp` lemma in `Mathlib` is foo far away?

eric-wieser:
There is a circular dependency here I'm not entirely sure how to resolve; this PR shouldn't be merged unless we're happy with the mathlib fallout, but the mathlib PR is marked "blocked by another PR" (this one) so will never hit the review queue.

digama0:
I would say this PR is unblocked as soon as you have determined that the accompanying mathlib changes are doable. So we would proceed with this one first and then that one becomes a bump PR.

Std/Data/List/Basic.lean
digama0: If you think the lemma stuff is going to take a while and don't want that to hold things up here, at least put in a TODO and/or open an issue tracking moving those lemmas to std.

eric-wieser:
> at least put in a TODO and/or open an issue tracking moving those lemmas to std.

Done

> I would say this PR is unblocked as soon as you have determined that the accompanying mathlib changes are doable

https://github.com/leanprover-community/mathlib4/pull/7746 passes CI, so I think this is ready to go.

joehendrix (approved):

PR #303, opened 2023-10-18
Fixes and clarifications to BitVec documentation.

joehendrix:


tobiasgrosser:
Hi @joehendrix, the new definitions look good to me (and I also just found the msb typo). I am wondering if the naming can be more descriptive. I feel the smt prefix really is rather indirect. I personally am more used to LLVM, which also has an IR where both operands are BitVectors (https://llvm.org/docs/LangRef.html#lshr-instruction). So I was a little bit surprised to that this format is the 'smt' format. I wonder if there is a mathlib/lean convention to make this based on the types, e.g. leftShiftNat vs. leftShiftVal or just leftShift vs. leftShift'? This is just my 5 cents, so don't feel held back by this comment.

joehendrix:
@tobiasgrosser 

I personally don't care for the use of prime in naming since I find it hard to remember why one is primed and other not.

I intentionally used `Nat` as the second argument, because it is pretty unambiguous.  Languages that use bitvector's for the index tend to have subtle differences that aren't captured in the types.

The LLVM semantics is different in that a shift value larger than the bitwidth of the first argument is poison.  We want to be careful that we don't ignore that when verifying LLVM or compiling Lean to LLVM.  Similarly, Java and x86 have shifts for specific bitwidths that mask off a specific number of bits, so e.g., a 32-bit shift in Java and x86 only looks at the least significant 5-bits in the shift amount.

I tend to think people writing Lean are typically going to be shifting by positive constants and in that case the type of the index should not matter.

tobiasgrosser:
From my perspective, the LLVM semantics have an additional layer that allow for poison to be modeled and exploited. AFAIU one conservative choice of modeling LLVM would be to interpretat of poison could exactly be the semantics you propose. We are currently playing with modeling the LLVM semantics and we plan to eventually add a poison layer on to of the current BitVec layer.

> 
@tobiasgrosser

I personally don't care for the use of prime in naming since I find it hard to remember why one is primed and other not.

I intentionally used Nat as the second argument, because it is pretty unambiguous. Languages that use bitvector's for the index tend to have subtle differences that aren't captured in the types.

The LLVM semantics is different in that a shift value larger than the bitwidth of the first argument is poison. We want to be careful that we don't ignore that when verifying LLVM or compiling Lean to LLVM. Similarly, Java and x86 have shifts for specific bitwidths that mask off a specific number of bits, so e.g., a 32-bit shift in Java and x86 only looks at the least significant 5-bits in the shift amount.

> I tend to think people writing Lean are typically going to be shifting by positive constants and in that case the type of the index should not matter.

I am still not sure why this would mean we need SMT in the name. I would prefer a name that actually states what the function does, rather than referring to SMT as interpretation. The semantics that SMT choose are in my understanding the obvious semantics for having the 2nd argument a BitVector. 

Anyway, just my two cents. So if you feel you prefer smt in the name, this works for me as well.

Std/Data/BitVec/Basic.lean
```diff
@@ -319,20 +319,44 @@ instance : HShiftLeft (BitVec w) Nat (BitVec w) := ⟨.shiftLeft⟩
...
 As a numeric operation, this is equivalent to `a.toInt >>> s`.
 
-SMT-Lib name: `bvashr`.
+See `smtAShr` for an SMT-Lib compatible version.
 -/
 def sshiftRight (a : BitVec n) (s : Nat) : BitVec n := .ofInt n (a.toInt >>> s)
 
+/--
+(Logical) right shift for bit vectors. The high bits are filled with zeros.
+As a numeric operation, this is equivalent to `a / 2^s`, rounding down.
```
digama0, line 354:
```suggestion
Arithmetic right shift for bit vectors. The high bits are filled with the
most-significant bit.
As a numeric operation, this is equivalent to `a.toInt >>> s`.
```

tobiasgrosser:
From my perspective, the LLVM semantics have an additional layer that allow for poison to be modeled and exploited. AFAIU one conservative choice of modeling LLVM would be to interpretat of poison could exactly be the semantics you propose. We are currently playing with modeling the LLVM semantics and we plan to eventually add a poison layer on to of the current BitVec layer.

> 
@tobiasgrosser

I personally don't care for the use of prime in naming since I find it hard to remember why one is primed and other not.

I intentionally used Nat as the second argument, because it is pretty unambiguous. Languages that use bitvector's for the index tend to have subtle differences that aren't captured in the types.

The LLVM semantics is different in that a shift value larger than the bitwidth of the first argument is poison. We want to be careful that we don't ignore that when verifying LLVM or compiling Lean to LLVM. Similarly, Java and x86 have shifts for specific bitwidths that mask off a specific number of bits, so e.g., a 32-bit shift in Java and x86 only looks at the least significant 5-bits in the shift amount.

> I tend to think people writing Lean are typically going to be shifting by positive constants and in that case the type of the index should not matter.

I am still not sure why this would mean we need SMT in the name. I would prefer a name that actually states what the function does, rather than referring to SMT as interpretation. The semantics that SMT choose are in my understanding the obvious semantics for having the 2nd argument a BitVector. 

Anyway, just my two cents. So if you feel you prefer smt in the name, this works for me as well.

Std/Data/BitVec/Basic.lean
```diff
@@ -53,7 +53,7 @@ protected def toNat (a : BitVec n) : Nat := a.toFin.val
 /-- Return the `i`-th least significant bit. -/
 @[inline] def getLsb (x : BitVec w) (i : Nat) : Bool := x.toNat &&& (1 <<< i) != 0
 
-/-- Return the `i`-th least significant bit. -/
+/-- Return the `i`-th most significant bit. -/
```
digama0, line 56: This implementation returns the lsb when `i` is out of range, while getLsb returns 0 when `i` is out of range. I suggest making them both return zero and document it as such.

digama0:
I also tend to agree that we should reserve `smt` in the name to the case when SMT-lib implements bespoke semantics that can't easily be explained another way.

tobiasgrosser:
Sorry, I did not want to imply that we should not have the "SMTLIB" shift functions. I feel they are very useful.

joehendrix:
After some reflection, I am now thinking that maybe it doesn't make sense to add the new shift operations after all.  I've submitted a patch that backs out of those changes in favor of just the corrections.

With the current implementation, a `BitVec w` is just a natural number with a proof that the natural number is less than the `2^w`.

I don't think there is a good reason from a Lean perspective that one provide a proof that the shift amount has the same `2^w` bound as the value being shifted.  If a person finds themselves needing to shift with a bitvector shift amount, then they can always just get the underlying `Nat` at no cost and minimal typing.

joehendrix:
@tobiasgrosser Did my comment make sense?  If somebody has `x <<< y` and `y` is a `BitVec` then they can just write `x <<< y.toNat`.

I'm open to reviewing a PR from someone else or adding this if it useful later, but would prefer to prioritize other matters.

tobiasgrosser:
I think this makes sense. I am not a lean expert, but feel that this is indeed trivial. I just ported our in-progress LLVM/MLIR formalization to BitVec (https://github.com/opencompl/ssa/blob/alive-statements-to-BitVec/SSA/Projects/InstCombine/AliveStatements.lean) and we can already state theorems such as:

```
theorem bitvec_InstCombineShift__422_1:
 ∀ (Y X C : BitVec 31), (Y + X >>> C) <<< C = Y <<< C + X &&& (-1) <<< C
:= by alive_auto
```

I guess this is because of coercion or BitVec.ofNat, maybe?

tobiasgrosser (approved):

digama0:
> I don't think there is a good reason from a Lean perspective that one provide a proof that the shift amount has the same `2^w` bound as the value being shifted. If a person finds themselves needing to shift with a bitvector shift amount, then they can always just get the underlying `Nat` at no cost and minimal typing.

To this point, I was thinking of making `toNat` be a `Coe`, so that it works out of the box. The drawback is that you may want to view a bitvector as either a signed or unsigned value, and having coes for both would be a bad idea because then you get a noncommuting diagram of coercions. IMO it is reasonable to make the unsigned view the default one though.

digama0 (approved):

tobiasgrosser:
> It is also possible to have a `<<<` instance with bitvectors on both sides without a name for the resulting function. It might make type inference worse though, I haven't tested. (EDIT: @tobiasgrosser 's example seems to indicate that this already works, somehow?)

We have a couple of theorems added on top of BitVec that are not yet upstream. See e.g. https://github.com/opencompl/ssa/blob/main/SSA/Projects/InstCombine/ForStd.lean#L18, which defines:

```lean4
instance {n} : ShiftLeft (BitVec n) := ⟨fun x y => x <<< y.toNat⟩
instance {n} : ShiftRight (BitVec n) := ⟨fun x y => x >>> y.toNat⟩
```

joehendrix:
I added slight generalizations of those @tobiasgrosser's instances since that example looks great.  Thank you!  

@digama0 I think this is ready to merge.

PR #305, opened 2023-10-19
fix: `repeat'` should roll back the state

digama0:
As [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Problem.20with.20.60repeat'.60/near/397447927).

### DIFF
--- a/Std/Lean/Meta/Basic.lean
+++ b/Std/Lean/Meta/Basic.lean
@@ -350,7 +350,7 @@ or until `maxIters` total calls to `f` have occurred.
 Returns a boolean indicating whether `f` succeeded at least once, and
 all the remaining goals (i.e. those on which `f` failed).
 -/
-def repeat'Core [Monad m] [MonadError m] [MonadMCtx m]
+def repeat'Core [Monad m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
     (f : MVarId → m (List MVarId)) (gs : List MVarId) (maxIters := 100000) :
     m (Bool × List MVarId) := do
   let (progress, acc) ← go maxIters false gs [] #[]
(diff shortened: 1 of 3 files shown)


semorrison (approved):

Std/Lean/Meta/Basic.lean
```diff
@@ -390,7 +390,7 @@ then runs `f` again on all of those goals, and repeats until `f` fails on all re
 or until `maxIters` total calls to `f` have occurred.
 Fails if `f` does not succeed at least once.
 -/
-def repeat1' [Monad m] [MonadError m] [MonadMCtx m]
+def repeat1' [Monad m] [MonadError m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
```
joehendrix, line 393: I noticed that `repeat1'` is never actually used.  Instead `Std.Tactic.Basic` defines a tactic `repeat1` (with no apostrophe) that uses `repeat1'` directly.

Should `repeat1'` here be removed?

Should the tactic `repeat1` be renamed to `repeat1'` for consistency since both apply to all subgoals?

digama0: This is a public API function (I think it is used by mathlib as well), so I would rather not remove it. On the contrary I think it would be better to just call it from `repeat1`, so I've done that now.

digama0: > Should the tactic repeat1 be renamed to repeat1' for consistency since both apply to all subgoals?

I've done this, although this is a breaking change, I'm sure `repeat1` gets a fair amount of usage. Maybe we should have an alias for it or something?

semorrison: Mathlib seems to compile fine against this branch as is.

PR #306, opened 2023-10-19
fix: version numbers in code actions 

bustercopley:
[Zulip topic](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/LSP.20code.20action.20document.20version)

VS Code ignores the version number in a WorkspaceEdit but does not allow it to be null. There is at least one LSP client which does validate the version number, namely the Eglot client included in Emacs. By sending the correct version, we can be compatible with both.

~~This is a self-contained (std4-only) version, instead of the two changes (to lean4 and std4) that I mentioned on Zulip.~~

### DIFF
--- a/Std/CodeAction/Deprecated.lean
+++ b/Std/CodeAction/Deprecated.lean
@@ -60,7 +60,7 @@ def deprecatedCodeActionProvider : CodeActionProvider := fun params snap => do
         let pos := doc.meta.text.leanPosToLspPos msg.pos
         let endPos' := doc.meta.text.leanPosToLspPos endPos
         pure { eager with
-          edit? := some <| .ofTextEdit params.textDocument.uri {
+          edit? := some <| .ofTextEdit doc.versionedIdentifier {
             range := ⟨pos, endPos'⟩
             newText := toString c'
           }
--- a/Std/CodeAction/Misc.lean
+++ b/Std/CodeAction/Misc.lean
@@ -85,7 +85,7 @@ instance : Monad Id := {
         else
           str := str ++ indent ++ "}"
       pure { eager with
-        edit? := some <| .ofTextEdit params.textDocument.uri {
+        edit? := some <| .ofTextEdit doc.versionedIdentifier {
           range := doc.meta.text.utf8RangeToLspRange ⟨holePos, info.stx.getTailPos?.get!⟩
           newText := str
         }
@@ -131,7 +131,7 @@ def foo : Expr → Unit := fun
 ```
 
 -/
-@[hole_code_action] def eqnStub : HoleCodeAction := fun params snap ctx info => do
+@[hole_code_action] def eqnStub : HoleCodeAction := fun _ snap ctx info => do
   let some ty := info.expectedType? | return #[]
   let .forallE _ dom .. ← info.runMetaM ctx (whnf ty) | return #[]
   let .const name _ := (← info.runMetaM ctx (whnf dom)).getAppFn | return #[]
@@ -156,7 +156,7 @@ def foo : Expr → Unit := fun
           str := str ++ if arg.hasNum || arg.isInternal then " _" else s!" {arg}"
         str := str ++ s!" => {holeKindToHoleString info.elaborator ctor}"
       pure { eager with
-        edit? := some <|.ofTextEdit params.textDocument.uri {
+        edit? := some <|.ofTextEdit doc.versionedIdentifier {
           range := doc.meta.text.utf8RangeToLspRange ⟨holePos, info.stx.getTailPos?.get!⟩
           newText := str
         }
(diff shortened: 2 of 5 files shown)


bustercopley:
Fixed unused variable linter errors

joehendrix:
@bustercopley Thanks, I'll find the right reviewer for this so we can figure out if it should belong in std or core.

bustercopley:
Thanks!
Fixed long line linter error.

david-christiansen:
From my reading of [the spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#optionalVersionedTextDocumentIdentifier), providing a null version means that the server has not yet heard of the file being open, rather than that the server simply doesn't want to send it, so this seems to strictly increase the conformance to the protocol.

Std/CodeAction/Basic.lean
```diff
@@ -9,6 +9,31 @@ import Std.Lean.Name
...
+
+In core lean, the version number is always 0.
+See Lean.Lsp.WorkspaceEdit.ofTextEdit in Lean.Data.Lsp.Basic.
+-/
+
+namespace Lean.Lsp.WorkspaceEdit
+
+/-- Construct a WorkspaceEdit from a VersionedTextDocumentIdentifier and a TextEdit --/
+def ofVersionedTextEdit (td : VersionedTextDocumentIdentifier) (te : TextEdit) : WorkspaceEdit :=
+  ofTextDocumentEdit { textDocument := td, edits := #[te]}
```
digama0, line 23:
```suggestion
  ofTextDocumentEdit { textDocument := td, edits := #[te] }
```

Std/CodeAction/Basic.lean
```diff
@@ -9,6 +9,31 @@ import Std.Lean.Name
...
+/-!
+# Helper constructors for WorkspaceEdit with document version
+
+In core lean, the version number is always 0.
+See Lean.Lsp.WorkspaceEdit.ofTextEdit in Lean.Data.Lsp.Basic.
+-/
+
+namespace Lean.Lsp.WorkspaceEdit
+
+/-- Construct a WorkspaceEdit from a VersionedTextDocumentIdentifier and a TextEdit --/
```
digama0, line 21: This should be in `Std.Lean`, maybe `Std.Lean.Lsp`

mhuisi: I think it may be better off in core as originally suggested by @bustercopley. Alternatively, if we keep it in std, we should delete `ofTextEdit` in core so that downstream users do not accidentally use the wrong function. I think I would personally prefer the former.

digama0: SGTM

Std/CodeAction/Basic.lean
```diff
@@ -9,6 +9,31 @@ import Std.Lean.Name
...
+def ofVersionedTextEdit (td : VersionedTextDocumentIdentifier) (te : TextEdit) : WorkspaceEdit :=
+  ofTextDocumentEdit { textDocument := td, edits := #[te]}
+
+end Lean.Lsp.WorkspaceEdit
+
+namespace Lean.Server.FileWorker.EditableDocument
+
+/-- Construct a VersionedTextDocumentIdentifier from an EditableDocument --/
+def textDocument (ed : EditableDocument) : Lsp.VersionedTextDocumentIdentifier := {
```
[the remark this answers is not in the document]
bustercopley: `textDocument` is the usual name in the LSP spec for JSON fields of this type (and the related types TextDocumentIdentifier and OptionalVersionedTextDocumentIdentifier), and also in Lean. For example in `Lean.Data.Lsp.TextSync.lean`,
```
structure DidChangeTextDocumentParams where
  textDocument : VersionedTextDocumentIdentifier
  contentChanges : Array TextDocumentContentChangeEvent
  deriving ToJson, FromJson
```

Std/CodeAction/Basic.lean
bustercopley: Yes, especially since the changes to core amount to a simplificiation/streamlining, that is the natural place to do it. (I'm a bit intimidated by the process for core Lean PRs.)

Std/CodeAction/Basic.lean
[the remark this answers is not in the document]
mhuisi: LSP uses `textDocument` to refer to both `TextDocumentItem`s and various kinds of identifiers, and the Lean LSP data structures mirror that convention so that we can derive the corresponding JSON handlers. Regardless, I don't think that confusing the two is something that we should do in our own code, especially in server code that usually works with an `EditableDocument` and refers to it as `doc`.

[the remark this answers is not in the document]
bustercopley: True, and my initial instinct was to use `.versionedTextDocumentIdentifier`; I'd be happy with that, but if it needs an abbreviation, `textDocument` seems natural. It needn't be construed as confusion.
Just my opinion. I won't quibble if you make a decision.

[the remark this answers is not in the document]
mhuisi: I would prefer `versionedIdentifier`, as the code using it will typically read as `doc.versionedIdentifier`.

Std/CodeAction/Basic.lean
mhuisi: > Yes, especially since the changes to core amount to a simplificiation/streamlining, that is the natural place to do it. (I'm a bit intimidated by the process for core Lean PRs.)

It's fine to PR this without a corresponding RFC since it is very small, uncontroversial and you discussed it with the Lean community and the respective code owner :)

bustercopley: @mhuisi :) Do you mind saying something along those lines in the Zulip topic?

Std/CodeAction/Basic.lean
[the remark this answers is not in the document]
bustercopley: Renamed to `.versionedIdentifier`.

bustercopley:
Force push: now depends on core [#2721](https://github.com/leanprover/lean4/pull/2721); squashed and rebased.

PR #312, opened 2023-10-21
chore: fixes for new simp default (`decide := false`) (lean4#2722)

collares:
Adapt to new default in Lean master.

### DIFF
--- a/Std/Data/Fin/Lemmas.lean
+++ b/Std/Data/Fin/Lemmas.lean
@@ -645,7 +645,7 @@ theorem exists_fin_two {p : Fin 2 → Prop} : (∃ i, p i) ↔ p 0 ∨ p 1 :=
   exists_fin_succ.trans <| or_congr_right exists_fin_one
 
 theorem fin_two_eq_of_eq_zero_iff : ∀ {a b : Fin 2}, (a = 0 ↔ b = 0) → a = b := by
-  simp [forall_fin_two]
+  simp only [forall_fin_two]; decide
 
 /--
 Define `motive i` by reverse induction on `i : Fin (n + 1)` via induction on the underlying `Nat`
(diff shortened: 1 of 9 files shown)


Std/Data/Fin/Lemmas.lean
```diff
@@ -644,7 +644,7 @@ theorem exists_fin_two {p : Fin 2 → Prop} : (∃ i, p i) ↔ p 0 ∨ p 1 :=
   exists_fin_succ.trans <| or_congr_right exists_fin_one
 
 theorem fin_two_eq_of_eq_zero_iff : ∀ {a b : Fin 2}, (a = 0 ↔ b = 0) → a = b := by
-  simp [forall_fin_two]
+  simp [forall_fin_two]; decide
```
semorrison, line 647: This probably counts as a nonterminal simp.

semorrison:
These changes seem pretty reasonable to me.

Std/Data/Fin/Lemmas.lean
digama0: I would say it's an exception, `decide` is one of those tactics that is not very sensitive to the result of `simp`. Although, it seems like this could just be `by decide` without the `simp`?

collares: Unfortunately just `by decide` returns
```
failed to synthesize
  Decidable (∀ {a b : Fin 2}, (a = 0 ↔ b = 0) → a = b)
```

collares: Either way, I replaced `simp` by `simp only` just to be sure.

collares:
awaiting-review

semorrison:
@digama0, @joehendrix, could this please be reviewed and merged? Note that this will go into the `bump/v4.4.0` branch, not `main`, which is intended to accummulate *approved* changes that are needed for the next release.

After this is merged I will merge https://github.com/leanprover-community/mathlib4/pull/7834 into Mathlib's `bump/v4.4.0` branch as well.

digama0 (approved):

PR #313, opened 2023-10-22
feat: Array.get_pop and List.get_dropLast

nomeata:
I needed these lemmas in a little project of mine, so maybe they are useful to
others.


Std/Data/Array/Lemmas.lean
```diff
@@ -136,6 +136,14 @@ theorem swapAt!_def (a : Array α) (i : Nat) (v : α) (h : i < a.size) :
...
 @[simp] theorem pop_push (a : Array α) : (a.push x).pop = a := by simp [pop]
 
+theorem get_pop (a : Array α) (i : Fin a.pop.size) :
+    a.pop.get i = a.get (i.castLE (a.size_pop ▸ Nat.sub_le _ _ )) :=
+  List.get_dropLast _ _
+
+@[simp] theorem getElem_pop (a : Array α) (i : Nat) (hi : i < a.pop.size) :
+    a.pop[i] = a[i]'(Nat.lt_of_lt_of_le (a.size_pop ▸ hi) (Nat.sub_le _ _)) :=
+  List.get_dropLast _ _
+
```
joehendrix, line 146: There are a lot of theorems like this potentially.  How about picking a`simp`-normal form for the element of an array?

I have a slight preference for `a[i]'h` over `a.get i` since only the proof is dependent, but I noticed `List` seems to simplify `l[i]'h` to `l.get <i, h>`.

nomeata: It seems for arrays it already has `a[i]` as the simp normal form, that's why the linter nudged me to add this lemma (I myself only needed the `get` lemma).

Or are you just talking about the RHS, `a[i]'h` vs. `a.get`?

I don't have an informed opinion here, happy to take advise.

semorrison:

semorrison:
awaiting-review

Std/Data/Array/Lemmas.lean
joehendrix: If `a[i]` is the simp normal form, would `getElem_pop` alone (no `get_pop` with a simp rule `a.get i = a[i]` meet your goals?

`get_pop` on it's own is only one lemma, but I am wondering about consistency across all operations that return an array.

nomeata: I guess you are saying that I should simply avoid `a.get i` in my project, and write `a[↑i]` everywhere instead? But that coercion and the hidden `h` makes me anxious if this is really a good idea. I can try at some point.

nomeata: Indeed I could just use `[i]` in my formalization, and even get some more proofs done by `simp`. So happy to drop `Array.get_pop` here.

semorrison (approved):

nomeata:
Thanks for golfing!

PR #314, opened 2023-10-22
feat: de-mathlib `Nat.binaryRec`

negiizhao:


Std/Data/Nat/Lemmas.lean
```diff
@@ -991,3 +1016,198 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
...
+
+theorem bit_decomp (n : Nat) : bit (bodd n) (div2 n) = n :=
+  (bit_val _ _).trans (bodd_add_div2 _)
+
+theorem bit_eq_zero_iff {n : Nat} {b : Bool} : bit b n = 0 ↔ n = 0 ∧ b = false := by
+  constructor
+  · cases b <;> simp [Nat.bit, Nat.add_eq_zero]
+  · simp (config := {contextual := true})
+
+/-! ### binary rec/cases -/
```
fgdorais, line 1134: The defs below (and possibly a few auxiliary lemmas for decreasing_by) should be in `Std/Data/Nat/Basic.lean`

Std/Data/Nat/Basic.lean
```diff
@@ -126,3 +126,92 @@ where
...
+  (1 &&& n) != 0
+
+/-- `div2 n = ⌊n/2⌋` the greatest integer smaller than `n/2`-/
+def div2 (n : Nat) : Nat :=
+  n >>> 1
+
+/-- `bit b` appends the digit `b` to the binary representation of
+  its natural number input. -/
+def bit (b : Bool) (n : Nat) : Nat :=
+  cond b (n + n + 1) (n + n)
```
joehendrix, line 141: I'm not a huge fan of `bit` in this namespace as is.  The meaning isn't obvious to me and `Nat`.  Could we replace `bit b n` with `2*n + ofBool b` where `abbrev ofBool b := cond b 1 0`?

joehendrix: I have lesser, but similar concerns about `div2`.  Perhaps just write `n / 2`?

digama0: Agreed, these definitions are holdovers from the lean 3 way of representing numbers, and I think we can just do without them. `bit` is a somewhat useful spec function (it is a sort of inductive constructor of binary `Nat`), but it could definitely have a longer name like `Nat.pushBit` or `Nat.shiftBit` or something along those lines.

`div2` we can just remove, there is no need for it in these days of kernel bignums.

fgdorais: > I'm not a huge fan of `bit` in this namespace as is. The meaning isn't obvious to me and `Nat`. Could we replace `bit b n` with `2*n + ofBool b` where `abbrev ofBool b := cond b 1 0`?

There would be no need for this abbrev if there was a coercion from Bool to Nat. I'm in favor of such a coercion but there is room for debate.

fgdorais: > Agreed, these definitions are holdovers from the lean 3 way of representing numbers, and I think we can just do without them. `bit` is a somewhat useful spec function (it is a sort of inductive constructor of binary `Nat`), but it could definitely have a longer name like `Nat.pushBit` or `Nat.shiftBit` or something along those lines.
> 
> `div2` we can just remove, there is no need for it in these days of kernel bignums.

I totally agree but there should be some more exposure of the kernel bignum so that people could implement and reason about efficient functions in Lean. Ultimately, Lean should only rely on a bignum library that only implements Lean-verified algorithms!

negiizhao: For big `n`, `n / 2` is still slower than `n >>> 1`, although they are often both relatively fast. Maybe we should let the compiler optimize it.

digama0: Are you sure (have you profiled)? Keep in mind that `/` has a kernel implementation but `>>>` does not.

negiizhao:
```lean
def div2 (x : Nat) : Nat := x / 2
def div2' (x : Nat) : Nat := x >>> 1

def run_div2 (n : Nat) (x : Nat) : IO Unit := do
  let mut t := 0
  for _ in [0 : n] do
    t := div2 x

def run_div2' (n : Nat) (x : Nat) : IO Unit := do
  let mut t := 0
  for _ in [0 : n] do
    t := div2' x

def n := 10^4
def x := 11^1000000

#eval timeit "x/2" (run_div2 n x)
#eval timeit "x>>>1" (run_div2' n x)
```
I'm not sure if it's the right way. In this example, `x >>> 1` is always significantly faster when `x` is large.

joehendrix: Thanks for the analysis!  I think if `x >>> 1` is faster than `x/2` then `x >>> 1` is good with me.  One could also implement it in the kernel if that leads to significant speedup.

How about renaming `bit` to `shiftBit` and making `shiftBit` an abbreviation so it is transparent to decision procedures?

I'd also favor reordering the arguments to `def shiftBit : Nat -> Bool -> Nat` so you have a visually consistent ordering (e.g., `shiftBit 0b101 true = 0b1011`).  That's a taste issue though.

digama0: heh, you know that order is deliberate...

joehendrix: Yes, I figured it was deliberate.  Lean has always had most-significant digit on the left in numeric literals, so I'm trying to be consistent with that.

That said, I'd recommend we merge with `bit` renamed to `shiftBit` and the definition made an abbreviation.

digama0: Why an abbreviation? That has specific consequences and I'm not sure they are relevant or appropriate here.

fgdorais: inline does make sense though.

joehendrix: I guess we could see.  My thinking is `shiftBit` would be something one would want a linear arithmetic decision procedure to see through.  Maybe that's premature.

digama0: I would rather we just taught said linear arithmetic decision procedure about shiftBit directly.

joehendrix: I'm not quite sure how to implement a linear arithmetic decision procedure without decoding `shiftBit` into a polynomial at some point, but I'm happy to table that discussion.  If there's a concrete need for a change we could make that later.  It still needs to be renamed from `bit` to `shiftBit`.

digama0: Yes, I'm saying that the linear arithmetic procedure would see `shiftBit` and apply a theorem saying it's a specific polynomial because it's `shiftBit` and it knows about this function just like it knows about `+` and `*`. It can do this without `shiftBit` needing to be an `abbrev`.

eric-wieser: If this consensus is to rename `bit ` to `shiftBit`, I think a PR to make the rename in mathlib first would reduce friction when bumping Std.

I'm not a huge fan of the `shiftBit` name; it doesn't really make it clear where the bit is being shifted; `pushLsb` seems clearer in this regard, at the expense of not including the word `bit`.

fgdorais: Indeed, we should try to harmonize with `Mathlib.Init.Data.Nat.Bitwise` and `Mathlib.Data.Nat.Bitwise`. They currently use [`shiftLeft'`](https://github.com/leanprover-community/mathlib4/blob/8f163221eaac63ae5e4d71fe84e34d68c727899c/Mathlib/Init/Data/Nat/Bitwise.lean#L180C1-L180C1) for this (not that this is better name). There's a couple more awkward conflicts there such as `shiftLeft_eq` and `shiftLeft_eq'`.

alexkeizer: > I'm not a huge fan of the shiftBit name; it doesn't really make it clear where the bit is being shifted; pushLsb seems clearer in this regard, at the expense of not including the word bit.

I'm also not a big fan of `shiftBit`; my suggestion would be `concatBit` or `concatMsb`, following the suggestion to reorder the arguments so that the new bit is the second argument. The operation, with that order, is somewhat like `List.concat`, so the name would be consistent with that.

eric-wieser: `concatBit` seem misleading to me, as its meaning is only correct if you think of a bit-vector as a *big*-endian list of bits; for *little* endian, it would be `consBit`.

negiizhao:
awaiting-review

eric-wieser:
What's the status of this PR in respect to #366?

Std/Data/Nat/Lemmas.lean
```diff
@@ -356,10 +356,10 @@ protected theorem eq_zero_of_add_eq_zero_right (h : n + m = 0) : n = 0 :=
 protected theorem eq_zero_of_add_eq_zero_left (h : n + m = 0) : m = 0 :=
   (Nat.eq_zero_of_add_eq_zero h).2
 
-protected theorem add_eq_zero {n m : Nat} : n + m = 0 ↔ n = 0 ∧ m = 0 :=
-  ⟨eq_zero_of_add_eq_zero, by simp (config := {contextual := true})⟩
+protected theorem add_eq_zero_iff : n + m = 0 ↔ n = 0 ∧ m = 0 :=
```
fgdorais: Not sure the name change is necessary. What else would `Nat.add_eq_zero` mean?

Std/Data/Nat/Lemmas.lean
```diff
@@ -356,10 +356,10 @@ protected theorem eq_zero_of_add_eq_zero_right (h : n + m = 0) : n = 0 :=
...
+  ⟨Nat.eq_zero_of_add_eq_zero, fun ⟨h₁, h₂⟩ => h₂.symm ▸ h₁⟩
 
-protected theorem add_left_cancel_iff {n m k : Nat} : n + m = n + k ↔ m = k :=
+protected theorem add_left_cancel_iff {n : Nat} : n + m = n + k ↔ m = k :=
```
fgdorais: I would actually be in favor of making `n` explicit here but that's debatable. Just voicing an opinion here, this is not a required change.

Std/Data/Nat/Lemmas.lean
```diff
@@ -1307,6 +1312,27 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
     rw [shiftRight_add, shiftRight_eq_div_pow m k]
     simp [Nat.div_div_eq_div_mul, ← Nat.pow_succ]
 
+theorem mul_add_div {m : Nat} (m_pos : m > 0) (x y : Nat) : (m * x + y) / m = x + y / m := by
```
fgdorais: It's also debatable whether `m` should be implicit here.

Std/Data/Nat/Lemmas.lean
eric-wieser: I don't know what diff you're commenting on here, but this name is in main already:

https://github.com/leanprover/std4/blob/fa17cba0459830906e99f0a81fb0edc106220380/Std/Data/Nat/Lemmas.lean#L358-L359

negiizhao:
I'm not sure where these diffs are. It might just be some mess from the merge.

negiizhao:
> What's the status of this PR in respect to #366? In that PR we got:
> 
> ```
> /--
> An induction principal that works on divison by two.
> -/
> noncomputable def div2Induction {motive : Nat → Sort u}
>     (n : Nat) (ind : ∀(n : Nat), (n > 0 → motive (n/2)) → motive n) : motive n := by
>   induction n using Nat.strongInductionOn with
>   | ind n hyp =>
>     apply ind
>     intro n_pos
>     if n_eq : n = 0 then
>       simp [n_eq] at n_pos
>     else
>       apply hyp
>       exact Nat.div_lt_self n_pos (Nat.le_refl _)
> ```
> 
> but these don't seem directly related.

I replaced it with `binaryRec`s and did some golf.

fgdorais:
Strange. Please ignore my comments then.

negiizhao:
Well, what should I do now? `Std.Data.Nat.Bitwise` is not there now.

semorrison:
> It seems that `Std.Data.Nat.Bitwise` is not there now. What should I do now? Should I create a PR for core?

Yes.

fgdorais:
Marking as `duplicate` while waiting for [lean4#3756](https://github.com/leanprover/lean4/pull/3756). Ask to remove that label if that PR doesn't get merged.

PR #321, opened 2023-10-25
feat: Migrate #print prefix from Mathlib

joehendrix:
This migrates #print prefix while doing some migration of functionality naming to std conventions.

This also restricts the long line check to Std.  I had some long lines in the test case file that aren't easy to fix.

### DIFF
--- a/Std/Lean/Util/EnvSearch.lean
+++ b/Std/Lean/Util/EnvSearch.lean
@@ -29,6 +29,7 @@ def getMatchingConstants {m} [Monad m] [MonadEnv m] (p : ConstantInfo → m Bool
                     pure #[]
   (← getEnv).constants.map₂.foldlM (init := matches_) check
 where
+  /-- Check constant should be returned -/
   check matches_ name cinfo := do
     if opts.checkPrivate || !isPrivateName name then
       if ← p cinfo then
--- a/Std/Lean/Util/EnvSearch.lean
+++ b/Std/Lean/Util/EnvSearch.lean
@@ -22,7 +22,10 @@ structure EnvironmentSearchOptions where
 /--
 Find constants in current environment that match find options and predicate.
 -/
-def getMatchingConstants {m} [Monad m] [MonadEnv m] (p : ConstantInfo → m Bool) (opts : EnvironmentSearchOptions := {}) : m (Array ConstantInfo) := do
+def getMatchingConstants {m} [Monad m] [MonadEnv m]
+    (p : ConstantInfo → m Bool)
+    (opts : EnvironmentSearchOptions := {})
+    : m (Array ConstantInfo) := do
   let matches_ ← if opts.stage1 then
                     (← getEnv).constants.map₁.foldM (init := #[]) check
                    else
--- a/Std/Lean/Util/EnvSearch.lean
+++ b/Std/Lean/Util/EnvSearch.lean
@@ -3,10 +3,7 @@ Copyright (c) 2021 Shing Tak Lam. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Shing Tak Lam, Daniel Selsam, Mario Carneiro
 -/
-
 import Lean.Modifiers
-
-open Lean
 
 namespace Lean
 
@@ -26,7 +23,7 @@ def getMatchingConstants {m} [Monad m] [MonadEnv m]
     (p : ConstantInfo → m Bool)
     (opts : EnvironmentSearchOptions := {})
     : m (Array ConstantInfo) := do
-  let matches_ ← 
+  let matches_ ←
     if opts.stage1 then
       (← getEnv).constants.map₁.foldM (init := #[]) check
     else
@@ -42,5 +39,3 @@ where
         pure matches_
     else
       pure matches_
-
-end Lean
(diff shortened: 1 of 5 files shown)


Std/Lean/Util/EnvSearch.lean
```diff
@@ -0,0 +1,45 @@
...
+Find constants in current environment that match find options and predicate.
+-/
+def getMatchingConstants {m} [Monad m] [MonadEnv m]
+    (p : ConstantInfo → m Bool)
+    (opts : EnvironmentSearchOptions := {})
+    : m (Array ConstantInfo) := do
+  let matches_ ← if opts.stage1 then
+                    (← getEnv).constants.map₁.foldM (init := #[]) check
+                   else
+                    pure #[]
```
semorrison, line 32: I would have formatted this as 
```suggestion
  let matches_ ← 
    if opts.stage1 then
      (← getEnv).constants.map₁.foldM (init := #[]) check
    else
      pure #[]
```
but I'm not sure there is actually a style rule against deep indenting like this?

Std/Lean/Util/EnvSearch.lean
```diff
@@ -0,0 +1,45 @@
...
+-/
+
+import Lean.Modifiers
+
+open Lean
+
+namespace Lean
+
+/--
+Options to control getMatchingConstants options below.
```
semorrison, line 14:
```suggestion
Options to control `getMatchingConstants` options below.
```
One day names in doc-strings will hyperlink. :-)

Std/Tactic/PrintPrefix.lean
```diff
@@ -0,0 +1,48 @@
...
+    let mut msg ← appendMatchingConstants "" fun cinfo => pure $ nameId.isPrefixOf cinfo.name
+    if msg.isEmpty then
+      if let [name] ← resolveGlobalConst name then
+        msg ← appendMatchingConstants msg fun cinfo => pure $ name.isPrefixOf cinfo.name
+    if !msg.isEmpty then
+      logInfoAt tk msg
+| _ => throwUnsupportedSyntax
+
+end Elab.Command
+end Lean
```
semorrison, line 48: Mario once told me he prefers not to close namespaces if it is only being done at the end of the file. I forget why, however.

I would actually prefer the opposite convention, as it makes it easier doing single-file-minimization (i.e. you can copy a file into a section of another, and only have to move the imports, not adjust namespaces too).

semorrison (approved):
Otherwise LGTM.

Std/Tactic/PrintPrefix.lean
fgdorais: Maybe these coding conventions should be recorded somewhere. I also did not know about that one!

Std/Tactic/PrintPrefix.lean
```diff
@@ -0,0 +1,48 @@
...
+  let cinfos := cinfos.qsort fun p q => p.name.lt q.name
+  let mut msg := msg
+  for cinfo in cinfos do
+    msg := msg ++ s!"{cinfo.name} : {← Meta.ppExpr cinfo.type}\n"
+  pure msg
+
+/--
+Command for #print prefix
+-/
+syntax (name := printPrefix) "#print prefix " ident : command
```
fgdorais, line 23: My personal version of used `"#print" &"prefix"` rather than `"#print prefix"`. I'm not sure my version is better but I am interested in learning more about the difference. (It's been so long I don't remember why I did that!)

fgdorais (approved):
This is better than my personal version, so I'm approving but I do have a question in the review comments.

Std/Tactic/PrintPrefix.lean
joehendrix: It looks like this dates back at least to when it was [checked into Mathlib](https://github.com/leanprover-community/mathlib4/commit/81e736ed58fe4f5ef6bba68f4aeb84916d7ea093) by @digama0

Std/Tactic/PrintPrefix.lean
digama0: The rule is that you shouldn't close the "main namespace" of the file, to make it easier to end-extend the file.

Std/Tactic/PrintPrefix.lean
digama0: ``"#print " "prefix "` is better if the intent is to allow more or less whitespace between the tokens. One probably needs to also add a `colGt` or something, or else unfinished `#print` statements cannot be differentiated from `#print prefix` where the `prefix` just happens to be the next command.

joehendrix: Interesting, I needed to learn this.  It seems like I should allow whitespace, but I noticed that i also need `withPosition`

Consider these two examples:

```
#print
  prefix
  Nat

#print
prefix
Nat
```

I think I want to parse the first, but not the second.  However the line below parses both:

```
syntax (name := printPrefix) "#print " (colGt "prefix ") (colGt ident) : command
```

This works though:

```
syntax (name := printPrefix) withPosition("#print " (colGt "prefix ") (colGt ident)) : command
```

Any feedback?  I was surprised I needed an explicit `withPosition` since other similar examples didn't seem to use it.

digama0: Yes, you will need a `withPosition` because commands are parsed at the top level and this doesn't have a position set. (Tactics, by contrast, appear somewhere after a `by` and this sets the position.)

PR #317, opened 2023-10-23
chore: automation to maintain a nightly-testing branch

semorrison:
Adds the following steps to CI:
* On every commit to `main`, merges `main` into `nightly-testing` (with all conflicts resolved in favour of `nightly-testing`)
* Once per day, updates `lean-toolchain` on the `nightly-testing` branch to point to the latest nightly release.
* On each CI failure on the `nightly-testing` branch, writes a message to the [mathlib reviewers](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers) > [CI failure on the nightly-testing branch](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/CI.20failure.20on.20the.20nightly-testing.20branch) topic.

I plan to merge `depends on core changes` labelled PRs into `nightly-testing` as the corresponding PRs to Lean land in nightly releases.

Finally, I am going to tweak the Mathlib CI for `nightly-testing` so that it runs `lake update std` daily (it will be pointing at the  `nightly-testing` branch of Std).

---

This will need two secrets installed in the repository, which I'll message @joehendrix and @digama0 about privately.

PR #322, opened 2023-10-26
chore: move workflows to correct directory

semorrison:
Oops, I put the new CI files from #317 in the wrong directory, so they weren't doing anything.


digama0:
@semorrison The master branch has been red since this was merged. Is the new workflow supposed to be gating new PRs?

PR #324, opened 2023-10-27
chore: upstream some lemmas about if-then-else

semorrison:
---

Trying to reduce Mathlib dependencies as they arise: I needed a few lemmas about ite.

### DIFF
--- a/Std/Logic.lean
+++ b/Std/Logic.lean
@@ -29,6 +29,10 @@ theorem not_of_not_imp {a : Prop} : ¬(a → b) → ¬b := mt fun h _ => h
 
 @[simp] theorem imp_not_self : (a → ¬a) ↔ ¬a := ⟨fun h ha => h ha ha, fun h _ => h⟩
 
+/-! ## ne -/
+
+theorem ne_comm {α} {a b : α} : a ≠ b ↔ b ≠ a := ⟨Ne.symm, Ne.symm⟩
+
 /-! ## iff -/
 
 -- This is needed for `calc` to work with `iff`.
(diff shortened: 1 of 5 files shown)


Std/Logic.lean
```diff
@@ -755,6 +777,123 @@ theorem apply_ite (f : α → β) (P : Prop) [Decidable P] (x y : α) :
 @[simp] theorem ite_not (P : Prop) [Decidable P] (x y : α) : ite (¬P) x y = ite P y x :=
   dite_not P (fun _ => x) (fun _ => y)
 
+variable {σ : α → Sort _} (f : α → β) {P Q : Prop} [Decidable P] [Decidable Q]
```
fgdorais, line 780: To facilitate maintenance, `variable` should be used sparingly in Std. Especially binder updates as seen below.

semorrison: Hmm.. I'm not sure what to do here.

I think the binders are all pretty erratic here.

Options:

1. Upstream as-is, to avoid having to deal with Mathlib breakages (which I'd prefer not to have to do!), and someone can curate the binders later?
2. I make a much smaller PR, that has exactly the handful of lemmas I need, rather than the whole section.
3. Go through and curate the binders now. My instinct is that these should basically all be `{}`, but there are presumably good reasons Mathlib has made some of them `()`. 
4. Curate the binders in Mathlib first, where breakages are easier to identify.

I'm happy to do 1. or 2., but would prefer to avoid the time sink of 3. or 4.

fgdorais: I understand the predicament, this is a mess to fix up! I think 2 is best for the time being.

semorrison: Okay, I may wait for an opinion from @digama0 or @joehendrix before doing anything here.

digama0: I think 2 is a good idea: mathlib has a lot of grandfathered theorems from lean 3 which have never really been used and are just being cargo culted around so I would like us to only upstream things with a manifest use case. (In fact, I believe these lemmas were deliberately skipped in my initial port of the file because of this.) That is, I'm not even sure mathlib uses these lemmas, and question the claim that the binder choice has been subject to deep thought.

If we determine that we need the theorems, I would suggest doing 3 in this PR. The mathlib breakage can be managed in the bump PR, and it would be best to get a sense of exactly what kind of things are liable to lead to major breakage before making any changes here.

PR #331, opened 2023-10-30
chore: changes required for leanprover/lean4#2783

semorrison:
These are the changes required for leanprover/lean4#2783. Please review skeptically!


digama0:
This seems to contain more than just https://github.com/leanprover/lean4/pull/2783 . Should this perhaps be a PR which targets the nightly-testing branch instead? I'm not really sure how the review process for these kind of PRs is supposed to work (or where I should push suggested changes), as they can't be merged as is to `main`.

Std/Classes/LawfulMonad.lean
```diff
@@ -93,7 +94,8 @@ protected theorem map [Functor m] [LawfulFunctor m] {x : m α}
     (hx : SatisfiesM p x) (hf : ∀ {a}, p a → q (f a)) : SatisfiesM q (f <$> x) := by
   let ⟨x', hx⟩ := hx
   refine ⟨(fun ⟨a, h⟩ => ⟨f a, hf h⟩) <$> x', ?_⟩
-  rw [← hx]; simp [← comp_map, Function.comp]
+  rw [← hx];
```
digama0, line 97:
```suggestion
  rw [← hx]
```

digama0:
We also need to restore a bunch of `@[simp]` attributes on List.Basic definitions and remove the named equation theorems (or keep them but remove the `@[simp]` attribute). This is that big `attribute [simp]` block in Data.List.Init.Lemmas that had to be broken up because of the issue that https://github.com/leanprover/lean4/pull/2783 fixes.

semorrison:
> We also need to restore a bunch of `@[simp]` attributes on List.Basic definitions and remove the named equation theorems (or keep them but remove the `@[simp]` attribute). This is that big `attribute [simp]` block in Data.List.Init.Lemmas that had to be broken up because of the issue that [leanprover/lean4#2783](https://github.com/leanprover/lean4/pull/2783) fixes.

Could we please defer that? It seems this could be done optionally/later, and I am not keen for this to become a blocker for releasing a new version.

Std/Classes/LawfulMonad.lean
```diff
@@ -74,7 +74,8 @@ namespace SatisfiesM
 /-- If `p` is always true, then every `x` satisfies it. -/
 theorem of_true [Applicative m] [LawfulApplicative m] {x : m α}
     (h : ∀ a, p a) : SatisfiesM p x :=
-  ⟨(fun a => ⟨a, h a⟩) <$> x, by simp [← comp_map, Function.comp]⟩
+  ⟨(fun a => ⟨a, h a⟩) <$> x,
+    by simp (config := { unfoldPartialApp := true }) [← comp_map, Function.comp]⟩
```
semorrison, line 78: @digama0, in Mathlib we have the lemma `Function.comp_def`, and there I have used that to avoid needing `unfoldPartialApp := true` every time we want to unfold `Function.comp`.

Do you think I should move that up to Std?

semorrison:
> This seems to contain more than just [leanprover/lean4#2783](https://github.com/leanprover/lean4/pull/2783) . Should this perhaps be a PR which targets the nightly-testing branch instead? I'm not really sure how the review process for these kind of PRs is supposed to work (or where I should push suggested changes), as they can't be merged as is to `main`.

I've fix this so it is now relative to `nightly-testing-2023-11-06`, i.e. the same nightly that the Lean PR was based off. It can't actually be merged as is: I think the right thing to do is just approve, and then when I do a bump PR for `main` that needs to include this I'll just link to this, and close after the bump PR has been merged.

Std/Classes/LawfulMonad.lean
semorrison: Okay, I've done this in #367 now.

semorrison:
Closing this in favour of #367

PR #332, opened 2023-10-30
feat: DiscrTree: Avoid unsafe/partial definitions

nomeata:
in #285 there was a comment that it is desirable to avoid `unsafe` and
`partial` definitions here. This PR explores making these definitions
not `unsafe` by proving termination.

This requires `Array.attach`, analogous to `List.attach`. Furthermore,
because the recursion also goes through another layer of `Prod`, a custom
lemma and recursion tactic seems to be required. I conjecture that a more
compositional default `recursion_tactic` setup might make that not
necessary.

The function `mergePreservingDuplicates` is particularly tricky, as it
recurses through `Array.mergeSortedMergingDuplicates`, which I have to
generalize to make this work. This examples show a benefit of extrinsic
termination proofs a la Isabelle (where one would have to prove a
separate `congr` lemma for `Array.mergeSortedMergingDuplicates`, but
would not have to change its definition), although I am not yet sure if
such an approach to well-founded recursion is possible in Lean.


nomeata:
@digama0, in #285 you asked if these definitions could be make non-partial. This is how (given my current understanding) I’d do that. Is this something we’d want in std?

There is a (small) performance penalty due to the generalization of `Array.mergeSortedMergingDuplicates`.

nomeata:
The functions changed here have been moved to core in https://github.com/leanprover/lean4/pull/3303; closing.

PR #330, opened 2023-10-28
Add Option.bind_none lemma

blaxill:

PR #334, opened 2023-10-30
Option seq/bind lemmas

joehendrix:
This adds a few simplification theorems to simplify Option seq and bind expressions.

It builds on the commits from #330.

### DIFF
--- a/Std/Data/Option/Lemmas.lean
+++ b/Std/Data/Option/Lemmas.lean
@@ -88,10 +88,21 @@ theorem ball_ne_none {p : Option α → Prop} : (∀ x (_ : x ≠ none), p x) �
       simp [some_get] at this ⊢
       exact this⟩
 
-@[simp] theorem bind_some (x : Option α) : x.bind some = x := by cases x <;> rfl
+/-! ### seq -/
 
-@[simp] theorem bind_none (x : Option α) : x.bind (fun _ => none (α := β)) = none := by 
+@[simp] theorem seq_some_some (h : α → β) a : Seq.seq (some h) (fun _ => some a) = some (h a) := rfl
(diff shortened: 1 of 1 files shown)


digama0:
My preference is to merge #330 and rebase this on top, as the author of #330 is a new contributor and we shouldn't be poaching their work without (github) attribution.

joehendrix:
I don't agree with labeling this as "poaching", but I am merging in their request now.

Std/Data/Option/Lemmas.lean
```diff
@@ -88,7 +88,22 @@ theorem ball_ne_none {p : Option α → Prop} : (∀ x (_ : x ≠ none), p x) 
...
+
+/-! ### bind -/
+
+@[simp] theorem none_bind (f : α → Option β) : Option.bind none f = none := rfl
+@[simp] theorem some_bind (x : α) (f : α → Option β): Option.bind (some x) f = f x := rfl
+
+@[simp] theorem bind_none (x : Option α)
+    : Option.bind x (fun _ => (none : Option β)) = none := by
+  cases x <;> rfl
+@[simp] theorem bind_some (x : Option α) : Option.bind x some = x := by cases x <;> rfl
```
digama0, line 106:
```suggestion
@[simp] theorem none_bind (f : α → Option β) : none.bind f = none := rfl
@[simp] theorem some_bind (x : α) (f : α → Option β): (some x).bind f = f x := rfl

@[simp] theorem bind_none (x : Option α)
    : x.bind (fun _ => (none : Option β)) = none := by
  cases x <;> rfl
@[simp] theorem bind_some (x : Option α) : x.bind some = x := by cases x <;> rfl
```

PR #339, opened 2023-10-31
chore: add Option.getD_eq_iff

semorrison:
The characteristic property of `getD`.


digama0:
FIXME: style lint for trailing whitespace

PR #340, opened 2023-10-31
feat: lemmas characterising List.minimum?

semorrison:
The full characterisation is only given for `List Nat`, as Std doesn't have the necessary order typeclasses to prove it generally. I'm using `List.minimum?` over `Nat`, so this suffices for my purposes for now. There's a comment about generalizing.

### DIFF
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -2167,3 +2167,48 @@ theorem reverse_range' : ∀ s n : Nat, reverse (range' s n) = map (s + n - 1 - 
 
 @[simp] theorem enum_map_fst (l : List α) : map Prod.fst (enum l) = range l.length := by
   simp only [enum, enumFrom_map_fst, range_eq_range']
+
+/-! ### minimum? -/
+
+@[simp] theorem minimum?_nil [Min α] : ([] : List α).minimum? = none := rfl
+
+-- We don't put `@[simp]` on `minimum?_cons`,
+-- because the definition in terms of `foldl` is not useful for proofs.
+theorem minimum?_cons [Min α] {xs : List α} : (x :: xs).minimum? = foldl min x xs := rfl
+
+@[simp] theorem minimum?_eq_none_iff {xs : List α} [Min α] : xs.minimum? = none ↔ xs = [] := by
+  cases xs <;> simp [minimum?]
+
+-- This could be generalized away from `Nat`.
+-- The standard library doesn't yet have the order typeclasses
+-- that would be necessary for such a generalization.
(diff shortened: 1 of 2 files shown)


Std/Data/List/Lemmas.lean
```diff
@@ -2167,3 +2167,48 @@ theorem reverse_range' : ∀ s n : Nat, reverse (range' s n) = map (s + n - 1 -
...
+theorem minimum?_cons [Min α] {xs : List α} : (x :: xs).minimum? = foldl min x xs := rfl
+
+@[simp] theorem minimum?_eq_none_iff {xs : List α} [Min α] : xs.minimum? = none ↔ xs = [] := by
+  cases xs <;> simp [minimum?]
+
+-- This could be generalized away from `Nat`.
+-- The standard library doesn't yet have the order typeclasses
+-- that would be necessary for such a generalization.
+theorem minimum?_eq_some_iff {xs : List Nat} :
+    xs.minimum? = some a ↔ (a ∈ xs ∧ ∀ b ∈ xs, a ≤ b) := by
```
digama0, line 2186: Could this be proven in general given a plain hypothesis saying the necessary assumptions about `min` and `le` interaction?

semorrison: Yes, but it's quite a few hypotheses!

`Nat.min_le_left`, `Nat.min_le_right`, `Nat.le_trans`, `Nat.min_def`, `Nat.le_refl`, `Nat.lt_of_le_of_ne`, at least.

digama0: My guess is that you don't need all of those to prove the theorem, but rather some combination of them. I suspect `le_total` and `le_antisymm` suffice.

semorrison: I needed:
```
[Min α] [LE α] [DecidableRel ((· : α) ≤ ·)]
    [Trans ((· : α) ≤ ·) (· ≤ ·) (· ≤ ·)] [Antisymm ((· : α) ≤ ·)]
    (le_refl : ∀ a : α, a ≤ a) (le_total : ∀ a b : α, a ≤ b ∨ b ≤ a)
    (le_min_iff : ∀ a b c : α, a ≤ min b c ↔ a ≤ b ∧ a ≤ c)
```

semorrison: Thanks!

fgdorais: Note that `le_refl` follows from `le_total`.

semorrison: (No longer relevant given Mario's refactor.)

PR #345, opened 2023-11-03
feat: upstream 'change at' tactic

semorrison:


Std/Tactic/Change.lean
```diff
@@ -0,0 +1,111 @@
...
+```
+changes the goal to `⊢ n + 1 + 1 = 2`. The tactic also applies to the local context.
+If `h : n + 2 = 2` and `h' : n + 3 = 4` are in the local context, then
+```lean
+change _ + 1 = _ at h h'
+```
+changes their types to be `h : n + 1 + 1 = 2` and `h' : n + 2 + 1 = 4`.
+
+Change is like `refine` in that every placeholder needs to be solved for by unification,
+but you can use named placeholders and `?_` where you want `change` to create new goals.
```
joehendrix, line 94: I like the examples.  Could you add an example illustrating the named placeholder and `?_` syntax.

Std/Tactic/Change.lean
```diff
@@ -0,0 +1,111 @@
...
+
+The tactic `show e` is interchangeable with `change e`, where the pattern `e` is applied to
+the main goal. -/
```
joehendrix, line 97: Would it be better to enhance `show` and migrate code using `change` to it?  I saw the comment how it was using `show` behind the scenes in a hacky way.

digama0: I like the idea of not having two synonymous tactics, but `show T at h` doesn't make a whole lot of sense linguistically.

joehendrix: Linguistically, one option would be `change to T` and `change h to T`.  It's a little bit more typing so the `to` could be optional if `h` is omitted.

digama0: The issue there is that there is *also* a clear pattern in lean tactics of using `at` for referring to hypotheses, so if we were to change the syntax I would rather we do something more syntactic like `change h : T` or similar.

joehendrix: I have no objection to the `change h : T` (or `show h : T`) syntax.  I also am ok with keeping the syntax `change T at h`.

My main concern is I'd like to not unintentional provide two ways of writing the same thing so that users aren't confused about which one to use.

digama0: Well, all of these tactics exist as logical completions of other things, so we end up with two tactics that mean the same thing (and say as much in the documentation). It might be good to bring this up on zulip.

semorrison: It seems like the options here are:
* leave as is, with overlap between the tactics
* remove `change` entirely (i.e. remove `change` from Lean, and `change at` from Mathlib/Std)

I would be happy with either, but I'd prefer to do it separately with the PR that moves things around.

Std/Tactic/Change.lean
semorrison: One of the tests already includes this, and I've added a comment that says so.

I think an example of using `?_` in the doc-string might actually be counterproductive: because it has to unify regardless of the value of the `?_`, the examples where this is useful are all a bit obscure; the one case I can find in Mathlib of this being used is itself a confusing porting hack!

joehendrix (approved):

PR #346, opened 2023-11-03
refactor(Std.Util.Cache): Introduce DeclCache2

nomeata:
in order to implement `#find` in https://github.com/leanprover-community/mathlib4/pull/6363
I needed some of the functionality that was in `DiscrTreeCache` (in
particular the separation of imported and local declaration), but with a
diffent data structure than `DiscrTree`, so I refactored this module a
bit.

Especially the introduction of `getImported` is useful: this goes along
with `mkFromCache`, and keeps the logic of what precisely to persist
local toether. Before there was a stray `.2` in `MathlibExtras.*` that,
if one would forget to change to `.1`, easily makes lean crash or yield
wrong results.


nomeata:
I can keep this in the loogle repo for now, no need to go into std at this point.

PR #347, opened 2023-11-04
feat: upstream @[simps] attribute

semorrison:


joehendrix (commented):
I don't have bandwidth for a full review this week, but I added a couple notes.

Std/Lean/KVMap.lean
```diff
@@ -0,0 +1,27 @@
...
+
+/-- erase an entry from the map -/
+def erase : KVMap → Name → KVMap
+  | ⟨m⟩, k => ⟨eraseCore m k⟩
+
+/-- update a Boolean entry based on its current value. -/
+def updateBool (m : KVMap) (k : Name) (f : Bool → Bool) : KVMap :=
+  m.insert k <| DataValue.ofBool <| f <| m.getBool k
+
+end Lean.KVMap
```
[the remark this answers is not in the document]
semorrison: https://github.com/leanprover/lean4/pull/2851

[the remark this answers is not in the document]
semorrison: They are not good reasons but:
1. it's extra effort
2. historically PRs to core have been discouraged :-)

Std/Lean/Linter.lean
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Lean.Linter
+
+/-- If `linterOption` is true, print a linter warning message at the position determined by `stx`.
+-/
+def logLintIf [Monad m] [MonadLog m] [AddMessageContext m] [MonadOptions m]
+    (linterOption : Lean.Option Bool) (stx : Syntax) (msg : MessageData) : m Unit := do
+  if linterOption.get (← getOptions) then logLint linterOption stx msg
+
+end Lean.Linter
```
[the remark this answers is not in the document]
semorrison: https://github.com/leanprover/lean4/pull/2852

For both of these, I'd prefer *not* to wait: it is trivial for me to remove functions again when bumping Lean versions, but there is lots of mental overhead to have PRs waiting on each other.

Std/Lean/KVMap.lean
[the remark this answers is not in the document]
semorrison: This is in core now, but can also go in Std in the meantime.

Std/Lean/Linter.lean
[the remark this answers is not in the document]
semorrison: This is in core now, but can also go in Std in the meantime.

Std/Lean/Expr.lean
```diff
@@ -40,6 +40,14 @@ def lambdaArity : Expr → Nat
   | lam _ _ b _ => 1 + lambdaArity b
   | _ => 0
 
+/-- If the expression is a constant, return that name. Otherwise return `Name.anonymous`. -/
+def constName (e : Expr) : Name :=
+  e.constName?.getD Name.anonymous
+
+/-- Return the function (name) and arguments of an application. -/
+def getAppFnArgs (e : Expr) : Name × Array Expr :=
+  withApp e λ e a => (e.constName, a)
```
fgdorais, line 49:
```suggestion
  withApp e fun e a => (e.constName, a)
```

Std/Lean/KVMap.lean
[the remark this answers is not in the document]
semorrison: We've bumped by now to a Lean version that includes these, so I've removed them again.

Std/Lean/Linter.lean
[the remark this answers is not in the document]
semorrison: As above, so removed.

Std/Lean/Expr.lean
```diff
@@ -40,6 +40,14 @@ def lambdaArity : Expr → Nat
...
   | _ => 0
 
+/-- If the expression is a constant, return that name. Otherwise return `Name.anonymous`. -/
+def constName (e : Expr) : Name :=
+  e.constName?.getD Name.anonymous
+
+/-- Return the function (name) and arguments of an application. -/
+def getAppFnArgs (e : Expr) : Name × Array Expr :=
+  withApp e fun e a => (e.constName, a)
+
```
joehendrix, line 50: Would these and the Std/Lean/Name operations be better as utility functions inside the Simps tactic?  They seem to make assumptions about the names that may hold in the context they are used in that tactic, but could mask errors in other contexts.

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1157 @@
...
+open Meta hiding Config
+open Elab.Term hiding mkConst
+
+/-- `updateName nm s isPrefix` adds `s` to the last component of `nm`,
+  either as prefix or as suffix (specified by `isPrefix`), separated by `_`.
+  Used by `simps_add_projections`. -/
+def updateName (nm : Name) (s : String) (isPrefix : Bool) : Name :=
+  nm.updateLast fun s' => if isPrefix then s ++ "_" ++ s' else s' ++ "_" ++ s
+
+-- move
```
joehendrix, line 65: What should be moved?

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1157 @@
...
+    toPrint ++
+      if nondefaults.isEmpty then [] else
+      [("No lemmas are generated for the projections: " : MessageData) ++ print2 ++ "."]
+  let toPrint := MessageData.joinSep toPrint ("\n" : MessageData)
+  m!"{pref} {str}:\n{toPrint}"
+
+/-- Find the indices of the projections that need to be applied to elaborate `$e.$projName`.
+Example: If `e : α ≃+ β` and ``projName = `invFun`` then this returns `[0, 1]`, because the first
+projection of `MulEquiv` is `toEquiv` and the second projection of `Equiv` is `invFun`. -/
+def findProjectionIndices (strName projName : Name) : MetaM (List Nat) := do
```
joehendrix, line 466: Comment example is from Mathlib and uses `e` insteas of `strName`.

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1157 @@
...
+  let .some baseStr := findField? env strName projName |
+    throwError "{strName} has no field {projName} in parent structure"
+  let .some fullProjName := getProjFnForField? env baseStr projName |
+    throwError "no such field {projName}"
+  let .some pathToField := getPathToBaseStructure? env baseStr strName |
+    throwError "no such field {projName}"
+  let allProjs := pathToField ++ [fullProjName]
+  return allProjs.map (env.getProjectionFnInfo? · |>.get!.i)
+
+/-- Auxiliary function of `getCompositeOfProjections`. -/
```
joehendrix, line 477: Could more detail be added either here to body of code?

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1157 @@
...
+  isSimp := false
+
+/-- `instantiateLambdasOrApps es e` instantiates lambdas in `e` by expressions from `es`.
+If the length of `es` is larger than the number of lambdas in `e`,
+then the term is applied to the remaining terms.
+Also reduces head let-expressions in `e`, including those after instantiating all lambdas.
+
+This is very similar to `expr.substs`, but this also reduces head let-expressions. -/
+partial def _root_.Lean.Expr.instantiateLambdasOrApps (es : Array Expr) (e : Expr) : Expr :=
+  e.betaRev es.reverse true -- check if this is what I want
```
joehendrix, line 857: Comment seems fishy and this should either be moved to `Std/Lean/Expr.lean` or renamed.

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1157 @@
...
+  will give the output
+  ```
+    [(`(id), `apply, (Equiv.toFun), [], true, false),
+     (`(id), `symm_apply, (fun e => e.symm.toFun), [], true, false),
+     ...,
+     ...]
+  ```
+-/
+def getProjectionExprs (stx : Syntax) (tgt : Expr) (rhs : Expr) (cfg : Config) :
+    MetaM <| Array <| Expr × ProjectionData := do
```
joehendrix, line 885: Use parens instead of `<|` for consistency

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1157 @@
...
+        ""}Note 1: `@[simps!]` also calls the `simp` tactic, and this can be expensive in certain {
+        ""}cases.\n{
+        ""}Note 2: `@[simps!]` is equivalent to `@[simps (config := \{rhsMd := .default, {
+        ""}simpRhs := true})]`. You can also try `@[simps (config := \{rhsMd := .default})]` {
+        ""}to still unfold the definitions, but avoid calling `simp` on the resulting statement.\n{
+        ""}Note 3: You need `simps!` if not all fields are given explicitly in this definition, {
+        ""}even if the definition is a constructor application. For example, if you give a {
+        ""}`MulEquiv` by giving the corresponding `Equiv` and the proof that it respects {
+        ""}multiplication, then you need to mark it as `@[simps!]`, since the attribute needs to {
+        ""}unfold the corresponding `Equiv` to get to the `toFun` field."
```
joehendrix, line 1064: I'd love to see this function and the one above broken up into smaller entities with comments about what they do.

Std/Lean/Expr.lean
semorrison: This one I would prefer to keep out "in public", as it is very thoroughly used all over tactics in Mathlib. (And e.g. I will use it in `omega`.) It's basically the standard interface for decomposing an expression, although probably should be deprecated in favour of Kyle's recent https://github.com/leanprover/lean4/pull/2900.

Std/Tactic/Simps/Basic.lean
semorrison: I've clarified the comment to explain the use of `e`. It's still in terms of Mathlib. (This is a bigger overhaul: there are lots of Mathlib specific examples in this file --- `Equiv` isn't even defined in Std.)

Std/Tactic/Simps/Basic.lean
semorrison: @fpvandoorn, could I perhaps interest you in this? :-)

Std/Lean/Expr.lean
joehendrix: Ok, that makes sense.  This does seem like a common need to get the head function and arguments.  Use of anonymous is appropriate given that need.

Std/Lean/Meta/Simp.lean
```diff
@@ -72,6 +120,7 @@ def mkSimpContext' (simpTheorems : SimpTheorems) (stx : Syntax) (eraseLocal : Bo
         simpTheorems ← simpTheorems.addTheorem (.fvar h) (← h.getDecl).toExpr
     return { ctx := { r.ctx with simpTheorems }, dischargeWrapper }
 
+
```
joehendrix, line 123:
```suggestion
```

Std/Tactic/Simps/Basic.lean
```diff
@@ -0,0 +1,1113 @@
...
+/-- `updateName nm s isPrefix` adds `s` to the last component of `nm`,
+  either as prefix or as suffix (specified by `isPrefix`), separated by `_`.
+  Used by `simps_add_projections`. -/
+private def updateName (nm : Name) (s : String) (isPrefix : Bool) : Name :=
+  nm.updateLast fun s' => if isPrefix then s ++ "_" ++ s' else s' ++ "_" ++ s
+
+/-- Tests whether `declName` has the `@[simp]` attribute in `env`. -/
+def hasSimpAttribute (env : Environment) (declName : Name) : Bool :=
+  simpExtension.getState env |>.lemmaNames.contains <| .decl declName
+
```
joehendrix, line 68:
```suggestion
```

This looks like it is only used in testing.

semorrison: I've moved it next to the (commented out) test that uses it.

Std/Tactic/Simps/Basic.lean
fpvandoorn: Yeah, this code is not pretty, since I wrote it in Lean 3, and then ported it to Lean 4 keeping the Lean 3 structure. I think the proper way to make this code better is to make use of monad extensions with a state and context, instead of these functions that take ~10 arguments. That makes it also much easier to split off small sub-functions.
I don't have time for that refactor at the moment.

It is a bit tricky to break up this function in the current state, since there are recursive calls in multiple places.

Std/Tactic/Simps/Basic.lean
fpvandoorn: I never tested whether this function satisfies the docstring exactly, and it doesn't do the second half of ``Also reduces head let-expressions in `e`, including those after instantiating all lambdas.``

In https://github.com/leanprover-community/mathlib4/compare/master...simps_upstream_tests I tried to see if doing that breaks anything in Mathlib. I expect not, given the low number of `let` expressions in definitions. 

We can also keep the current functionality. It would be nice if `Lean.Expr.beta` supports the optional flags of `Lean.Expr.betaRev`, then this function doesn't need to exist.

Std/Tactic/Simps/Basic.lean
fpvandoorn: Oh, I don't have push access to this repo. Can I get access, or shall I make a PR into this PR?

Std/Tactic/Simps/Basic.lean
semorrison: Thanks.

While it would be really nice to get `@[simps]` into Std, it is not essential: users can always write the boilerplate by hand!

I propose that we shop this project around a bit, and see if anyone on zulip might be interested in taking on a refactor?

Std/Tactic/Simps/Basic.lean
semorrison: @fpvandoorn, for now, would you mind a PR to the PR? Really I should have made this PR from a fork so I could have given you access myself.

PR #348, opened 2023-11-04
feat: have simpa using TryThis

kmill:


girving:
Here is an MWE, though it's probably unnecessary:

```
import Mathlib.Data.Real.Basic

theorem simple {p : ℝ → Prop} {x : ℝ} (px : p x) : p (- -x) := by
  simpa?
```

PR #351, opened 2023-11-05
feat: upstream norm_cast

semorrison:


scripts/nolints.json
digama0, line 1: `add_decl_doc`

Std/Tactic/NormCast.lean
```diff
@@ -0,0 +1,333 @@
...
+
+/-!
+# The `norm_cast` family of tactics.
+-/
+
+open Lean Meta Simp
+open Std.Tactic.NormCast
+open Std.Tactic.Coe
+
+namespace Tactic.NormCast
```
digama0, line 20:
```suggestion
namespace Std.Tactic.NormCast
```

Std/Tactic/NormCast.lean
```diff
@@ -0,0 +1,333 @@
...
+    Simp.mkCongrFun (← Simp.mkCongr {expr := op} x_x2) y
+  catch _ =>
+    return {expr}
+
+/--
+Discharging function used during simplification in the "squash" step.
+
+TODO: normCast takes a list of expressions to use as lemmas for the discharger
+TODO: a tactic to print the results the discharger fails to prove
+-/
```
digama0, line 127:
```suggestion
-/
-- TODO: normCast takes a list of expressions to use as lemmas for the discharger
-- TODO: a tactic to print the results the discharger fails to prove
```

Std/Tactic/NormCast.lean
```diff
@@ -0,0 +1,333 @@
...
+macro_rules
+  | `(tactic| rw_mod_cast $[$config]? [$rules,*] $[$loc]?) => do
+    let tacs ← rules.getElems.mapM fun rule =>
+      `(tactic| (norm_cast at *; rw $[$config]? [$rule] $[$loc]?))
+    `(tactic| ($[$tacs]*))
+
+/--
+Normalize the goal and the given expression, then close the goal with exact.
+-/
+macro "exact_mod_cast " e:term : tactic => `(tactic| exact mod_cast ($e : _))
```
digama0, line 290: Is the empty type ascription needed here? Besides the part where this annotation does not work in lean 4 (which is why `(e :)` was added), if we can get away with just `exact mod_cast $e` then perhaps we don't need `exact_mod_cast` at all as it's just syntax for something more compositional with the same number of characters.

Std/Tactic/NormCast.lean
```diff
@@ -0,0 +1,333 @@
...
+  push_cast,
+  push_cast at h1,
+  push_cast [int.add_zero] at h2,
+end
+```
+
+The implementation and behavior of the `norm_cast` family is described in detail at
+<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
+-/
+syntax (name := pushCast) "push_cast " (config)? (discharger)? (&" only")?
```
digama0, line 325:
```suggestion
syntax (name := pushCast) "push_cast" (config)? (discharger)? (&" only")?
```

test/norm_cast.lean
```diff
@@ -0,0 +1,81 @@
+import Std.Tactic.NormCast
+import Std
+
+/-!
+# Tests for norm_cast
+
+Since the behaviour of `norm_cast` depends on the `@[norm_cast]` attribute being applied correctly,
+this test file imports all of `Std` as an integration test.
```
digama0, line 8: I'd prefer you didn't. Just import the specific files containing `@[norm_cast]` attributes executed in the test.

semorrison: done.

Std/Tactic/NormCast.lean
semorrison: Unfortunately a small fraction of the `exact_mod_cast`s in Mathlib can not be replaced by `exact mod_cast`. See #8222.

PR #354, opened 2023-11-08
feat: Upstream the coercion arrows from mathlib, and use them in the coe delaborator

eric-wieser:
This means that it becomes easier to distinguish function coercion from "normal" coercion in a goal state.

This remains opt-in downstream; the `@[coe]` attribute does not provide syntax for the `coeFun` and `coeSort` versions.

This also does not upstream the unapplied versions.

Some more discussion is [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Bringing.20back.20.E2.87.91/near/401155616).

### DIFF
--- a/Std/Tactic/CoeExt.lean
+++ b/Std/Tactic/CoeExt.lean
@@ -26,7 +26,7 @@ namespace Std.Tactic.Coe
 /-- `⇑ t` coerces `t` to a function. -/
 -- We increase the right precedence so this goes above most binary operators.
 -- Otherwise `⇑f = g` will parse as `⇑(f = g)`.
-elab "⇑" m:term:80 : term => do
+elab:1024 "⇑" m:term:1024 : term => do
   let x ← elabTerm m none
   if let some ty ← coerceToFunction? x then
     return ty
@@ -34,7 +34,7 @@ elab "⇑" m:term:80 : term => do
     throwError "cannot coerce to function{indentExpr x}"
 
 /-- `↥ t` coerces `t` to a type. -/
-elab "↥" t:term:80 : term => do
+elab:1024 "↥" t:term:1024 : term => do
   let x ← elabTerm t none
   if let some ty ← coerceToSort? x then
     return ty
--- a/Std/Tactic/CoeExt.lean
+++ b/Std/Tactic/CoeExt.lean
@@ -18,15 +18,16 @@ def A.toB (a : A) : B := sorry
 instance : Coe A B where coe := A.toB
 ```
 is used, then `A.toB a` will be pretty-printed as `↑a`.
+
+This file also provides `⇑f` and `↥t` notation, which are syntax for `Lean.meta.coerceToFunction?`
+and `Lean.Meta.coerceToSort?` respectively.
 -/
 
 namespace Std.Tactic.Coe
 
-
 /-- `⇑ t` coerces `t` to a function. -/
--- We increase the right precedence so this goes above most binary operators.
--- Otherwise `⇑f = g` will parse as `⇑(f = g)`.
-elab:1024 "⇑" m:term:1024 : term => do
+-- the precendence matches that of `coeNotation`
+elab:1024 (name := coeFunNotation) "⇑" m:term:1024 : term => do
   let x ← elabTerm m none
   if let some ty ← coerceToFunction? x then
     return ty
@@ -34,13 +35,12 @@ elab:1024 "⇑" m:term:1024 : term => do
     throwError "cannot coerce to function{indentExpr x}"
 
 /-- `↥ t` coerces `t` to a type. -/
-elab:1024 "↥" t:term:1024 : term => do
+elab:1024 (name := coeSortNotation) "↥" t:term:1024 : term => do
   let x ← elabTerm t none
   if let some ty ← coerceToSort? x then
     return ty
   else
     throwError "cannot coerce to sort{indentExpr x}"
-
 
 /-- The different types of coercions that are supported by the `coe` attribute. -/
 inductive CoeFnType
--- a/Std/Tactic/CoeExt.lean
+++ b/Std/Tactic/CoeExt.lean
@@ -5,9 +5,29 @@ Authors: Gabriel Ebner, Mario Carneiro
 -/
 import Lean.PrettyPrinter.Delaborator.Builtins
 import Std.Lean.Delaborator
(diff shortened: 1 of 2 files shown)


Std/Tactic/CoeExt.lean
```diff
@@ -5,10 +5,30 @@ Authors: Gabriel Ebner, Mario Carneiro
...
-open Lean Meta Std
+open Lean Elab.Term Meta Std
 
 namespace Std.Tactic.Coe
 
+
+/-- `⇑ t` coerces `t` to a function. -/
+-- We increase the right precedence so this goes above most binary operators.
+-- Otherwise `⇑f = g` will parse as `⇑(f = g)`.
+elab "⇑" m:term:80 : term => do
```
digama0, line 16: This precedence doesn't sound high enough. I would expect that `⇑f x` means `(⇑f) x` which means that the precedence needs to be higher than application.

joehendrix:
Is there documentation available that describes this file and the rational for these notations?  

Std is going to need much better documentation and discoverability.  Every time features are added (including this notation), we need to be considering how are we going to help people learn that and use it effectively.  As we need to make sure this is important enough that all Lean users should be prepared to learn it (since they may see it in their files).

I know we don't have great documentation tools yet, but that is being worked on.  Could we an overview of the coercion notations `Std` has either in the `CoeExt.lean` file directly or in documentation that is referenced by this file?

Std/Tactic/CoeExt.lean
eric-wieser: This precedence is copied from mathlib, evidently it was never used!

test/coe.lean
```diff
@@ -0,0 +1,44 @@
...
+#guard_msgs in #check n.val
+/-- info: ↑n : Nat -/
+#guard_msgs in #check (↑n : Nat)
+
+/-- info: ⇑f : Nat → Nat-/
+#guard_msgs in #check f.fn
+/-- info: ⇑f : Nat → Nat-/
+#guard_msgs in #check ⇑f
+/-- info: ⇑f 1 : Nat-/
+#guard_msgs in #check ⇑f 1
```
eric-wieser, line 39: I made a thread about this at https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Bringing.20back.20.E2.87.91/near/401155616

test/coe.lean
```diff
@@ -0,0 +1,44 @@
...
+  typ : Type
+
+attribute [coe] WrappedNat.val
+instance : Coe WrappedNat Nat where coe := WrappedNat.val
+
+#eval Std.Tactic.Coe.registerCoercion ``WrappedFun.fn (some ⟨1, 0, .coeFun⟩)
+instance : CoeFun WrappedFun (fun _ => Nat → Nat) where coe := WrappedFun.fn
+
+#eval Std.Tactic.Coe.registerCoercion ``WrappedType.typ (some ⟨1, 0, .coeSort⟩)
+instance : CoeSort WrappedType Type where coe := WrappedType.typ
```
eric-wieser, line 24: These could potentially be spelt `@[coe coeFun]` and `@[coe coeSort]`

test/coe.lean
```diff
@@ -0,0 +1,45 @@
...
+/-- info: ↑n : Nat -/
+#guard_msgs in #check (↑n : Nat)
+
+/-- info: ⇑f : Nat → Nat -/
+#guard_msgs in #check f.fn
+/-- info: ⇑f : Nat → Nat -/
+#guard_msgs in #check ⇑f
+-- applied functions do not need the `⇑`
+/-- info: f 1 : Nat -/
+#guard_msgs in #check ⇑f 1
```
fpvandoorn, line 40: Can you add tests for some more complicated cases, like 
```
fun f : Nat -> WrappedFun => f 2
fun f : Nat -> WrappedFun => f 2 3
``` 
and a test of a term that would in Lean 3 be written as `⇑(⇑f x) y`

eric-wieser: Done; the output is `(f x) y`, which isn't perfect but is still an improvement over what we currently have.

Std/Tactic/CoeExt.lean
eric-wieser: Now merged in mathlib, so this change won't cause any issues with an Std bump.

test/coe.lean
eric-wieser: ... but I'd strongly prefer to declare that out of scope, Mathlib already uses this approach:

https://github.com/leanprover-community/mathlib4/blob/15ce6b63385ff165118b075a86a21a9bfabb4664/Mathlib/Data/FunLike/Basic.lean#L152-L154

eric-wieser:
@joehendrix, I added some documentation; would you mind taking another look? This would be a nice usability improvement for mathlib.

eric-wieser:
awaiting-review

Std/Tactic/CoeExt.lean
```diff
@@ -5,10 +5,43 @@ Authors: Gabriel Ebner, Mario Carneiro
...
+When writing a coercion, if the pattern
+```
+@[coe]
+def A.toB (a : A) : B := sorry
+
+instance : Coe A B where coe := A.toB
+```
+is used, then `A.toB a` will be pretty-printed as `↑a`.
+
+This file also provides `⇑f` and `↥t` notation, which are syntax for `Lean.meta.coerceToFunction?`
```
fgdorais, line 22:
```suggestion
This file also provides `⇑f` and `↥t` notation, which are syntax for `Lean.Meta.coerceToFunction?`
```

Std/Tactic/CoeExt.lean
```diff
@@ -5,10 +5,43 @@ Authors: Gabriel Ebner, Mario Carneiro
...
+and `Lean.Meta.coerceToSort?` respectively.
+-/
 
 namespace Std.Tactic.Coe
 
+/-- `⇑ t` coerces `t` to a function. -/
+-- the precendence matches that of `coeNotation`
+elab:1024 (name := coeFunNotation) "⇑" m:term:1024 : term => do
```
fgdorais, line 30: Is there a reason to use 1024 explicitly instead of `max`?

eric-wieser: I just copied core, which also writes 1024 explicitly

joehendrix (approved):

PR #355, opened 2023-11-08
feat: Improve #print prefix filtering.

joehendrix:
This fixes some annoyances I had with `#print prefix`.

It filters out internal and unstable auto-generated names by default, but has an options to control that.  It also prints out private declarations if the scope is explicitly named.

### DIFF
--- a/Std/Tactic/PrintPrefix.lean
+++ b/Std/Tactic/PrintPrefix.lean
@@ -4,12 +4,12 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Shing Tak Lam, Daniel Selsam, Mario Carneiro
 -/
 import Std.Lean.Util.EnvSearch
-import Lean.Elab.Command
+import Lean.Elab.Tactic.Config
 
 namespace Lean.Elab.Command
 
 private def appendMatchingConstants (msg : String)
-    (ϕ : ConstantInfo → MetaM Bool) (opts : EnvironmentSearchOptions := {}) : MetaM String := do
+    (ϕ : ConstantInfo → MetaM Bool) (opts : EnvironmentSearchOptions) : MetaM String := do
   let cinfos ← getMatchingConstants ϕ opts
   let cinfos := cinfos.qsort fun p q => p.name.lt q.name
   let mut msg := msg
@@ -17,23 +17,27 @@ private def appendMatchingConstants (msg : String)
     msg := msg ++ s!"{cinfo.name} : {← Meta.ppExpr cinfo.type}\n"
   pure msg
 
+/-- Function elaborating `Config`. -/
+declare_config_elab elabEnvironmentSearchOptions EnvironmentSearchOptions
+
 /--
 Command for #print prefix
 -/
-syntax (name := printPrefix) "#print prefix " ident : command
+syntax (name := printPrefix) "#print prefix " (Lean.Parser.Tactic.config)? ident : command
 
 /--
 The command `#print prefix foo` will print all definitions that start with
 the namespace `foo`.
 -/
 @[command_elab printPrefix] def elabPrintPrefix : CommandElab
-| `(#print prefix%$tk $name:ident) => do
+| `(#print prefix%$tk $[$cfg:config]? $name:ident) => do
   let nameId := name.getId
   liftTermElabM do
-    let mut msg ← appendMatchingConstants "" fun cinfo => pure $ nameId.isPrefixOf cinfo.name
+    let opts ← elabEnvironmentSearchOptions (mkOptionalNode cfg)
+    let mut msg ← appendMatchingConstants "" (opts:=opts) fun cinfo => pure $ nameId.isPrefixOf cinfo.name
     if msg.isEmpty then
       if let [name] ← resolveGlobalConst name then
-        msg ← appendMatchingConstants msg fun cinfo => pure $ name.isPrefixOf cinfo.name
+        msg ← appendMatchingConstants msg (opts:=opts) fun cinfo => pure $ name.isPrefixOf cinfo.name
     if !msg.isEmpty then
       logInfoAt tk msg
 | _ => throwUnsupportedSyntax
--- a/Std/Tactic/PrintPrefix.lean
+++ b/Std/Tactic/PrintPrefix.lean
@@ -63,12 +63,13 @@ private def takeNameSuffix (cnt : Nat) (name : Name) (pre : Name := .anonymous) 
   match cnt, name with
   | .succ cnt, .str q s => takeNameSuffix cnt q (.str pre s)
   | .succ cnt, .num q n => takeNameSuffix cnt q (.num pre n)
-  | _, name => (name, reverseName prev)
+  | _, name => (name, reverseName pre)
 
 /--
 `matchName opts pre cinfo` returns true if the search options should include the constant.
 -/
-private def matchName (opts : PrintPrefixConfig) (pre : Name) (cinfo : ConstantInfo) : MetaM Bool := do
+private def matchName (opts : PrintPrefixConfig)
+                      (pre : Name) (cinfo : ConstantInfo) : MetaM Bool := do
   let name := cinfo.name
   let preCnt := pre.getNumParts
   let nameCnt := name.getNumParts
@@ -113,7 +114,6 @@ By default, it will included imported names and filter out auto-generated
 and internal names.  These options can be controlled by passing in config flags.
 If the prefix itself contains internal components,
 
-
 For example, the command below will print out all non-internal names in the
 `List namespace.
 
@@ -132,6 +132,9 @@ The command below will exclude imported names:
 ```lean
 #print prefix (config:={imported:=false}) List
 ```
+
+The complete set of flags can be seen in the documentation
+for `Lean.Elab.Command.PrintPrefixConfig`.
 -/
 @[command_elab printPrefix] def elabPrintPrefix : CommandElab
 | `(#print prefix%$tk $[$cfg:config]? $name:ident) => do
--- a/Std/Tactic/PrintPrefix.lean
+++ b/Std/Tactic/PrintPrefix.lean
@@ -14,7 +14,7 @@ Options to control `getMatchingConstants` options below.
 structure PrintPrefixConfig where
   /-- Include declarations in imported environment. -/
   imported : Bool := true
-  /-- Include declarations who types are propositions. -/
+  /-- Include declarations whose types are propositions. -/
   propositions : Bool := true
   /-- Exclude definitions who types are not propositions. -/
   propositionsOnly : Bool := false
--- a/Std/Tactic/PrintPrefix.lean
+++ b/Std/Tactic/PrintPrefix.lean
@@ -18,7 +18,7 @@ structure PrintPrefixConfig where
   propositions : Bool := true
   /-- Exclude definitions who types are not propositions. -/
   propositionsOnly : Bool := false
-  /-- Flag to control whether we should print the type of a declartion. -/
+  /-- Print the type of a declaration. -/
   showTypes : Bool := true
   /--
   Include internal declarations (names starting with `_`, `match_` or `proof_`)
--- a/Std/Tactic/PrintPrefix.lean
+++ b/Std/Tactic/PrintPrefix.lean
@@ -29,7 +29,31 @@ structure PrintPrefixConfig where
 declare_config_elab elabPrintPrefixConfig PrintPrefixConfig
 
 /--
-Command for #print prefix
+The command `#print prefix foo` will print all definitions that start with
+the namespace `foo`.
+
+For example, the command below will print out definitions in the `List` namespace:
+
+```lean
+#print prefix List
+```
+
+`#print prefix` can be controlled by flags in `PrintPrefixConfig`.  These provide
+options for filtering names and formatting.   For example,
+`#print prefix` by default excludes internal names, but this can be controlled
+via config:
+```lean
+#print prefix (config:={internals:=true}) List
+```
+
+By default, `#print prefix` prints the type after each name.  This can be controlled
+by setting `showTypes` to `false`:
+```lean
+#print prefix (config:={showTypes:=false}) List
+```
+
+The complete set of flags can be seen in the documentation
+for `Lean.Elab.Command.PrintPrefixConfig`.
 -/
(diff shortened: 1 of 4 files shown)


Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+/--
+`takeNameSuffix n name` returns a pair `(pre, suf)` where `suf` contains the last `n` components
+of the name and `pre` contains the rest.
+-/
+private def takeNameSuffix (cnt : Nat) (name:Name) (prev:Name := .anonymous) : Name × Name :=
+  match cnt, name with
+  | .succ cnt, .str q s => takeNameSuffix cnt q (.str prev s)
+  | .succ cnt, .num q n => takeNameSuffix cnt q (.num prev n)
+  | _, name => (name, reverseName prev)
```
fgdorais, line 66: Match docstring name:
```suggestion
private def takeNameSuffix (cnt : Nat) (name : Name) (pre : Name := .anonymous) : Name × Name :=
  match cnt, name with
  | .succ cnt, .str q s => takeNameSuffix cnt q (.str pre s)
  | .succ cnt, .num q n => takeNameSuffix cnt q (.num pre n)
  | _, name => (name, reverseName prev)
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+/--
+`matchName opts pre cinfo` returns true if the search options should include the constant.
+-/
+private def matchName (opts : PrintPrefixConfig) (pre:Name) (cinfo:ConstantInfo) : MetaM Bool := do
```
fgdorais, line 71:
```suggestion
private def matchName (opts : PrintPrefixConfig) (pre : Name) (cinfo : ConstantInfo) : MetaM Bool := do
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+private def lexNameLt : Name -> Name -> Bool
+| _, .anonymous => false
+| .anonymous, _ => true
+| .num p m, .num q n => m < n || m == n && lexNameLt p q
+| .num _ _, .str _ _ => true
+| .str _ _, .num _ _ => false
+| .str p m, .str q n => m < n || m == n && lexNameLt p q
+
+private def appendMatchingConstants (msg : String) (opts : PrintPrefixConfig) (pre:Name)
```
fgdorais, line 94:
```suggestion
private def appendMatchingConstants (msg : String) (opts : PrintPrefixConfig) (pre : Name)
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+      || s.startsWith "match_"
+      || s.startsWith "proof_"
+      || p.isInternal
+  | .num p _     => p.isInternal
+  | _            => false
+
+/--
+`reverseName name` reverses the components of a name.
+-/
+private def reverseName : Name → (pre:Name := .anonymous) → Name
```
fgdorais, line 53:
```suggestion
private def reverseName : Name → (pre : Name := .anonymous) → Name
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
-  for cinfo in cinfos do
-    msg := msg ++ s!"{cinfo.name} : {← Meta.ppExpr cinfo.type}\n"
-  pure msg
+/--
+Options to control `getMatchingConstants` options below.
+-/
+structure PrintPrefixConfig where
+  /-- Include declarations in imported environment. -/
+  imported : Bool := true
+  /-- Include declarations who types are propositions. -/
```
fgdorais, line 17:
```suggestion
  /-- Include declarations whose types are propositions. -/
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+  propositions : Bool := true
+  /-- Exclude definitions who types are not propositions. -/
```
fgdorais, line 19:
```suggestion
  /-- Exclude declarations whose types are not propositions. -/
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+  propositionsOnly : Bool := false
+  /-- Flag to control whether we should print the type of a declartion. -/
```
fgdorais, line 21:
```suggestion
  /-- Print the type of a declaration. -/
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+  showTypes : Bool := true
+  /--
+  Set to true to include internal declarations (names with "_" or ending with match_ or proof_)
```
fgdorais, line 24:
```suggestion
  Include internal declarations (names starting with `_`, `match_` or `proof_`)
```

test/print_prefix.lean
```diff
@@ -44,14 +41,39 @@ TestStruct.bar : TestStruct → Int
...
 TestStruct.mk.injEq : ∀ (foo bar foo_1 bar_1 : Int),
   ({ foo := foo, bar := bar } = { foo := foo_1, bar := bar_1 }) = (foo = foo_1 ∧ bar = bar_1)
 TestStruct.mk.sizeOf_spec : ∀ (foo bar : Int), sizeOf { foo := foo, bar := bar } = 1 + sizeOf foo + sizeOf bar
+TestStruct.noConfusion : {P : Sort u} → {v1 v2 : TestStruct} → v1 = v2 → TestStruct.noConfusionType P v1 v2
+TestStruct.noConfusionType : Sort u → TestStruct → TestStruct → Sort u
+TestStruct.rec : {motive : TestStruct → Sort u} → ((foo bar : Int) → motive { foo := foo, bar := bar }) → (t : TestStruct) → motive t
+TestStruct.recOn : {motive : TestStruct → Sort u} → (t : TestStruct) → ((foo bar : Int) → motive { foo := foo, bar := bar }) → motive t
+-/
+#guard_msgs in
+#print prefix (config:={internals:=true}) TestStruct
```
fgdorais, line 54: This test case doesn't seem to actually test the `internals` flag.

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+    msg := msg ++ (← ppInfo cinfo)
+  pure msg
 
 /--
 The command `#print prefix foo` will print all definitions that start with
 the namespace `foo`.
+
+By default, it will included imported names and filter out auto-generated
+and internal names.  These options can be controlled by passing in config flags.
+If the prefix itself contains internal components,
```
fgdorais, line 114: Incomplete sentence?

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
-private def appendMatchingConstants (msg : String)
-    (ϕ : ConstantInfo → MetaM Bool) (opts : EnvironmentSearchOptions := {}) : MetaM String := do
-  let cinfos ← getMatchingConstants ϕ opts
-  let cinfos := cinfos.qsort fun p q => p.name.lt q.name
-  let mut msg := msg
-  for cinfo in cinfos do
-    msg := msg ++ s!"{cinfo.name} : {← Meta.ppExpr cinfo.type}\n"
-  pure msg
+/--
+Options to control `getMatchingConstants` options below.
```
fgdorais, line 12:
```suggestion
Options to control `#print prefix` command and `getMatchingConstants`.
```

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,144 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+```lean
+#print prefix (config:={internals:=true}) List
+```
+
+The command below will exclude imported names:
+
+```lean
+#print prefix (config:={imported:=false}) List
+```
```
fgdorais, line 134: Missing descriptions of some flags.

fgdorais (changes requested):
Some test cases are missing. When testing a flag, do both the `true` and `false` cases for the same declaration so we can see the difference.

test/print_prefix.lean
joehendrix: It does, but it's a small delta.  `TestStruct._sizeOf_1 : TestStruct → Nat` is internal.

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,147 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+  internals : Bool := false
+
+/-- Function elaborating `Config`. -/
+declare_config_elab elabPrintPrefixConfig PrintPrefixConfig
 
 /--
 Command for #print prefix
 -/
-syntax (name := printPrefix) "#print prefix " ident : command
+syntax (name := printPrefix) "#print prefix " (Lean.Parser.Tactic.config)? ident : command
```
fgdorais, line 34: Do we really want to require exactly one space between `#print` and `prefix`? That seems excessively rigid.

Std/Tactic/PrintPrefix.lean
```diff
@@ -4,36 +4,147 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+/--
+Returns true if any part of name starts with underscore or uses a num.
+
+This can be used to hide internally generated names.
+-/
+def isInternalDetail : Name → Bool
```
fgdorais, line 41: I think this should be "higher up". If not in core then in `Std.Lean`?

Std/Tactic/PrintPrefix.lean
fgdorais: Fixed below.

test/print_prefix.lean
```diff
@@ -35,11 +46,34 @@ structure TestStruct where
...
   bar : Int
 
+/-
+  /-- Include declarations whose types are propositions. -/
+  propositions : Bool := true
+  /-- Exclude declarations whose types are not propositions. -/
+  propositionsOnly : Bool := false
+  /-- Print the type of a declaration. -/
+  showTypes : Bool := true
+-/
```
fgdorais, line 56: Stray comment?

fgdorais (commented):
Looks great to me, just a few questions and minor edits.

Std/Tactic/PrintPrefix.lean
joehendrix: Thanks, I've reworked the documentation a bit, and moved it so it comes up when hovering in VSCode.

Std/Tactic/PrintPrefix.lean
joehendrix: I've added it to `Lean.Name` in `Std/Data/Lean` for now   I made [an issue in core](https://github.com/leanprover/lean4/issues/2850), but it didn't come to a clear resolution.

I've put this upstream a bit in `Std` and if other tactics adopt `isInternalDetail` predicate then it would be easy to upstream.

Std/Tactic/PrintPrefix.lean
joehendrix: Fixed with commenting change.

Std/Tactic/PrintPrefix.lean
fgdorais: Sounds like a good temporary fix to me!

Std/Tactic/PrintPrefix.lean
joehendrix: Good catch.  I fixed this.  I'm also not a huge fan of the config syntax I used from `Tactic`, but some small changes I tried didn't make it better, so I am keeping as is.

joehendrix:
Thanks for all the comments @fgdorais.  I'll plan to merge this tonight unless I see more things to address.

Std/Tactic/PrintPrefix.lean
fgdorais: FWIW, I think the fix allows things like:
```lean
#print 

prefix 

Empty
```
I don't think this is a major problem though, no fix necessary.

fgdorais (approved):
Thanks for putting up with all my nitpicks!

test/print_prefix.lean
```diff
@@ -48,10 +74,72 @@ TestStruct.noConfusion : {P : Sort u} → {v1 v2 : TestStruct} → v1 = v2 → T
...
   ({ foo := foo, bar := bar } = { foo := foo_1, bar := bar_1 }) = (foo = foo_1 ∧ bar = bar_1)
 TestStruct.mk.sizeOf_spec : ∀ (foo bar : Int), sizeOf { foo := foo, bar := bar } = 1 + sizeOf foo + sizeOf bar
+TestStruct.noConfusion : {P : Sort u} → {v1 v2 : TestStruct} → v1 = v2 → TestStruct.noConfusionType P v1 v2
+TestStruct.noConfusionType : Sort u → TestStruct → TestStruct → Sort u
+TestStruct.rec : {motive : TestStruct → Sort u} → ((foo bar : Int) → motive { foo := foo, bar := bar }) → (t : TestStruct) → motive t
+TestStruct.recOn : {motive : TestStruct → Sort u} → (t : TestStruct) → ((foo bar : Int) → motive { foo := foo, bar := bar }) → motive t
 -/
 #guard_msgs in
-#print prefix TestStruct
+#print prefix (config:={internals:=true}) TestStruct
```
fgdorais, line 125: Absolute final nitpick! It would be nice if this example also tested `proof_` and `match_`.

joehendrix: I've added a more comprehensive test case for printing internals, but I am not sure the positive test of checking the output of `#print prefix (config:={internals:=true})` is a good idea.  The problem is that internal names are inherently subject to change.

I've done it anyways with a warning this may be volatile.  If maintaining the test case is a chore, then I think it should be removed.

PR #356, opened 2023-11-09
feat: infer Decidable instances in simp lemmas by unification / adaptation for leanprover/lean4#2816

semorrison:
I made some similar changes in leanprover/lean4#2816. There are then two breakages downstream in https://github.com/leanprover-community/mathlib4/pull/8285 which are then fixed by this PR.


digama0:
I can't review or merge this as is. Also, it seems to have some changes for earlier PRs which I had modified in those previous PRs and the modifications don't show up here, so I'm really losing track of what is merged where.

semorrison:
Very sorry, I forgot to set the base branch, so the diff was a mess. I've fixed it now.

semorrison:
Actually, I'm pretty sure these changes will work for Std and for Mathlib whether or not we make the change to the Lean lemmas first, so I will now reconfigure this as a PR for `main`.

semorrison:
Closing this in favour of #367

PR #357, opened 2023-11-09
feat: use notation as simp-normal form for BitVec operations

alexkeizer:
Relatively minor change factored out from leanprover-community/mathlib4#5920


Std/Data/BitVec/Lemmas.lean
```diff
@@ -0,0 +1,26 @@
...
+
+import Std.Data.BitVec.Basic
+
+namespace Std.BitVec
+variable {w v : Nat}
+
+/-! We add simp-lemmas that rewrite bitvector operations into the equivalent notation -/
+@[simp] theorem append_eq (x : BitVec w) (y : BitVec v)   : BitVec.append x y = x ++ y        := rfl
+@[simp] theorem shiftLeft_eq (x : BitVec w) (n: Nat)      : BitVec.shiftLeft x n = x <<< n    := rfl
+@[simp] theorem ushiftRight_eq (x : BitVec w) (n: Nat)    : BitVec.ushiftRight x n = x >>> n  := rfl
```
eric-wieser, line 16:
```suggestion
@[simp] theorem shiftLeft_eq (x : BitVec w) (n : Nat)     : BitVec.shiftLeft x n = x <<< n    := rfl
@[simp] theorem ushiftRight_eq (x : BitVec w) (n : Nat)   : BitVec.ushiftRight x n = x >>> n  := rfl
```

eric-wieser (commented):
Potentially these could even go in `Basic`, since these don't contain any proofs and are really just API. I'll leave that for Joe or Mario to decide.

digama0:
The policy I use is that only equation lemmas and other rfl lemmas go in Basic, and I think this would qualify.

eric-wieser:
@digama0, note you need to click a button for CI to run at all

alexkeizer:
I've just moved the lemmas to the `Basic` file

digama0:
> @digama0, note you need to click a button for CI to run at all

Actually I have to re-click the button after every commit.

eric-wieser:
Once this PR is merged, you won't have to do that for any future PRs from Alex, at least...

joehendrix (approved):

alexkeizer:
Thanks for the quick review turnaround!

PR #359, opened 2023-11-10
feat: define `cons` and `concat` on bitvectors.

alexkeizer:
Define specializations of `BitVec.append` to add a single bit to a bitvector, either as the new LSB or the new MSB.

I've opted to name these functions `cons` and `concat`, to be consistent with the naming of (morally) similar functions on `List`, and following the big-ending convention set in `append`.

There has previously been some discussion about not assuming a specific order. Some alternative naming (in order of my preference):
* `consMsb` / `concatLsb` : keep the `cons`/`concat` contrast, and the big-endian assumption it carries, but make explicit whether the new bit is the new MSB or LSB.
* `consMsb` / `consLsb`: to be truly neutral w.r.t. to the big vs little endian question.

The intention is to eventually define recursion principles in terms of these functions, so it's desirable to have them even though they are somewhat simple wrappers over the existing `append`

### DIFF
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -475,6 +475,10 @@ def concat {n} (msbs : BitVec n) (lsb : Bool) : BitVec (n+1) := msbs ++ (ofBool 
 def cons {n} (msb : Bool) (lsbs : BitVec n) : BitVec (n+1) :=
   ((ofBool msb) ++ lsbs).cast (Nat.add_comm ..)
 
+/-- All empty bitvectors are equal -/
+instance : Subsingleton (BitVec 0) where
+  allEq := by intro ⟨0, _⟩ ⟨0, _⟩; rfl
+
 /-- Every bitvector of length 0 is equal to `nil`, i.e., there is only one empty bitvector -/
 theorem eq_nil : ∀ (x : BitVec 0), x = nil
   | ofFin ⟨0, _⟩ => rfl
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -479,12 +479,10 @@ def cons {n} (msb : Bool) (lsbs : BitVec n) : BitVec (n+1) :=
 theorem eq_nil : ∀ (x : BitVec 0), x = nil
   | ofFin ⟨0, _⟩ => rfl
 
-@[simp]
 theorem append_ofBool (msbs : BitVec w) (lsb : Bool) :
     msbs ++ ofBool lsb = concat msbs lsb :=
   rfl
 
-@[simp]
 theorem ofBool_append (msb : Bool) (lsbs : BitVec w) :
     ofBool msb ++ lsbs = (cons msb lsbs).cast (Nat.add_comm ..) :=
   rfl
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -433,6 +433,7 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
 @[simp] theorem add_eq (x y : BitVec w)                   : BitVec.add x y = x + y            := rfl
 @[simp] theorem sub_eq (x y : BitVec w)                   : BitVec.sub x y = x - y            := rfl
 @[simp] theorem mul_eq (x y : BitVec w)                   : BitVec.mul x y = x * y            := rfl
+@[simp] theorem zero_eq                                   : BitVec.zero n = 0#n               := rfl
 
 @[simp]
 theorem cast_ofNat {n m : Nat} (h : n = m) (x : Nat) :
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -436,8 +436,8 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
 
 /-- Turn a `Bool` into a bitvector of length `1` -/
 def ofBool : Bool → BitVec 1
-  | false => 1
-  | true  => 0
+  | false => 0
+  | true  => 1
 
 /-- The empty bitvector -/
 def nil : BitVec 0 :=
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -434,6 +434,21 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
 @[simp] theorem sub_eq (x y : BitVec w)                   : BitVec.sub x y = x - y            := rfl
 @[simp] theorem mul_eq (x y : BitVec w)                   : BitVec.mul x y = x * y            := rfl
 
+@[simp]
+theorem cast_ofNat {n m : Nat} (h : n = m) (x : Nat) :
+    cast h (BitVec.ofNat n x) = BitVec.ofNat m x := by
+  subst h; rfl
+
+@[simp]
+theorem cast_cast {n m k : Nat} (h₁ : n = m) (h₂ : m = k) (x : BitVec n) :
+    cast h₂ (cast h₁ x) = cast (h₁ ▸ h₂) x :=
+  rfl
+
+@[simp]
+theorem cast_eq {n : Nat} (h : n = n) (x : BitVec n) :
+    cast h x = x :=
+  rfl
+
 /-- Turn a `Bool` into a bitvector of length `1` -/
 def ofBool : Bool → BitVec 1
   | false => 0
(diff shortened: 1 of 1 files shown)


Std/Data/BitVec/Basic.lean
```diff
@@ -433,3 +433,21 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
...
 @[simp] theorem mul_eq (x y : BitVec w)                   : BitVec.mul x y = x * y            := rfl
+
+/-- Turn a `Bool` into a bitvector of length `1` -/
+def ofBool : Bool → BitVec 1
+  | false => 0
+  | true  => 1
+
+/-- The empty bitvector -/
+def nil : BitVec 0 :=
+  BitVec.zero 0
```
eric-wieser, line 444:
```suggestion
abbrev nil : BitVec 0 :=0
```

eric-wieser: It looks like we're missing `zero_eq` in the lemmas above.

Std/Data/BitVec/Basic.lean
```diff
@@ -433,3 +433,21 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
...
+
+/-- Append a single bit to the end of a bitvector, using big endian order (see `append`).
+    That is, the new bit is the least significant bit. -/
+def concat {n} (msbs : BitVec n) (lsb : Bool) : BitVec (n+1) := msbs ++ (ofBool lsb)
+
+/-- Prepend a single bit to the front of a bitvector, using big endian order (see `append`).
+    That is, the new bit is the most significant bit. -/
+def cons {n} (msb : Bool) (lsbs : BitVec n) : BitVec (n+1) :=
+  ((ofBool msb) ++ lsbs).cast (Nat.add_comm ..)
```
eric-wieser, line 453: It's not clear to me that doing the cast is a good idea here

Std/Data/BitVec/Basic.lean
```diff
@@ -433,3 +433,21 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
...
+  | false => 0
+  | true  => 1
+
+/-- The empty bitvector -/
+def nil : BitVec 0 :=
+  BitVec.zero 0
+
+/-- Append a single bit to the end of a bitvector, using big endian order (see `append`).
+    That is, the new bit is the least significant bit. -/
+def concat {n} (msbs : BitVec n) (lsb : Bool) : BitVec (n+1) := msbs ++ (ofBool lsb)
```
eric-wieser, line 448: I think a `lemma append_ofBool : msbs ++ ofBool lsb = concat msbs lsb` would be reasonable

Std/Data/BitVec/Basic.lean
alexkeizer: What do you mean by `zero_eq`, what would be its lhs and rhs?

eric-wieser: ` theorem zero_eq : BitVec.zero n = 0 := rfl`

Std/Data/BitVec/Basic.lean
alexkeizer: Can you elaborate on why you think the cast might not be a good idea?
This way, we can easily reuse the existing implementation of append, and have a clear connection with that function.
Note that this is `BitVec.cast`, not the generic casting operation (which you are right to be wary of), and, in particular, we have 
```
@[simp]
theorem cast_ofNat {n m : Nat} (h : n = m) (x : Nat) :
    cast h (BitVec.ofNat n x) = BitVec.ofNat m x := by
  subst h; rfl
```

So its not too hard to show equivalence to an alternative definition that does not use cast
```lean
theorem cons_alt (msb : Bool) (lsbs : BitVec w) :
    cons msb lsbs = BitVec.ofNat (w + 1) (BitVec.toNat (ofBool msb) <<< w ||| lsbs.toNat) := by
  simp [cons, (· ++ ·), Append.append, BitVec.append]
```

Of course, that also means it's not too hard to pick this alternative definition as the main definition and proof a theorem that relates `cons` to `append` and `cast`.
I somewhat prefer the current definition, but I'm happy to switch if you insist.

Std/Data/BitVec/Basic.lean
alexkeizer: Added

Std/Data/BitVec/Basic.lean
alexkeizer: Added it (with `0#n` as the rhs, since we also have a `0 = 0#n` simp lemma already)

Std/Data/BitVec/Basic.lean
eric-wieser: To be clear, what I'm suggesting as the simpler alternative is
```suggestion
def cons {n} (msb : Bool) (lsbs : BitVec n) : BitVec (1+n) :=
  ofBool msb ++ lsbs
```

alexkeizer: I would be somewhat wary of having the resulting length be `1+n`. The precedent with, for example, `Vector.cons`/`Vector.concat` and `Fin.cons`/`Fin.snoc` is to have the result of both operations be `n+1`-sized

eric-wieser: Ah, that's a good argument. Perhaps that's worth putting in a docstring or at least a comment.

Std/Data/BitVec/Basic.lean
```diff
@@ -433,3 +433,50 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
...
+
+/-- Prepend a single bit to the front of a bitvector, using big endian order (see `append`).
+    That is, the new bit is the most significant bit. -/
+def cons {n} (msb : Bool) (lsbs : BitVec n) : BitVec (n+1) :=
+  ((ofBool msb) ++ lsbs).cast (Nat.add_comm ..)
+
+/-- Every bitvector of length 0 is equal to `nil`, i.e., there is only one empty bitvector -/
+theorem eq_nil : ∀ (x : BitVec 0), x = nil
+  | ofFin ⟨0, _⟩ => rfl
```
eric-wieser, line 472: This would usually be stated as `Unique (BitVec 0)`

alexkeizer: Is [Unique](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Unique.html#Unique) not a Mathlib thing? Here again I'm just following the precedent of [`Vector.eq_nil`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Vector.html#Vector.eq_nil)

Std/Data/BitVec/Basic.lean
joehendrix: `concat` sounds like a synonym for `append` or `join`. What about `snoc` since that makes the naming clear?   When vectors are in `std` they could use the same naming convention.

Std/Data/BitVec/Basic.lean
```diff
@@ -433,3 +433,58 @@ def signExtend (v : Nat) (x : BitVec w) : BitVec v := .ofInt v x.toInt
...
+
+@[simp]
+theorem append_ofBool (msbs : BitVec w) (lsb : Bool) :
+    msbs ++ ofBool lsb = concat msbs lsb :=
+  rfl
+
+@[simp]
+theorem ofBool_append (msb : Bool) (lsbs : BitVec w) :
+    ofBool msb ++ lsbs = (cons msb lsbs).cast (Nat.add_comm ..) :=
+  rfl
```
joehendrix, line 490: I think we should rewrite `cons` and `concat` (or whatever they are renamed to) to append, so the order should be reversed.  I also don't think a rule that introduces a cast should be labeled `@[simp]`.

joehendrix (changes requested):

Std/Data/BitVec/Basic.lean
alexkeizer: `Vector.concat` was chosen to be consistent with `List.concat`, which is part of core.
I like both `concat` and `snoc`, but I'm not sure this is worth the churn of a rename in core, and I do care about being consistent (it's already somewhat annoying that `Fin.snoc` deviates from these other two).

Std/Data/BitVec/Basic.lean
eric-wieser: Ah, in which case you should probe `Subsingleton`, which isn't.

Std/Data/BitVec/Basic.lean
alexkeizer: The reverse order would just be unfolding the definitions of `concat` and `cons`, which I don't think we should mark as `simp`. The `cast` is inevitable, given the decision to have `cons` return a `n + 1` sized bitvector, rather than `1 + n`. 
For now, I've just removed the `simp` attributes from both lemmas

Std/Data/BitVec/Basic.lean
alexkeizer: Added the appropriate instance!
I still like the `eq_nil` as a special case, though, since I think of `nil` as the canonical empty bitvector to be used in lemmas, as is the case for vectors.

Std/Data/BitVec/Basic.lean
joehendrix: Oh I see.  That looks good to me.

joehendrix (approved):
I'll plan to merge later today (Pacific) if there are no further discussions.

alexkeizer:
Perfect, thanks for the quick review!

PR #361, opened 2023-11-11
feat: introduce balanced mod

semorrison:
I will need `bmod` for the omega algorithm, and Aymeric Fromherz on [zulip](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Signed.20Modulo.20Operator/near/401346626) also wanted some of this.

Some of the proofs have already become lovely test cases for `omega`!

- [x] depends on: #291 
- [x] depends on #345 
- [x] depends on #360


Std/Data/Int/DivMod.lean
```diff
@@ -864,6 +878,174 @@ theorem eq_one_of_mul_eq_one_right {a b : Int} (H : 0 ≤ a) (H' : a * b = 1) :
...
+
+/--
+Balanced mod, taking values in the range [- m/2, (m - 1)/2].
+-/
+def bmod (x : Int) (m : Nat) : Int :=
+  let r := x % m
+  if r < (m + 1) / 2 then
+    r
+  else
+    r - m
```
eric-wieser, line 896: To make things more confusing; C++ provides a kind of "balanced modulo" called [`std::remainder`](https://en.cppreference.com/w/cpp/numeric/math/remainder), but it rounds to even instead of rounding down!

eric-wieser, line 896: Is this better or worse as:
```suggestion
  let offset := m / 2; (x + offset) % m - offset
```

The fact it avoids a branch means it might be both faster and easier to prove things about

semorrison: Oooh... That is probably much better. :-)

semorrison: I had a bit of a look. Some of the inequalities do get easier, but the annoying lemma at the end `bmod_natAbs_plus_one`, looks more annoying: the proof I have in mind splits into cases along the same lines as the definition here.

I'll keep this suggestion in mind, but it is a low priority to replace the proofs for now.

PR #364, opened 2023-11-15
feat: Array.attach

nomeata:
This adds `Array.attach`, an analogue to `List.attach`, to facilitate
well-founded recursion through arrays.


nomeata:
This is extracted from #332, but probably useful on its own. I just told David Christiansen to use it, only to find that it was still on a PR.

I had to put it into its own file due to the dependency on `List.attach`, it seems.

david-christiansen:
Thanks!

nomeata:
awaiting-review

Std/Data/Array/Attach.lean
```diff
@@ -0,0 +1,16 @@
+import Std.Data.List.Basic
```
digama0, line 1: Can this go in `Std.Data.Array.Basic`? Assuming it doesn't increase the imports.

digama0:
It seems that it did require a cross-cutting import after all. I've rejiggered things to avoid that. (Yes, some documentation regarding the allowed imports is overdue. The short version is that `Std.A.Basic` should not import `Std.B.Basic` unless it is doing so already, and requirements of std definitions within itself should be moved to a `Init` folder.)

nomeata:
Thanks for that help!

PR #372, opened 2023-11-17
feat: simple lemmas about LE and LT

semorrison:


Std/Classes/Order.lean
```diff
@@ -101,3 +101,15 @@ instance (f : α → β) (cmp : β → β → Ordering) [TransCmp cmp] : TransCm
 
 instance (f : α → β) (cmp : β → β → Ordering) [TransCmp cmp] : TransCmp (byKey f cmp) where
   le_trans h₁ h₂ := TransCmp.le_trans (α := β) h₁ h₂
+
+@[simp] theorem ge_iff_le [LE α] {x y : α} : x ≥ y ↔ y ≤ x := Iff.rfl
+
+@[simp] theorem gt_iff_lt [LT α] {x y : α} : x > y ↔ y < x := Iff.rfl
+
+theorem le_of_eq_of_le {a b c : α} [LE α] (h₁ : a = b) (h₂ : b ≤ c) : a ≤ c := by subst h₁; exact h₂
```
Ruben-VandeVelde, line 109: These have dot notation aliases as well in mathlib (e.g. `Eq.trans_le`). How do you feel about adding those?

semorrison: I'm not convinced they make sense without also upstreaming the primed versions e.g. `le_of_eq_of_le'`, we could have `Eq.trans_ge` as well as `Eq.trans_le`. 
 
More generally, I think with upstreaming PRs the right course of action instead of proposing "how about also upstreaming these?" is to make a second PR. I'd prefer to be able to move stuff that I know I need, without necessarily having to have a discussion about how much more to move at the same time.

There's also the danger of upstreaming cruft from Mathlib that isn't actually going to be used. :-)

fgdorais: `Eq.trans_le`/`Eq.trans_lt` are both special cases of `Eq.substr`. I don't think we need to clutter the `Eq` namespace with special cases of `Eq.subst`/`Eq.substr`.

fgdorais: I can see the utility for `.trans_eq` but I would prefer `.congr_right` or similar since I can't think of a precedent for `trans_eq` in std or core.

semorrison: There's no way I'm going to include in this PR a lemma I don't care about, moved from Mathlib, but changing it's name.

As there is needless controversy, I'm removing the aliases from this PR. If someone needs them in Std, or has opinions about their names in Mathlib, please PR those separately!

Std/Classes/Order.lean
```diff
@@ -5,6 +5,7 @@ Authors: Mario Carneiro
 -/
 import Std.Data.Ord
 import Std.Tactic.Simpa
+import Std.Tactic.Alias
```
digama0, line 8:
```suggestion
```

PR #373, opened 2023-11-18
feat: lemmas for ReaderT and StateT

fgdorais:
Handy lemmas extracted from #85 before closing that old PR.

### DIFF
--- a/Std/Control/Lemmas.lean
+++ b/Std/Control/Lemmas.lean
@@ -1,3 +1,8 @@
+/-
+Copyright (c) 2023 François G. Dorais. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
 import Std.Tactic.Ext
 
 namespace ReaderT
--- a/Std/Control/Lemmas.lean
+++ b/Std/Control/Lemmas.lean
@@ -10,10 +10,10 @@ namespace ReaderT
 attribute [ext] ReaderT.ext
 
 @[simp] theorem run_failure [Monad m] [Alternative m] (ctx : ρ) :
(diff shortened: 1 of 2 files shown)


Std/Control/Lemmas.lean
```diff
@@ -0,0 +1,25 @@
+import Std.Tactic.Ext
+
+namespace ReaderT
+
+attribute [ext] ReaderT.ext
+
+@[simp] theorem run_failure [Monad m] [Alternative m] (ctx : ρ) :
+  (failure : ReaderT ρ m α).run ctx = failure := rfl
```
semorrison, line 8: As far as I can tell the naming conventions as [written](https://leanprover-community.github.io/contribute/naming.html) are silent on whether this should be `run_failure` or `failure_run`.

I think we should start becoming more consistent about this, and really hope that we can go with `failure_run` here:
* We all use dot notation, and basic principle is "important fragments of the statement, read left to right".
* This is obviously a theorem about `failure` more than it is a theorem about `run`, so `failure` deserves top billing.

Obviously we should not hold up this PR for such a discussion, of course. :-)

semorrison: @digama0 @joehendrix

semorrison (approved):

Std/Control/Lemmas.lean
digama0: The [last time](https://github.com/leanprover/std4/pull/329#issuecomment-1786193983) this came up, we ended up with "mostly prefix, but binary operators are infix" for a not-completely-defined notion of what constitutes a binary operator. Here `run` could conceivably be interpreted as a binary operator, but I lean toward not doing so because it doesn't have an associated notation.

PR #374, opened 2023-11-19
feat: `Fin.foldl` and `Fin.foldr`

fgdorais:


Std/Data/Fin/Basic.lean
```diff
@@ -48,3 +49,22 @@ def natAdd (n) (i : Fin m) : Fin (n + m) := ⟨n + i, Nat.add_lt_add_left i.2 _
 
 /-- `min n m` as an element of `Fin (m + 1)` -/
 def clamp (n m : Nat) : Fin (m + 1) := ⟨min n m, Nat.lt_succ_of_le (Nat.min_le_right ..)⟩
+
+/-- Folds over `Fin n` from the left: `foldl 3 f x = f (f (f x 0) 1) 2` -/
+@[inline] def foldl (n) (f : α → Fin n → α) (init : α) : α :=
```
semorrison, line 54: I was going to ask if this warranted a characterising theorem, relating it to folding over the `List` of `Fin`s... but it seems at this point we don't even have a way to obtain the list of `Fin`s!?

Std/Data/Fin/Basic.lean
```diff
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Robert Y. Lewis, Keeley Hoek, Mario Carneiro
 -/
 import Std.Data.Nat.Init.Lemmas
+import Std.Tactic.Lint
```
semorrison, line 7:
```suggestion
```

test/fin_fold.lean
```diff
@@ -0,0 +1,5 @@
+import Std.Data.Fin.Basic
+import Std.Tactic.GuardMsgs
```
semorrison, line 2:
```suggestion
import Std.Tactic.GuardExpr
```
suffices.

Not sure how I feel about having a test file for these, when we really want a lemma!

test/fin_fold.lean
```diff
@@ -0,0 +1,5 @@
+import Std.Data.Fin.Basic
+import Std.Tactic.GuardExpr
+
+#guard Fin.foldl 4 (fun x i => List.cons i x) [] == [3,2,1,0]
+#guard Fin.foldr 4 (fun i x => List.cons i x) [] == [0,1,2,3]
```
semorrison, line 5: Not sure how I feel about having a test file for these, when we really want a lemma! I wouldn't insist on the lemma, but I think I'd prefer removing the test file in the meantime.

Std/Data/Fin/Basic.lean
fgdorais: Yes, I probably wouldn't have made this PR if there was an easy way to get the list of `Fin`. I think `Fin.foldr` is the easiest way to get that list.

If a characterization theorem is essential, I will probably make these `foldlTR`/`foldrTR` and use `csimp`.

test/fin_fold.lean
fgdorais: Deleted. It's easy to recreate if necessary.

Std/Data/Fin/Basic.lean
fgdorais: I tried it and it's kind of a mess to do that characterizing theorem. I have other fishes to fry so I'm hoping we can do without for now :/

semorrison (approved):

PR #375, opened 2023-11-19
chore: fix failure on main; something went wrong after #291

semorrison:


digama0:
That's weird, these look like the opposite of some changes from the rc2 lean bump.

semorrison:
Yes, I double checked the toolchain before committing these! Not sure what's going on there.

PR #247, opened 2023-09-07
chore: upstream run_cmd / run_tac / by_elab

semorrison:

PR #376, opened 2023-11-19
chore: quiet some tests

semorrison:
This does not complete silence the test:

* `#guard_msgs` can check messages like ``test/case.lean:196:7: warning: unused variable `x` [linter.unusedVariables]``, but it doesn't seem to be able to suppress them. Not sure why.
* It seems silly to write
```
/-- info: -/
#guard_msgs in
```
in front of every `#eval guard ...`. Should we replace `#eval guard` with a new `#guard` command?

Once we've dealt with these I can have CI maintain silence. :-)


digama0:
`#guard` already exists, doesn't it?

semorrison:
> `#guard` already exists, doesn't it?

It does, but isn't useful for replacing
```
#eval do guard (← simpComm.test ``unop_inj).isNone
```

digama0:
Another alternative is `run_cmd`, which does not do an empty print at the end.

semorrison:
> Another alternative is `run_cmd`, which does not do an empty print at the end.

PR'd two months ago: #247 :-)

joehendrix (approved):

semorrison:
> Another alternative is `run_cmd`, which does not do an empty print at the end.

@digama0, it seems the really useful command is `run_meta`, which would be equivalent to `#eval show CommandElabM Unit from do`.

How would you feel about adding:
```
/--
Helper function for `run_meta` command.
-/
def _root_.Lean.Elab.Command.runMeta (x : MetaM α) : CommandElabM α := do
  (·.1) <$> Core.CoreM.toIO (MetaM.run' x) { (← read) with } { env := (← get).env }

/--
The `run_meta doSeq` command executes code in `MetaM Unit`.
This is almost the same as `#eval show MetaM Unit from do doSeq`,
except that it doesn't print an empty diagnostic.
-/
elab (name := runMeta) "run_meta " elems:doSeq : command => do
  ← liftTermElabM <|
    unsafe evalTerm (CommandElabM Unit)
      (mkApp (mkConst ``CommandElabM) (mkConst ``Unit))
      (← `(Command.runMeta <| discard do $elems))

run_meta guard <| 1 = 1
```
?

joehendrix:
`run_meta` seems useful to me.

semorrison:
I have added a less hacky version of `Command.runMeta` (it tries to do a better but still imperfect job; unfortunately `CoreM` and `CommandElabM` are awkwardly similar without quite being the same).

digama0:
Why cant you use `runTermElabM` to run a `MetaM` inside `CommandElabM`?

semorrison:
> Why cant you use `runTermElabM` or `liftTermElabM` to run a `MetaM` inside `CommandElabM`?

🤦‍♀️ Done.

Std/Tactic/RunCmd.lean
```diff
@@ -32,6 +32,17 @@ elab (name := runCmd) "run_cmd " elems:doSeq : command => do
       (mkApp (mkConst ``CommandElabM) (mkConst ``Unit))
       (← `(discard do $elems))
 
+/--
+The `run_meta doSeq` command executes code in `MetaM Unit`.
+This is almost the same as `#eval show MetaM Unit from do doSeq`,
+except that it doesn't print an empty diagnostic.
+-/
+elab (name := runMeta) "run_meta " elems:doSeq : command => do
```
digama0, line 40: This is misnamed and mis-documented now since `$elems` is expecting a TermElabM.

Std/Tactic/RunCmd.lean
```diff
@@ -32,6 +32,17 @@ elab (name := runCmd) "run_cmd " elems:doSeq : command => do
...
+  ← liftTermElabM <|
+    unsafe evalTerm (CommandElabM Unit)
+      (mkApp (mkConst ``CommandElabM) (mkConst ``Unit))
+      (← `(Command.liftTermElabM <| discard do $elems))
```
digama0, line 44:
```suggestion
    liftTermElabM <|
      unsafe evalTerm (MetaM Unit)
        (mkApp (mkConst ``MetaM) (mkConst ``Unit))
        (← `(discard do $elems))
```

Std/Tactic/RunCmd.lean
semorrison: With your proposed change I think this is back to being correct?

digama0: assuming I got the syntax correct...

Std/Tactic/RunCmd.lean
```diff
@@ -32,6 +32,17 @@ elab (name := runCmd) "run_cmd " elems:doSeq : command => do
       (mkApp (mkConst ``CommandElabM) (mkConst ``Unit))
       (← `(discard do $elems))
 
+/--
+The `run_meta doSeq` command executes code in `MetaM Unit`.
+This is almost the same as `#eval show MetaM Unit from do doSeq`,
+except that it doesn't print an empty diagnostic.
```
semorrison, line 38:
```suggestion
The `run_meta doSeq` command executes code in `TermElabM Unit`.
This is almost the same as `#eval show TermElabM Unit from do doSeq`,
except that it doesn't print an empty diagnostic.
```

semorrison:
@digama0, your suggestion above didn't work, and I couldn't make it work, so I've kept the current implementation.

I made `run_term_elab` and `run_meta` as two separate commands: `run_meta` could always be replaced by `run_term_elab`, I think, but I think it's more readable and accessible.

PR #378, opened 2023-11-20
feat: add a few missing lemmas about array size

dupuisf:


Std/Data/Array/Lemmas.lean
```diff
@@ -227,6 +227,26 @@ theorem mapIdx_induction' (as : Array α) (f : Fin as.size → α → β)
   simp only [reverse]; split <;> simp [go]
 termination_by _ => j - i
 
+@[simp] theorem size_range {n : Nat} : (range n).size = n := by
```
semorrison, line 230: I would prefer `range_size` here, since we all use dot notation now? As I understand it, the naming rules are flexible on this.

semorrison (approved):
Otherwise, LGTM.

Std/Data/Array/Lemmas.lean
digama0: No, the naming convention says to write these in prefix, although I certainly understand the sentiment. To do otherwise we would need some relatively well defined way to differentiate prefix and postfix functions (and we don't want it to depend on how the author happened to have written the theorem since the user can't predict that).

PR #379, opened 2023-11-20
chore: remove unnecessary files

fgdorais:
These files are not necessary since `Std.lean` imports everything.

### DIFF
--- a/Std/Lean/Meta/UnusedNames.lean
+++ b/Std/Lean/Meta/UnusedNames.lean
@@ -5,7 +5,7 @@ Authors: Jannis Limperg
 -/
 
 import Lean.Meta
-import Std.Data.String
+import Std.Data.String.Basic
 
 open Lean Lean.Meta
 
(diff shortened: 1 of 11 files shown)


digama0:
These files exist so that people can import parts of Std without importing the whole thing. This is an intended use-case. They also exist so that Std can be backward-compatible when splitting files.

Std/Lean/Meta/UnusedNames.lean
```diff
@@ -5,7 +5,7 @@ Authors: Jannis Limperg
 -/
 
 import Lean.Meta
-import Std.Data.String
+import Std.Data.String.Basic
```
digama0, line 8: (that said, within Std itself we should be using minimal imports, so this could go in a separate PR)

fgdorais:
I submitted preemptive patches [aesop#86](https://github.com/leanprover-community/aesop/pull/86) and [mathlib#8538](https://github.com/leanprover-community/mathlib4/pull/8538). Once these are applied, this PR can be merged with no downstream effects to Aesop and Mathlib.

PR #381, opened 2023-11-20
feat: missing ext lemmas

fgdorais:


Std/Tactic/Ext.lean
```diff
@@ -202,7 +202,9 @@ syntax "ext?" (colGt ppSpace rintroPat)* (" : " num)? : tactic
 
 end Std.Tactic.Ext
 
-attribute [ext] funext propext
+attribute [ext] funext propext Subtype.eq
+attribute [ext] Prod PProd Sigma PSigma
```
semorrison, line 206: This changes the signature of `Prod.ext` relative to what is in Mathlib, so this will need a Mathlib patch reviewed before it can be merged.

semorrison: Similarly for `Sigma`.

Mathlib, correctly I think, uses implicit arguments for the terms being compared.

PR #385, opened 2023-11-23
feat: Eq.ndrec lemma

fgdorais:
Not sure about the name nor where to put this but it is definitely handy to get rid of `▸`:
```lean
example (h : n = m) (x : Fin n) : (Eq.ndrec x h).val = x.val := congr_ndrec ..
```


fgdorais:
Hm. It doesn't pass the simp test because the lhs head symbol is a variable. Still, it *sometimes* works and it objectively simplifies when it does. Is this worth an exception?

digama0:
> Hm. It doesn't pass the simp test because the lhs head symbol is a variable. Still, it sometimes works and it objectively simplifies when it does. Is this worth an exception?

The reason these are disallowed is not because it doesn't work, but because it's a performance issue - they cannot be meaningfully indexed so they apply literally everywhere. I think there should not be an exception made here since it is a rarely used lemma.

fgdorais:
Interesting idea. How would that work?

joehendrix (approved):
I don't have any recent experience with dependent congruence elimination, but this seems reasonable to me.

fgdorais:
This lemma could actually be superseded by #465. I'm not sure though.

fgdorais:
False alarm, it's not quite covered by #465.

PR #386, opened 2023-11-24
feat: lemmas characterising Array.all

semorrison:
The corresponding lemmas about `Array.any` already exist in `Std.Data.Array.Init.Lemmas`.

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -336,3 +336,26 @@ theorem forIn_eq_data_forIn [Monad m]
       simp [← this]; congr; funext x; congr; funext b
       rw [loop (i := i)]; rw [← ij, Nat.succ_add]; rfl
   simp [forIn, Array.forIn]; rw [loop (Nat.zero_add _)]; rfl
+
+theorem all_iff_forall (p : α → Bool) (as : Array α) (start stop) :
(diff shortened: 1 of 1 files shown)


fgdorais (approved):

Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +336,26 @@ theorem forIn_eq_data_forIn [Monad m]
       simp [← this]; congr; funext x; congr; funext b
       rw [loop (i := i)]; rw [← ij, Nat.succ_add]; rfl
   simp [forIn, Array.forIn]; rw [loop (Nat.zero_add _)]; rfl
+
+theorem all_iff_forall (p : α → Bool) (as : Array α) (start stop) :
+    all as p start stop ↔ ∀ i : Fin as.size, start ≤ i.1 ∧ i.1 < stop → p as[i] := by
+  have := SatisfiesM_anyM_iff_exists (m := Id) (fun a => ! p a) as start stop (! p as[·]) (by simp)
+  rw [SatisfiesM_Id_eq] at this
+  dsimp [all, allM, Id.run]
+  rw [show ∀ b : Bool, (!b) = true ↔ !(b = true) by decide]
```
digama0, line 345: this looks like a lemma

semorrison: I replaced it with `rw [Bool.not_eq_true', Bool.eq_false_iff, Ne]`.

PR #387, opened 2023-11-24
feat: lemmas about Array.filter and Array.filterMap

semorrison:


Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +338,58 @@ theorem forIn_eq_data_forIn [Monad m]
...
+  dsimp only [filter]
+  rw [foldl_eq_foldl_data]
+  generalize l.data = l
+  suffices ∀ a, (List.foldl (fun r a => if p a = true then push r a else r) a l).data =
+      a.data ++ List.filter p l by
+    simpa using this #[]
+  induction l
+  · simp_all
+  · simp only [List.filter_cons]
+    split <;> simp_all
```
fgdorais, line 360:
```suggestion
  induction l with simp
  | cons => split <;> simp [*]
```

Std/Data/List/Lemmas.lean
```diff
@@ -1279,6 +1279,12 @@ end erase
 @[simp] theorem filter_cons_of_neg {p : α → Bool} {a : α} (l) (pa : ¬ p a) :
     filter p (a :: l) = filter p l := by rw [filter, eq_false_of_ne_true pa]
 
+theorem filter_cons :
+    (x :: xs : List α).filter p = if p x then x :: (xs.filter p) else xs.filter p := by
+  split <;> rename_i h
+  · rw [filter_cons_of_pos _ h]
+  · rw [filter_cons_of_neg _ h]
```
fgdorais, line 1286:
```suggestion
  split <;> simp [*]
```

fgdorais (changes requested):

Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +338,56 @@ theorem forIn_eq_data_forIn [Monad m]
...
+  generalize l.data = l
+  have this : ∀ a : Array β, (Id.run (List.foldlM (m := Id) ?_ a l)).data =
+    a.data ++ List.filterMap f l := ?_
+  exact this #[]
+  induction l
+  · simp_all [Id.run]
+  · simp_all [Id.run]
+    split <;> simp_all
+
+@[simp] theorem mem_filterMap (f : α → Option β) (l : Array α) {b : β} :
```
joehendrix, line 391:
```suggestion
theorem mem_filterMap (f : α → Option β) (l : Array α) {b : β} :
```

I'd let people opt into this.

Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +338,56 @@ theorem forIn_eq_data_forIn [Monad m]
       simp [← this]; congr; funext x; congr; funext b
       rw [loop (i := i)]; rw [← ij, Nat.succ_add]; rfl
   simp [forIn, Array.forIn]; rw [loop (Nat.zero_add _)]; rfl
+
+/-! ### map -/
+
+@[simp] theorem mem_map {f : α → β} {l : Array α} : b ∈ l.map f ↔ ∃ a, a ∈ l ∧ f a = b := by
```
joehendrix, line 344:
```suggestion
theorem mem_map {f : α → β} {l : Array α} : b ∈ l.map f ↔ ∃ a, a ∈ l ∧ f a = b := by
```

Opt in?

Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +338,56 @@ theorem forIn_eq_data_forIn [Monad m]
...
+  rw [foldl_eq_foldl_data]
+  generalize l.data = l
+  suffices ∀ a, (List.foldl (fun r a => if p a = true then push r a else r) a l).data =
+      a.data ++ List.filter p l by
+    simpa using this #[]
+  induction l with simp
+  | cons => split <;> simp [*]
+
+@[simp] theorem filter_filter (q) (l : Array α) :
+    filter p (filter q l) = filter (fun a => p a ∧ q a) l := by
```
joehendrix, line 361:
```suggestion
theorem filter_filter (q) (l : Array α) :
    filter p (filter q l) = filter (fun a => p a ∧ q a) l := by
```

I think this is fairly reasonable as a simp rule, but am suggesting it for consistency with `mem_map` and `mem_filterMap` suggestions.

Std/Data/Array/Lemmas.lean
semorrison: Interesting. I think I disagree on all three suggestions here (removing `@[simp]` from `mem_map`, `mem_filterMap`, and `filter_filter`).

A few points:
* The right hand sides of `mem_map` and `mem_filterMap` do look "more complicated" at first, and hence you should be suspicious of these as simp lemmas
* However, a very standard next step is to run `subst` on the resulting `f a = b` expression (e.g. via `rcases` or `obtain`), and so the net result is that you have "simplified" `b` into `f a`, and `b ∈ l.map` into `a ∈ l`.
* The `List` versions of these are `@[simp]`. It's fine if the simp normal form for operations on `List` and `Array` are different, but they should be clear reasons why they diverge.
* These lemmas are already a bit strange: `∈` is not really part of the primary API for `Array`. The fact that we're in the context of "membership of arrays" means we are probably working adjacent to the `List` API, and so it's good to stay parallel.
* `filter_filter` seems pretty safe to me: you've traded nesting of API operations in logical connectives.
* Generally, I think lemmas that trade "an interaction of API functions" into "logical operations or lambdas" are great `simp` lemmas. In some sense all functions like `filter`, `erase`, etc. are black boxes, and we want to handle proofs about them with: `simp`, followed if necessary by "pure logic".
* Even more generally, I am in favour of making `simp` as powerful as possible! It's annoying to have to locally set up your simp set in a downstream file, and many users just won't do it.
* I would prefer to see lots of `by simp` and the occasional `by simp only [X, Y, Z]` (and maybe a very rare `by simp [-X]`) over seeing frequent `by simp [X, Y]`.
* If we had infinite capacity to review these things, one might ask to see evidence in the form of a proof that is nicer with the `@[simp]` attribute, and some argument that the `@[simp]` attribute will never or rarely go wrong. Obviously we don't have that. But while we're getting starting perhaps it's helpful? I could dig up the locations I needed these simp lemmas.

At the end of the day, of course it's okay to remove them: we can always add them back in later.

joehendrix: Thanks for the thoughtful response.

I agree `Array` and `List` should be consistent, and if it's already using this as `@[simp]` rule, let's keep it the same.  Even if this does make the theorem slightly worse it's a constrained expansion since `a` and `b` don't appear more times on the right-hand side.

Conversely, I am a little surprised `mem_filter` is not annotated as simp.  Any reason not to make that `simp`?

semorrison: Hmm, not that I can recall. I've added `@[simp]` to `mem_filter` now.

joehendrix (approved):

PR #388, opened 2023-11-24
feat: Array.join

semorrison:


Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +337,31 @@ theorem forIn_eq_data_forIn [Monad m]
...
+  dsimp [join]
+  simp only [foldl_eq_foldl_data]
+  generalize l.data = l
+  have : ∀ a : Array α, (List.foldl ?_ a l).data = a.data ++ ?_ := ?_
+  exact this #[]
+  induction l with
+  | nil => simp_all
+  | cons h t ih =>
+    simp only [List.foldl_cons, ih, List.map_cons, List.join_cons, ← List.append_assoc]
+    induction h.data <;> simp_all
```
fgdorais, line 353:
```suggestion
  | nil => simp
  | cons h => induction h.data <;> simp [*]
```
Might depend on new `join` def.

Std/Data/Array/Basic.lean
```diff
@@ -117,6 +117,13 @@ Unsafe implementation of `attach`, taking advantage of the fact that the represe
 @[implemented_by attachImpl] def attach (xs : Array α) : Array {x // x ∈ xs} :=
   ⟨xs.data.pmap Subtype.mk fun _ => Array.Mem.mk⟩
 
+/--
+`O(|join L|)`. `join L` concatenates all the arrays in `L` into one array.
+* `join #[#[a], #[], #[b, c], #[d, e, f]] = #[a, b, c, d, e, f]`
+-/
+def join (l : Array (Array α)) : Array α :=
+  l.foldl (init := #[]) fun a m => m.foldl (init := a) fun a' x => a'.push x
```
fgdorais, line 125:
```suggestion
@[inline] def join (l : Array (Array α)) : Array α := l.foldl (.++.) #[]
```

fgdorais (changes requested):

Std/Data/Array/Lemmas.lean
semorrison: Haha, worked both before and after, unlike my proof.

joehendrix (approved):

fgdorais (approved):

PR #389, opened 2023-11-24
feat: miscellaneous Array lemmas

semorrison:


Std/Data/Array/Lemmas.lean
```diff
@@ -37,6 +37,14 @@ attribute [simp] isEmpty uget
 @[simp] theorem toArray_data : (a : Array α) → a.data.toArray = a
   | ⟨l⟩ => ext' (data_toArray l)
 
+@[simp] theorem mem_data {a : α} {l : Array α} : a ∈ l.data ↔ a ∈ l := (mem_def _ _).symm
```
digama0, line 40:
```suggestion
theorem mem_data {a : α} {l : Array α} : a ∈ l.data ↔ a ∈ l := (mem_def _ _).symm
```
Although it is simpler, this simp lemma will interfere with the pattern of proving theorems about arrays by introducing `.data` and using simp lemmas which distribute this. For example `a ∈ arr.pop.data` becomes non-confluent with this.

semorrison: Hmm, okay. That's a pity.

PR #390, opened 2023-11-24
feat: ensure HashMap computes in the kernel

semorrison:
It seems a poor tradeoff to eke out a little more performance at the cost of losing the ability to compute with `HashMap` in the kernel.

### DIFF
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -29,18 +29,18 @@ def mk (buckets := 8) (h : 0 < buckets := by decide) : Buckets α β :=
   ⟨mkArray buckets .nil, by simp [h]⟩
 
 /-- Update one bucket in the bucket array with a new value. -/
-def update (data : Buckets α β) (i : USize)
+def update (data : Buckets α β) (i : UInt64)
     (d : AssocList α β) (h : i.toNat < data.1.size) : Buckets α β :=
-  ⟨data.1.uset i d h, (Array.size_uset ..).symm ▸ data.2⟩
+  ⟨data.1.set ⟨i.toNat, h⟩ d, (Array.size_set ..).symm ▸ data.2⟩
 
 /--
 The number of elements in the bucket array.
 Note: this is marked `noncomputable` because it is only intended for specification.
 -/
-noncomputable def size (data : Buckets α β) : Nat := .sum (data.1.data.map (·.toList.length))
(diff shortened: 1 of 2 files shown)


Std/Data/HashMap/Basic.lean
```diff
@@ -29,18 +29,18 @@ def mk (buckets := 8) (h : 0 < buckets := by decide) : Buckets α β :=
...
     (d : AssocList α β) (h : i.toNat < data.1.size) : Buckets α β :=
-  ⟨data.1.uset i d h, (Array.size_uset ..).symm ▸ data.2⟩
+  ⟨data.1.set ⟨i.toNat, h⟩ d, (Array.size_set ..).symm ▸ data.2⟩
 
 /--
 The number of elements in the bucket array.
 Note: this is marked `noncomputable` because it is only intended for specification.
 -/
-noncomputable def size (data : Buckets α β) : Nat := .sum (data.1.data.map (·.toList.length))
+def size (data : Buckets α β) : Nat := .sum (data.1.data.map (·.toList.length))
```
JamesGallicchio, line 40: per comment in code this was noncomputable to discourage use outside of specification. does it have to be computable?

semorrison: Oops, sorry, this was from when I was working out what was going on. I'll revert.

JamesGallicchio (approved):

Std/Data/HashMap/Basic.lean
```diff
@@ -369,3 +369,7 @@ def ofListWith [BEq α] [Hashable α] (l : List (α × β)) (f : β → β → 
     match m.find? p.1 with
     | none   => m.insert p.1 p.2
     | some v => m.insert p.1 <| f v p.2
+
+-- Verify that basic operations are computable in the kernel, and not blocked by
+-- `System.Platform.numBits`.
+example : (HashMap.empty.insert 0 0).isEmpty = false := rfl
```
digama0, line 375: Tests don't go in `Std/`

digama0:
Note, this PR may be rejected or modified pending discussion [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/no.20kernel.20reduction.20of.20HashMap.2Einsert/near/403887177).

semorrison:
Yes, I agree this is a bad idea now. Thanks for the explanations, very helpful.

(Sadly, using RBMap in my application turned out to be even slower that just traversing lists...)

PR #392, opened 2023-11-26
feat: tail-recursive and short-circuiting version of List.any/all

semorrison:


Std/Data/List/Basic.lean
```diff
@@ -13,7 +13,49 @@ namespace List
...
+@[csimp] theorem any_eq_anyTR : @any = @anyTR := by
+  funext α xs p
+  dsimp [any, anyTR]
+  induction xs with
+  | nil => simp [anyTR.go]
+  | cons h t ih =>
+    simp_all only [foldr_cons, anyTR.go]
+    by_cases w : p h
+    · simp [w, anyTR.go]
+    · simp [w]
```
fgdorais, line 35:
```suggestion
@[csimp] theorem any_eq_anyTR : @any = @anyTR := by
  funext _ xs p
  induction xs with simp_all [any, anyTR, anyTR.go]
  | cons h => cases p h <;> simp [anyTR.go]
```

Std/Data/List/Basic.lean
```diff
@@ -13,7 +13,49 @@ namespace List
...
+@[csimp] theorem all_eq_allTR : @all = @allTR := by
+  funext α xs p
+  dsimp [all, allTR]
+  induction xs with
+  | nil => simp [allTR.go]
+  | cons h t ih =>
+    simp_all only [foldr_cons, allTR.go]
+    by_cases w : p h
+    · simp [w]
+    · simp [w, allTR.go]
```
fgdorais, line 56:
```suggestion
@[csimp] theorem all_eq_allTR : @all = @allTR := by
  funext _ xs p
  induction xs with simp_all [all, allTR, allTR.go]
  | cons h => cases p h <;> simp [allTR.go]
```

fgdorais (approved):
Looks great! Just a couple rounds of golf...

Std/Data/List/Basic.lean
```diff
@@ -13,7 +13,37 @@ namespace List
...
-/-- Tail recursive version of `erase`. -/
+/-- Tail recursive version of `List.any`, with short-circuiting. -/
+def anyTR (xs : List α) (p : α → Bool) : Bool :=
+  go false xs
+where
+  /-- Auxiliary for `anyTR`. -/
+  go : Bool → List α → Bool
+  | true, _ => true
+  | false, [] => false
+  | false, h :: t => go (p h) t
```
joehendrix, line 24: How about the following?

```suggestion
def anyTR : List α -> (α → Bool) -> Bool
| [], _ => false
| h :: t, p => p h || anyTR t p
```

Std/Data/List/Basic.lean
```diff
@@ -13,7 +13,37 @@ namespace List
...
+
+/-- Tail recursive version of `List.all`, with short-circuiting. -/
+def allTR (xs : List α) (p : α → Bool) : Bool :=
+  go true xs
+where
+  /-- Auxiliary for `allTR`. -/
+  go : Bool → List α → Bool
+  | false, _ => false
+  | true, [] => true
+  | true, h :: t => go (p h) t
```
joehendrix, line 39:
```suggestion
def allTR : List α -> (α → Bool) -> Bool
| [], _ => true
| h :: t, p => p h && allTR t p
```

Std/Data/List/Basic.lean
fgdorais: After inspecting the C code in stage0, it seems that `List.or`, which is the specialization of `List.any` where `p=id`, compiles as short-circuiting tail-recursive code. I'm not sure about `List.any` though since it has a lot more C crud.

I wonder if the short-circuiting tail-recursive version should actually be in core?

semorrison: 💡 Interesting. I wish we had an annotation to require / check tail-recursiveness...

digama0: @fgdorais my [tail recursion detector](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/tail.20recursion/near/404515501) says that `List.any`/`List.all` is not tail-recursive, but `List.or`/`List.and` is, which is a bit mysterious.

joehendrix: I'm ok with merging `anyTR`/`allTR` if that would unblock something, but making the change to core seems preferable.

semorrison:
I'll close this in favour of leanprover/lean4#2972. Reviews there welcome.

PR #393, opened 2023-11-26
feat: lemmas about List.indexOf/indexesOf

semorrison:
Adds `indexOf_mem_indexesOf (m : x ∈ xs) : xs.indexOf x ∈ xs.indexesOf x`.

Currently `List.indexOf` and `List.indexesOf` use `(a == ·)` and `(· == a)` respectively as the test function to locate elements.

The fact that they are inconsistent makes proofs relating them slightly inconvenient (although for the main proof here, we need a `LawfulBEq` hypothesis anyway so it would be possible to preserve the current behaviour if preferred).

Thus this PR proposes switching both to using `(· == a)`.

Std is still quite inconsistent in this choice in other places, and it would be good if someone could do a global review of this issue.

For these two functions, either choice is fine.


Std/Data/List/Basic.lean
```diff
@@ -361,7 +361,7 @@ drop_while (· != 1) [0, 1, 2, 3] = [1, 2, 3]
   | a :: l, n => bif p a then n else go l (n + 1)
 
 /-- Returns the index of the first element equal to `a`, or the length of the list otherwise. -/
-def indexOf [BEq α] (a : α) : List α → Nat := findIdx (a == ·)
+def indexOf [BEq α] (a : α) : List α → Nat := findIdx (· == a)
```
digama0, line 364: This needs some justification. I think there was some discussion earlier about switching one or more of these which resulted in a TODO comment in mathlib, but I forget the details.

semorrison: I've updated the PR description.

I couldn't find the TODO in mathlib, unfortunately. Mathlib is consistent however about always using `· == a`, so maybe we should just do that globally here too.

joehendrix (approved):

PR #396, opened 2023-11-27
feat: move lemmas about lists from mathlib + add lemmas about `Array.zipWith`

dupuisf:
Lemmas from `Mathlib.Data.List.Basic`:
`drop_one`, `drop_add`, `drop_left`, `drop_left'`, `drop_eq_get_cons`

The rest of the list lemmas are from `Mathlib.Data.List.Zip`.

The array lemmas are new.


Std/Data/List/Lemmas.lean
```diff
@@ -206,6 +229,66 @@ theorem zipWith_get? {f : α → β → γ} :
     | nil => simp
     | cons b bs => cases i <;> simp_all
 
+@[simp]
+theorem zipWith_eq_nil_iff {f : α → β → γ} {l l'} : zipWith f l l' = [] ↔ l = [] ∨ l' = [] := by
+  cases l <;> cases l' <;> simp
+
+theorem lt_length_left_of_zipWith {f : α → β → γ} {i : Nat} {l : List α} {l' : List β}
```
semorrison, line 236: Or `lt_length_left_of_lt_zipWith_length`? Or `lt_length_of_lt_zipWith_left_length`? :-)

dupuisf: Indeed that sounds better. I just changed it, but we'll have to remember to still delete it from mathlib though!

Std/Data/Array/Lemmas.lean
```diff
@@ -336,3 +338,78 @@ theorem forIn_eq_data_forIn [Monad m]
...
+    (as.zipWith bs f).size = min as.size bs.size := by
+  simp [size_eq_length_data, zipWith_eq_zipWith_data, List.length_zipWith]
+
+theorem size_zipWith_left (as : Array α) (bs : Array β) (h : as.size ≤ bs.size)
+    (f : α → β → γ) : (as.zipWith bs f).size = as.size := by
+  rw [size_zipWith, Nat.min_eq_left h]
+
+theorem size_zipWith_right (as : Array α) (bs : Array β) (h : bs.size ≤ as.size)
+    (f : α → β → γ) : (as.zipWith bs f).size = bs.size := by
+  rw [size_zipWith, Nat.min_eq_right h]
```
joehendrix, line 396: I feel like it's reasonable to just have `size_zipWith` and rely on min lemmas rather than have these special cases.

dupuisf: Done.

Std/Data/List/Lemmas.lean
```diff
@@ -240,6 +323,55 @@ theorem zip_map_left (f : α → γ) (l₁ : List α) (l₂ : List β) :
...
+  | l₁, r₁, [], r₂, h => by simp only [eq_nil_of_length_eq_zero h]; rfl
+  | a :: l₁, r₁, b :: l₂, r₂, h => by
+    simp only [cons_append, zip_cons_cons, zip_append (Nat.succ.inj h)]
+
+theorem zip_map' (f : α → β) (g : α → γ) :
+    ∀ l : List α, zip (l.map f) (l.map g) = l.map fun a => (f a, g a)
+  | [] => rfl
+  | a :: l => by simp only [map, zip_cons_cons, zip_map']
+
+theorem mem_zip {a b} : ∀ {l₁ : List α} {l₂ : List β}, (a, b) ∈ zip l₁ l₂ → a ∈ l₁ ∧ b ∈ l₂
```
semorrison, line 347: `of_mem_zip`

semorrison (approved):

Std/Data/List/Lemmas.lean
```diff
@@ -206,6 +229,66 @@ theorem zipWith_get? {f : α → β → γ} :
...
+
+theorem lt_length_of_lt_zipWith_left_length {f : α → β → γ} {i : Nat} {l : List α} {l' : List β}
+    (h : i < (zipWith f l l').length) : i < l.length := by
+  rw [length_zipWith, Nat.lt_min] at h
+  exact h.left
+
+theorem lt_length_of_lt_zipwith_right_length {f : α → β → γ} {i : Nat} {l : List α} {l' : List β}
+    (h : i < (zipWith f l l').length) : i < l'.length := by
+  rw [length_zipWith, Nat.lt_min] at h
+  exact h.right
```
joehendrix, line 244: Sorry for delayed review.

It seems like it would be better to just make a lemma `(zipWith f l l').length = min l.length l'.length` and do the same for similar theorems such as `lt_length_left_of_zip`.

dupuisf: I just deleted these theorems -- we indeed already have `length_zipWith` and `length_zip`.

dupuisf:
awaiting-review

joehendrix (approved):

PR #397, opened 2023-11-27
feat: MLList.ofTaskList

semorrison:
Take a `List` of `Tasks`, and produce a monadic lazy list that returns the values as their become available.

(prerequisite for `hint`)


semorrison:
I've added a simple test.

joehendrix (approved):

PR #399, opened 2023-11-27
feat: run `List (MetaM x)` asynchronously, returning an `MLList MetaM x` in the order they return.

kim-em:
Depends on:
* [x] https://github.com/leanprover/std4/pull/398
* [x] https://github.com/leanprover/std4/pull/397

Prerequisite for the `hint` tactic.


Std/Data/MLList/Meta.lean
```diff
@@ -0,0 +1,224 @@
...
+
+namespace IO
+
+/--
+Given a list of values in `IO α`, executes them all in parallel as tasks, and returns
+* a cancellation hook and
+* a monadic lazy list which returns the values in the order they complete.
+
+Note that the cancellation hook merely requests cooperative cancellation:
+the tasks must call `IO.checkCanceled` themselves.
```
semorrison, line 46:
```suggestion
Note that the cancellation hook merely requests cooperative cancellation:
the tasks must call `IO.checkCanceled` themselves.
The tactic framework does so automatically, at the same time it checks for exceeding `maxHeartbeats`.
```

joehendrix (approved):
It'd be good to have some tests, but I guess if this is largely to support another commit we can test it once that's in place.

joehendrix (commented):
`(prio := .max)` seems suspicious.  Why is it needed?

Std/Data/MLList/Meta.lean
```diff
@@ -0,0 +1,226 @@
...
+After calling the cancellation hook the behaviour of the monadic lazy list should not be relied on.
+It may continue returning values, or fail.
+Recommended usage is to take a prefix of the list
+(e.g. with `MLList.takeUpToFirst` followed by `MLList.force`, or `MLList.takeAsList`)
+and then call the cancellation hook to request cancellation of later unwanted tasks.
+
+(Similarly also for `CoreM`, `TermElabM`, and `TacticM`.)
+
+## Implementation notes:
+Calling `IO.cancel` on `t.map f` does not cancel `t`,
```
semorrison, line 26:
```suggestion
## Implementation notes:
We have not thoroughly tested this approach to parallelization,
and remain concerned that in some applications tasks may get stuck waiting for each other.
This is currently used to implement parallelization for the `hint` tactic.
We recommend using this elsewhere only with caution,
and particular caution combining it with other code that manipulates tasks!

Calling `IO.cancel` on `t.map f` does not cancel `t`,
```

semorrison:
> `(prio := .max)` seems suspicious. Why is it needed?

The `prio := .max` only appears on `map` tasks which are doing minor work after the main workload has been completed. Setting the priority higher makes it less likely that returning the final value will get stuck at the `map` stage behind any long-running tasks which have yet to be started.

Std/Data/MLList/Meta.lean
```diff
@@ -0,0 +1,226 @@
...
+so we have to be careful throughout this file
+to construct cancellation hooks connected to the underlying task,
+rather than the various maps of it that we construct to pass state.
+
+Thomas Murrills has a suggestion to significantly refactor this code,
```
semorrison, line 31:
```suggestion
In several places below we use `prio := .max` to set the priority of tasks higher. This is only used on `map` tasks which run after the main workload has finished. This helps avoid the situation where the main workload has finished, but the value is not returned to the consumer because other long-running tasks are scheduled before the `map`.

Thomas Murrills has a suggestion to significantly refactor this code,
```

Std/Data/MLList/Meta.lean
```diff
@@ -0,0 +1,238 @@
...
+Given a list of monadic values in `TacticM`, runs them all as tasks,
+and returns
+* a cancellation hook and
+* the monadic lazy list which returns the values in the order they complete.
+
+See the doc-string for `IO.runGreedily` for details about the cancellation hook behaviour.
+-/
+def runGreedily (jobs : List (TacticM α)) : TacticM (BaseIO Unit × MLList TacticM α) := do
+  let (cancels, tasks) := (← jobs.mapM asTask).unzip
+  return (cancels.forM id, .squash fun _ => return MLList.ofTaskList tasks |>.liftM.mapM id)
```
joehendrix, line 230: What is. the intended semantics of `runGreedily`?  When you get a value from the tactic list, does retrieving it affect the tactic state?

Perhaps the `MLList` should have type `MLList BaseIO (Except Exception (α × Tactic.SavedState)`?

joehendrix: Oh I wanted to see if I could avoid the need to double the number of tasks (via Task.map) and did that in [this branch](https://github.com/leanprover/std4/tree/jhx/MLList.Meta).

That preserves the behavior of restoring state implicitly though which I am not sure if the right approach.

semorrison: The intention is that the state is exposed via the MLList, so one could obtain `α × Tactic.SavedState` just by `.mapM fun a => return (a, ← saveState)`.

I can add a doc-string explaining this and a test.

semorrison: I added a test, and then tried merging your branch. Your branch failed the test, with a curiously strange error message: with a goal of `Nat`, `exact 0` failed with `failed to synthesize instance OfNat (_ : True) 0`! Something messed up in the metavar context, I think. So I haven't commited the merge of your branch.

semorrison: How important is avoiding the extra tasks? I can try again at this, or we could not worry. :-)

joehendrix: I think this interface is not intuitive to me due to backtracking.  It's counter-intuitive to me, for example, that I could create a meta variable, get a value from a MLList generated by `runGreedily` in Meta and then the variable is no longer valid (because of `restoreState`).

Perhaps this patch should introduce a notion of a suspended computation (perhaps as a class) that can be resumed?  The suspended computation would have the state+context.  It's not quite as concise, but would make the code more clear.

semorrison:
I'm not going to work on this in the near future. If anyone would like to adopt, and think about a better design, please go ahead!

fgdorais:

PR #401, opened 2023-11-27
feat: declaration form to document welcome proofs

david-christiansen:
Add proof_wanted, a declaration form that is elaborated as an axiom and then thrown away. This can serve as a guide for new contributors who are looking for high-impact work that can be more quickly reviewed.

This idea came up during a chat with @semorrison, who mentioned that he had a list of lemmas that would have been useful in a recent project, but didn't really know where to put them.


fgdorais:
Such a great idea! Thanks!

Std/Util/ProofWanted.lean
```diff
@@ -0,0 +1,28 @@
...
+
+
+open Lean Parser Elab Command
+
+/-- This proof would be a welcome contribution to the library!
+
+The syntax of `proof_wanted` declarations is just like that of `theorem`, but without `:=` or the
+proof. Lean checks that `proof_wanted` declarations are well-formed (e.g. it ensures that all the
+mentioned names are in scope, and that the theorem statement is a valid proposition), but they are
+discarded afterwards. This means that they cannot be used as axioms. -/
```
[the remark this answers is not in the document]
fgdorais: Scott's examples resolve this but I would still prefer a sample in the docstring. Since this is new notation, I would like users to have an example to copy-paste rather than think about usage in basic cases.

Std/Data/HashMap/Basic.lean
```diff
@@ -369,3 +370,9 @@ def ofListWith [BEq α] [Hashable α] (l : List (α × β)) (f : β → β → 
     match m.find? p.1 with
     | none   => m.insert p.1 p.2
     | some v => m.insert p.1 <| f v p.2
+
+@[simp] proof_wanted Std.HashMap.empty_find? [BEq α] [Hashable α] {a : α} :
+    (∅ : HashMap α β).find? a = none
+
+proof_wanted insert_find? [BEq α] [Hashable α] (m : HashMap α β) (a a' : α) (b : β) :
+    (m.insert a b).find? a' = if a' == a then some b else m.find? a'
```
digama0, line 378: These should be in `Std.Data.HashMap.Lemmas`

semorrison: Fixed.

joehendrix (approved):

PR #406, opened 2023-11-29
chore: moreServerOptions instead of moreServerArgs

semorrison:
This is a PR to the `bump/v4.4.0`, updating the `lakefile.lean` to account for a deprecated option.

It is slightly clunky that we need the `.map (·.asCliArg)` step to translate from `moreServerOptions` to `moreLeanArgs`.

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -3,14 +3,14 @@ import Lake
 open Lake DSL
 
 package std where
-  moreLeanArgs := moreServerArgs.push "-DwarningAsError=true"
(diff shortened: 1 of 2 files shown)


semorrison:
@tydeu, could you please review?

digama0:
The new version of lake supports `leanOptions`, which means we no longer have to do this copy-paste-paste-map-asCliArg job anymore. I've updated the PR accordingly, and mathlib will also want to do the same thing.

tydeu (approved):

PR #407, opened 2023-11-29
fix: Use `MLList.uncons?` to avoid stack overflow

digama0:
Fixes an issue [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/MLList.20crash/near/404755610).


semorrison:
The Mathlib tests for `exact?` / `rw?` etc that exercise `MLList` still run with this change, so looks good to me! I've added a test file.

semorrison (approved):

fgdorais (approved):

PR #408, opened 2023-11-30
fix: DiscrTreeCache.mk init argument should be lazy

semorrison:
This is causing a terrible problem at Mathlib: the `exact?` and `rw?` caches are being memory mapped at every single file that imports these tactics, rather than just the ones that use these tactics.

Moreover, after nightly-2023-11-29, this causes a build failure on every file if that memory mapped cache has the wrong git hash. The "rebuild-on-failure" mode in this PR means that in the event a memory mapped cache is rejected as invalid, we (silently) rebuilt the cache.

### DIFF
--- a/Std/Util/Cache.lean
+++ b/Std/Util/Cache.lean
@@ -128,7 +128,7 @@ from a function that returns a collection of keys and values for each declaratio
 def DiscrTreeCache.mk [BEq α] (profilingName : String)
     (processDecl : Name → ConstantInfo → MetaM (Array (Array DiscrTree.Key × α)))
     (post? : Option (Array α → Array α) := none)
-    (init : Option (DiscrTree α) := none) :
+    (init : Option (MetaM (DiscrTree α)) := none) :
     IO (DiscrTreeCache α) :=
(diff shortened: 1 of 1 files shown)


Std/Util/Cache.lean
```diff
@@ -88,19 +88,35 @@ structure DeclCache (α : Type) where mk' ::
...
+If `pre` is empty, or it fails,
+the cached structure `α` is initialized with `empty`,
+and then `addLibraryDecl` is called for every imported constant.
 After all imported constants have been added, we run `post`.
+Finally, the result is cached.
+
 When `get` is called, `addDecl` is also called for every constant in the current file.
 -/
-def DeclCache.mk (profilingName : String) (empty : α)
+def DeclCache.mk (profilingName : String) (pre : Option (MetaM α)) (empty : α)
```
digama0, line 103:
```suggestion
def DeclCache.mk (profilingName : String) (pre : MetaM α := failure) (empty : α)
```

semorrison: Thanks.

PR #412, opened 2023-12-01
feat: Add proof to proof_wanted theorems of Array get?_size and get?_push

casavaca:


Std/Data/Array/Lemmas.lean
```diff
@@ -85,16 +85,24 @@ theorem get!_eq_getD [Inhabited α] (a : Array α) : a.get! n = a.getD n default
...
 
-@[simp] proof_wanted get?_size {a : Array α} : a[a.size]? = none
+theorem get?_push {a : Array α} : (a.push x)[i]? = if i = a.size then some x else a[i]? := by
+  split
+  . next heq => rw [heq, getElem?_pos, get_push_eq]
+  · next hne => simp only [getElem?, size_push]
+                split <;> split <;> try simp only [*, get_push_lt]
+                · next p q => have tmp := Nat.le_of_lt_succ p
+                              apply Or.elim (Nat.eq_or_lt_of_le tmp) <;> assumption
+                · next p q => exact p (Nat.lt.step q)
```
semorrison, line 102: Please just two extra space when adding indenting: no need to line up with anything on the previous line.

semorrison: (Newlines immediately after `=>` are also good.)

nomeata: Is that really a good simp lemma, with the if-then-else on the other side? It kinda blows up the term.

My (incomplete, tainted by Isabelle's simp) intuition is that you'd want the conditional two lemmas above as simp lemmas, if any at all?

casavaca: Here, trichotomy of Nat is needed, but it's tricky to derive that from very limited theorems from lean4 itself.
Thus, I just called `split` all the way and then looked for contradictions.

Only split / simp only / exact / rw are used. So, it won't blow up the term, right?

nomeata: Sorry, I wasn't clear, I'm wondering if the `@[simp]` annotation is a good default

casavaca: I see. I have no idea about that.

nomeata: In that case maybe just leave it out in this PR

joehendrix (approved):

PR #413, opened 2023-12-01
chore: 'lemma' command elaborates as 'theorem' but gives warning + codeaction to use 'theorem' instead

jcommelin:


Std/Tactic/Basic.lean
```diff
@@ -162,3 +162,19 @@ example (P : (Nat → Nat) → Prop) : P (fun n => n - n) := by
 -/
 macro (name := Conv.equals) "equals " t:term " => " tac:tacticSeq : conv =>
   `(conv| tactic => show (_ = $t); next => $tac)
+
+/-- `lemma` is not supported, please use `theorem` or import `Mathlib.Tactic.Basic` -/
+syntax (name := lemma) declModifiers
+  group("lemma " declId ppIndent(declSig) declVal Parser.Command.terminationSuffix) : command
+
+/-- `lemma` is not supported, please use `theorem` or import `Mathlib.Tactic.Basic` -/
+@[command_elab «lemma»] def elabLemma : CommandElab := fun stx => do
```
eric-wieser, line 171: Should we do something like
```suggestion
@[command_elab «lemma»] def elabLemma (allow : Bool := false) : CommandElab := fun stx => do
```
so that in mathlib we can just call the same elaborator, with the warning code path skipped?

jcommelin: I made an attempt at implementing this.

Std/Tactic/Basic.lean
```diff
@@ -162,3 +162,27 @@ example (P : (Nat → Nat) → Prop) : P (fun n => n - n) := by
 -/
 macro (name := Conv.equals) "equals " t:term " => " tac:tacticSeq : conv =>
   `(conv| tactic => show (_ = $t); next => $tac)
+
+/-- `lemma` is not supported, please use `theorem` or import `Mathlib.Tactic.Basic` -/
```
fgdorais, line 166: Why should something in Std instruct to use Mathlib?

Std/Tactic/Basic.lean
```diff
@@ -162,3 +162,27 @@ example (P : (Nat → Nat) → Prop) : P (fun n => n - n) := by
...
+/-- Elaborator for the `lemma` command, with a boolean flag to control
+whether the command emits a warning and code action
+instructing the user to use `theorem` instead.-/
+def elabLemma' (allow : Bool) : CommandElab := fun stx => do
+  if !allow then
+    let lemmaStx := stx[1][0]
+    Elab.Command.liftTermElabM <|
+      Std.Tactic.TryThis.addSuggestion lemmaStx { suggestion := "theorem" }
+    logWarningAt lemmaStx
+      "`lemma` is not supported, please use `theorem` or import `Mathlib.Tactic.Basic`"
```
fgdorais, line 179: Same but more important since this is not just documentation.

fgdorais (commented):
I'm very skeptical about this. What's the motivation?

Std/Tactic/Basic.lean
eric-wieser: The idea here is:
* Some users will type `lemma` from muscle memory, and the first half of the warning tells them that they should be writing `theorem` instead
* Some of these users might have just forgotten to import mathlib (where `lemma` is encouraged), and pointing out that this is allowed in mathlib is helpful to them.

Of course it would all be simpler if mathlib didn't have `lemma` either, but that ship has long since sailed.

digama0: > Some of these users might have just forgotten to import mathlib (where lemma is encouraged), and pointing out that this is allowed in mathlib is helpful to them.

For the "forgot to import mathlib" case, it would be more helpful to have this in Init rather than Std.

fgdorais: I think the confusion for me is that it makes it sound like Std is against using lemma. If Mathlib.Init doesn't work then I would rather have the plain implementation of lemma in Std for people to use, and simply have a style rule not to use it internally.

digama0: Std *is* against using `lemma`, or more precisely it does not want the syntax to diverge from core and if you want lemma everywhere you should take it upstream. Non-implementing `lemma`  here is a way to improve the error message while still not changing the official grammar or causing lean code in the wild to accrue uses of `lemma`. 

`rustc` does this kind of thing a lot, for example if you write function implementations inside a `struct` it will parse the syntax even though it is not legal, specifically so it can tell you that unlike languages like C# and C++, function implementations go in a separate impl block. [(demo)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b5dfe53c49452c52936cb1ee682442b2)

Std/Tactic/Basic.lean
digama0:
```suggestion
      "`lemma` is not supported, please use `theorem` instead.\n{""
      }note: Mathlib defines a `lemma` command, did you mean to `import Mathlib.Tactic.Basic`?"
```

Std/Tactic/Basic.lean
```diff
@@ -162,3 +162,27 @@ example (P : (Nat → Nat) → Prop) : P (fun n => n - n) := by
...
+
+/-- Elaborator for the `lemma` command, with a boolean flag to control
+whether the command emits a warning and code action
+instructing the user to use `theorem` instead.-/
+def elabLemma' (allow : Bool) : CommandElab := fun stx => do
+  if !allow then
+    let lemmaStx := stx[1][0]
+    Elab.Command.liftTermElabM <|
+      Std.Tactic.TryThis.addSuggestion lemmaStx { suggestion := "theorem" }
+    logWarningAt lemmaStx
```
digama0, line 178:
```suggestion
    logErrorAt lemmaStx
```

Std/Tactic/Basic.lean
fgdorais: There's a miscommunication here. I was trying to say that: (1) Std doesn't use `lemma` and (b) Std doesn't have an opinion whether or not `lemma` should be used as a synonym for `theorem`.

My objection is that: (1) this PR makes it seem that (b) is not true and (2) it makes it harder for users to implement `lemma` in their own project.

In my opinion, the best solution is for Std to have no handling for `lemma` at all (satus quo). However, this is a "paper cut" and, as such, the next best idea is probably to implement `lemma` as a convenient synonym for `theorem` for downstream users but _not_ use it internally.

digama0: > (2) it makes it harder for users to implement `lemma` in their own project.

How do you figure? AFAIK if the user has an implementation of `lemma` like mathlib's then it will continue to work in the same way. The std non-implementation does not prevent downstream implementations. It even makes them slightly easier since it's now just `elabLemma' true`

jcommelin, line 166:
```suggestion
/-- `lemma` is not supported, please use `theorem` instead -/
```

semorrison:
I've set this back to `awaiting-review`. I am keen to have this in.

Std/Tactic/Basic.lean
```diff
@@ -162,3 +162,28 @@ example (P : (Nat → Nat) → Prop) : P (fun n => n - n) := by
...
+      }note: Mathlib defines a `lemma` command, did you mean to `import Mathlib.Tactic.Basic`?"
+  let out ← Elab.liftMacroM <| do
+    let stx := stx.modifyArg 1 fun stx =>
+      let stx := stx.modifyArg 0 (mkAtomFrom · "theorem" (canonical := true))
+      stx.setKind ``Parser.Command.theorem
+    pure <| stx.setKind ``Parser.Command.declaration
+  Elab.Command.elabCommand out
+
+/-- `lemma` is not supported, please use `theorem` or import `Mathlib.Tactic.Basic` -/
+@[command_elab «lemma»] def elabLemma : CommandElab := elabLemma' false
```
eric-wieser, line 189: This docstring isn't the same as the one on the `syntax`; do we actually need both anyway? Which one is used by vscode?

fgdorais:
Superseded by #625

jcommelin:
Since #625 has been merged, this can now certainly be closed.

PR #415, opened 2023-12-01
perf: use a faster implementation of BitVec.ofInt

eric-wieser:
Maybe this will be irrelevant if a GMP overload appears for this...

### DIFF
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -66,12 +66,6 @@ theorem isLt (x : BitVec w) : x.toNat < 2^w := x.toFin.isLt
 /-- Return most-significant bit in bitvector. -/
 @[inline] protected def msb (a : BitVec n) : Bool := getMsb a 0
 
-/-- The `BitVec` with value `(2^n + (i mod 2^n)) mod 2^n`.  -/
-protected def ofInt (n : Nat) (i : Int) : BitVec n :=
-  match i with
-  | Int.ofNat a => .ofNat n a
-  | Int.negSucc a => ~~~.ofNat n a
-
 /-- Interpret the bitvector as an integer stored in two's complement form. -/
 protected def toInt (a : BitVec n) : Int :=
   if a.msb then Int.ofNat a.toNat - Int.ofNat (2^n) else a.toNat
@@ -318,6 +312,12 @@ SMT-Lib name: `bvnot`.
 -/
 protected def not (x : BitVec n) : BitVec n := -(x + .ofNat n 1)
 instance : Complement (BitVec w) := ⟨.not⟩
+
+/-- The `BitVec` with value `(2^n + (i mod 2^n)) mod 2^n`.  -/
+protected def ofInt (n : Nat) (i : Int) : BitVec n :=
+  match i with
+  | Int.ofNat a => .ofNat n a
+  | Int.negSucc a => ~~~.ofNat n a
 
 /--
 Left shift for bit vectors. The low bits are filled with zeros. As a numeric operation, this is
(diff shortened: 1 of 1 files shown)


eric-wieser:
awaiting-review

Std/Data/BitVec/Basic.lean
```diff
@@ -316,9 +310,18 @@ Bitwise NOT for bit vectors.
 ```
 SMT-Lib name: `bvnot`.
 -/
-protected def not (x : BitVec n) : BitVec n := -(x + .ofNat n 1)
+protected def not (x : BitVec n) : BitVec n :=
+  .ofFin ⟨(1 <<< n).pred,
+    n.one_shiftLeft.symm ▸
+      (Nat.pred_lt <| Nat.ne_of_gt <| Nat.pos_pow_of_pos _ <| Nat.zero_lt_succ _)⟩ ^^^ x
```
joehendrix, line 316: Seems like the first term could be made into a separate definition -- maybe `ones : BitVec n`?

eric-wieser: My hesitation there is that we now have three ways to spell the vector of all ones; `-1`, `~~~0`, and `ones`.

joehendrix: That seems ok to me.  It has good algebraic properties and couldn't be expressed with the direct property you wanted in the code above.

Another possibility would be to also modify `neg` (and introduce `one`) so that `-1` could be used in place of the code above.

Std/Data/BitVec/Basic.lean
```diff
@@ -316,9 +310,18 @@ Bitwise NOT for bit vectors.
...
 instance : Complement (BitVec w) := ⟨.not⟩
 
+/-- The `BitVec` with value `(2^n + (i mod 2^n)) mod 2^n`.  -/
+protected def ofInt (n : Nat) (i : Int) : BitVec n :=
+  match i with
+  | Int.ofNat a => .ofNat n a
+  | Int.negSucc a => ~~~.ofNat n a
```
alexkeizer, line 323: It's not immediately obvious to me that this new definition is indeed faster.
The old version has some extraneous modulus, but could be optimized to `.ofFin \<2^n - 1 - a, _\>`, since `2^n - 1 - a < 2^n` holds for all `n` and all `a`, this is is two subtractions.
Your version is one bitshift, one subtraction (by one) and one xor.

Do we have a good story for micro-benchmarks in Lean? I'd love to have some clear numbers on the relative speeds here

Std/Data/BitVec/Basic.lean
```diff
@@ -316,9 +310,18 @@ Bitwise NOT for bit vectors.
...
 SMT-Lib name: `bvnot`.
 -/
-protected def not (x : BitVec n) : BitVec n := -(x + .ofNat n 1)
+protected def not (x : BitVec n) : BitVec n :=
+  .ofFin ⟨(1 <<< n).pred,
+    n.one_shiftLeft.symm ▸
+      (Nat.pred_lt <| Nat.ne_of_gt <| Nat.pos_pow_of_pos _ <| Nat.zero_lt_succ _)⟩ ^^^ x
 instance : Complement (BitVec w) := ⟨.not⟩
 
+/-- The `BitVec` with value `(2^n + (i mod 2^n)) mod 2^n`.  -/
```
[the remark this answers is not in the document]
joehendrix: It doesn't simplify when `i < -2^n`.  Maybe could be reworded as "bit vector congruent to `i` modulo `2^n`".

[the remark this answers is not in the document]
tobiasgrosser: This comment seems to be resolved.

Std/Data/BitVec/Basic.lean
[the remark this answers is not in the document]
alexkeizer: Let's go forward with this suggestion, to have the `ones` term as a let. This PR is blocking https://github.com/leanprover-community/mathlib4/pull/8711 so I would prefer we merge it sooner rather than later.
If it turns out later that we find another use-case for this `ones` term, we can always extract it out to a separate def then.

PR #417, opened 2023-12-02
Pattern-based unfolding tactic

0art0:
This is a tactic that allows targeted unfolding of expressions in the proof state by specifying patterns and occurrences (an idea that is due to Yaël Dillies).

The file `Utils.lean` contains utilities that can be re-used for other pattern-based tactics.

This is coupled with a point-and-click widget (which will be in a different PR to `Mathlib`) for automatically generating the tactic call with the appropriate configuration from the infoview selection.

Co-authored-by: Jovan Gerbscheid <jovan.gerbscheid@gmail.com>

---


0art0:
This PR is ready for review. In a set of future PRs, we would like to introduce the following additional features:

- [ ] Using the same kinds of brackets in both `hypLoc` and `goalLoc`
- [ ] If no location is specified, set it to `at ⊢` by default
- [ ] Recursive unfolding, including unfolding to a specified depth

JovanGerb:
To be honest, I'm not yet convinced that the new location syntax is needed, because if you want to do something at specific locations in multiple different hypotheses, then you may as well do it with separate tactic calls.

JovanGerb:
I have code that allows these patterns to also include bound variables. Would it be better to add it to this PR or to put it in a later PR?

I also think that the code in this PR should go in 2 separate folders: one with the code related to patterns, and one with the code for the unfolding tactic (which actually is just one file). Because the code for the patterns is to be used by more future tactics, and the unfold tactic is just here as a demonstration of the usefulness of the patterns.

0art0:
> To be honest, I'm not yet convinced that the new location syntax is needed, because if you want to do something at specific locations in multiple different hypotheses, then you may as well do it with separate tactic calls.

I think I will remove it then.

0art0:
I have modified the `locs` syntax to something simpler that targets a single location. I think I am much happier with it in this form. I have updated the checklist from my previous comment erasing the items related to the `locs` syntax.

0art0:
The `replaceLocalLetDefEq`, `defineAfter` and `replaceLocalLetDecl` definitions in `Pattern/Location` are functions that may be of general utility, so I could also try to PR them to Lean core (say into the file `Lean/Meta/Tactic/Replace.lean`), if that would be better.

0art0:
> I have code that allows these patterns to also include bound variables. Would it be better to add it to this PR or to put it in a later PR?

I am fine either way. Perhaps the reviewers could weigh in here.

> I also think that the code in this PR should go in 2 separate folders: one with the code related to patterns, and one with the code for the unfolding tactic (which actually is just one file). Because the code for the patterns is to be used by more future tactics, and the unfold tactic is just here as a demonstration of the usefulness of the patterns.

Thanks for the suggestion, I've restructured the files according to this.

Std/Tactic/Pattern/Utils.lean
```diff
@@ -0,0 +1,163 @@
...
+-/
+open Parser.Tactic Conv
+
+/-- Refer to a set of subexpression by specifying a pattern.
+
+For example, if hypothesis `h` says that `1 + (2 + 3) = 1 + (2 + 3)`, then
+`(occs := 2 3) _ + _ at h` refers to the two expression `2 + 3`,
+because it first skips `1 + (2 + 3)`, and matches with `2 + 3`,
+which instantiates the pattern to be `2 + 3`, so the next match is the
+second instance of `2 + 3`. -/
```
semorrison, line 34: It is rather unhelpful here that you use `2` and `3` here to mean different things! This would be immensely clearer if you used `17 + (37 + 19)` or something.

Also, the phrase "the two expression `2 + 3`" doesn't make sense to me. Do you mean "the two occurrences of the expression `2 + 3`"?

Std/Tactic/Pattern/Utils.lean
```diff
@@ -0,0 +1,163 @@
...
+      visitChildren ()
+    else if e.toHeadIndex != pHeadIdx || e.headNumArgs != pNumArgs then
+      visitChildren ()
+    else if (← isDefEq e pattern) then
+      let i ← get
+      set (i+1)
+      if p = position then
+        return ()
+      else
+        visitChildren ()
```
semorrison, line 123: This code in `kabstract` has been updated in the meantime to give better behaviour!

Std/Tactic/PatternTactics/Unfold.lean
```diff
@@ -0,0 +1,76 @@
...
+The targeted sub-expression is selected using a pattern.
+
+example use case:
+```
+@[irreducible] def f (n : Nat) := n + 1
+example : ∀ n : Nat, n + 1 = f n := by
+  unfold' f _ at ⊢ --∀ (n : Nat), n + 1 = n + 1
+  intro n
+  rfl
+```
```
semorrison, line 25: Examples should explain use of `occs`.

fgdorais:
Looks like this wasn't switched back to `awaiting-review` after the last updates. It could use another update to adapt Std -> Batteries.

JovanGerb:
I cleaned up this PR a while ago, but I didn't put it on awaiting-review, because I wasn't sure whether this should really be included into batteries. It seems to me like the recent (last week merged) mathlib PR (https://github.com/leanprover-community/mathlib4/pull/12016#event-13586997181) for rewriting things into unfolded form is an improvement over this PR.

fgdorais:
Thanks for the quick response @JovanGerb ! Should this PR be closed then?

JovanGerb:
I think so. The main observation making this PR redundant is that if you want to do a rewrite which holds definitionally, then you can do this via `rw [show old = new from rfl]`.

fgdorais:
That makes sense, thanks! I'm removing `awaiting-review` and adding `duplicate`. That tag is not exactly right but it's the closest fit. Anyone can ask for this tag to be removed, ping me or a maintainer if that's the case.

PR #420, opened 2023-12-06
feat: avoid termination_by', introduce WellFounded.wrap

nomeata:
I’d like to remove support for the `termination_by'` annotation. Until
https://github.com/leanprover/std4/pull/371 it wasn't used anywhere else in
lean, std, mathlib, so this PR removes the single use of it.

It does so using the pattern that can be used to replace uses of
`termination_by'`: Using `WellFounded.wrap` one can indicate an explicit
`WellFounded` relation to use.

This also un-privates the `wfRel` instance for `WellFoundedRelation { val // Acc r val }`
which looks useful enough.

So this PR uses that pattern to avoid the use of `termination_by'` here,
and at the same time provides the necessary definitions for others, so
when Lean drops support for `termination_by'`
(https://github.com/leanprover/lean4/pull/3033), we can tell users how
migrate.

### DIFF
--- a/Std/WF.lean
+++ b/Std/WF.lean
@@ -41,7 +41,7 @@ def log2p1 : Nat → Nat := -- works!
 
 namespace Acc
 
-private local instance wfRel {r : α → α → Prop} : WellFoundedRelation { val // Acc r val } where
+instance wfRel {r : α → α → Prop} : WellFoundedRelation { val // Acc r val } where
   rel := InvImage r (·.1)
   wf  := ⟨fun ac => InvImage.accessible _ ac.2⟩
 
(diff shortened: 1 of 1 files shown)


digama0:
We also need a way to get a well founded relation out of a value of type `WellFoundedRelation A`  (i.e. the thing that `termination_by'` is expecting).

nomeata:
Would that be `wfr.wf.wrap`?

digama0:
Yes that works. Next challenge: how to get the type inference working correctly when there are multiple arguments involved?

digama0:
By the way, the `wfRel` instance in this file is very similar to `Wrapped`, but it only requires `Acc r x` instead of `WellFounded r`. (It is also private so that we didn't have to do the API work for it.)

nomeata:
> Yes that works. Next challenge: how to get the type inference working correctly when there are multiple arguments involved?

What example do you have in mind? I'd expect once the `WellFounded` is given, all is known?

Note that users will have to manually select and tuple the arguments they care about. But the target audience (users of `termination_by'`) should already know how do deal with that without too much handholding, I'd think?

nomeata:
> By the way, the `wfRel` instance in this file is very similar to `Wrapped`, but it only requires `Acc r x` instead of `WellFounded r`. (It is also private so that we didn't have to do the API work for it.)

Interesting, thanks for the pointer. So an alternative would be to make that not private and add
```
def wrap {α : Sort u} {r : α → α → Prop} (h : WellFounded r) (x : α) : {x : α // Acc r x} :=
  ⟨_, h.apply x⟩
``
which is nice, because there are no additional definitions with unused arguments to write and explain, and the instance looks kinda useful anyways.

I have updated the PR accordingly, WDYT?

nomeata:
> Lean needs quite a lot of handholding to infer dependent sigma types.

I’d expect that the user has to define the `Wellfounded (PSigma …)` type explicitly ahead of times anyways, so by the time they write `wf.wrap …`, the type is well known? [Quite like we see in the wild already with `termination_by'`](https://github.com/leanprover-community/con-nf/blob/688a8ab014bafbef47553e81e50de7db3f1d080e/ConNF/Foa/Complete/HypAction.lean#L32-L52).

PR #421, opened 2023-12-06
Incremental Library Search

joehendrix:
This is a large PR that builds upon #343 to rename the Mathlib specific cache path and improve performance of library initialization when used without a cache.  It does this primarily by introducing a lazy discriminator tree that defers population of tree branches from initialization to first lookup.

Testing on OSX is currently showing that initializing the tree using Mathlib takes under 2 seconds and a simply query adds an additional 2-3 seconds.  It would be good to get additional information more machines (particularly Windows), but I think the timings are low enough that most users will not need caching.

### DIFF
--- a/Std/Lean/Meta/LazyDiscrTree.lean
+++ b/Std/Lean/Meta/LazyDiscrTree.lean
@@ -312,7 +312,7 @@ end MatchClone
 export MatchClone (Key Key.const)
 
 /--
-An unprocessed entry in the lazy discriminator tree.
+An unprocessed entry in the lazy discrimination tree.
 -/
 private abbrev LazyEntry α := Array Expr × ((LocalContext × LocalInstances) × α)
 
--- a/Std/Lean/Meta/LazyDiscrTree.lean
+++ b/Std/Lean/Meta/LazyDiscrTree.lean
@@ -16,7 +16,7 @@ lazy initialization strategy.
 
 The discrimination tree can be created through
 `createImportedEnvironment`. This creates a discrimination tree from all
-imported modules in an environment using a callback that proviees the
+imported modules in an environment using a callback that provides the
 entries as `InitEntry` values.
 
 The function `getMatch` can be used to get the values that match the
--- a/Std/Lean/Meta/LazyDiscrTree.lean
+++ b/Std/Lean/Meta/LazyDiscrTree.lean
@@ -140,7 +140,6 @@ private def isOffset (fName : Name) (e : Expr) : MetaM Bool := do
 private def shouldAddAsStar (fName : Name) (e : Expr) : MetaM Bool := do
   isOffset fName e
 
-
 /--
   Try to eliminate loose bound variables by performing beta-reduction.
   We use this method when processing terms in discrimination trees.
@@ -446,34 +445,34 @@ private partial def evalLazyEntries (config : WhnfCoreConfig)
     (values : Array α) (starIdx : TrieIndex) (children : HashMap Key TrieIndex)
     (entries : Array (LazyEntry α)) :
     MatchM α (Array α × TrieIndex × HashMap Key TrieIndex) := do
-  if h : entries.size = 0 then
-    pure (values, starIdx, children)
-  else
-    let p : entries.size - 1 < entries.size := Nat.sub_lt (Nat.pos_of_ne_zero h) (Nat.le_refl _)
-    let (todo, lctx, v) := entries[entries.size - 1]
-    let entries := entries.pop
-    if todo.isEmpty then
-      let values := values.push v
-      evalLazyEntries config values starIdx children entries
-    else
-      let e    := todo.back
-      let todo := todo.pop
-      let (k, todo) ← withLCtx lctx.1 lctx.2 $ pushArgs false todo e config
-      if k == .star then
-        if starIdx = 0 then
-          let starIdx ← newTrie (todo, lctx, v)
-          evalLazyEntries config values starIdx children entries
+  let rec iter values starIdx children (i : Nat) : MatchM α _ := do
+        if p : i < entries.size then
+          let (todo, lctx, v) := entries[i]
+          if todo.isEmpty then
+            let values := values.push v
+            iter values starIdx children (i+1)
+          else
+            let e    := todo.back
+            let todo := todo.pop
+            let (k, todo) ← withLCtx lctx.1 lctx.2 $ pushArgs false todo e config
+            if k == .star then
+              if starIdx = 0 then
+                let starIdx ← newTrie (todo, lctx, v)
+                iter values starIdx children (i+1)
+              else
+                addLazyEntryToTrie starIdx (todo, lctx, v)
+                iter values starIdx children (i+1)
+            else
+              match children.find? k with
+              | none =>
+                let children := children.insert k (← newTrie (todo, lctx, v))
+                iter values starIdx children (i+1)
+              | some idx =>
+                addLazyEntryToTrie idx (todo, lctx, v)
+                iter values starIdx children (i+1)
         else
-          addLazyEntryToTrie starIdx (todo, lctx, v)
-          evalLazyEntries config values starIdx children entries
-      else
-        match children.find? k with
-        | none =>
-          let children := children.insert k (← newTrie (todo, lctx, v))
-          evalLazyEntries config values starIdx children entries
-        | some idx =>
-          addLazyEntryToTrie idx (todo, lctx, v)
-          evalLazyEntries config values starIdx children entries
+          pure (values, starIdx, children)
+  iter values starIdx children 0
 
 private def evalNode (c : TrieIndex) :
     MatchM α (Array α × TrieIndex × HashMap Key TrieIndex) := do
@@ -521,7 +520,6 @@ private partial def getMatchLoop (todo : Array Expr) (c : TrieIndex) (result : M
   else
     let e     := todo.back
     let todo  := todo.pop
-    let (k, args) ← MatchClone.getMatchKeyArgs e (root := false) (←read)
     /- We must always visit `Key.star` edges since they are wildcards.
         Thus, `todo` is not used linearly when there is `Key.star` edge
         and there is an edge for `k` and `k != Key.star`. -/
@@ -535,6 +533,7 @@ private partial def getMatchLoop (todo : Array Expr) (c : TrieIndex) (result : M
       | none   => return result
       | some c => getMatchLoop (todo ++ args) c result
     let result ← visitStar result
+    let (k, args) ← MatchClone.getMatchKeyArgs e (root := false) (←read)
     match k with
     | .star  => return result
     /-
--- a/Std/Lean/Meta/LazyDiscrTree.lean
+++ b/Std/Lean/Meta/LazyDiscrTree.lean
@@ -1,3 +1,9 @@
+/-
+Copyright (c) 2023 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joe Hendrix, Scott Morrison
+-/
+
 import Lean.Meta.DiscrTree
 import Std.Data.Nat.Init.Lemmas
 
--- a/Std/Lean/Meta/LazyDiscrTree.lean
+++ b/Std/Lean/Meta/LazyDiscrTree.lean
@@ -788,7 +788,7 @@ private def createImportedEnvironmentSeq (env : Environment)
               toFlat d tree
   termination_by go _ idx stop => stop - idx
 
-/-- Get the rests of each task and merge using combining function -/
+/-- Get the results of each task and merge using combining function -/
 private def combineGet [Append α] (z : α) (tasks : Array (Task α)) : α :=
   tasks.foldl (fun x t => x ++ t.get) (init := z)
 
--- a/Std/Lean/Meta/LazyDiscrTree.lean
+++ b/Std/Lean/Meta/LazyDiscrTree.lean
@@ -394,17 +394,17 @@ private def initCapacity := 8
 Adds a key and
 -/
 def addEntry' : LazyDiscrTree α → LocalContext × LocalInstances → Key → Array Expr → α →
-    MetaM (LazyDiscrTree α)
-| { config := c, array := a, root := r }, lctx, k, todo, v => withReducible $ do
+    LazyDiscrTree α
+| { config := c, array := a, root := r }, lctx, k, todo, v =>
   let rest := (todo, lctx, v)
   match r.find? k with
   | none =>
     let r := r.insert k a.size
     let a := a.push (.node #[] 0 {} #[rest])
-    pure { config := c, array := a, root := r }
+    { config := c, array := a, root := r }
   | some idx =>
     let a := a.modify idx fun t => t.pushPending rest
-    pure { config := c, array := a, root := r }
+    { config := c, array := a, root := r }
 
 /--
 Get the root key of an expression using the config from the lazy dicriminator tree.
@@ -415,13 +415,21 @@ def rootKey (d:LazyDiscrTree α) (lctx : LocalContext × LocalInstances) (e : Ex
   withReducible $ withLCtx lctx.1 lctx.2 $ pushArgs true todo e d.config
 
 /--
+Get the root key of an expression using the config from the lazy dicriminator tree.
+-/
+def rootKey' (d:LazyDiscrTree α) (e : Expr) :
+    MetaM (Key × Array Expr) :=
+  pushArgs true (Array.mkEmpty initCapacity) e d.config
+
+
+/--
 Adds an association between the given expression.  Free variables are used to
 denote paramters that may be matched against.
 -/
 def addEntry (d : LazyDiscrTree α) (lctx : LocalContext × LocalInstances) (type : Expr) (v : α) :
     MetaM (LazyDiscrTree α) := do
   let (k, todo) ← rootKey d lctx type
-  addEntry' d lctx k todo v
+  pure $ addEntry' d lctx k todo v
 
 private partial def mkPathAux (root : Bool) (todo : Array Expr) (keys : Array Key)
     (config : WhnfCoreConfig) : MetaM (Array Key) := do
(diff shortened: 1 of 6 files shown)


digama0:
> the cacheless approach has the advantage that it only selects candidates in the current environment

I'm not sure I would call that an advantage necessarily. For a search tool it is useful to be made aware of theorems from files you haven't imported.

joehendrix:
> > the cacheless approach has the advantage that it only selects candidates in the current environment
> 
> I'm not sure I would call that an advantage necessarily. For a search tool it is useful to be made aware of theorems from files you haven't imported.

That's a good point that may impact operations other than `exact?` and `apply?`.   I've made a [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Cacheless.20library.20search/near/406442450) since there might be more discussion on this and other changes.

joehendrix:
I finally got back to this and fixed the lint message.  I think this could use some testing on more machines.

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,822 @@
+import Lean.Meta.DiscrTree
+import Std.Data.Nat.Init.Lemmas
+
```
semorrison, line 3: Copyright header and module doc missing.

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,822 @@
...
+namespace Lean.Meta
+
+namespace LazyDiscrTree
+
+namespace MatchClone
+open DiscrTree
+
+private def tmpMVarId : MVarId := { name := `_discr_tree_tmp }
```
semorrison, line 11: Could we have a summary of which parts of this are copy-pasted from Lean and which parts diverge?

I am tempted to suggest that this is a valid use case for `open private`!

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,822 @@
...
+tree.  When first adding patterns to the tree, only the root
+discriminator key is computed and processing the remaining
+terms is deferred until demanded by a match.
+-/
+structure LazyDiscrTree (α : Type) where
+  /-- Configuration for normalization. -/
+  config : Lean.Meta.WhnfCoreConfig := {}
+  /-- Backing array of trie entries.  Should be owned by this trie. -/
+  tries : Array (LazyDiscrTree.Trie α) := #[default]
+  /-- Map from disriminator trie roots to the index. -/
```
semorrison, line 320:
```suggestion
  /-- Map from discriminator trie roots to the index. -/
```

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,822 @@
...
+  (fun (r, d) => (r.toArrayRev, d)) <$> runMatch (getMatchCore d.roots e) d
+
+/--
+Structure for quickly initializing a lazy discrimination tree with a large number
+of elements using concurrent functions for generating entries.
+-/
+private structure PreDiscrTree (α : Type) where
+  /-- Maps keys to index in tries array. -/
+  roots : HashMap Key Nat := {}
+  /-- Lazy entries for root of treie-/
```
semorrison, line 592:
```suggestion
  /-- Lazy entries for root of trie. -/
```

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,822 @@
...
+    (f : Array (LazyEntry α) → Array (LazyEntry α)) : PreDiscrTree α :=
+  let { roots, tries } := d
+  match roots.find? k with
+  | .none =>
+    let roots := roots.insert k tries.size
+    { roots, tries := tries.push (f #[]) }
+  | .some i =>
+    { roots, tries := tries.modify i f }
+
+/-- Add an entries to the pre-discrimination tree.-/
```
semorrison, line 608:
```suggestion
/-- Add an entry to the pre-discrimination tree.-/
```

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,822 @@
...
+            if start < stop then
+              let mname := env.header.moduleNames[start]!
+              let mdata := env.header.moduleData[start]!
+              let tree ← loadImportedModule env act d tree mname mdata
+              go d tree (start+1) stop
+            else
+              toFlat d tree
+  termination_by go _ idx stop => stop - idx
+
+/-- Get the rests of each task and merge using combining function -/
```
semorrison, line 791:
```suggestion
/-- Get the results of each task and merge using combining function -/
```

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,533 @@
...
+open LazyDiscrTree (InitEntry isBlackListed createImportedEnvironment)
+
+namespace DiscrTreeFinder
+
+open System (FilePath)
+
+
+/--
+Once we reach Mathlib, and have `cache` available,
+it will be essential that we load a precomputed cache for `exact?` from a `.olean` file.
```
semorrison, line 71:
```suggestion
we may still want to load a precomputed cache for `exact?` from a `.olean` file.
```

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,533 @@
...
+| .ok (some _) => crossEmoji
+| .ok none => checkEmoji
+
+/--
+Try to solve the goal either by:
+* calling `solveByElim`
+* or applying a library lemma then calling `solveByElim` on the resulting goals.
+
+If it successfully closes the goal, returns `none`.
+Otherwise, it returns `some a`, where `a : Array (MetavarContext × List MVarId)`,
```
semorrison, line 411:
```suggestion
Otherwise, it returns `some a`, where `a : Array (List MVarId × MetavarContext)`,
```

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,533 @@
...
+with an entry for each library lemma which was successfully applied,
+containing the metavariable context after the application, and a list of the subsidiary goals.
```
semorrison, line 413:
```suggestion
containing a list of the subsidiary goals, and the metavariable context after the application.
```

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,533 @@
...
+  | .none => pure lem
+  | .mp => mapForallTelescope (fun e => mkAppM ``Iff.mp #[e]) lem
+  | .mpr => mapForallTelescope (fun e => mkAppM ``Iff.mpr #[e]) lem
+
+/--
+A library search candidate using symmetry includes the goal to solve, the metavar
+context for that goal, and the name and orientation of a rule to try using with goal.
+-/
+@[reducible]
+def Candidate :=  (MVarId × MetavarContext) × (Name × DeclMod)
```
semorrison, line 218: I'm confused why the `MVarId × MetavarContext` is here. These are always the same, as the second factor varies.

semorrison: Oh, I see, because of the way you're handling `symm` now, half the time it will be the `symm` of the goal.

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,533 @@
...
+    pure imports
+
+end IncDiscrTreeFinder
+
+private unsafe def mkImportFinder : IO CandidateFinder := do
+  let path ← DiscrTreeFinder.cachePath
+  if ← path.pathExists then
+    let (imports, _) ← unpickle (DiscrTree (Name × DeclMod)) path
+    -- `DiscrTree.getMatch` returns results in batches, with more specific lemmas coming later.
+    -- Hence we reverse this list, so we try out more specific lemmas earlier.
```
semorrison, line 185: This comment is stray?

Std/Lean/Meta/LazyDiscrTree.lean
joehendrix: The `MatchClone` namespace only contains copied definitions.  I've added a comment.

I'm trying to minimize forward dependencies if discrimination trees for Simp change so I've also copied over Key.

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+/-!
+# Lazy Discrimination Tree
+
+This file defines a new type of discrimination tree optimized for
+rapidly population of imported modules for use in tactics.  It uses a
+lazy initialization strategy.
+
+The discrimination tree can be created through
+`createImportedEnvironment`. This creates a discrimination tree from all
+imported modules in an environment using a callback that proviees the
```
semorrison, line 19:
```suggestion
imported modules in an environment using a callback that provides the
```

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+private abbrev getMatchKeyArgs (e : Expr) (root : Bool) (config : WhnfCoreConfig) :
+    MetaM (Key × Array Expr) :=
+  getKeyArgs e (isMatch := true) (root := root) (config := config)
+
+end MatchClone
+
+export MatchClone (Key Key.const)
+
+/--
+An unprocessed entry in the lazy discriminator tree.
```
semorrison, line 313:
```suggestion
An unprocessed entry in the lazy discrimination tree.
```

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+-/
+private abbrev LazyEntry α := Array Expr × ((LocalContext × LocalInstances) × α)
+
+/--
+Index identifying trie in a discriminator tree.
```
semorrison, line 318:
```suggestion
Index identifying trie in a discrimination tree.
```

Std/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+/-- Create a new trie with the given lazy entry. -/
+private def newTrie [Monad m] [MonadState (Array (Trie α)) m] (e : LazyEntry α) : m TrieIndex := do
+  modifyGet fun a => let sz := a.size; (sz, a.push (.node #[] 0 {} #[e]))
+
+/-- Add a lazy entry to an existing trie. -/
+private def addLazyEntryToTrie (i:TrieIndex) (e : LazyEntry α) : MatchM α Unit :=
+  modify (·.modify i (·.pushPending e))
+
+/--
+This evaluates all lazy entries in a trie and adds updates values starIdx and children
```
semorrison, line 485:
```suggestion
This evaluates all lazy entries in a trie and updates `values`, `starIdx`, and `children`
```

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,531 @@
...
+      return tree
+
+/--
+Constructs an discriminator tree from the current environment.
+-/
+def buildImportCache (config : WhnfCoreConfig) : MetaM (DiscrTree (Name × DeclMod)) := do
+  let profilingName := "apply?: init cache"
+  -- Sort so lemmas with longest names come first.
+  let post (A : Array (Name × DeclMod)) :=
+        A.map (fun (n, m) => (n.toString.length, n, m)) |>.qsort (fun p q => p.1 > q.1) |>.map (·.2)
```
semorrison, line 115: There's no analogue of this in the no-cache-file version, right?

It's not essential, but this does seem to be a genuinely helpful way of organising results. We can revisit later.

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,531 @@
...
+-/
+private def constantsPerTask : Nat := 6500
+
+private def addImport (name : Name) (constInfo : ConstantInfo) :
+    MetaM (Array (InitEntry (Name × DeclMod))) :=
+  forallTelescope constInfo.type fun _ type => do
+    let e ← InitEntry.fromExpr type (name, DeclMod.none)
+    let a := #[e]
+    let biff := e.key == .const ``Iff 2
+    if biff then
```
semorrison, line 156:
```suggestion
    if e.key == .const ``Iff 2 then
```

Std/Tactic/LibrarySearch.lean
semorrison: You resolved, but I'm still confused here!

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,531 @@
...
+    let (imports, _) ← unpickle (DiscrTree (Name × DeclMod)) path
+    -- `DiscrTree.getMatch` returns results in batches, with more specific lemmas coming later.
+    -- Hence we reverse this list, so we try out more specific lemmas earlier.
+    pure <| DiscrTreeFinder.mkImportFinder {} imports
+  else do
+    IncDiscrTreeFinder.mkImportFinder
+
+
+/--
+Retrieve the current current of lemmas.
```
semorrison, line 190:
```suggestion
The preferred candidate finding function.
```
?

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,531 @@
...
+and current results returned.
+-/
+private initialize abortSpeculationId : InternalExceptionId ←
+  registerInternalExceptionId `Std.Tactic.LibrarySearch.abortSpeculation
+
+/--
+Called to abort speculative execution in library search.
+-/
+def abortSpeculation [MonadExcept Exception m] : m α :=
+  throw (Exception.internal abortSpeculationId {})
```
semorrison, line 323: I can't help but be a little sad to be doing exception handling here rather than just using `MLList`, but no big deal. :-)

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,531 @@
...
+
+/--
+Sequentially invokes a tactic `act` on each value in candidates on the current state.
+
+The tactic `act` should return a list of meta-variables that still need to be resolved.
+If this list is empty, then no variables remain to be solved, and `tryOnEach` returns
+`none` with the environment set so each goal is resolved.
+
+If the action throws an internal exception with the `abortSpeculationId` id then
+further computation is stoped and intermediate results returned. If any other
```
semorrison, line 338:
```suggestion
further computation is stopped and intermediate results returned. If any other
```

test/library_search/basic.lean
```diff
@@ -0,0 +1,224 @@
...
+
+/-- info: Try this: exact Nat.add_pos_left ha b -/
+#guard_msgs in
+example (a b : Nat) (ha : 0 < a) (_hb : 0 < b) : 0 < a + b := by std_apply?
+
+/-- info: Try this: exact Nat.add_pos_left ha b -/
+#guard_msgs in
+-- Verify that if maxHeartbeats is 0 we don't stop immediately.
+set_option maxHeartbeats 0 in
+example (a b : Nat) (ha : 0 < a) (_hb : 0 < b) : 0 < a + b := by std_apply?
```
semorrison, line 89: While I remember that deterministic testing of the maxHeartbeats feature was hard to arrange in the original version, I presume that you've verified observationally at least once that the new implementation of this feature works. :-)

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,531 @@
...
+
+private def librarySearchEmoji : Except ε (Option α) → String
+| .error _ => bombEmoji
+| .ok (some _) => crossEmoji
+| .ok none => checkEmoji
+
+/--
+Try to solve the goal either by:
+* calling `solveByElim`
+* or applying a library lemma then calling `solveByElim` on the resulting goals.
```
semorrison, line 405:
```suggestion
Try to solve the goal either by:
* calling `tactic true`
* or applying a library lemma then calling `tactic false` on the resulting goals.

Typically here `tactic` is `solveByElim`,
with the `Bool` flag indicating whether it may retry with `exfalso`.
```

semorrison:
I think that's pretty much the last round of comments from me on this; I'm otherwise happy!

Std/Tactic/LibrarySearch.lean
joehendrix: The code that tries out candidates would be a good place for concurrency, but results should be deterministic.

Std/Tactic/LibrarySearch.lean
joehendrix: I don't like using the names for ordering as a general principle.

The lazy discrimination tree orders results in reverse order they are found which is an approximation of match size with patterns in the same location being ordered in the order they are imported.

By "match size" I mean the number of keys other than star that are shared between the term and pattern.

I'll see if I can use a heap to return actual largest match first with import order for ties.

Std/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,530 @@
...
+/-- Sort the incomplete results from `librarySearchSymm` according to
+* the number of local hypotheses used (the more the better) and
+* the number of remaining subgoals (the fewer the better).
+-/
+def sortResults (goal : MVarId) (R : Array (List MVarId × MetavarContext)) :
+    MetaM (Array (List MVarId × MetavarContext)) := do
+  let R' ← R.mapM fun (gs, ctx) => do
+    return (← withMCtx ctx (subgoalRanking goal gs), gs, ctx)
+  let R'' := R'.qsort fun a b => compare a.1 b.1 = Ordering.gt
+  return R''.map (·.2)
```
joehendrix, line 310: I've done some testing and at least with `LazyDiscrTree`, `sortResults` seems to make the results worse.

I think I'll remove it for now, but may want to add it back.

Std/Tactic/LibrarySearch.lean
joehendrix: I added the ability to use longest match first.  We can adjust as needed in response to feedback.

test/library_search/basic.lean
joehendrix: I did a quick test.  I am going to leave this as is, but I think it'd be good to clarify the purpose of this, and see if there is a better way to achieve it.  My impression is the feature is so users get back some results.

PR #422, opened 2023-12-06
feat: Control warningAsError (default off but stays on in CI)

joehendrix:
This also has the new compiler to default off and uses a more declarative approach to keyword flags as I'd like `lakefile.lean` to be as declarative as possible (maybe we can upstream those parts).

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -2,13 +2,26 @@ import Lake
 
 open Lake DSL
 
+/-- Turn a config option into a Boolean -/
+def enabled (val : Option String) (default : Bool := false) : Bool :=
+  match val with
+  | .some "off" => false
+  | .some "false" => false
+  | .some "on" => true
+  | .some "true" => true
+  | .some _ => not default -- This allows -K=opt to flip the default.
+  | _ => default
+
+-- Selectively choose an option if feature is enabled.
+macro "optArg?" x:ident ws d:ident v:term : term =>
+  `(if enabled (get_config? $x) $d then $v else default)
+
 package std where
-  moreLeanArgs := #["-DwarningAsError=true"]
-  leanOptions := Id.run do
-    let mut opts := #[⟨`linter.missingDocs, true⟩]
-    if get_config? disable_new_compiler |>.isSome then
-      opts := opts.push ⟨`compiler.enableNew, false⟩
-    opts
+  moreLeanArgs :=
+    optArg? werror false #["-DwarningAsError=true"]
+  leanOptions :=
+    #[⟨`linter.missingDocs, true⟩] ++
+    optArg? disable_new_compiler true #[⟨`compiler.enableNew, false⟩]
 
 @[default_target]
 lean_lib Std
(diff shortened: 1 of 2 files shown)


lakefile.lean
```diff
@@ -2,13 +2,26 @@ import Lake
...
+  | .some "off" => false
+  | .some "false" => false
+  | .some "on" => true
+  | .some "true" => true
+  | .some _ => not default -- This allows -K=opt to flip the default.
+  | _ => default
+
+-- Selectively choose an option if feature is enabled.
+macro "optArg?" x:ident ws d:ident v:term : term =>
+  `(if enabled (get_config? $x) $d then $v else default)
```
digama0, line 17: This is too much code to go in the lakefile. This should be moved into lake if required.

digama0: For boolean options I suggest just normalizing everything to `get_config? foo |>.isSome`

lakefile.lean
```diff
@@ -2,13 +2,26 @@ import Lake
...
-  leanOptions := Id.run do
-    let mut opts := #[⟨`linter.missingDocs, true⟩]
-    if get_config? disable_new_compiler |>.isSome then
-      opts := opts.push ⟨`compiler.enableNew, false⟩
-    opts
+  moreLeanArgs :=
+    optArg? werror false #["-DwarningAsError=true"]
+  leanOptions :=
+    #[⟨`linter.missingDocs, true⟩] ++
+    optArg? disable_new_compiler true #[⟨`compiler.enableNew, false⟩]
```
digama0, line 24: `compiler.enableNew` will soon be disabled by default, I think we can just remove this option from the lakefile when that lands

lakefile.lean
```diff
@@ -2,13 +2,26 @@ import Lake
...
+
 package std where
-  moreLeanArgs := #["-DwarningAsError=true"]
-  leanOptions := Id.run do
-    let mut opts := #[⟨`linter.missingDocs, true⟩]
-    if get_config? disable_new_compiler |>.isSome then
-      opts := opts.push ⟨`compiler.enableNew, false⟩
-    opts
+  moreLeanArgs :=
+    optArg? werror false #["-DwarningAsError=true"]
```
digama0, line 21: I would also like lake to gain specific knowledge of this flag so that it can be controlled without having to reimplement this logic in every project that uses CI

lakefile.lean
```diff
@@ -2,26 +2,14 @@ import Lake
...
-  | .some "false" => false
-  | .some "on" => true
-  | .some "true" => true
-  | .some _ => not default -- This allows -K=opt to flip the default.
-  | _ => default
-
--- Selectively choose an option if feature is enabled.
-macro "optArg?" x:ident ws d:ident v:term : term =>
-  `(if enabled (get_config? $x) $d then $v else default)
+macro "optArg?" x:ident v:term : term =>
```
digama0, line 5:
```suggestion
macro "opt_arg?" x:ident v:term : term =>
```
macros / keywordish tokens are snake cased

lakefile.lean
joehendrix: I'd prefer a simple fix for now rather than more work.  I stripped it down though.

lakefile.lean
digama0: No need to block this PR on it, just keep it as is for now. Just wanted to let you know regarding the future of this flag

PR #432, opened 2023-12-12
chore: deprecate Int.ofNat_natAbs_eq_of_nonneg

### LINKED ISSUE #419: Int.ofNat_natAbs_eq_of_nonneg and Int.natAbs_of_nonneg seem duplicates

semorrison:
Fixes #419.


Std/Data/Int/Lemmas.lean
```diff
@@ -1339,6 +1339,7 @@ theorem eq_one_of_mul_eq_self_right {a b : Int} (Hpos : b ≠ 0) (H : b * a = b)
 
 /-! ### nat abs -/
 
+@[deprecated natAbs_of_nonneg]
 theorem ofNat_natAbs_eq_of_nonneg : ∀ a : Int, 0 ≤ a → Int.natAbs a = a
```
digama0, line 1343:
```suggestion
@[deprecated] alias ofNat_natAbs_eq_of_nonneg := natAbs_of_nonneg
```
that way we get the code action. Unless the statement isn't exactly the same?

Issue #431, opened 2023-12-11
Array.minX functions are buggy


Blei:
There are two basic bugs: the default value of `minD` is always used and compared against all entries in the array, instead of only being used as default value if the array is empty:

```lean
#eval #[1, 2, 3].minD 0
```
Expected: `1`, actual: `0`.

And the case that `startPos` <= `endPos` should signify an empty range is not considered:

```lean
#eval #[1, 2, 3].min? 1 0
```
Expected: `none`, actual: `some 2`.

semorrison:
@Blei, I agree, both of these are bugs. Would you be interested in making a PR to fix them?

Blei:
Yes, I would be interested. As long as you don't care so much about latency in getting this in. :)

So, the plan would be to add new minX functions and rename the ones with the current behaviour to minWithX?

General question: are there plans for unit tests in this standard library?

digama0:
> General question: are there plans for unit tests in this standard library?

Yes, there are tests in the `tests/` directory and you are welcome to add tests for regular functions too, not just tactics. (Then again, for regular functions we usually have theorems in lieu of tests - we want these anyway and they have better coverage than any test suite.) We also (should) have some illustrative tests in the documentation itself.

PR #433, opened 2023-12-12
Add minWith, maxWith; fix bugs in other min_ and max_ functions.

Blei:
This is meant to be a fix for #431.

This is my first time writing lean for anyone but horrible hacky advent of code scripts, so please don't hold back with the feedback.

### DIFF
--- a/Std/Data/Array/Basic.lean
+++ b/Std/Data/Array/Basic.lean
@@ -38,6 +38,18 @@ def qsortOrd [ord : Ord α] (xs : Array α) : Array α :=
 
 set_option linter.unusedVariables.funArgs false in
 /--
+Returns the first minimal element among `d` and elements of the array.
+If `start` and `stop` are given, only the subarray `xs[start:stop]` is
+considered (in addition to `d`).
+-/
+@[inline]
+protected def minWith [ord : Ord α]
+    (xs : Array α) (d : α) (start := 0) (stop := xs.size) : α :=
+  xs.foldl (init := d) (start := start) (stop := stop) fun min x =>
+    if compare x min |>.isLT then x else min
+
+set_option linter.unusedVariables.funArgs false in
+/--
 Find the first minimal element of an array. If the array is empty, `d` is
 returned. If `start` and `stop` are given, only the subarray `xs[start:stop]` is
 considered.
@@ -45,8 +57,10 @@ considered.
 @[inline]
 protected def minD [ord : Ord α]
     (xs : Array α) (d : α) (start := 0) (stop := xs.size) : α :=
-  xs.foldl (init := d) (start := start) (stop := stop) fun min x =>
-    if compare x min |>.isLT then x else min
+  if h: start < xs.size ∧ start < stop then
+    xs.minWith (xs.get ⟨start, h.left⟩) (start + 1) stop
+  else
+    d
 
 set_option linter.unusedVariables.funArgs false in
 /--
@@ -57,8 +71,8 @@ considered.
 @[inline]
 protected def min? [ord : Ord α]
     (xs : Array α) (start := 0) (stop := xs.size) : Option α :=
-  if h : start < xs.size then
-    some $ xs.minD (xs.get ⟨start, h⟩) start stop
+  if h : start < xs.size ∧ start < stop then
+    some $ xs.minD (xs.get ⟨start, h.left⟩) start stop
   else
     none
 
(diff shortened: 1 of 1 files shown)


Std/Data/Array/Basic.lean
```diff
@@ -36,6 +36,18 @@ Sort an array using `compare` to compare elements.
...
   xs.qsort λ x y => compare x y |>.isLT
 
+set_option linter.unusedVariables.funArgs false in
+/--
+Returns the first minimal element among `d` and elements of the array.
+If `start` and `stop` are given, only the subarray `xs[start:stop]` is
+considered (in addition to `d`).
+-/
+@[inline]
+protected def minWith [ord : Ord α]
```
semorrison, line 46:
```suggestion
/--
Returns the first minimal element among `d` and elements of the array.
If `start` and `stop` are given, only the subarray `xs[start:stop]` is
considered (in addition to `d`).
-/
@[inline]
protected def minWith [Ord α]
```
No need to name typeclasses.

Std/Data/Array/Basic.lean
```diff
@@ -45,8 +57,10 @@ considered.
 @[inline]
 protected def minD [ord : Ord α]
     (xs : Array α) (d : α) (start := 0) (stop := xs.size) : α :=
-  xs.foldl (init := d) (start := start) (stop := stop) λ min x =>
-    if compare x min |>.isLT then x else min
+  if h: start < xs.size ∧ start < stop then
```
semorrison, line 60:
```suggestion
  if h : start < xs.size ∧ start < stop then
```

semorrison (approved):
Otherwise, looks good to me.

It would be nice to have some basic theorems specifying the behaviour of this function
(even if they are not proved, you can state them using `proof_wanted`).

But I don't think those specification lemmas need to prevent the PR being merged.

Std/Data/Array/Basic.lean
semorrison: Oh, I missed that you were using the `(ord := ...)` named argument syntax below.

An alternative is
```
  @Array.minWith _ ord.opposite xs d start stop
```

but I'm ambivalent about which is better. Feel free to just click "resolve" without any action if you prefer!

Std/Data/Array/Basic.lean
```diff
@@ -57,8 +71,8 @@ considered.
 @[inline]
 protected def min? [ord : Ord α]
     (xs : Array α) (start := 0) (stop := xs.size) : Option α :=
-  if h : start < xs.size then
-    some $ xs.minD (xs.get ⟨start, h⟩) start stop
+  if h : start < xs.size ∧ start < stop then
+    some $ xs.minD (xs.get ⟨start, h.left⟩) start stop
```
digama0, line 75:
```suggestion
    some <| xs.minD (xs.get ⟨start, h.left⟩) start stop
```

Std/Data/Array/Basic.lean
```diff
@@ -36,6 +36,18 @@ Sort an array using `compare` to compare elements.
...
+set_option linter.unusedVariables.funArgs false in
+/--
+Returns the first minimal element among `d` and elements of the array.
+If `start` and `stop` are given, only the subarray `xs[start:stop]` is
+considered (in addition to `d`).
+-/
+@[inline]
+protected def minWith [ord : Ord α]
+    (xs : Array α) (d : α) (start := 0) (stop := xs.size) : α :=
+  xs.foldl (init := d) (start := start) (stop := stop) λ min x =>
```
digama0, line 48:
```suggestion
  xs.foldl (init := d) (start := start) (stop := stop) fun min x =>
```

Std/Data/Array/Basic.lean
Blei: I like your suggestion more TBH, but I think this might be an externally visible API change. So I think we should be consistent in the whole file about this whether we want these to be named or not. We should probably treat this in a separate issue/PR.

Std/Data/Array/Basic.lean
Blei: Done.

Std/Data/Array/Basic.lean
Blei: Done.

Std/Data/Array/Basic.lean
Blei: Looks like someone cleaned this up in the mean time :) Done.

semorrison (approved):

Blei:
Do you still expect to review this, or are there still problems with the PR?

semorrison:
@Blei, ah, sorry, the `awaiting-author` label may have deterred the maintainers from looking at it. Anyone can change the labels by commenting, as follows. (Sorry this is not at all obvious!)

semorrison:
awaiting-review

semorrison:
(And I'll ping about this on the zulip channel!)

Blei:
Thank you! Good to know.

joehendrix (approved):

PR #434, opened 2023-12-13
feat: (Lawful)BEq for AssocList

semorrison:
I worry slightly that someone will say they want the relation that *doesn't* care about the ordering of key-value pairs, but that is rather harder to calculate, not `Lawful`, and ... I want this one!

### DIFF
--- a/Std/Data/AssocList.lean
+++ b/Std/Data/AssocList.lean
@@ -4,7 +4,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
 import Std.Data.List.Basic
-import Std.Util.ProofWanted
 
 namespace Std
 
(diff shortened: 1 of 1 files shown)


digama0:
Why wouldn't it be lawful? I think it is reasonable for the `==` implementation to be order-dependent, but we could give it a `≈` implementation using `List.isPerm`.

Std/Data/AssocList.lean
```diff
@@ -236,3 +236,35 @@ instance : Stream (AssocList α β) (α × β) := ⟨pop?⟩
...
+@[simp] theorem beq_nil_cons [BEq α] [BEq β] : ((.nil : AssocList α β) == .cons a b t) = false :=
+  rfl
+@[simp] theorem beq_cons_nil [BEq α] [BEq β] : ((.cons a b t : AssocList α β) == .nil) = false :=
+  rfl
+@[simp] theorem beq_cons₂ [BEq α] [BEq β] :
+    ((.cons a b t : AssocList α β) == .cons a' b' t') = (a == a' && b == b' && t == t') := rfl
+
+instance [BEq α] [LawfulBEq α] [BEq β] [LawfulBEq β] : LawfulBEq (AssocList α β) where
+  rfl := @fun L => by induction L <;> simp_all
+  eq_of_beq := @fun L M => by
```
digama0, line 262:
```suggestion
  rfl {L} := induction L <;> simp_all
  eq_of_beq {L M} := by
```

Std/Data/AssocList.lean
```diff
@@ -236,3 +236,35 @@ instance : Stream (AssocList α β) (α × β) := ⟨pop?⟩
 
 @[simp] theorem toList_toAssocList (l : AssocList α β) : l.toList.toAssocList = l := by
   induction l <;> simp [*]
+
+private def beq [BEq α] [BEq β] : AssocList α β → AssocList α β → Bool
```
digama0, line 240:
```suggestion
def beq [BEq α] [BEq β] : AssocList α β → AssocList α β → Bool
```

digama0:
There should be a theorem proving it is equivalent to `a.toList == b.toList`

PR #435, opened 2023-12-13
feat: AssocList.size

semorrison:
I will need this for some termination proofs for operations on coefficients in `omega`.

### DIFF
--- a/Std/Data/AssocList.lean
+++ b/Std/Data/AssocList.lean
@@ -52,7 +52,7 @@ def length (L : AssocList α β) : Nat :=
 @[simp] theorem length_nil : length (nil : AssocList α β) = 0 := rfl
 @[simp] theorem length_cons : length (cons a b t) = length t + 1 := rfl
 
-theorem toList_length (l : AssocList α β) : l.toList.length = l.length := by
+theorem length_toList (l : AssocList α β) : l.toList.length = l.length := by
   induction l <;> simp_all
 
 /-- `O(n)`. Fold a monadic function over the list, from head to tail. -/
@@ -94,11 +94,11 @@ def toListTR (as : AssocList α β) : List (α × β) :=
   | nil        => nil
   | cons k v t => cons (f k) v (mapKey f t)
 
-@[simp] theorem mapKey_toList (f : α → δ) (l : AssocList α β) :
+@[simp] theorem toList_mapKey (f : α → δ) (l : AssocList α β) :
     (mapKey f l).toList = l.toList.map (fun (a, b) => (f a, b)) := by
   induction l <;> simp [*]
 
-@[simp] theorem mapKey_length : (mapKey f l).length = l.length := by
+@[simp] theorem length_mapKey : (mapKey f l).length = l.length := by
   induction l <;> simp_all
 
 /-- `O(n)`. Map a function `f` over the values of the list. -/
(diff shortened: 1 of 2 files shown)


Std/Data/AssocList.lean
```diff
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
 import Std.Data.List.Basic
+import Std.Util.ProofWanted
```
digama0, line 7:
```suggestion
```
you didn't use it

Std/Data/AssocList.lean
```diff
@@ -43,6 +44,18 @@ def isEmpty : AssocList α β → Bool
 @[simp] theorem isEmpty_eq (l : AssocList α β) : isEmpty l = l.toList.isEmpty := by
   cases l <;> simp [*, isEmpty, List.isEmpty]
 
+/-- The number of entries in an `AssocList`. -/
+def size (L : AssocList α β) : Nat :=
+  match L with
```
digama0, line 49:
```suggestion
def length : AssocList α β → Nat
```
The naming difference between `List` and other data structures is unfortunate, but `AssocList` takes its API mainly from `List` so I think it should use `length` here. Or maybe one could say that `length` is for the O(n) operation and `size` is when it is O(1) because of some kind of caching? In which case this should also be `length`.

Std/Data/AssocList.lean
semorrison: I guess we should import this early somewhere, to avoid having this happen too often?

digama0: Yes, I think `Tactic.Basic` has a few others like this.

Std/Data/AssocList.lean
```diff
@@ -43,6 +43,18 @@ def isEmpty : AssocList α β → Bool
...
+/-- The number of entries in an `AssocList`. -/
+def length (L : AssocList α β) : Nat :=
+  match L with
+  | .nil => 0
+  | .cons _ _ t => t.length + 1
+
+@[simp] theorem length_nil : length (nil : AssocList α β) = 0 := rfl
+@[simp] theorem length_cons : length (cons a b t) = length t + 1 := rfl
+
+theorem toList_length (l : AssocList α β) : l.toList.length = l.length := by
```
digama0, line 55:
```suggestion
theorem length_toList (l : AssocList α β) : l.toList.length = l.length := by
```
ditto for all the other theorems below

semorrison: Are we sure? Should I rename (the already existing) `mapKey_toList` to `toList_mapKey` as well? I really don't like not being able to name things as they appear with dot notation. :-(

digama0: Well it has lead to inconsistent naming even in this very PR (`length_nil` and `length_cons` differ from the others because you decided not to use dot notation there). I think the current status is that this is a general rule, but if you want to make a specific exception for this function I'm okay with that provided you do it consistently and name the others in postfix too.

semorrison: Interesting.

I would argue that both `length_cons` and `toList_length` are simultaneously desirable: the thing the theorem is "about" should come first, and theorems should mostly be about functions rather than terms.

Hence `length_cons` is preferable to `cons_length`, because this is telling us about the behaviour of the function `length` on the constructors.

Similarly `toList_length` is preferable to `length_toList`, because it is describing a property of the function `toList`.

joehendrix: I prefer `length_toList` as well because of the naming convention [mentioned here](https://github.com/leanprover/std4/pull/329#issuecomment-1786193983).

semorrison: Okay. Under mild protest, I have changed all the names in this file.

joehendrix (approved):

PR #438, opened 2023-12-13
feat: lemmas about if-then-else

semorrison:


Std/Logic.lean
```diff
@@ -784,6 +784,18 @@ theorem apply_ite (f : α → β) (P : Prop) [Decidable P] (x y : α) :
 @[simp] theorem ite_eq_right_iff {P : Prop} [Decidable P] : ite P a b = b ↔ P → a = b :=
   dite_eq_right_iff
 
+/-- A `dite` whose results do not actually depend on the condition may be reduced to an `ite`. -/
+@[simp] theorem dite_eq_ite [Decidable P] : (dite P (fun _ => a) fun _ => b) = ite P a b :=
+  rfl
```
digama0, line 789:
```suggestion
@[simp] theorem dite_eq_ite [Decidable P] : (dite P (fun _ => a) fun _ => b) = ite P a b := rfl
```

Std/Logic.lean
```diff
@@ -784,6 +784,18 @@ theorem apply_ite (f : α → β) (P : Prop) [Decidable P] (x y : α) :
...
+
+theorem ite_some_none_eq_none [Decidable P] :
```
digama0, line 791:
```suggestion
@[simp] theorem ite_some_none_eq_none [Decidable P] :
```
Is there a reason you marked the other one as `@[simp]` but not this one?

semorrison: Yes, the other is already handled by `ite_eq_right_iff`. I'll add a note.

PR #439, opened 2023-12-13
feat: Option.min/max

semorrison:


Std/Data/Option/Basic.lean
```diff
@@ -157,3 +157,24 @@ instance (α) [BEq α] [LawfulBEq α] : LawfulBEq (Option α) where
     match x, y with
     | some x, some y => rw [LawfulBEq.eq_of_beq (α := α) h]
     | none, none => rfl
+
+@[simp] theorem all_none : Option.all p none = true := rfl
+@[simp] theorem all_some : Option.all p (some x) = p x := rfl
+
+/-- The minimum of two optional values. -/
+@[simp] def min [Min α] : Option α → Option α → Option α
```
digama0, line 165:
```suggestion
protected def min [Min α] : Option α → Option α → Option α
```
Is the issue with `@[simp] def` doing something stupid fixed? Last I checked I think we still have to write equation lemmas by hand.

semorrison: It's less stupid, because of the new `simp (config := { unfoldPartialApp := true })` flag. It will no longer just unfold an unapplied `min` without arguments.

However `example [Min α] {a b : Option α} : min a b = sorry := by simp` will still unfold here.

I'll replace these with equation lemmas.

PR #440, opened 2023-12-13
chore: relax List.insert to only rely on BEq

semorrison:


digama0:
needs a mathlib bump PR

digama0 (approved):

semorrison:
> needs a mathlib bump PR

The mathlib bump PR is compiling, changes look fairly harmless. (Currently stacked on top of another Std bump PR, however.)

PR #442, opened 2023-12-13
feat: lemmas about Int.sign

semorrison:
Also gather together the existing lemmas about `sign`.


Std/Data/Int/Lemmas.lean
```diff
@@ -1337,6 +1279,80 @@ theorem eq_one_of_mul_eq_self_left {a b : Int} (Hpos : a ≠ 0) (H : b * a = a)
...
+  | .ofNat (_ + 1) => rfl
+  | .negSucc _ => rfl
+
+@[simp] theorem sign_nonneg : 0 ≤ sign x ↔ 0 ≤ x := by
+  match x with
+  | 0 => rfl
+  | .ofNat (_ + 1) =>
+    simp (config := { decide := true }) only [sign, true_iff]
+    exact Int.le_add_one (ofNat_nonneg _)
+  | .negSucc _ => simp (config := {decide := true}) [sign]
```
digama0, line 1353:
```suggestion
  | .negSucc _ => simp (config := { decide := true }) [sign]
```
we aren't particularly consistent about this but we should probably do one style uniformly. grep suggests that the spaces form is more common in std.

PR #447, opened 2023-12-13
Port: solve_by_elim from Mathlib

joehendrix:
This ports solve_by_elim from Mathlib and backtracking as part of that.

### DIFF
--- a/Std/Tactic/SolveByElim.lean
+++ b/Std/Tactic/SolveByElim.lean
@@ -11,6 +11,20 @@ import Std.Tactic.SolveByElim.Backtrack
 
 /-!
 # `solve_by_elim`, `apply_rules`, and `apply_assumption`.
+
+`solve_by_elim` takes a collection of facts from the local context or
+supplied as arguments by the user, and performs a backtracking
+depth-first search by attempting to `apply` these facts to the goal.
+
+It is a highly configurable tactic, with options to control the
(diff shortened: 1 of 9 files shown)


Std/Tactic/SolveByElim.lean
```diff
@@ -0,0 +1,536 @@
...
+-/
+import Std.Tactic.Backtrack
+import Lean.Meta.Tactic.Apply
+import Std.Tactic.LabelAttr
+import Std.Data.Sum.Basic
+import Std.Tactic.Relation.Symm
+import Std.Data.Option.Basic
+
+/-!
+# `solve_by_elim`, `apply_rules`, and `apply_assumption`.
```
semorrison, line 14:
```suggestion
# `solve_by_elim`, `apply_rules`, and `apply_assumption`.

`solve_by_elim` takes a collection of facts from the local context or supplied as arguments by the user, and performs a backtracking depth-first search by attempting to `apply` these facts to the goal.

It is a highly configurable tactic, with options to control the backtracking or to supply a discharging tactic for unprovable goals.

`apply_rules` and `apply_assumption` are much simpler tactics which do not perform backtracking, but are currently implemented in terms of `solve_by_elim` with backtracking disabled, in order to be able to share the front-end customisation and parsing of user options. It would be reasonable to further separate these in future.
```

PR #449, opened 2023-12-14
chore: ToExpr Int

semorrison:
I'm not sure in which file this belongs.

(I'm also still a bit sad that we [decided](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/ToExpr.20derive.20handler/near/386476438) not to upstream the `ToExpr` derive handler from Mathlib...!)


joehendrix:
I'm not sure about the location either, but merging it seems to fine to me.

PR #451, opened 2023-12-14
feat: Int.add_le_iff_le_sub and variants

semorrison:


Std/Data/Int/Lemmas.lean
```diff
@@ -841,6 +841,29 @@ protected theorem add_neg_of_nonpos_of_neg {a b : Int} (ha : a ≤ 0) (hb : b <
...
+  rw [Int.add_le_iff_le_sub, Int.zero_sub]
+
+protected theorem add_le_zero_iff_le_neg' (a b : Int) : a + b ≤ 0 ↔ b ≤ -a := by
+  rw [Int.add_comm, Int.add_le_zero_iff_le_neg]
+
+protected theorem add_nonnneg_iff_neg_le (a b : Int) : 0 ≤ a + b ↔ -b ≤ a := by
+  rw [Int.le_add_iff_sub_le, Int.zero_sub]
+
+protected theorem add_nonnneg_iff_neg_le' (a b : Int) : 0 ≤ a + b ↔ -a ≤ b := by
+  rw [Int.add_comm, Int.add_nonnneg_iff_neg_le]
```
joehendrix, line 865: I know these aren't `simp`, but I am surprised by the direction of these `iff` rules, I would have expected them to normalize the expressions in some way perhaps to `0 <= x + ... + y`.

Could you say a little about why the orientation of the left and right-hand sides?

semorrison: Just the direction these turned up:
```
w: c + Coeffs.dot x y ≤ 0 ∧ 0 ≤ c + Coeffs.dot x y
⊢ -c ≤ Coeffs.dot x y ∧ Coeffs.dot x y ≤ -c
```
No objection to flipping them, or making them `omega` private.

I'm a little skeptical that there's any general sense in which one side is simpler than the other (possibly why these lemmas don't already exist).

joehendrix: In that case, I'd just be inclined to keep them with omega for now.

PR #453, opened 2023-12-14
chore: forward direction of not_or and Decidable.not_and_iff_or_not

semorrison:
Useful to have these when constructing proof `Expr`s by hand, but I can hide them away somewhere if they are not wanted.

### DIFF
--- a/Std/Logic.lean
+++ b/Std/Logic.lean
@@ -348,6 +348,8 @@ theorem or_imp : (a ∨ b → c) ↔ (a → c) ∧ (b → c) :=
 
 theorem not_or : ¬(p ∨ q) ↔ ¬p ∧ ¬q := or_imp
 
+alias ⟨and_not_not_of_not_or, _⟩ := not_or
+
 theorem not_and_of_not_or_not (h : ¬a ∨ ¬b) : ¬(a ∧ b) := h.elim (mt (·.1)) (mt (·.2))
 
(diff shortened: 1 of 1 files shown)


Std/Logic.lean
```diff
@@ -649,6 +651,8 @@ theorem Decidable.or_congr_left' [Decidable c] (h : ¬c → (a ↔ b)) : a ∨ c
 theorem Decidable.or_congr_right' [Decidable a] (h : ¬a → (b ↔ c)) : a ∨ b ↔ a ∨ c := by
   rw [or_iff_not_imp_left, or_iff_not_imp_left]; exact imp_congr_right h
 
+alias ⟨Decidable.or_not_not_of_not_and, _⟩ := Decidable.not_and_iff_or_not
+
```
joehendrix, line 655: This seems like a niche use.  I'm not in favor putting these definitions here.

One side note, I don't see a reason to prefer `Decidable.not_and_iff_or_not` over `not_and`.  Am I missing something?

joehendrix (changes requested):

Std/Logic.lean
semorrison: Are you asking about the `iff_or_not` suffix (which I agree could be dropped), or the `Decidable` namespace (which I think is important to distinguish that it is not a classical theorem and takes hidden decidability arguments).

PR #456, opened 2023-12-15
feat: Expr.nat?/int? to check for numbers in normal form

semorrison:
These are used by various Mathlib tactics, and now by `omega`, so will need to move to Std.

### DIFF
--- a/Std/Lean/Expr.lean
+++ b/Std/Lean/Expr.lean
@@ -164,11 +164,14 @@ def nat? (e : Expr) : Option Nat := do
 
 /--
 Checks if an expression is an "integer in normal form",
-i.e. either a natural number in normal form, or the negation of one,
+i.e. either a natural number in normal form, or the negation of a positive natural number,
 and if so returns the integer.
 -/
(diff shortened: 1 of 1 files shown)


Std/Lean/Expr.lean
```diff
@@ -135,3 +135,40 @@ def constName (e : Expr) : Name :=
...
+def nat? (e : Expr) : Option Nat := do
+  guard <| e.isAppOfArity ``OfNat.ofNat 3
+  let lit (.natVal n) := e.appFn!.appArg! | none
+  n
+
+/--
+Checks if an expression is an "integer in normal form",
+i.e. either a natural number in normal form, or the negation of one,
+and if so returns the integer.
+-/
```
digama0, line 169:
```suggestion

Note: this considers `-0` to be a normal form integer, although tactics will not produce them.
-/
```
or should we screen it out?

semorrison: Maybe more prudent to screen it out. I've pushed a change.

PR #457, opened 2023-12-15
chore: upstream getLocalHyps

semorrison:
Another utility function that is used by various Mathlib tactics, and now by `omega`.

### DIFF
--- a/Std/Lean/Meta/Basic.lean
+++ b/Std/Lean/Meta/Basic.lean
@@ -436,3 +436,10 @@ where
       match ← tac goal with
       | none => acc.modify λ s => s.push goal
       | some goals => goals.forM (go acc)
+
+/-- Return local hypotheses which are not "implementation detail", as `Expr`s. -/
+def getLocalHyps [Monad m] [MonadLCtx m] : m (Array Expr) := do
+  let mut hs := #[]
+  for d in ← getLCtx do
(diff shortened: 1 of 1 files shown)


Std/Lean/Meta/Basic.lean
```diff
@@ -436,3 +436,10 @@ where
       match ← tac goal with
       | none => acc.modify λ s => s.push goal
```
digama0, line 437:
```suggestion
      | none => acc.modify fun s => s.push goal
```

semorrison: No, we have mapsto, but never did lambda.

semorrison: Do you really just want to forbid the character `λ` entirely? That seems overly restrictive.

semorrison: Also, the `λ` you've noticed here isn't even part of this PR. Perhaps do it separately?

digama0: > Do you really just want to forbid the character `λ` entirely? That seems overly restrictive.

It's a style rule within std. It doesn't seem particularly restrictive since `fun` is entirely interchangeable with `λ`, although a linter that just bans the character might have false positives in documentation.

PR #459, opened 2023-12-15
feat: mkAppN_macro

semorrison:
This is @kmill's macro allowing using `mkAppN` in pattern matches.

They have [PR'd that to Lean](https://github.com/leanprover/lean4/pull/2900), but the PR has stalled and I would like to be using this in `omega` PRs coming very soon, so I propose temporarily installing it in Std.


digama0:
I don't like this kind of PR. I suggest you work to try to fix the bug flagged in the lean PR (or induce others to do so) rather than pushing it to a different repo. If you need to use it in `omega`, I would suggest adding it only in an "internal" way, e.g. using `local macro_rules`, and confining it to the affected file.

semorrison:
Sure, I'll make it local for now.

PR #460, opened 2023-12-15
feat: false_or_by_contra tactic

semorrison:
This is the very first layer of pre-processing for `omega`, but is perhaps independently useful so I'll put it in a separate file.

* [x] depends on #458
* [x] depends on #459


Std/Tactic/FalseOrByContra.lean
```diff
@@ -0,0 +1,43 @@
...
+If the goal is `False`, do nothing.
+If the goal is `¬ P`, introduce `P`.
+If the goal is `x ≠ y`, introduce `x = y`.
+Otherwise, for a goal `P`, replace it with `¬ ¬ P` and introduce `¬ P`.
+-/
+syntax (name := false_or_by_contra) "false_or_by_contra" : tactic
+
+@[inherit_doc false_or_by_contra]
+def falseOrByContra (g : MVarId) : MetaM (List MVarId) := do
+  match ← g.getType with
```
digama0, line 34:
```suggestion
  match ← whnfR (← g.getType) with
```

semorrison: Seems not.

joehendrix (approved):

PR #462, opened 2023-12-15
feat: MVarId.cases₂, convenience wrapper for cases with two single argument constructors

semorrison:
If this is too specialized I can inline.


joehendrix:
This feels too specialized to me at the moment unless there are multiple tactics that want it.

PR #464, opened 2023-12-15
feat: more `Bool` lemmas

negiizhao:


Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+theorem bne_true_right : ∀ (x : Bool), (x != true) = !x := by decide
+
+theorem bne_not_left : ∀ (x y : Bool), ((!x) != y) = (x == y) := by decide
+
+theorem bne_not_right : ∀ (x y : Bool), (x != (!y)) = (x == y) := by decide
+
+theorem bne_not_not : ∀ (x y : Bool), ((!x) != (!y)) = (x != y) := by decide
+
+theorem bne_eq_xor : ∀ (x y : Bool), (x != y) = xor x y := by decide
```
joehendrix, line 83: There are so many potential rules here and normal forms may be computationally intractable.

Is there a criteria used for these rules in particular?

negiizhao: Currently `xor` is an alias of `bne`, so its normal form is the same as `xor`.

negiizhao:
awaiting-review

Std/Data/Bool.lean
fgdorais: Since `xor` is an abbreviation for `bne`, simp sees through the abbreviation and all simp lemmas for `xor` work for `bne`. However, the seemingly duplicate lemmas are useful for `rw` which doesn't see through abbreviations as well. Because none of the proposed `bne` lemmas are marked simp there is no problem with adding these.

There is a bigger issue though about the simp normal form for `Bool`. Since `x != y` is `!( x == y)`, there is a chance that simp might run into issues where `==` leads one way but `xor` leads another way. I think that figuring out what the right simp normal form for `Bool` can wait a bit though and shouldn't bog down this PR.

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+@[simp] theorem beq_true_left : ∀ (x : Bool), (true == x) = x := by decide
+
+@[simp] theorem beq_true_right : ∀ (x : Bool), (x == true) = x := by decide
+
+theorem beq_not_left : ∀ (x y : Bool), ((!x) == y) = !(x == y) := by decide
+
+theorem beq_not_right : ∀ (x y : Bool), (x == (!y)) = !(x == y) := by decide
+
+@[simp] theorem beq_not_not : ∀ (x y : Bool), ((!x) == (!y)) = (x == y) := by decide
```
fgdorais, line 63: Use "infix" names.

```suggestion
@[simp] theorem false_beq : ∀ (x : Bool), (false == x) = !x := by decide

@[simp] theorem beq_false : ∀ (x : Bool), (x == false) = !x := by decide

@[simp] theorem beq_true_left : ∀ (x : Bool), (true == x) = x := by decide

@[simp] theorem beq_true : ∀ (x : Bool), (x == true) = x := by decide

theorem not_beq : ∀ (x y : Bool), ((!x) == y) = !(x == y) := by decide

theorem beq_not : ∀ (x y : Bool), (x == (!y)) = !(x == y) := by decide

@[simp] theorem not_beq_not : ∀ (x y : Bool), ((!x) == (!y)) = (x == y) := by decide
```

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+theorem bne_true_left : ∀ (x : Bool), (true != x) = !x := by decide
+
+theorem bne_true_right : ∀ (x : Bool), (x != true) = !x := by decide
+
+theorem bne_not_left : ∀ (x y : Bool), ((!x) != y) = (x == y) := by decide
+
+theorem bne_not_right : ∀ (x y : Bool), (x != (!y)) = (x == y) := by decide
+
+theorem bne_not_not : ∀ (x y : Bool), ((!x) != (!y)) = (x != y) := by decide
```
fgdorais, line 81: Use "infix" names.
```suggestion
theorem false_bne : ∀ (x : Bool), (false != x) = x := by decide

theorem bne_false : ∀ (x : Bool), (x != false) = x := by decide

theorem true_bne : ∀ (x : Bool), (true != x) = !x := by decide

theorem bne_true : ∀ (x : Bool), (x != true) = !x := by decide

theorem not_bne : ∀ (x y : Bool), ((!x) != y) = (x == y) := by decide

theorem bne_not : ∀ (x y : Bool), (x != (!y)) = (x == y) := by decide

theorem not_bne_not : ∀ (x y : Bool), ((!x) != (!y)) = (x != y) := by decide
```

fgdorais (changes requested):
The current convention is to prefer infix names unless there is a good reason not to. (Aliases can be used if non-infix names are commonly used.)

negiizhao:
awaiting-review

fgdorais (approved):

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
 instance : Max Bool := ⟨or⟩
 instance : Min Bool := ⟨and⟩
 
+theorem false_ne_true : false ≠ true := Bool.noConfusion
+
+theorem eq_false_or_eq_true : (b : Bool) → b = true ∨ b = false := by decide
+
+theorem eq_false_iff : {b : Bool} → b = false ↔ b ≠ true := by decide
+
+theorem eq_true_iff : {b : Bool} → b = true ↔ b ≠ false := by decide
```
joehendrix, line 47: Sorry for the delay, what's the motivation for these four lemmas?  It seems like they all reduce or simp or otherwise handled through tactics.

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+@[simp] theorem beq_true : ∀ (x : Bool), (x == true) = x := by decide
+
+theorem not_beq : ∀ (x y : Bool), ((!x) == y) = !(x == y) := by decide
+
+theorem beq_not : ∀ (x y : Bool), (x == (!y)) = !(x == y) := by decide
+
+@[simp] theorem not_beq_not : ∀ (x y : Bool), ((!x) == (!y)) = (x == y) := by decide
+
+theorem beq_eq_not_xor : ∀ (x y : Bool), (x == y) = !xor x y := by decide
```
joehendrix, line 65:
```suggestion
@[simp] theorem not_xor_eq_beq : ∀ (x y : Bool), !(xor x y) = (x == y)  := by decide
```

It seems like flipping the orientation and adding simp will strictly reduce term size.

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+@[simp] theorem beq_false : ∀ (x : Bool), (x == false) = !x := by decide
+
+@[simp] theorem true_beq : ∀ (x : Bool), (true == x) = x := by decide
+
+@[simp] theorem beq_true : ∀ (x : Bool), (x == true) = x := by decide
+
+theorem not_beq : ∀ (x y : Bool), ((!x) == y) = !(x == y) := by decide
+
+theorem beq_not : ∀ (x y : Bool), (x == (!y)) = !(x == y) := by decide
```
joehendrix, line 61:
```suggestion
theorem not_beq : ∀ (x y : Bool), ((!x) == y) = xor x y := by decide

theorem beq_not : ∀ (x y : Bool), (x == (!y)) = xor x y := by decide
```

We already a lot of theorems about xor, so I suggest just rewriting to it.

Std/Data/Bool.lean
joehendrix: Given all the `xor` theorems, the `bne` theorems feel unnecessary.  I'd suggest removing the bne theorems or if there are ones not found in the xor theorems replacing bne with xor.

joehendrix (commented):
Appology for delay.  I have more bandwidth now, and will try to move this through.

fgdorais:
The `bne` theorems are still needed for `rw` where `xor` won't work. The alternative is to systematically normalize `bne` to `xor` by adding a simp `bne x y = xor x y`. I'm a bit hesitant to do that but it might be an interesting experiment to try it out.

digama0:
I believe this was originally written with `xor` being an alias for `bne`, so it is weird to have this duplication. I'm not even sure simp lemmas on `xor` will trigger because it's reducible.

fgdorais:
The simps for `xor` trigger on both `xor` and `bne`. The annoying thing is that `bne_eq_ne` triggers on `xor`. So, for example, `xor (!x) (!y && x)` simps to `¬(!x) = (!y && x)`, which isn't useful. It's probably best to not make `xor` reducible.

digama0:
I'm a bit perplexed why you are using `xor` in the first place. It is not intended to be used in proofs, that's why it is reducible, and I don't see what's wrong with using Ne instead in that example.

fgdorais:
I think the issue is that `Bool` has two roles: values of decidable propositions vs a single bit of data. `bne` and `xor` are the same but they belong to opposite roles of `Bool`: `xor` is about bits while `bne` is a decidable proposition. It makes perfect sense to simp `bne` to `Ne` but it doesn't make sense for `xor`. At least, converting `xor` to `Ne` should be on the same level as converting `and` to `And`, `or` to `Or`. Not automatic since it's not always desirable but also not difficult since it's often used.

digama0:
But converting `and` to `And` is automatic...
```lean
import Std

example (x y : Bool) : x && y := by
  simp
  -- ⊢ x = true ∧ y = true
```

fgdorais:
But the correct translation of `xor a b = true` should be `¬(a = true ↔ b = true)`, not `Ne`.

digama0:
Agreed, but the trouble is that we can't have a simp lemma `a = b <-> (a = true <-> b = true)` because it would loop. See the zulip discussions about `IsTrue`.

fgdorais:
But you can have a simp lemma `xor a b = true ↔ ¬(a = true ↔ b = true)` if `xor` is not reducible to `bne`.

Std/Data/Bool.lean
fgdorais: I think some of these are moved from `Std.Logic` but `eq_true_iff` is not as useful as `ne_false_iff`.

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+/-! ### beq -/
+
+@[simp] theorem false_beq : ∀ (x : Bool), (false == x) = !x := by decide
+
+@[simp] theorem beq_false : ∀ (x : Bool), (x == false) = !x := by decide
+
+@[simp] theorem true_beq : ∀ (x : Bool), (true == x) = x := by decide
+
+@[simp] theorem beq_true : ∀ (x : Bool), (x == true) = x := by decide
```
fgdorais, line 57: The simpNF  for `bne` is `Eq`, so I don't think these simps are effective.

Std/Data/Bool.lean
```diff
@@ -38,6 +38,50 @@ instance (x y : Bool) : Decidable (x < y) := inferInstanceAs (Decidable (!x && y
...
+
+theorem not_beq : ∀ (x y : Bool), ((!x) == y) = !(x == y) := by decide
+
+theorem beq_not : ∀ (x y : Bool), (x == (!y)) = !(x == y) := by decide
+
+@[simp] theorem not_beq_not : ∀ (x y : Bool), ((!x) == (!y)) = (x == y) := by decide
```
fgdorais, line 63: This is also not an effective simp. This is basically `not_inj_iff`.

fgdorais:
I created a separate PR #512 to untangle the discussion on the reducibility of `xor` from this PR.

negiizhao:
I found that I didn't need as much lemmas as I remembered. So I've decided to keep only what I need now and leave the controversial parts for later.

PR #466, opened 2023-12-16
feat: "Try this:" custom code action title

thorimur:
Currently, the code action title in the lightbulb menu is hardcoded to `Try this: <prettyPrintedSuggestionText>`. This needs to be customized in various cases, such as teaching in non-English languages. This PR:

* Adds a field `toCodeActionTitle? : Option (String -> String)` to `Suggestion`, which, if provided, can be applied to the pretty-printed suggestion text and transform it as desired (note that this application occurs prior to encoding, so we only pass the resulting code action title `String`, if any, to the Json version of the `Suggestion` structure; note also that we don't want to pretty-print twice, so we take in the pretty-printed `String` as an argument instead of the `SuggestionText` itself)
* Adds an argument `codeActionPrefix? : Option String := none` to `addSuggestions` to conveniently provide a default code action title prefix to use for all elements in an `Array Suggestion` which do not have a nontrivial `toCodeAction?` value.

See [zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Trythis.20code.20action.20title) for the feature request and discussion on the choice of default behavior.

### DIFF
--- a/Std/Tactic/TryThis.lean
+++ b/Std/Tactic/TryThis.lean
@@ -96,9 +96,12 @@ apply the replacement.
     let .ok suggestions := suggestions.getArr? | panic! "bad type"
     let mut result := result
     for h : i in [:suggestions.size] do
-      let .ok newText := (suggestions[i]'h.2).getObjValAs? String "suggestion" | panic! "bad type"
+      let suggestion := suggestions[i]'h.2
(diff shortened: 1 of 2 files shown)


thorimur:
awaiting-review

semorrison (approved):

PR #470, opened 2023-12-20
chore: fixes for lean4#2973

semorrison:
This is a PR to the `bump/v4.5.0` branch. It relies on the toolchain `leanprover/lean4:nightly-2023-12-19`.

### DIFF
--- a/Std/Tactic/Lint/Misc.lean
+++ b/Std/Tactic/Lint/Misc.lean
@@ -221,14 +221,12 @@ with rfl when elaboration results in a different term than the user intended. -/
       return none
 
 /--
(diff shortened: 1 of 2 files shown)


joehendrix (approved):

kmill (approved):
I wrote this originally, but I figured I'd put in an approval to signify that I believe this to be the correct adaptation for the change to the letFun encoding!

Correctness relies on the fact that `forEachExpr` will visit every prefix to a nested application, so even if the letFun is applied to additional arguments, the body of `forEachExpr` will eventually see the letFun with the correct arity. This analysis only matters for rare situations like `(have x := v; notUsingX) a1 a2 a3`.

PR #473, opened 2023-12-20
feat(Data/Option/Basic): Add `StrictRel` instance

linesthatinterlace:
`Option.Rel` exists and is suitable for lifting a relation to `Option`. However, it is not suitable for lifting a strict relation like `<`, because it makes it so that `none ~ none`. `StrictRel` is similar, but avoids this.

### DIFF
--- a/Std/Data/Option/Basic.lean
+++ b/Std/Data/Option/Basic.lean
@@ -84,6 +84,12 @@ inductive Rel (r : α → β → Prop) : Option α → Option β → Prop
   /-- `none ~ none` -/
   | none : Rel r none none
 
+/-- Lifts a relation `α → β → Prop` to a relation `Option α → Option β → Prop` by making
+`none` incomparable. -/
+inductive StrictRel (r : α → β → Prop) : Option α → Option β → Prop
+  /-- If `a ~ b`, then `some a ~ some b` -/
+  | some {a b} : r a b → StrictRel r (some a) (some b)
+
 /--
 Partial bind. If for some `x : Option α`, `f : Π (a : α), a ∈ x → Option β` is a
 partial function defined on `a : α` giving an `Option β`, where `some a = x`,
--- a/Std/Data/Option/Basic.lean
+++ b/Std/Data/Option/Basic.lean
@@ -85,7 +85,7 @@ inductive Rel (r : α → β → Prop) : Option α → Option β → Prop
   | none : Rel r none none
 
 /-- Lifts a relation `α → β → Prop` to a relation `Option α → Option β → Prop` by making
-`none` incomparable. -/
+`none` unrelated to anything. -/
 inductive StrictRel (r : α → β → Prop) : Option α → Option β → Prop
   /-- If `a ~ b`, then `some a ~ some b` -/
   | some {a b} : r a b → StrictRel r (some a) (some b)


eric-wieser (approved):

Std/Data/Option/Basic.lean
```diff
@@ -84,6 +84,12 @@ inductive Rel (r : α → β → Prop) : Option α → Option β → Prop
   /-- `none ~ none` -/
   | none : Rel r none none
 
+/-- Lifts a relation `α → β → Prop` to a relation `Option α → Option β → Prop` by making
+`none` incomparable. -/
```
fgdorais, line 88:
```suggestion
`none` unrelated to anything. -/
```

linesthatinterlace: I'm not minded to resist this suggestion but me myself I don't really think it matters too much.

fgdorais: Reviews aren't only about things that matter. Not changing the status to `awaiting-author` is what signaled that this is not important enough to block merging.

linesthatinterlace: For sure! I just wanted to say I was accepting the change because I do quite like it, but I'm probably not minded to pull my hair out trying to find the perfect wording.

fgdorais (approved):

eric-wieser:
@fgdorais, can you approve the workflow so that CI runs?

fgdorais:
I wish I had that ability! Maintainers can do it: @joehendrix  @semorrison @digama0

joehendrix (changes requested):
There's are multiple possibilities for lifting a relation over `Option` and nothing in standard is currently using `Option.Rel`.  It does look like there is only one theorem in Mathlib referencing `Option.Rel` (`List.rel_filterMap`).  

Rather than adding another unused lifting operation, I'd instead propose removing `Option.Rel`.

I like new contributions, but I'd propose we not merge this until the usefulness is more clear.

linesthatinterlace:
I actually have a pull request in mathlib right now that uses both Rel and StrictRel. It's leanprover/mathlib4#9181 over there.

linesthatinterlace:
I also think that this is good for being a "minimal" lift for the relation - you're adding in the minimal data you need. But if you really don't want this to exist I suppose it could be deleted. I think it would be a shame and that partially why there are no theorems about it is just that nobody has bothered.

eric-wieser:
I think Rel and StrictRel should exist, and would happily merge them into mathlib if Std doesn't want them (we have similar constructions for `Sum` in mathlib)

linesthatinterlace:
Indeed, earlier today we merged a PR removing an accidental duplication of Option.Rel in Mathlib. I'm perfectly happy to put it back along with its cousin!

linesthatinterlace:
Incidentally, I think Sum.LiftRel is actually in standard...

digama0:
I think we could remove that one too, it's also unused in std and looks like a better fit for mathlib, unless we find some other use for it.

linesthatinterlace:
I don't disagree with that - it's a lot more arcane. Though I would like to make sure we do add these into Mathlib4 before they disappear from std!

joehendrix:
Yes, it looks like `Sum` has both `LiftRel` and `Lex` so things like this are compatible with existing definitions.

I'm personally inclined to say that Mathlib would be a better home for this, but only recently involved with Std and PRs like #251 predate me.  It'd be great to have others weigh in.

fgdorais:
I tend to agree with Mario and Joe but I don't feel very strongly about removing existing definitions. Here are my thoughts.

The main use case where this type of construction is essential is for wellfounded recursion. Having a standard def for a common construction like these is useful to make lemmas for use in automating `decreasing_by` arguments. Everybody hates having to write one of these. This could be used to motivate `Option.StrictRel` but it doesn't appear to be necessary.

A secondary use case is for setoids/quotients. There aren't that many of those in core and std but I would consider that a valid motivation. This could be used to motivate `Option.Rel` but I would like to see more use cases.

Some data structures do use this type of construction. This could be a motivator but I personally would tend to build my own ad-hoc construction with the lemmas I need rather than looking up a similar construction elsewhere. In such cases, I would expect the definition to come along with the data structure implementation.

digama0:
I think we should definitely keep `Sum.Lex`, as you say this is useful in well founded recursion. The others are not needed for this purpose though, since they are strictly weaker orders. I should probably have considered `Sum.LiftRel` more carefully in #251, but it is difficult to give every definition due weight when they are being bulk imported in this way, so you shouldn't put too much stock in it.

I am not opposed to having `Option.Rel` or `Option.StrictRel` in std if it is either facilitating another definition in std of more obvious value, or if it is part of some pattern we wish to promote, and this is not irreversible, we can add it back later if a use case comes up. That's why I'm in favor of removing the definitions for now and waiting for the use cases to become clearer.

linesthatinterlace:
Would you prefer I abandon this PR, and then we subsequently make a PR to remove Option.Rel? How do we synchronise moving it over to mathlib at the same time?

semorrison:
@linesthatinterlace, yes, let's abandon this PR, and separately remove `Option.Rel`.

The preferred way to do this is to create a PR removing it, and then create a PR at Mathlib which uses the Std PR as its Std dependency. As long as the Mathlib PR clearly explains that it shouldn't be merged until the Std PR is merged, it usually works!

PR #476, opened 2023-12-21
feat: lift StateM to StateT

JamesGallicchio:
I feel like this might already exist somewhere, but with a not too old Std/Mathlib imported it does not exist yet.

### DIFF
--- a/Std/Control/State.lean
+++ b/Std/Control/State.lean
@@ -1,0 +1,8 @@
+/-
+Copyright (c) 2023 James Gallicchio.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: James Gallicchio
+-/
+
+instance [Monad m] : MonadLift (StateM σ) (StateT σ m) where
+  monadLift s := pure ∘ s


fgdorais:
Maybe I'm not thinking about this right, but `StateM σ` is just `StateT σ Id`. Is there a reason why `MonadLift` doesn't happen automatically in such cases?

Std/Control/State.lean
```diff
@@ -0,0 +1,11 @@
...
+Copyright (c) 2023 James Gallicchio.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: James Gallicchio
+-/
+
+instance [Monad m] : MonadLift (StateM σ) (StateT σ m) where
+  monadLift s := pure ∘ s
+
+instance [MonadLift m n] : MonadLift (StateT σ m) (StateT σ n) where
+  monadLift s := fun state => liftM (s state)
```
joehendrix, line 11: Does this have performance or error message implications?  I think `MonadLift` is intended from outer to inner while this would potentially enable a search process.

fgdorais:

PR #481, opened 2023-12-22
feat: upstream mathlib definitions used for `ImportGraph`

joneugster:
In leanprover-community/mathlib4#9169 we move `ImportGraph` to its own package [lean-graph](https://github.com/leanprover-community/lean-graph).

To reduce its dependency on mathlib, some definitions need to be upstreamed to `Std`.

The mirror PR of this one is: leanprover-community/mathlib4#9195


Std/Lean/IO/Process.lean
```diff
@@ -0,0 +1,43 @@
...
+open System (FilePath)
+
+/--
+Pipe `input` into stdin of the spawned process,
+then return `(exitCode, stdout, stdErr)` upon completion.
+-/
+-- TODO We could reduce some code duplication by centralising some functions like this,
+-- which are used here, in `cache`, and in https://github.com/leanprover-community/llm.
+def runCmdWithInput' (cmd : String) (args : Array String)
+    (input : String := "") (throwFailure := true) : IO (UInt32 × String × String) := do
```
digama0, line 22:
```suggestion
    (input : String := "") (throwFailure := true) : IO Output := do
```
Isn't there a struct specifically for this?

digama0 (approved):

Std/Lean/Name.lean
```diff
@@ -33,3 +36,44 @@ def isInternalDetail : Name → Bool
...
+# Additional functions on `Lean.Name`.
+-/
+
+private def isBlackListed (declName : Name) : CoreM Bool := do
+  if declName.toString.startsWith "Lean" then return true
+  let env ← getEnv
+  pure $ declName.isInternalDetail
+   || isAuxRecursor env declName
+   || isNoConfusion env declName
+  <||> isRec declName <||> isMatcher declName
```
digama0, line 54: this definition is too tailored to the use case. I don't think it should go in std, and dependent functions should either be parameterized over it or they should also be removed from std.

Std/Lean/IO/Process.lean
joneugster: done

Std/Lean/Name.lean
joneugster: I removed it. can easily be added in the function `p : Name → Bool` (in mathlib) which is provided to the two commands.

joneugster: acutally, I moved all three declarations back into mathlib. Mainly because I realised it used `Std.HashMap` instead of `Lean.HashMap` which would complicate this PR further. Might upstream it in a future PR.

PR #485, opened 2023-12-27
chore: potential nontermination at `simp`

leodemoura:
Using `simp [forIn]` may cause nontermination with different `simp` reduction strategies. We have the following theorem
```
@[simp] theorem forIn_eq_forIn [Monad m] : @List.forIn α β m _ = forIn := rfl
```
Then, executing `simp [forIn]` in a term containing `forIn` iterating over a `List` may cause nontermination since the theorem above will fold it again.

Remark: the reduction strategy implemented
https://github.com/leanprover/lean4/pull/3118 triggers the nontermination.

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -354,14 +354,14 @@ theorem forIn_eq_data_forIn [Monad m]
       Array.forIn.loop as f i h b = forIn (as.data.drop j) b f
     | 0, _, _, _, rfl => by rw [List.drop_length]; rfl
     | i+1, _, _, j, ij => by
-      simp [forIn.loop]
+      simp only [forIn.loop, Nat.add]
       have j_eq : j = size as - 1 - i := by simp [← ij, ← Nat.add_assoc]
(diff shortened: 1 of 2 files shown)


digama0:
Why wasn't this already nonterminating before?

digama0:
In general this seems like it will be a problem with any use of simp lemmas for re-folding an unfolded notation, e.g. `Nat.add a b = a + b` (which are important for cleaning up a goal state when these appear due to some other tactic). Is there something we can do with tagging them specially or detecting that they are the inverse of a definitional equation and turning them off when the user explicitly uses `simp [(. + .)]` or the like?

joehendrix:
I noticed these lemmas also had non-terminal simp, so I went ahead and cleaned that up.

I think there are a ways this could be improved.  Ultimately, I think it'd be nice if we had some sort of simp modulo some limited equations so that, for example, `Nat.add` and `@Add.add Nat instAddNat` were equal for `simp`.  That's a ways off though.

joehendrix (approved):

PR #488, opened 2023-12-29
feat: add a `TacticM Unit` frontend for omega that can be used as an aesop rule

dupuisf:
This adds two `TacticM Unit` frontends for omega: `omegaTactic` which takes an `OmegaConfig` parameter, and `omegaDefault` which uses the default configuration. The latter can be used as an aesop rule (which must be a declaration of type `TacticM Unit`).

I've experimented a bit with this as a `get_elem_tactic` replacement, where I add `omega` as a safe rule together with some `forward` rules to bring useful facts into context before the omega call, and it works remarkably well!


dupuisf:
awaiting-review

semorrison (approved):
LGTM, and excited to see `aesop` integration!

joehendrix (approved):

PR #489, opened 2023-12-29
feat: implement union-find

fgdorais:
Migrating union-find implementation from Mario Carneiro's [Lean4Lean](https://github.com/digama0/lean4lean/blob/master/Lean4Lean/UnionFind.lean).

### DIFF
--- a/Std/Data/UnionFind/Basic.lean
+++ b/Std/Data/UnionFind/Basic.lean
@@ -165,52 +165,52 @@ def push (self : UnionFind) : UnionFind where
   rankD_lt := by simp [push_parentD, push_rankD]; exact self.rank_lt
 
 /-- Root of a union-find node -/
-def root' (self : UnionFind) (x : Fin self.size) : Fin self.size :=
+def root (self : UnionFind) (x : Fin self.size) : Fin self.size :=
   let y := (self.arr.get x).parent
   if h : y = x then
     x
   else
     have := Nat.sub_lt_sub_left (self.lt_rankMax x) (self.rank'_lt _ h)
-    self.root' ⟨y, self.parent'_lt x⟩
+    self.root ⟨y, self.parent'_lt x⟩
 termination_by _ => self.rankMax - self.rank x
 
 @[nolint unusedHavesSuffices]
-theorem parent'_root' (self : UnionFind) (x : Fin self.size) :
-    (self.arr.get (self.root' x)).parent = self.root' x := by
-  rw [root']; split <;> [assumption; skip]
+theorem parent_root (self : UnionFind) (x : Fin self.size) :
+    (self.arr.get (self.root x)).parent = self.root x := by
+  rw [root]; split <;> [assumption; skip]
   have := Nat.sub_lt_sub_left (self.lt_rankMax x) (self.rank'_lt _ ‹_›)
-  apply parent'_root'
+  apply parent_root
 termination_by _ => self.rankMax - self.rank x
 
 /-- Root of a union-find node -/
-def root (self : UnionFind) (x : Nat) : Nat :=
-  if h : x < self.size then self.root' ⟨x, h⟩ else x
+def root! (self : UnionFind) (x : Nat) : Nat :=
+  if h : x < self.size then self.root ⟨x, h⟩ else panicWith x "index out of bounds"
 
-theorem parent_root (self : UnionFind) (x : Nat) :
-    self.parent (self.root x) = self.root x := by
-  rw [root]; split <;>
-    [simp [parentD, parent'_root', -Array.get_eq_getElem]; simp [parentD_of_not_lt, *]]
+theorem parent_root! (self : UnionFind) (x : Nat) :
+    self.parent (self.root! x) = self.root! x := by
+  rw [root!]; split <;>
+    [simp [parentD, parent_root, -Array.get_eq_getElem]; simp [parentD_of_not_lt, *]]
 
 @[nolint unusedHavesSuffices]
-theorem root_parent (self : UnionFind) (x : Nat) : self.root (self.parent x) = self.root x := by
-  simp [root, parent_lt]; split <;> simp [parentD, parentD_of_not_lt, *, -Array.get_eq_getElem]
-  (conv => rhs; rw [root']); split
-  · rw [root', dif_pos] <;> simp [*, -Array.get_eq_getElem]
+theorem root_parent (self : UnionFind) (x : Nat) : self.root! (self.parent x) = self.root! x := by
+  simp [root!, parent_lt]; split <;> simp [parentD, parentD_of_not_lt, *, -Array.get_eq_getElem]
+  (conv => rhs; rw [root]); split
+  · rw [root, dif_pos] <;> simp [*, -Array.get_eq_getElem]
   · simp
 
-theorem root_lt {self : UnionFind} {x : Nat} : self.root x < self.size ↔ x < self.size := by
-  simp [root]; split <;> simp [*]
+theorem root_lt {self : UnionFind} {x : Nat} : self.root! x < self.size ↔ x < self.size := by
+  simp [root!]; split <;> simp [*]
 
 @[nolint unusedHavesSuffices]
-theorem root_eq_self {self : UnionFind} {x : Nat} : self.root x = x ↔ self.parent x = x := by
-  refine ⟨fun h => by rw [← h, parent_root], fun h => ?_⟩
-  rw [root]; split <;> [rw [root', dif_pos (by rwa [parent, parentD_eq' ‹_›] at h)]; rfl]
+theorem root_eq_self {self : UnionFind} {x : Nat} : self.root! x = x ↔ self.parent x = x := by
+  refine ⟨fun h => by rw [← h, parent_root!], fun h => ?_⟩
+  rw [root!]; split <;> [rw [root, dif_pos (by rwa [parent, parentD_eq' ‹_›] at h)]; rfl]
 
-theorem root_root {self : UnionFind} {x : Nat} : self.root (self.root x) = self.root x :=
-  root_eq_self.2 (parent_root ..)
+theorem root_root {self : UnionFind} {x : Nat} : self.root! (self.root! x) = self.root! x :=
+  root_eq_self.2 (parent_root! ..)
 
 theorem root_ext {m1 m2 : UnionFind}
-    (H : ∀ x, m1.parent x = m2.parent x) {x} : m1.root x = m2.root x := by
+    (H : ∀ x, m1.parent x = m2.parent x) {x} : m1.root! x = m2.root! x := by
   if h : m2.parent x = x then
     rw [root_eq_self.2 h, root_eq_self.2 ((H _).trans h)]
   else
@@ -218,7 +218,7 @@ theorem root_ext {m1 m2 : UnionFind}
     rw [← root_parent, H, root_ext H, root_parent]
 termination_by _ => m2.rankMax - m2.rank x
 
-theorem le_rank_root {self : UnionFind} {x : Nat} : self.rank x ≤ self.rank (self.root x) := by
+theorem le_rank_root {self : UnionFind} {x : Nat} : self.rank x ≤ self.rank (self.root! x) := by
   if h : self.parent x = x then
     rw [root_eq_self.2 h]; exact Nat.le_refl ..
   else
@@ -228,7 +228,7 @@ theorem le_rank_root {self : UnionFind} {x : Nat} : self.rank x ≤ self.rank (s
 termination_by _ => self.rankMax - self.rank x
 
 theorem lt_rank_root {self : UnionFind} {x : Nat} :
-    self.rank x < self.rank (self.root x) ↔ self.parent x ≠ x := by
+    self.rank x < self.rank (self.root! x) ↔ self.parent x ≠ x := by
   refine ⟨fun h h' => Nat.ne_of_lt h (by rw [root_eq_self.2 h']), fun h => ?_⟩
   rw [← root_parent]
   exact Nat.lt_of_lt_of_le (self.rank_lt h) le_rank_root
@@ -255,8 +255,8 @@ termination_by _ => self.rankMax - self.rank x
 
 @[nolint unusedHavesSuffices]
 theorem findAux_root {self : UnionFind} {x : Fin self.size} :
-    (findAux self x).root = self.root' x := by
-  rw [findAux, root']; simp; split <;> simp
+    (findAux self x).root = self.root x := by
+  rw [findAux, root]; simp; split <;> simp
   have := Nat.sub_lt_sub_left (self.lt_rankMax x) (self.rank'_lt _ ‹_›)
   exact findAux_root
 termination_by _ => self.rankMax - self.rank x
@@ -265,11 +265,11 @@ termination_by _ => self.rankMax - self.rank x
 theorem findAux_s {self : UnionFind} {x : Fin self.size} :
     (findAux self x).s = if (self.arr.get x).parent = x then self.arr else
       (self.findAux ⟨_, self.parent'_lt x⟩).s.modify x fun s =>
-        { s with parent := self.root x } := by
-  rw [show self.root _ = (self.findAux ⟨_, self.parent'_lt x⟩).root from _]
+        { s with parent := self.root! x } := by
+  rw [show self.root! _ = (self.findAux ⟨_, self.parent'_lt x⟩).root from _]
   · rw [findAux]; split <;> rfl
   · rw [← root_parent, parent, parentD_eq]
-    simp [findAux_root, root]
+    simp [findAux_root, root!]
     apply dif_pos
     exact parent'_lt ..
 
@@ -287,7 +287,7 @@ termination_by _ => self.rankMax - self.rank x
 
 theorem parentD_findAux {self : UnionFind} {x : Fin self.size} :
     parentD (findAux self x).s i =
-    if i = x then self.root x else parentD (self.findAux ⟨_, self.parent'_lt x⟩).s i := by
+    if i = x then self.root! x else parentD (self.findAux ⟨_, self.parent'_lt x⟩).s i := by
   rw [findAux_s]; split <;> [split; skip]
   · subst i; rw [root_eq_self.2 _] <;> simp [parentD_eq, *, -Array.get_eq_getElem]
   · rw [findAux_s]; simp [*, -Array.get_eq_getElem]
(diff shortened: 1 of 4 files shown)


Std/Data/UnionFind/Basic.lean
```diff
@@ -0,0 +1,438 @@
...
+namespace UnionFind
+
+/-- Size of union-find structure -/
+@[inline] abbrev size (self : UnionFind) := self.arr.size
+
+/-- Empty union-find structure -/
+def empty : UnionFind where
+  arr := #[]
+  parentD_lt := (fun.)
+  rankD_lt := fun.
```
joehendrix, line 73: Maybe `empty` should be defined as `mkEmpty 0`?

Std/Data/UnionFind/Basic.lean
```diff
@@ -0,0 +1,438 @@
...
+open UnionFind
+
+/-- Union-find data structure -/
+structure UnionFind where
+  /-- Array of union-find nodes -/
+  arr : Array UFNode
+  /-- Validity for parent nodes -/
+  parentD_lt : ∀ {i}, i < arr.size → parentD arr i < arr.size
+  /-- Validity for rank -/
+  rankD_lt : ∀ {i}, parentD arr i ≠ i → rankD arr i < rankD arr (parentD arr i)
```
joehendrix, line 62: Could you add more documentation to the structure here?

I think it should include names for the operations to:
* Create an empty UnionFind.
* Add a new union find node.
* Assert two union find nodes are equal.
* Check if two union find nodes are equal.

I don't think we've consistently done this for other operations, but at least in this case, I found it a bit difficult to find those operations.

digama0: The operations provided are:

1. `empty`: Create an empty UnionFind.
2. `push: Push a new UnionFind node not equal to anything else.
3. `find`: Find the canonical representative of a UnionFind node, and update the data structure to make subsequent lookups faster
4. `root`: Find the canonical representative of a UnionFind node without changing the structure
5. `union`: Join the equivalence classes of two UnionFind nodes.

Operations (3) and (5) give the data structure its name. The reason why this is not more abstracted is because *finding is a mutating operation*, and this fact is key to the performance of the data structure. You can use operation (4) to find instead of (3), but it would make the performance much worse and would be a performance pitfall because it has a nicer signature. Your two operations "Assert two union find nodes are equal" and "check if two union find nodes are equal" have possible implementations using (3) or (4), but only the version using (4) can be used for specification purposes and we don't want people to stumble into doing that instead of using (3).

Std/Data/UnionFind/Basic.lean
```diff
@@ -52,7 +52,21 @@ end UnionFind
 
 open UnionFind
 
-/-- Union-find data structure -/
+/-- Union-find data structure
```
digama0, line 55:
```suggestion
/-- ### Union-find data structure
```
or put a period at the end of the phrase

Std/Data/UnionFind/Basic.lean
```diff
@@ -52,7 +52,21 @@ end UnionFind
...
+
+The `UnionFind` structure is an implementation of disjoint-set data structure
+that uses path compression to make the primary operations run in amortized
+nearly linear time. The key operations for `UnionFind` are:
+
+* `empty`/`mkEmpty` are used to create a new empty structure.
+* `push` adds a new node to a structure, unlinked to any other node.
+* `union` links two nodes of the data structure, joining their equivalence
```
digama0, line 63:
```suggestion
* `size` returns the number of elements in the structure.
```

Std/Data/UnionFind/Basic.lean
fgdorais: Let me know if I missed something.

fgdorais: I moved `UnionFind.Equiv` to `Basic` even though its (deliberately) noncomputable. That just made more sense to me.

Std/Data/UnionFind/Basic.lean
```diff
@@ -0,0 +1,465 @@
...
+/-- Find root of a union-find node, updating the structure using path compression -/
+def find (self : UnionFind) (x : Fin self.size) :
+    (s : UnionFind) × {_root : Fin s.size // s.size = self.size} :=
+  let r := self.findAux x
+  { 1.arr := r.s
+    2.1.val := r.root
+    1.parentD_lt := fun h => by simp [FindAux.size_eq] at *; exact parentD_findAux_lt h
+    1.rankD_lt := fun h => by rw [rankD_findAux, rankD_findAux]; exact lt_rankD_findAux h
+    2.1.isLt := show _ < r.s.size by rw [r.size_eq]; exact r.root.2
+    2.2 := by simp [size, r.size_eq] }
```
joehendrix, line 349: The use of `Fin s.size` to denote elements in the union-find seems fairly painful.

I'd prefer something simpler like `def Node := Nat` and define a nodes outside the range `s.arr.size` as single element sets.  Alternatively, one could grow the array as needed if union is passed nodes outside the size.

With this end up with signatures like:

```lean
def Node := Nat
def find (self : UnionFind) (x : Node) : UnionFind ×  Node := ...
def union (self : UnionFind) (x y : Node) : UnionFind  := ...
```

Other alternatives that may be worth considering are a parameter for the size or automatically growing the array if `union` is passed an index that is larger than the current size.  Then it's effectively a equivalence relation over `Nat`.

digama0: I agree, and some of these functions already exist in `Lemmas` I think. I don't like the idea of dynamically increasing the size of the structure if you pass out of bounds elements though, because that makes it hard to keep track of the size and it also makes the runtime of the function complicated. Only `union` has this problem though, `find` can validly be extended outside the `Fin` domain. In lean4lean I extended both functions outside the domain by doing nothing when passed invalid indexes, but this is only correct for `find`. Maybe `union` should panic and then do that?

joehendrix: Yeah, I agree on the auto-resizing.  How about just `Nat` based `find!` and `union!` and `find` and `union` take proofs:

```lean
def find (self : UnionFind) (x : Fin self.size) : UnionFind × Nat := ...
def union (self : UnionFind) (x y : Fin self.size) : UnionFind  := ...

def find! (self : UnionFind) (x : Nat) : UnionFind × Nat := ...
def union! (self : UnionFind) (x y : Nat) : UnionFind  := ...
```

If `find!` is passed an invalid index it panics and returns its arguments, if `union!` is passed an invalid index it panics and returns its input.  I suspect most users would prefer the `Nat` versions (unless they get hurt by panics).

I'd also recommend a `checkEquals!` (maybe I overlooked an equivalent version).
```lean
def checkEquals! (self : UnionFind) (x y : Nat) : UnionFind × Bool := do
  let (self, x) := self.find! x
  let (self, y) := self.find! y
  pure (self, x == y)
```lean

A `Fin` version would be useful too; maybe `proof_wanted` could be used if that's too much right now.

Comment (digama0) [reply Std/Data/UnionFind/Basic.lean:349]: The extra arguments in the Fin version are there for the convenience of working *with* the Fin version; it is somewhat more cumbersome to write `union` without it, for example.

Comment (fgdorais) [reply Std/Data/UnionFind/Basic.lean:349]: I like a lot of these ideas! I'm happy to implement them once this discussion reaches a conclusion.

I think `checkEquals!` should panic before any work is done, and I would call it `checkEquiv!'.

Comment (joehendrix) [reply Std/Data/UnionFind/Basic.lean:349]: I think the panic up front sounds good.  I'm excited to have this in Std.

Given that there will be at least one auxiliary array, perhaps the variable count should be codified as an external value, e.g:
```lean
def push : UnionFind n -> UnionFind (n + 1) := ...
def find (self : UnionFind n) (x : Fin n) : UnionFind n × Fin n := ...
def union (self : UnionFind n) (x y : Fin n) : UnionFind n := ...
def checkEquiv (self : UnionFind n) (x y : Fin n) : UnionFind n × Bool := ...

-- or just using individual operations:

def find (self : UnionFind) (x : Fin n)  (_ : n = self.size) := UnionFind × Fin n := ...
def union (self : UnionFind) (x y : Fin n) (_ : n = self.size) : UnionFind := ...
def checkEquiv (self : UnionFind) (x y : Fin n)  (_ : n = self.size) : UnionFind × Bool := ...
```

I think this will simplify proofs - particularly with `union` and `checkEquiv`.  A default proof strategy could be used for the `n = self.size` proofs.

I have a slight personal preference for the explicit parameter in `UnionFind n`, but I'd accept this either way.  `find` should have a clearly readable type signature given it's in the name of the type.

Independent of the signatures, it may be good to clarify performance expectations.  I haven't audited it carefully, but presumedly the goal is to have destructive updates on the array when the `UnionFind` is not shared.  Assuming that's the case, I'd document that in the above operations.

We might eventually also eventually want a `PersistentUnionFind` or backtrackable union find (where union is destructive but can be reverted).

fgdorais: I added the easiest parts. Adding the parameter `n` is a more major task so I'll wait for Mario to chime in before starting on that.

PS: Where should I put `panicWith`? Or perhaps it already exists somewhere...

digama0: Having an index `n` would simplify proofs, but it would also cause additional indirections and redundant data storage because you will usually have to wrap this in something else that stores the `n`, even though the `n` is already stored in the structure. If we want this I think we should first pursue it for `Array A`, which has much the same concerns (indeed `UnionFind` is just a wrapper around a `Array UFNode`, and `UnionFind n` should be a wrapper around `Array UFNode n`). I think there is a zulip thread about this.

Std/Data/UnionFind/Basic.lean
```diff
@@ -165,60 +165,60 @@ def push (self : UnionFind) : UnionFind where
...
   have := Nat.sub_lt_sub_left (self.lt_rankMax x) (self.rank'_lt _ ‹_›)
-  apply parent'_root'
+  apply parent_root
 termination_by _ => self.rankMax - self.rank x
 
 /-- Root of a union-find node -/
-def root (self : UnionFind) (x : Nat) : Nat :=
-  if h : x < self.size then self.root' ⟨x, h⟩ else x
+def root! (self : UnionFind) (x : Nat) : Nat :=
+  if h : x < self.size then self.root ⟨x, h⟩ else panicWith x "index out of bounds"
```
digama0, line 187: No I want `root` too, maybe it should be called `rootD`? It is mathematically meaningful (as is `findD`), it represents the root operation when you extend it with an infinity of singleton classes / self-loops. It plays a pretty central role in a lot of the theorems here.

Std/Data/UnionFind/Basic.lean
fgdorais: I looked into adding a parameter `n` but it looks like a lot of work, so I added operations as in Joe's alternative suggestion. They're called `findN`, `unionN`, etc. Let me know if you have a better naming scheme for these.

joehendrix: Thanks, I'm happy with the naming of these.

I don't think I will have time to review the proofs carefully in the next week, but I think it's worth adding this type.

joehendrix (approved):

Std/Data/UnionFind/Basic.lean
fgdorais: I don't think there's pressure to merge this very soon. I will be relatively busy for the next couple of weeks (it's the start of the spring semester over here). I will still be responsive but I might not have time for major stuff. Of course, Mario should feel free to edit/audit/manage/etc this PR as needed since he is the true author!

Vtec234: I am late to the discussion, but curious: why the flat array that special-cases the equivalence to be over `Nat`?

It is also possible to implement union-find over an arbitrary type `α` using a map from `α` to `UFNode α`, e.g. with a hashmap if `Hashable α`. The array is effectively special-casing this map to be `ℕ → UFNode Nat`. Way back when I used the mathlib4 `UnionFind` for something, I immediately wrapped it with a `HashMap α (Fin uf.size)` that allowed me to compute equivalence classes of `α`s rather than `Nat`s/`Fin`s. I would guess that a polymorphic API like this would be preferrable for most users who want to equate things that aren't numbers, and the structure here could serve as an implementation detail of that. I worry that the additional indirection introduces some overhead that would not be there if the base structure was already polymorphic; but of course it's impossible to tell without benchmarking. Am I missing something as to the rationale?

(This is building on an [earlier comment](https://github.com/leanprover-community/mathlib4/pull/139#issuecomment-1023516152).)

fgdorais: The union-find structure here is designed for a wide variety of use cases. Since this is a widespread algorithm that is hard to get right (nevermind provably right!) it's preferable to supply a bare bones yet flexible interface rather than one that fits a specific use pattern.

Vtec234: Sorry, I am having some trouble understanding this answer. What other use cases do you have in mind besides storing some number of disjoint subsets of a type? If there aren't any others, wouldn't it be more general to store disjoint subsets of arbitrary `α` than only of `Nat`?

fgdorais: The prototypical use case works the other way than you describe: the `UnionFind` structure is coupled with an array of actual objects and the nodes are indices for that array.

Vtec234: Still not following I'm afraid. If you are referring to Joe Hendrix's example
```lean
structure UnionEqRel (α : Type) where
  elts : Array α
  uf : UnionFind
```

it can be implemented with a polymorphic `UnionFind` as

```lean
structure UnionEqRel (α : Type) where
  elts : Array α
  uf : UnionFind Nat -- or UnionFind (Fin elts.size)
```
where the nodes are indeed indices into `elts`.

joehendrix: My inclination is to use performance to guide this decision.

It'd be greatl to get some data on relative performance of a polymorphic `UnionFind Nat` vs `UnionFind` as is.   Additionally one could compare `UnionFind α` vs the above `UnionEqRel α`.

I know this is time consuming, but we should really be using some numbers to inform decisions like this.

fgdorais: I guess I don't understand how your polymorphism idea works. Would you be happy with just enlarging the node type to contain a value? 
```
/-- Union-find node type -/
structure UFNode (α) where
  /-- Value of node -/
  value : α
  /-- Parent of node -/
  parent : Nat
  /-- Rank of node -/
  rank : Nat
```
This is not different than maintaining a parallel `Array α` as in Joe's example.

Or do you envision something where `find (self : UnionFind α) (x : α) : UnionFind α × α` and `union (self : UnionFind α) (x y : α) : UnionFind α`? If so, what kind of mapping `α → α` are you envisioning for the underlying forest structure?

digama0: > I guess I don't understand how your polymorphism idea works. Would you be happy with just enlarging the node type to contain a value?
> 
> ```
> /-- Union-find node type -/
> structure UFNode (α) where
>   /-- Value of node -/
>   value : α
>   /-- Parent of node -/
>   parent : Nat
>   /-- Rank of node -/
>   rank : Nat
> ```
> 
> This is not different than maintaining a parallel `Array α` as in Joe's example.

Note that this is how the mathlib `UnionFind` was designed, and I found that in use cases I often had to set `α` to `Unit` and then handle the auxiliary data in a different way, which is to say the gambit that this is how the user data is stored didn't really pan out.

The current design is quite minimalistic in terms of other data, it can be wrapped in a variety of ways and you probably want to do so for your application. But I don't see a lot of commonality in that wrapping and it's better to have a primitive and wrap it rather than have some default wrapping, stub it out and then do some additional wrapping for what you actually wanted, especially if that adds additional operations on the hot path like hashmap lookups.

Vtec234: > Or do you envision something where `find (self : UnionFind α) (x : α) : UnionFind α × α` and `union (self : UnionFind α) (x y : α) : UnionFind α`? If so, what kind of mapping `α → α` are you envisioning for the underlying forest structure?

Exactly this one! Regarding the mapping from `α`s to `UFNode α`s, a `Hashable α` requirement would be my choice, but you could do with just `BEq α`: the usual choice one has when implementing maps out of `α`. I had implemented a dead-simple version a while ago [here](https://github.com/Vtec234/mathlib4/blob/9db22e5fc9df541168805a06a5aba062415a2d2b/Mathlib/Data/DisjointSet.lean).

But actually, even though I started talking about implementation details, this isn't really what my question was about; sorry for not making that clearer. I only mean to point out that the API that takes in `α`s and consumes `α`s has been easier to use in my experience, as well as in others' experiences that they reported to me. The one here feels like a special-case, potentially faster (though I completely agree with Joe that we can't really say such things without actually benchmarking a realistic workload) version; this is not to say it shouldn't be merged, I would just say that std4 ought to have the simple, general version together with the specialized one. It could probably be implemented on top the specialized one by also storing some representation of a map `α → Nat` recording the index of each `α` in the `UnionFind`.

Vtec234: > We can certainly provide a variant of this data structure that combines it with extra data like `Array α`, or `HashMap α Nat` or what have you, but these should be separate types because the core type here is also useful as a building block for those.

I missed this when writing my response above. This approach makes sense to me. Thanks for the explanation!

fgdorais: Here is an implementation for the polymorphic version you need:
```lean
structure PUnionFind (α : Type _) [Hashable α] [BEq α] where
  uf : UnionFind
  value : Array α
  index : HashMap α Nat := .empty

namespace PUnionFind
variable {α : Type _} [Hashable α] [BEq α]

abbrev size (s : PUnionFind α) := s.uf.size

def push (s : PUnionFind α) (x : α) : PUnionFind α :=
  if s.index.contains x then s else {
    uf := s.uf.push
    value := s.value.push x
    index := s.index.insert x s.size
  }

def find! (s : PUnionFind α) (x : α) : PUnionFind α × α :=
  let _ : Inhabited α := ⟨x⟩
  if s.index.contains x then
    match s.uf.find! (s.index.find! x) with
    | (uf, i) => ({s with uf}, s.value[i]!)
  else UnionFind.panicWith (s, x) "key not found"

def union! (s : PUnionFind α) (x y : α) : PUnionFind α :=
  let _ : Inhabited α := ⟨x⟩
  if s.index.contains x ∧ s.index.contains y then
    {s with uf := s.uf.union! (s.index.find! x) (s.index.find! y)}
  else UnionFind.panicWith s "key not found"
```
We can definitely add something like this after this PR gets through.

PR #491, opened 2023-12-31
feat: three array lemmas

fgdorais:
These three lemmas are useful for induction on array size.

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -182,7 +182,7 @@ theorem eq_push_pop_back_of_size_ne_zero [Inhabited α] {as : Array α} (h : as.
       cases heq; rw [get_push_eq, back, ←size_pop, get!_eq_getD, getD, dif_pos h]; rfl
 
 theorem eq_push_of_size_ne_zero {as : Array α} (h : as.size ≠ 0) :
-  ∃ (bs : Array α) (c : α), as = bs.push c :=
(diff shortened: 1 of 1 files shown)


Std/Data/Array/Lemmas.lean
```diff
@@ -164,6 +164,23 @@ theorem swapAt!_def (a : Array α) (i : Nat) (v : α) (h : i < a.size) :
     a.pop[i] = a[i]'(Nat.lt_of_lt_of_le (a.size_pop ▸ hi) (Nat.sub_le _ _)) :=
   List.get_dropLast ..
 
+theorem eq_empty_of_size_eq_zero {as : Array α} (h : as.size = 0) : as = #[] := by
+  apply ext
+  · simp [h]
+  · intros; contradiction
+
+theorem eq_push_of_size_ne_zero [Inhabited α] {as : Array α} (h : as.size ≠ 0) :
+    as = as.pop.push as.back := by
```
digama0, line 173: Might be nice to have a version of this with an existential too, so that rewrites don't get confused

fgdorais: Do you mean something like this?
```lean
theorem eq_push_of_size_ne_zero' {as : Array α} (h : as.size ≠ 0) : ∃ (bs : Array α) (c : α), as = bs.push c := sorry
```

fgdorais: Ah! Right, the `Inhabited` instance will cause rewrite issues.

joehendrix (approved):

PR #494, opened 2024-01-02
fix: don't panic when linting in empty environments

david-christiansen:
The "unreachable tactic" linter would previously assume that various parser categories were in the environment, and panic if they weren't. This caused problems when elaborating declarations in a fresh, empty environment.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,5 +1,8 @@
 TESTS = $(wildcard test/*.lean)
 
+# Ensure that panics actually cause the tests to fail
+export LEAN_ABORT_ON_PANIC=1
+
 .PHONY: all build test lint
 
 all: build test
--- a/Std/Linter/UnreachableTactic.lean
+++ b/Std/Linter/UnreachableTactic.lean
@@ -94,8 +94,11 @@ def unreachableTacticLinter : Linter where run := withSetOptionIn fun stx => do
   if (← get).messages.hasErrors then
     return
   let cats := (Parser.parserExtension.getState (← getEnv)).categories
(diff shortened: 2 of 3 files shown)


david-christiansen:
Where "this" means "the fact that the linters are running" rather than "find! was used", right?

Kha:
Yes!

joehendrix (approved):

joehendrix:
I'm not opposed to fixing this particular issue.  I suspect we'll run into more issues when running linters (or other meta-programming) in restricted environments that don't have the same environment as the main environment.  

It'd be good to have better regression testing.

david-christiansen:
I'll see if I can cook up a little test for it.

david-christiansen:
[2288140](https://github.com/leanprover/std4/pull/494/commits/228814099130cbe53821ab968862e13562d662f9) adds a regression test. I've manually verified that it would have caught the bug that's fixed in this PR, but I do suspect that there are many other bugs it wouldn't have caught.

david-christiansen:
One side effect of the way that I implemented the regression test is that panics during elaboration of any of the tests will now cause them to fail, rather than just write to stderr. I think this is a good thing, but I thought I should point it out. No tests currently panic, so it doesn't have an effect on existing tests.

joehendrix:
The changes looked good to me.  If we ever want to test  intentional panics, then we may need to support unsetting `LEAN_ABORT_ON_PANIC`.  However, that can be done if/when needed.

david-christiansen:
Thanks!

We can definitely cross that bridge when we get to it, I think it'd be a pretty small change to the Makefile.

PR #495, opened 2024-01-02
feat: add `String.splitOn_of_valid`

chabulhwi:
* Add lemmas for `Nat.le` and `List.append`.
* Add functions on `List`:
  * `List.rdropSublist` drops a sublist from the tail end of a list.
  * `List.splitOnSublist` splits a list at every occurrence of a separator sublist. The separators are not in the result.
* Add `simp` lemmas for `List.rdropSublist`.
* State and prove two theorems about `String`: `splitOnAux_of_valid` and `splitOn_of_valid`.

---

Before this pull request gets merged, I need to replace `String.splitOnAux` and `String.splitOn` in Lean core with the modified versions I made:

```lean
def splitOnAux' (s sep : String) (b : Pos) (i : Pos) (j : Pos) (r : List String) : List String :=
  if sep == "" then
    let r := (s.extract b s.endPos)::r
    r.reverse
  else if h : s.atEnd i then
    let r := (s.extract b i)::r
    r.reverse
  else
    have := Nat.sub_lt_sub_left (Nat.gt_of_not_le (mt decide_eq_true h)) (lt_next s _)
    if sep.atEnd j then
      if s.get i == sep.get 0
        splitOnAux' s sep i (s.next i) (sep.next 0) (s.extract b (i - j)::r)
      else
        splitOnAux' s sep i (s.next i) 0 (s.extract b (i - j)::r)
    else if s.get i == sep.get j then
      splitOnAux' s sep b (s.next i) (sep.next j) r
    else
      splitOnAux' s sep b (s.next i) 0 r
termination_by _ => s.endPos.1 - i.1

def splitOn' (s : String) (sep : String := " ") : List String :=
  splitOnAux' s sep 0 0 0 []
```

The function `splitOnAux'` differs from its original version in two ways. First, it checks whether the separator (`sep`) is empty. Second, it checks whether `sep.atEnd j` *before* it checks whether `s.get i == sep.get j`. For these reasons, the outputs of the original `splitOnAux` and its modified version are different for some inputs, as shown in the `#eval` commands below:

```lean
import Std.Data.String.Lemmas

namespace Test

open String
set_option linter.missingDocs false

private def l := "A12a".1
private def m := "B12b".1
private def r := "C12c".1

private def sep₁ := "1".1
private def sep₂ := "2".1
private def acc := ["X", "x"]

#eval splitOnAux  ⟨l++m++sep₁++r⟩ ⟨sep₁++sep₂⟩ ⟨utf8Len l⟩ ⟨utf8Len (l++m++sep₁)⟩
  ⟨utf8Len (sep₁++sep₂)⟩ acc -- output: ["x", "X", "B12b1C", "c"]
#eval splitOnAux' ⟨l++m++sep₁++r⟩ ⟨sep₁++sep₂⟩ ⟨utf8Len l⟩ ⟨utf8Len (l++m++sep₁)⟩
  ⟨utf8Len (sep₁++sep₂)⟩ acc -- ["x", "X", "B12", "C", "c"]

#eval splitOnAux  ⟨l++m++"A".1++r⟩ ⟨"A".1++sep₂⟩ ⟨utf8Len l⟩ ⟨utf8Len (l++m++"A".1)⟩
  ⟨utf8Len ("A".1++sep₂)⟩ acc -- ["x", "X", "B12bAC12c"]
#eval splitOnAux' ⟨l++m++"A".1++r⟩ ⟨"A".1++sep₂⟩ ⟨utf8Len l⟩ ⟨utf8Len (l++m++"A".1)⟩
  ⟨utf8Len ("A".1++sep₂)⟩ acc -- ["x", "X", "B12", "C12c"]

end Test
```

I'll open an issue on [Lean 4 GitHub issues][0] after I get feedback about `splitOnAux'` and `splitOn'`.

[0]: https://github.com/leanprover/lean4/issues

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -219,6 +219,18 @@ theorem replicateTR_loop_eq : ∀ n, replicateTR.loop a n acc = replicate n a ++
 @[csimp] theorem dropLast_eq_dropLastTR : @dropLast = @dropLastTR := by
   funext α l; simp [dropLastTR]
 
+/-- Drop a sublist from the tail end of a list. -/
+def rdropSublist [BEq α] : List α → List α → List α
+  | l, [] => l
+  | [], _ => []
+  | a::as, b::bs =>
+    if a::as == b::bs then
+      []
+    else if as == b::bs then
+      [a]
+    else
+      a::rdropSublist as (b::bs)
+
 /-- Tail recursive version of `intersperse`. -/
 def intersperseTR (sep : α) : List α → List α
   | [] => []
@@ -520,6 +532,29 @@ Split a list at every occurrence of a separator element. The separators are not 
 ```
 -/
 @[inline] def splitOn [BEq α] (a : α) (as : List α) : List (List α) := as.splitOnP (· == a)
+
+/--
+Split a list at every occurrence of a separator sublist. The separators are not in the result.
+```
+[1, 1, 2, 3, 2, 4, 4].splitOnSublist [2, 3] = [[1, 1], [2, 4, 4]]
+```
+-/
+def splitOnSublist [BEq α] (l sep : List α) : List (List α) := go l sep sep [] where
+  /-- Auxiliary for `splitOnSublist`. The first explicit argument is a list we want to split, the
+  second a separator sublist (`sep`), the third a part of `sep` that we'll use to split the first
+  argument, and the last an accumulated list of elements of the first argument (`acc`). -/
+  go : List α → List α → List α → List α → List (List α)
+  | a::as, sep@(b::bs), [], acc =>
+    if a == b then
+      (acc.reverse.rdropSublist sep)::go as sep bs [a]
+    else
+      (acc.reverse.rdropSublist sep)::go as sep sep [a]
+  | a::as, sep@(_::_), c::cs, acc =>
+    if a == c then
+      go as sep cs (a::acc)
+    else
+      go as sep sep (a::acc)
+  | l, _, _, acc => [acc.reverse++l]
 
 /--
 Apply a function to the nth tail of `l`. Returns the input without
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -161,6 +161,18 @@ theorem append_eq_append_iff {a b c d : List α} :
   | nil => simp; exact (or_iff_left_of_imp fun ⟨_, ⟨e, rfl⟩, h⟩ => e ▸ h.symm).symm
   | cons a as ih => cases c <;> simp [eq_comm, and_assoc, ih, and_or_left]
 
+theorem append_left_eq_self (l m : List α) : l ++ m = m ↔ l = [] := by
+  constructor <;> intro h
+  · conv at h => rhs; rw [← nil_append m]
+    rwa [append_cancel_right_eq] at h
+  · rw [h, nil_append]
+
+theorem append_right_eq_self (l m : List α) : l ++ m = l ↔ m = [] := by
+  constructor <;> intro h
+  · conv at h => rhs; rw [← append_nil l]
+    rwa [append_cancel_left_eq] at h
+  · rw [h, append_nil]
+
 @[simp] theorem mem_append {a : α} {s t : List α} : a ∈ s ++ t ↔ a ∈ s ∨ a ∈ t := by
   induction s <;> simp_all [or_assoc]
 
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -732,6 +732,29 @@ are often used for theorems about `Array.pop`.  -/
   | _::_::_, ⟨0, _⟩ => rfl
   | _::_::_, ⟨i+1, _⟩ => get_dropLast _ ⟨i, _⟩
 
+/-! ### rdropSublist -/
+
+@[simp] theorem rdropSublist_self [BEq α] [LawfulBEq α] (l : List α) :
+    l.rdropSublist l = [] := by
+  cases l <;> simp [rdropSublist]
+
+@[simp] theorem rdropSublist_append [BEq α] [LawfulBEq α] (l m : List α) :
+    (l++m).rdropSublist m = l := by
+  match l, m with
+  | [], _ => simp [rdropSublist_self]
+  | a::as, [] => simp [rdropSublist]
+  | a::as, b::bs =>
+    simp [rdropSublist]
+    split
+    · next h =>
+      simp at h
+      rw [append_cons, append_left_eq_self] at h
+      simp at h
+    · split <;> rename_i h <;> simp at h ⊢
+      · simp [append_left_eq_self] at h
+        exact h.symm
+      · apply rdropSublist_append
+
 /-! ### nth element -/
 
 @[simp] theorem get_cons_succ {as : List α} {h : i + 1 < (a :: as).length} :
--- a/Std/Data/Nat/Lemmas.lean
+++ b/Std/Data/Nat/Lemmas.lean
@@ -370,6 +370,16 @@ protected theorem add_le_add_iff_left {n : Nat} : n + m ≤ n + k ↔ m ≤ k :=
 protected theorem add_le_add_iff_right {n : Nat} : m + n ≤ k + n ↔ m ≤ k :=
   ⟨Nat.le_of_add_le_add_right, fun h => Nat.add_le_add_right h _⟩
 
+protected theorem add_left_le_self {n m : Nat} : n + m ≤ m ↔ n = 0 := by
+  conv => lhs; rhs; rw [← Nat.zero_add m]
+  rw [Nat.add_le_add_iff_right]
+  apply Nat.le_zero
+
+protected theorem add_right_le_self {n m : Nat} : n + m ≤ n ↔ m = 0 := by
+  conv => lhs; rhs; rw [← Nat.add_zero n]
+  rw [Nat.add_le_add_iff_left]
+  apply Nat.le_zero
+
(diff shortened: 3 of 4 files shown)


digama0:
But why? You say that `splitOnAux` has different behavior, but this is an auxiliary function, its behavior doesn't matter. Does `splitOn` have different behavior?

Checking the separator for emptiness in `splitOnAux` is a bad idea, this is the inner loop of `splitOn` and `sep` is a constant. Just check it outside the loop and pass in an appropriate hypothesis.

digama0:
Looking at the definitions of `splitOn` and `splitOnAux`, I think they are correct. `splitOnAux` has the invariant that `sep != ""`, and `j < sep.endPos`, which justifies the two aspects that you felt the need to change.

chabulhwi:
@digama0 There are some errors in `splitOnAux'`, which make `splitOn'` behave differently from `splitOn`. I'll reply to your questions after I fix those.

```lean
#eval "".splitOn "_" == "".splitOn' "_" -- output: true
#eval "My<>life<>for<>Aiur!".splitOn "<>" == "My<>life<>for<>Aiur!".splitOn' "<>" -- true
#eval "BABABA".splitOn  "A" -- ["B", "B", "B", ""]
#eval "BABABA".splitOn' "A" -- ["B", "B", "BA"]
```

fgdorais:
Why not use the KMP matcher at Std.Data.Array.Match?

digama0:
@chabulhwi Based on your example, it seems like `splitOn` is correct and `splitOn'` is not, no? If you split `"BABABA"` by `"A"` the correct answer is `["B", "B", "B", ""]`, and if you split `"ABABAB"` by `"A"` the correct answer is `["", "B", "B", "B"]`. `splitOn` gets both right, `splitOn'` only gets the second one right.

chabulhwi:
Right. But I think I can fix this error.

digama0:
Why not just use and prove properties of the existing function?

Std/Data/List/Basic.lean
```diff
@@ -521,6 +533,29 @@ Split a list at every occurrence of a separator element. The separators are not
 -/
 @[inline] def splitOn [BEq α] (a : α) (as : List α) : List (List α) := as.splitOnP (· == a)
 
+/--
+Split a list at every occurrence of a separator sublist. The separators are not in the result.
+```
+[1, 1, 2, 3, 2, 4, 4].splitOnSublist [2, 3] = [[1, 1], [2, 4, 4]]
+```
+-/
+def splitOnSublist [BEq α] (l sep : List α) : List (List α) := go l sep sep [] where
```
digama0, line 542: This name is confusing because we currently use `sublist` to mean a non-contiguous subsequence, see `List.Sublist`. How about just `splitOnList`?

chabulhwi: Sounds great. I'll rename it after I finish other work.

Std/Data/List/Basic.lean
```diff
@@ -521,6 +533,29 @@ Split a list at every occurrence of a separator element. The separators are not
...
+    if a == b then
+      (acc.reverse.rdropSublist sep)::go as sep bs [a]
+    else
+      (acc.reverse.rdropSublist sep)::go as sep sep [a]
+  | a::as, sep@(_::_), c::cs, acc =>
+    if a == c then
+      go as sep cs (a::acc)
+    else
+      go as sep sep (a::acc)
+  | l, _, _, acc => [acc.reverse++l]
```
digama0, line 557: This implementation is also too complex to be a good "naive algorithm" for specification purposes and not efficient enough to be the "optimized version" of the algorithm. Since this is already intended as specification for `String.splitOn`, my recommendation is to lean into the "naive algorithm" side of things for now and add an "optimized version" later (possibly using KMP):
```lean
def splitOnList [BEq α] (l sep : List α) : List (List α) :=
  if _h : sep.isEmpty then
    [l]
  else
    let rec go (l acc : List α) : List (List α) :=
      if h' : l.isPrefixOf sep then
        have : length l - sep.length < length l := sorry
        acc.reverse :: go (l.drop sep.length) []
      else
        match l with
        | a :: l => go l (a :: acc)
        | [] => h'.elim (by simp [List.isPrefixOf])
    go l []
termination_by go => l.length
```
This is still a bit too imperative for direct usage; I think you actually want to use the lemmas:
```lean
theorem splitOnList_of_infix [DecidableEq α] (hsep : sep ≠ []) (h : sep <:+: l) :
    ∃ s t, s ++ sep ++ t = l ∧ splitOnList l sep = s :: splitOnList t sep := sorry

theorem splitOnList_of_not_infix [DecidableEq α] (hsep : sep ≠ []) (h : ¬sep <:+: l) :
    splitOnList l sep = [l] := sorry
```

chabulhwi:
(I'll write comments within a few hours...)

Std/Data/List/Basic.lean
digama0: Actually this might be a better / more useful factoring:
```lean
/-- Returns `(l₁, l₂)` for the first split `l = l₁ ++ l₂` such that `P l₂` returns true. -/
def splitOnceP (P : List α → Bool) (l : List α) : Option (List α × List α) :=
  if P l then
    some ([], l)
  else
    match l with
    | [] => none
    | a :: l => (splitOnceP P l).map fun (l, r) => (a :: l, r)

def splitOnList [BEq α] (l sep : List α) : List (List α) :=
  if _h : sep.isEmpty then
    [l]
  else
    match e : splitOnceP sep.isPrefixOf l with
    | none => [l]
    | some (l₁, l₂) =>
      have : length l₂ - sep.length < length l := sorry
      splitOnList (l₂.drop sep.length) sep
termination_by _ => l.length
```

chabulhwi:
> But why? You say that `splitOnAux` has different behavior, but this is an auxiliary function, its behavior doesn't matter. Does `splitOn` have different behavior?
>
> Checking the separator for emptiness in `splitOnAux` is a bad idea, this is the inner loop of `splitOn` and `sep` is a constant. Just check it outside the loop and pass in an appropriate hypothesis.

I think the current definition of `splitOnAux` makes it more difficult to prove the following two theorems:

```lean
theorem splitOnAux_of_valid (sep₁ sep₂ l m r acc) :
    splitOnAux ⟨l++m++sep₁++r⟩ ⟨sep₁++sep₂⟩ ⟨utf8Len l⟩ ⟨utf8Len (l++m++sep₁)⟩ ⟨utf8Len sep₁⟩ acc =
      acc.reverse++(List.splitOnList.go r (sep₁++sep₂) sep₂ (m++sep₁).reverse).map mk := by
  sorry

theorem splitOn_of_valid (s sep) : splitOn s sep = (List.splitOnList s.1 sep.1).map mk := by
  simpa [splitOn] using splitOnAux_of_valid [] sep.1 [] [] s.1 []
```

When proving `splitOnAux_of_valid`, we need to consider the case where the separator is empty. Having `splitOnAux` check the separator for emptiness helps simplify both sides of the equation we need to prove.

chabulhwi:
> Looking at the definitions of `splitOn` and `splitOnAux`, I think they are correct. `splitOnAux` has the invariant that `sep != ""`, and `j < sep.endPos`, which justifies the two aspects that you felt the need to change. (Doing the `s.get i = sep.get pos` check first is a better design, because most characters are not going to be the same as the first character in `sep`, so this saves one test per character.)

In order to prove `splitOnAux_of_valid`, we also have to deal with the case where it's true that `sep.atEnd j`; this is where the problem with the current `splitOnAux` lies.

When `sep.atEnd j` is true, the value of `sep.get j` is `default`, which is `'A'`. In that event, checking whether `s.get i == sep.get j` is equivalent to checking whether `s.get i == 'A'`. This leads to the inconsistent behavior of `splitOnAux` shown below:

```lean
#eval "AB".splitOnAux "C" ⟨0⟩ ⟨0⟩ ⟨1⟩ []
#eval "AB".splitOnAux "C" ⟨1⟩ ⟨1⟩ ⟨0⟩ [""]
#eval "AB".splitOnAux "C" ⟨1⟩ ⟨2⟩ ⟨0⟩ [""]
#eval ("B"::[""]).reverse
#eval ["", "B"]

#eval "DB".splitOnAux "C" ⟨0⟩ ⟨0⟩ ⟨1⟩ []
#eval "DB".splitOnAux "C" ⟨0⟩ ⟨1⟩ ⟨0⟩ []
#eval "DB".splitOnAux "C" ⟨0⟩ ⟨2⟩ ⟨0⟩ []
#eval ("DB"::[]).reverse
#eval ["DB"]
```

digama0:
> In order to prove `splitOnAux_of_valid`, we also have to deal with the case where it's true that `sep.atEnd j`; this is where the problem with the current `splitOnAux` lies.

Like I said earlier, `splitOnAux` has a precondition that `j < sep.endPos`, which eliminates this case.

chabulhwi:
> Like I said earlier, `splitOnAux` has a precondition that `j < sep.endPos`, which eliminates this case. `splitOnAux_of_valid` should likewise have this as a hypothesis.

Adding `j < sep.endPos` to `splitOnAux_of_valid` as a hypothesis will solve the problems I had. I'll close this pull request and make another one. But I need to take a break first. 🫠

Std/Data/List/Basic.lean
chabulhwi: I'll use this implementation.

chabulhwi:
```diff
def splitOnList [BEq α] (l sep : List α) : List (List α) :=
  if _h : sep.isEmpty then
    [l]
  else
    match e : splitOnceP sep.isPrefixOf l with
    | none => [l]
    | some (l₁, l₂) =>
      have : length l₂ - sep.length < length l := sorry
-     splitOnList (l₂.drop sep.length) sep
+     l₁::splitOnList (l₂.drop sep.length) sep
termination_by _ => l.length
```

I fixed the definition as above.

PR #497, opened 2024-01-02
feat: array lemmas for append and extract

fgdorais:


Std/Data/Array/Lemmas.lean
```diff
@@ -497,11 +497,174 @@ theorem mem_join : ∀ {L : Array (Array α)}, a ∈ L.join ↔ ∃ l, l ∈ L 
...
 
+theorem size_append (as bs : Array α) : (as ++ bs).size = as.size + bs.size := by
+  simp only [size, append_data, List.length_append]
+
+theorem get_append_left {as bs : Array α} {h : i < (as ++ bs).size} (hlt : i < as.size) :
+    (as ++ bs)[i] = as[i] := by
+  simp only [getElem_eq_data_get]
+  have h' : i < (as.data ++ bs.data).length := by rwa [← data_length, append_data] at h
+  conv => rhs; rw [← List.get_append_left (bs:=bs.data) (h':=h')]
+  apply List.get_of_eq; rw [append_data]
```
joehendrix, line 521: Any particular reason for `h` to be implicit?

Std/Data/Array/Lemmas.lean
```diff
@@ -497,11 +497,174 @@ theorem mem_join : ∀ {L : Array (Array α)}, a ∈ L.join ↔ ∃ l, l ∈ L 
   · rintro ⟨s, h₁, h₂⟩
     refine ⟨s.data, ⟨⟨s, h₁, rfl⟩, h₂⟩⟩
 
+/-! ### empty -/
+
+theorem size_empty : (#[] : Array α).size = 0 := rfl
+
+theorem empty_data : (#[] : Array α).data = [] := rfl
```
joehendrix, line 504: Should the above two lemmas be `@simp`?

Std/Data/Array/Lemmas.lean
```diff
@@ -497,11 +497,174 @@ theorem mem_join : ∀ {L : Array (Array α)}, a ∈ L.join ↔ ∃ l, l ∈ L 
...
+  | succ size ih =>
+    if h : start < as.size then
+      rw [extract_loop_succ (h:=h), ih, size_push, Nat.add_assoc, ←Nat.add_min_add_left,
+        Nat.sub_succ, Nat.one_add, Nat.one_add, Nat.succ_pred_eq_of_pos (Nat.sub_pos_of_lt h)]
+    else
+      have h := Nat.le_of_not_gt h
+      rw [extract_loop_of_ge (h:=h), Nat.sub_eq_zero_of_le h, Nat.min_zero, Nat.add_zero]
+
+theorem size_extract (as : Array α) (start stop : Nat) :
+    (as.extract start stop).size = min stop as.size - start := by
```
joehendrix, line 584:
```suggestion
@[simp] theorem size_extract (as : Array α) (start stop : Nat) :
    (as.extract start stop).size = min stop as.size - start := by
```

Std/Data/Array/Lemmas.lean
```diff
@@ -497,11 +497,174 @@ theorem mem_join : ∀ {L : Array (Array α)}, a ∈ L.join ↔ ∃ l, l ∈ L 
...
+      have hge : bs.size + 1 ≤ i := Nat.lt_of_le_of_ne hge hi
+      rw [ih (bs.push as[start]) (start+1) ((size_push ..).symm ▸ hge)]
+      congr 1; rw [size_push, Nat.add_right_comm, Nat.add_sub_add_right]
+
+theorem get_extract {as : Array α} {start stop : Nat} (h : i < (as.extract start stop).size)
+    (h' : start + i < as.size := Nat.add_lt_of_lt_sub' (Nat.lt_of_lt_of_le (size_extract .. ▸ h)
+      (Nat.sub_le_sub_right (Nat.min_le_right ..) _))) :
+    (as.extract start stop)[i] = as[start + i] :=
+  show (extract.loop as (min stop as.size - start) start #[])[i] = as[start + i] by
+  rw [get_extract_loop_ge]; rfl; exact Nat.zero_le _
```
joehendrix, line 648: Can `get_extract` be made `simp`?  One might need to move the definition of `h'` to the right-hand side.

Std/Data/Array/Lemmas.lean
fgdorais: Simp can already prove them.

Std/Data/Array/Lemmas.lean
fgdorais: It follows from `hlt`. Perhaps I should make it optional?

joehendrix: Thanks, I think it's fine as is.  It should be there from the context.

PR #498, opened 2024-01-03
chore: Provide consistent Std.Data modules.

joehendrix:
This commit adds a check that there is a module for each directory dir in `Std/Data/dir` and that each of those
modules import all of the submodules.

It has an optional flag to automatically fix this as well as regenerate `Std.lean`.

The flag should be disabled in commits.


digama0:
Why did you remove the docs for updating Std.lean? I guess the process is more complicated now which is why the lean script does it, but we still want docs on how to do it in the readme. Std.lean should import everything in the package.

joehendrix:
It seems like most people would notice the failing test case and the documentation there.  I could move that to the README I  suppose and then refer people to the README in the test case.

`Std.lean` still imports everything, just indirectly, do you prefer the approach of directly importing everything?

joehendrix:
Nevermind, I should just make `scripts/updateStd.sh` do the right thing.  I'll do that in the morning.

digama0:
I don't mind if the spelling of `scripts/updateStd.sh` changes, but there should be some way to do the equivalent action to what `updateStd.sh` did. If the new plan is for `Std.lean` to not contain direct imports of `Std.Data.X.Y` because they are transitive through `Std.Data.X` that's fine by me but then the equivalent tool will need to be smarter to handle that case.

joehendrix:
Yes, changing `updateStd.sh` is exactly what the [commit](https://github.com/leanprover/std4/pull/498/commits/803b5e2a17c0f471e724f73e744a5861a3603267) above your comment does.  Let me know if you see a bug.

test/check_imports_std_data.lean
digama0, line 1: maybe this should be in `scripts/` instead?

scripts/updateStd.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+# This command updates the `Std.lean` file to include a list of all files
+# in the `Std` directory.
+__LEAN_STD_AUTOFIX_IMPORTS=true lake env lean test/check_imports_std_data.lean
```
digama0, line 4: This looks horrible, although I understand the reason you ended up with this way of passing arguments to lean elaboration. It looks like a lake feature request to me, although I would prefer we not have to put anything extra in the lakefile and just have a way of running scripts written in lean files.

joehendrix: It is a weird way of doing this, but I don't have the bandwidth to make changes to lake, and would prefer some other features in lake are prioritized at the moment.

Do you see a preferable way of passing this in?

I also considered registering an option, but I can't register an option in a file and also pass it into `lean`.
   
I also briefly considered using sed to change a line `autofix = false` to `autofix = true`, but that seemed even stranger.

Just to be clear, the origin of this PR is an in-person user request to be able to write things like `import Std.Data.Array` (and get all array theorems), and the changes to the imports are just a side effect of wanting to automate that since it's error prone to manually maintain that.

joehendrix:
It looks like you've moved this out of the test directory and now run it separately.

Here's my understanding of what you've changed build.yml to do:
1. Check out Std.
2. Run `lake env lean scripts/check_imports_std_data.lean` to regenerate `Std` inside `scripts/updateStd.sh`.  This will print out warnings.
3. Run `lake build` inside `scripts/updateStd.sh`.
4. Run `lake build -Kwerrror` from `build.yml`.
5. Run tests including a check whether `check_imports_std_data.lean` has changed anything.

It seems like at a minimum, `scripts/updateStd.sh` should not run `lake build`.

What are you trying to achieve?

digama0:
Oh I see, I did not realize `check_imports_std_data.lean` builds Std but of course it does. We should probably interchange those two steps.

Here's what I would like to happen:

1. Check out Std.
2. Run `lake build -Kwerror` from build.yml.
3. Run `lake env lean scripts/check_imports_std_data.lean` to regenerate Std inside `scripts/updateStd.sh`.
4. Run a check whether `check_imports_std_data.lean` has changed anything. This will be nonempty if there were missing imports, but it is superior to giving warnings in `check_imports_std_data.lean` because it comes with an actual diff suggesting what changes to make to what files.
5. Run tests.

scripts/updateStd.sh
fgdorais: Perhaps a naive idea, couldn't that be automated in lake?

joehendrix:
I know when I first encountered a failure in that step, I found it confusing since it didn't report an error, but had a diff.  That could be addressed through a comment.

I think we should just encourage running `scripts/updateStd.sh` locally if Github CI fails.  I could add a mention of running the script to the output of `check_imports_std_data.lean` when it finds errors.  Then we just merge your step 3 and 4.  I think that's simplest approach.

It'd be really neat if CI could add suggestions to the PR to fix missing imports.  That should be conceptually feasible through the Github REST API, but I don't have time to investigate that personally.

fgdorais:
For what it's worth, I don't see this change as very useful. One undesirable side effect is that it would lead contributors into using non-minimal imports, so we would probably need to add a linter for that.

The only non-total import that I think should exist is `StdBasic.lean` which imports just the `**/Basic.lean`, files dropping all the lemmas. The Basic/Lemmas structure has been ubiquitous since the beginning but we've been lax lately. I think we should work harder on that division. This would let "regular programmers" just use the programming content while just relying on the existence of the proofs for "certified" correctness.

digama0:
> The Basic/Lemmas structure has been ubiquitous since the beginning but we've been lax lately.

Do you have something in particular in mind? I've been doing my best to keep this line clearly marked.

fgdorais:
> Do you have something in particular in mind? I've been doing my best to keep this line clearly marked.

By "we" I really meant "I". I just recognized the importance of this structure.

joehendrix:
On the partial import side of things, another requested feature is `Std.Init` which focus on adding lemmas to core types but doesn't add additional datatypes.  

I agree a linter that checks imports to see if they are needed is useful.

digama0:
> I agree a linter that checks imports to see if they are needed is useful.

Do you mean like [`lake exe shake`](https://github.com/leanprover-community/mathlib4/pull/9346)?

joehendrix:
The reason this was requested was so that if a user is working with a datatype (use `Array` for example purposes) and they want all the features in Std related to `Array` by via `import Std.Data.Array`.

They don't have to separately import the definitions and lemmas or conversely import all of `Std`.  It's also consistent with how Lean core is organized.

joehendrix:
I think I have decent error reporting now.  Thanks for feedback!

scripts/updateStd.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
+set -e
 # This command updates the `Std.lean` file to include a list of all files
 # in the `Std` directory.
-find Std -name "*.lean" | env LC_ALL=C sort \
-  | sed 's/\.lean//;s,/,.,g;s/^/import /' > Std.lean
+__LEAN_STD_AUTOFIX_IMPORTS=true lake env lean test/check_imports.lean
```
digama0, line 5: I really do want this program to be in `scripts`. It's not like any other test, and it is also intended to run standalone.

digama0:
Do you have an example run? From what I can tell, it doesn't look like you are displaying a diff, which as I said was an advantage of the original version. Maybe you should call `diff` from the script?

test/check_imports.lean
```diff
@@ -129,9 +135,23 @@ def checkStdDataImports : MetaM Unit := do
         checkStdDataDir (autofix := autofix) modMap entry
     let stdImports ← expectedStdImports
     let .some stdMod := modMap.find? `Std
-        | warn "Missing Std module!"
+        | warn false "Missing Std module!; Run `lake build`."
     let warned ← checkMissingImports `Std stdMod stdImports
     if autofix && warned then
       writeImportModule "Std.lean" stdImports
+    match ← getWarningInfo with
```
digama0, line 142: Honestly this looks overcomplicated, all of the below and more are expressed by `diff`.

joehendrix: You may be correct on the complicated nature.  The one non-trivial thing to do from shell is avoiding overwriting definitions, e.g., here's output when one module has spurious definitions:

```
/repos/std4 lake build && ./scripts/updateStd.sh
[192/203] Building Std.Data.Array
[255/256] Building Std
Expected Std.Data.Array to not contain additional declarations.
Declarations should be moved out.
This error cannot be automatically fixed.
Found errors that cannot be automatically fixed.
Address unfixable issues and rerun lake build && ./scripts/updateStd.sh.
test/check_imports.lean:157:0: error: Fatal error
```

Here's the usual missing imports:
```
~/repos/std4 lake build && ./scripts/updateStd.sh
[255/256] Building Std
Missing import Std.Data.Array.Lemmas in Std.Data.Array
Generating Std/Data/Array.lean
Missing import Std.CodeAction.Deprecated in Std
Missing import Std.CodeAction.Basic in Std
Generating Std.lean
Found missing imports and attempted fixes.
Run lake build && ./scripts/updateStd.sh to verify.
Multiple runs may be needed.
~/repos/std4 lake build && ./scripts/updateStd.sh
[224/237] Building Std.Data.Array
[255/256] Building Std
```

This is partially an exercise for me to use Lean for file system manipulation to see what should be better.

digama0: Another thing I wanted to mention but maybe not pursue in this PR is to consider moving this script to a script with a `main` function instead of (ab)using compile-time elaboration execution which necessitates the strange argument passing. That would make it a bit easier to run directly as a script, so we could eliminate the `updateStd.sh` wrapper. (I'm imagining that we will have many scripts like this one eventually, and I would like to be able to write them in lean and run them without additional declarations in the lakefile.)

scripts/updateStd.sh
digama0: Another reason to make it a script is because it should be separate from the other tests in CI. That way you can see at a glance that what failed is a missing std import, not a functionality test.

joehendrix: I made it as a test, because it's good to have fewer things a developer needs to remember to run before pushing.

However, I already have a script that does that for me on the other CI checks (and forget to run occasionally).  I'll move it back to scripts, but I can't spend more iterations on preferenes.

test/check_imports.lean
fgdorais: This is really nice code! I think it may be useful to generalize a bit so that it can be reused by other packages. (Obviously, this is not required for this PR.)

joehendrix: Thanks, I will merge as is.  I like the idea of making this standalone eventually.

The next checker of a similar nature I have planned is one that looks for missing `simp` lemmas (e.g., checking operations that return a list and are meant for specifications have `simp` lemmas for `List.length` and other core introspection primitives).

PR #503, opened 2024-01-04
feat: omega handles iff and implications

semorrison:
While looking at theorems that "ought" to be proved by `omega`, I found a number of `iff` goals where one would need `constructor <;> (intros; omega)` to close the goal. This PR slightly powers up `omega` so it can do this by itself.

### DIFF
--- a/Std/Logic.lean
+++ b/Std/Logic.lean
@@ -612,7 +612,8 @@ theorem Decidable.not_iff [Decidable b] : ¬(a ↔ b) ↔ (¬a ↔ b) := by
 theorem Decidable.iff_not_comm [Decidable a] [Decidable b] : (a ↔ ¬b) ↔ (b ↔ ¬a) := by
   rw [@iff_def a, @iff_def b]; exact and_congr imp_not_comm not_imp_comm
 
-theorem Decidable.iff_iff_and_or_not_and_not [Decidable b] : (a ↔ b) ↔ (a ∧ b) ∨ (¬a ∧ ¬b) :=
+theorem Decidable.iff_iff_and_or_not_and_not {a b : Prop} [Decidable b] :
+    (a ↔ b) ↔ (a ∧ b) ∨ (¬a ∧ ¬b) :=
   ⟨fun e => if h : b then .inl ⟨e.2 h, h⟩ else .inr ⟨mt e.1 h, h⟩,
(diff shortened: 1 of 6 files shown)


Std/Logic.lean
```diff
@@ -612,7 +612,8 @@ theorem Decidable.not_iff [Decidable b] : ¬(a ↔ b) ↔ (¬a ↔ b) := by
 theorem Decidable.iff_not_comm [Decidable a] [Decidable b] : (a ↔ ¬b) ↔ (b ↔ ¬a) := by
   rw [@iff_def a, @iff_def b]; exact and_congr imp_not_comm not_imp_comm
 
-theorem Decidable.iff_iff_and_or_not_and_not [Decidable b] : (a ↔ b) ↔ (a ∧ b) ∨ (¬a ∧ ¬b) :=
+theorem Decidable.iff_iff_and_or_not_and_not {a b : Prop} [Decidable b] :
+    (a ↔ b) ↔ (a ∧ b) ∨ (¬a ∧ ¬b) :=
```
semorrison, line 616: This previously had implicit arguments `{b a}`.

joehendrix (approved):

Std/Tactic/Omega/Frontend.lean
```diff
@@ -245,30 +243,45 @@ def addIntInequality (p : MetaProblem) (h y : Expr) : OmegaM MetaProblem := do
     problem := ← (p.problem.addInequality lc.const lc.coeffs
       (some do mkAppM ``le_of_le_of_eq #[h, (← prf)])) |>.solveEqualities }
 
+
+
```
digama0, line 247:
```suggestion
```

test/false_or_by_contra.lean
```diff
@@ -18,7 +18,7 @@ example (h : False) : ¬ P := by
   guard_target = False
   simp_all
 
-example (h : False) : P := by
+example {P  : Prop} (h : False) : P := by
```
digama0, line 21: ?

semorrison: I've made the test clearer. `false_or_by_contra` behaves differently when the goal is a Type vs a Prop. Previously when `P` was an autoimplicit, it was being specialised to `Prop` *by the tactic* in the proof.

PR #504, opened 2024-01-04
feat: omega understands multiplication on the right by a constant

### LINKED ISSUE #499: `omega` only wants constants on one side.
```lean
import Std.Tactic.Omega

example (b : Nat) (hm : 2 = b * 2) : b = 1 := by
  -- omega -- omega did not find a contradiction: trivial
  rw [Nat.mul_comm] at hm
  omega
```

semorrison:
Closes #499.

### DIFF
--- a/Std/Tactic/Omega/Frontend.lean
+++ b/Std/Tactic/Omega/Frontend.lean
@@ -162,7 +162,9 @@ partial def asLinearComboImpl (e : Expr) : OmegaM (LinearCombo × OmegaM Expr ×
     let (yl, yprf, yfacts) ← asLinearCombo y
     if xl.coeffs.isZero ∨ yl.coeffs.isZero then
       let prf : OmegaM Expr := do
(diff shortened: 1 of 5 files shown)


Std/Tactic/Omega/Frontend.lean
```diff
@@ -157,17 +157,30 @@ partial def asLinearComboImpl (e : Expr) : OmegaM (LinearCombo × OmegaM Expr ×
         (← mkAppM ``Int.neg_congr #[← prf])
         (← mkEqSymm neg_eval)
     pure (-l, prf', facts)
-  | (``HMul.hMul, #[_, _, _, _, n, e']) =>
-    match intCast? n with
-    | some n' =>
-      let (l, prf, facts) ← asLinearCombo e'
+  | (``HMul.hMul, #[_, _, _, _, x, y]) =>
+    -- Check if the first factor is an integer literal
```
digama0, line 161: This still doesn't handle `1 * 2 * x`, right? My suggestion would be to evaluate both sides to `LinearCombo`, and then allow multiplication of a constant `LinearCombo` with any `LinearCombo` and vice versa.

semorrison: Okay, I've implemented this now.

ericrbg:
Worth noting that this closes #499.

joehendrix (approved):

PR #506, opened 2024-01-04
feat: make Except.emoji public

semorrison:
Multiple tactics in Mathlib use this, so it is inconvenient to have `private`.

### DIFF
--- a/Std/Tactic/SolveByElim.lean
+++ b/Std/Tactic/SolveByElim.lean
@@ -65,8 +65,6 @@ initialize registerTraceClass `Meta.Tactic.solveByElim
 
 namespace SolveByElim
 
-private def emoji (e:Except ε α) := if e.toBool then checkEmoji else crossEmoji
-
 /--
 `applyTactics lemmas goal` will return a list of tactics,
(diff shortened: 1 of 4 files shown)


joehendrix (approved):

Std/Tactic/SolveByElim.lean
```diff
@@ -80,7 +78,7 @@ calls to `apply` succeeded or failed.
 def applyTactics (cfg : ApplyConfig := {}) (transparency : TransparencyMode := .default)
     (lemmas : List Expr) (g : MVarId) : Nondet MetaM (List MVarId) :=
   (Nondet.ofList lemmas).filterMapM fun e => observing? do
-    withTraceNode `Meta.Tactic.solveByElim (return m!"{emoji ·} trying to apply: {e}") do
+    withTraceNode `Meta.Tactic.solveByElim (return m!"{Except.emoji ·} trying to apply: {e}") do
```
tydeu, line 81: It seems to me that dot notation would be preferable for these `Except.emoji ·` expressions (i.e., `·.emoji`).

PR #508, opened 2024-01-04
feat: more `ByteArray` lemmas

fgdorais:


Std/Data/ByteArray.lean
```diff
@@ -1,6 +1,6 @@
 /-
 Copyright (c) 2023 François G. Dorais. All rights reserved.
-Released under Apache 2.0 license as described in the file LICENSE.
+Released under Apache 2.0 license a described in the file LICENSE.
```
digama0, line 3:
```suggestion
Released under Apache 2.0 license as described in the file LICENSE.
```

digama0:
As a general remark, it would be nice if `ByteArray` was setup something like `AssocList` where we just rewrite everything in terms of arrays and then we don't need to have a separate ByteArray lemma collection.

fgdorais:
That would indeed be nice. I think core did it this way because `copySlice` is just a wrapper around `memcpy`, so it gives efficient implementations for `append` and `extract`.

digama0:
That's fine, it doesn't have to be defeq. I am imagining that `copySlice` has a rewrite rule in terms of `append` and `extract` (you already have this), and `append` and `extract` have rewrite rules connecting them to the corresponding array methods. (We can also have lemmas connecting them to lists, if people would prefer to rewrite in that direction instead.)

fgdorais:
Isn't that what the `*_data` theorems do? Or is there more to it?

joehendrix (approved):

PR #510, opened 2024-01-05
chore: omega consume Expr metadata when required

semorrison:
Closes #509.

PR #511, opened 2024-01-05
feat: omega understands Prod.Lex

semorrison:
This is useful for termination proofs.

---

- [x] depends on #510


digama0:
> This is useful for termination proofs.

Do you have an example of this for the test? I would have expected lex order in termination proofs to more often use `PSigma` because that's what is used in the automatic tupler. But maybe you mean in manual `termination_by _ x y => (x, y)` clauses?

semorrison:
> > This is useful for termination proofs.
> 
> Do you have an example of this for the test? I would have expected lex order in termination proofs to more often use `PSigma` because that's what is used in the automatic tupler. But maybe you mean in manual `termination_by _ x y => (x, y)` clauses?

The first of the new tests was actually extracted from a termination proof in Mathlib; I've tracked it down again and added a comment giving the reference (and confirmed that this branch really allows `omega` to work in Mathlib there!)

digama0:
Well I meant actually using it in a termination proof. As is it's just a direct `Prod.Lex` proof, which isn't really motivated in terms of termination. Maybe you could write the original definition with `0` or `sorry` for the match arms but the same recursion structure?

semorrison:
> Well I meant actually using it in a termination proof. As is it's just a direct `Prod.Lex` proof, which isn't really motivated in terms of termination. Maybe you could write the original definition with `0` or `sorry` for the match arms but the same recursion structure?
> 
> More specifically, I would like to also test that the magic that `termination_by` / `decreasing_by` does produces a goal that is compatible with what `omega` is proving. Ideally you can just write `decreasing_by omega` here.

My preference is not to do that quite yet in this PR. I have a branch in which I'm using `omega` in many places in Mathlib, and I'd prefer to work on that a bit more, and then be able to select the nicest examples to install as tests here in Std.

(But absolutely, yes, we want seamless integration of `omega` into termination proofs, probably even getting to the point that core knows about `omega` at least as a shim and it is not even necessary to write `decreasing_by omega`.)

joehendrix (approved):

joehendrix (approved):

PR #513, opened 2024-01-05
doc: how to depend on the library

david-christiansen:
Give actionable instructions for depending on `std4`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,16 @@
 
 Work in progress standard library for Lean 4. This is a collection of data structures and tactics intended for use by both computer-science applications and mathematics applications of Lean 4.
 
+# Using `std4`
+
+To use `std4` in your project, add the following to your `lakefile.lean`:
+
+```lean
+require std from git "git@github.com:leanprover/std4.git" @ "main"
+```
+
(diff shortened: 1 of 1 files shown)


david-christiansen:
I received feedback from people I know here in CPH that they tried to use Lean for a project but couldn't figure out how to actually use `std4`. I figured a README entry was a quick way to help.

README.md
```diff
@@ -2,6 +2,16 @@
 
 Work in progress standard library for Lean 4. This is a collection of data structures and tactics intended for use by both computer-science applications and mathematics applications of Lean 4.
 
+# Using `std4`
+
+To use `std4` in your project, add the following to your `lakefile.lean`:
+
+```lean
+require std from git "git@github.com:leanprover/std4.git" @ "main"
```
fgdorais, line 10: I think this only works for users that use ssh. I think this works for everyone:
```suggestion
require std from git "https://github.com/leanprover/std4" @ "main"
```

fgdorais:
Of course, the ideal situation is for std4 to ship with core (with an opt-out).

semorrison:
> Of course, the ideal situation is for std4 to ship with core (with an opt-out).

We're working on it. :-)

PR #514, opened 2024-01-07
feat(Data/List): deprecate `last'` and `ilast'`

urkud:
Also add 3 lemmas about `getLast?`, `getLast`, and `getLastD`.


urkud:
awaiting-review

fgdorais (approved):
One has to be very careful with deprecation, the replacement should ideally be an exact drop-in substitution for the original. In this case, `ilast'` and `getLastD` reverse the order of their arguments. I think this is minor enough that deprecation is still warranted.

urkud:
The replacement also has slightly different definitional equalities but we should definitely use functions from Lean core instead of functions from mathlib3.

PR #515, opened 2024-01-07
feat: `trySetOptions` / `tryEraseAttrs` command elab helpers

tydeu:
Helpers which I wrote for @hrmacbeth to use in her Mechanics of Proof [package](https://github.com/hrmacbeth/math2001/tree/main) to help with initializing options and attributes for a teaching environment. She suggested I PR these helpers to Std as they quite helpful when configuring custom environments and are thus useful to many who teach Lean.

### DIFF
--- a/Std/Lean/Elab/Command.lean
+++ b/Std/Lean/Elab/Command.lean
@@ -1,0 +1,53 @@
+/-
+Copyright (c) 2023 Mac Malone. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mac Malone
+-/
+import Lean.Elab.Command
+
+/-!
+# Command elaboration helpers
+-/
+
+namespace Lean.Elab.Command
+
+/--
+For each named option which exists, set its value.
+Will still fail if the set value does not match the option's proper type
+(e.g., setting a string value to a boolean option).
+-/
+def trySetOptions (settings : Array (Name × DataValue)) : CommandElabM PUnit := do
+  let opts ← getOptions
+  let ⟨_, opts⟩ ← StateT.run (s := opts) <| settings.forM fun ⟨optName, val⟩ => do
+    if let .ok decl ← getOptionDecl optName |>.toBaseIO then
+      unless decl.defValue.sameCtor val do
+        throwError "type mismatch for '{optName}'"
+      modify (·.insert optName val)
+  modifyScope ({· with opts})
+
+/--
+If `attrName` exists, try to erase it from the specified declaration names,
+skipping any that do not exist (or, due to implementation constraints,
+fail to erase).
+
+Thus, ``tryEraseAttr `my_attr  #[`foo, `bar]`` is essentially equivalent to
+the command `attribute [-my_attr] foo bar`, except it will not error if the
+attribute or declaration names do not exist.
+-/
+def tryEraseAttr (attrName : Name) (declNames : Array Name) : CoreM PUnit := do
+  let attrState := attributeExtension.getState (← getEnv)
+  if let some attr := attrState.map.find? attrName then
+    declNames.forM fun declName =>
+      try
+        attr.erase declName
+      catch _ =>
+        -- This consumes all types of errors, rather than just existence ones,
+        -- but there does not appear to be a better general way to achieve this
+        pure ()
+
+/--
+for each named attribute which exists,
+erase from it the specified declaration names, ignoring failures.
+-/
+def tryEraseAttrs (attrs :  Array (Name × Array Name)) : CommandElabM PUnit := do
+  liftCoreM <| attrs.forM fun ⟨attrName, declNames⟩ => tryEraseAttr attrName declNames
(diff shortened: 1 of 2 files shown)


tydeu:
awaiting-review

joehendrix:
Could you say a little bit about why one would wish to set options/remove attributes that may not exist?

Is this for customizing an environment when one isn't quite sure what has been imported?

hrmacbeth:
@joehendrix, here is the setting in which I use this: I have a [custom command](https://github.com/hrmacbeth/math2001/blob/5dfc9f866e668a7b84a958b8f68b62bfbd3a3550/Library/Basic.lean#L38) `math2001_init` which calls these two commands to set a list of options and remove a list of attributes.  This command is run at the beginning of every student-facing file:
[section 1.2](https://github.com/hrmacbeth/math2001/blob/5dfc9f866e668a7b84a958b8f68b62bfbd3a3550/Math2001/01_Proofs_by_Calculation/02_Proving_Equalities_in_Lean.lean#L5)
[section 1.3](https://github.com/hrmacbeth/math2001/blob/5dfc9f866e668a7b84a958b8f68b62bfbd3a3550/Math2001/01_Proofs_by_Calculation/03_Tips_and_Tricks.lean#L5)
[section 1.4](https://github.com/hrmacbeth/math2001/blob/5dfc9f866e668a7b84a958b8f68b62bfbd3a3550/Math2001/01_Proofs_by_Calculation/04_Proving_Inequalities.lean#L5)
etc

Bundling it all into a single command ensures that the environment is consistent (the same in every file), which is particularly helpful when I wish to change or experiment course-wide with the precise list of modified options/attributes.

Regarding your specific question:

> Could you say a little bit about why one would wish to set options/remove attributes that may not exist?
> 
> Is this for customizing an environment when one isn't quite sure what has been imported?

Yes -- some of the course files have more imports than others, so in some of them the option/attribute may not exist yet.

I have been very happy with Mac's setup, but if there is a different preferred way of achieving these goals I am (of course) happy to switch.

Std/Lean/Elab/Command.lean
```diff
@@ -0,0 +1,53 @@
...
+-/
+
+namespace Lean.Elab.Command
+
+/--
+For each named option which exists, set its value.
+Will still fail if the set value does not match the option's proper type
+(e.g., setting a string value to a boolean option).
+-/
+def trySetOptions (settings : Array (Name × DataValue)) : CommandElabM PUnit := do
```
joehendrix, line 19: How about also introduce `def trySetOption (name : Name) (val : DataValue) : CommandElabM PUnit` and define this in terms of that?  It looks like there wouldn't be a significant performance cost.

Std/Lean/Elab/Command.lean
```diff
@@ -0,0 +1,53 @@
...
+/--
+If `attrName` exists, try to erase it from the specified declaration names,
+skipping any that do not exist (or, due to implementation constraints,
+fail to erase).
+
+Thus, ``tryEraseAttr `my_attr  #[`foo, `bar]`` is essentially equivalent to
+the command `attribute [-my_attr] foo bar`, except it will not error if the
+attribute or declaration names do not exist.
+-/
+def tryEraseAttr (attrName : Name) (declNames : Array Name) : CoreM PUnit := do
```
joehendrix, line 37: Could this just take a single `declName`?

joehendrix:
@hrmacbeth Thanks for the explanation and @tydeu for the PR.

I have some minor suggestions above, but this seems like a useful capability.  A project may want to pull in a library that makes global changes (e.g., simp rules) that the project doesn't want, and meta programming provides an automated mechanism for fixing that.

Std/Lean/Elab/Command.lean
tydeu: @joehendrix IMO, I think it is better for the option to mirror the command which can take multiple declaration names (e.g., `attribute [-my_attr] foo bar`). It is also slightly more efficient for multiple declarations names this way.

Std/Lean/Elab/Command.lean
tydeu: @joehendrix I added `trySetOption`, but it did not seem worth it to define `trySetOptions` in terms of it  (the number of references loads would be `O(n)` instead of `O(1)`).

tydeu:
awaiting-review

fgdorais:
Heartbeat check: is this PR still active?

tydeu:
@fgdorais From my view, that seems like a Batteries maintainer question. I addressed the review comments, so the next step was to hear back whether my response / changes were sufficient. If so, this PR would be good to go. Unfortunately, the PR has rotted since then, so it would need some reorganization (i.e., `Std` -> `Batteries`) to be mergeable at this point.

fgdorais:
Thank you for the quick response! Just to be clear: (1) this PR is still needed, (2) all issues have been resolved (but not necessarily cleared) and (3) it still needs further review.

Please correct me if my (quick and probably incorrect) assessment is wrong.

fgdorais (approved):

digama0:
@tydeu I've fixed the main migration changes to this PR, but it no longer seems to work as designed, so it would help to have another look at this and make sure the design still makes sense with current lean.

PR #518, opened 2024-01-10
chore: allow omega to use autoparam wrapped hypotheses / consume metadata

semorrison:


Std/Tactic/Omega/Frontend.lean
```diff
@@ -282,7 +282,7 @@ partial def addFact (p : MetaProblem) (h : Expr) : OmegaM (MetaProblem × Nat) :
   if ! p.problem.possible then
     return (p, 0)
   else
-    let t ← instantiateMVars (← inferType h)
+    let t ← instantiateMVars (← inferType h).cleanupAnnotations
```
digama0, line 285:
```suggestion
    let t ← instantiateMVars (← whnfR <| ← inferType h)
```
isn't this a similar case as last time?

semorrison: Indeed.

digama0: note that to do this properly you also have to call `whnfR` on each subterm for the nested matches here (this is the sort of thing that `Qq` would ideally be automating)

joehendrix (approved):

PR #522, opened 2024-01-11
feat: omega handles double negation and implication hypotheses

semorrison:
---

- [ ] depends on #518


Std/Tactic/Omega/Frontend.lean
```diff
@@ -298,6 +298,9 @@ def pushNot (h P : Expr) : MetaM (Option Expr) := do
       return some (mkApp5 (.const ``Decidable.or_not_not_of_not_and []) P₁ P₂
         (.app (.const ``Classical.propDecidable []) P₁)
         (.app (.const ``Classical.propDecidable []) P₂) h)
+    | (``Not, #[P']) =>
+      return some (mkApp3 (.const ``Decidable.of_not_not []) P'
+        (.app (.const ``Classical.propDecidable []) P') h)
```
digama0, line 303: this unnecessary `Classical` makes me vaguely grumble but I won't ask for any changes in this PR since there are other uses in the same function.

test/omega/test.lean
```diff
@@ -326,5 +326,8 @@ example (a b : Int) (h : a > 7) (w : b > 2) : a > 0 ↔ b > 0 := by omega
 -- Verify that we can prove implications:
 example (a : Int) : a > 0 → a > -1 := by omega
 
+-- Verify that we can handle double negation:
+example (x y : Int) : x + 1 ≤ y → ¬ y + 1 ≤ x := by omega
```
digama0, line 330: why is this a double negation example? Shouldn't `false_or_by_contra` end up just introducing the hypotheses?

semorrison: No, `false_or_by_contra` doesn't do any intros. Probably that should change!

semorrison: Done.

Std/Tactic/Omega/Frontend.lean
semorrison: Very happy to approve PRs removing `Classical` from `omega`!

semorrison:
Merging by Joe's request after resolving conflicts.

PR #525, opened 2024-01-12
docs: extend docstrings for `ext` and `ext1`

kmill:
Makes sure to mention that the patterns are processed using `rintro`, and makes sure `ext` mentions `ext1`.


kmill:
awaiting-review

kmill:
Modifying this documentation was inspired by [this Zulip question](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/controlling.20how.20ext.20introduces.20variables/near/412498786), where the discoverability of `ext1` from `ext` would have helped.

PR #526, opened 2024-01-13
feat: add `ByteArray.ofFn`

fgdorais:


Std/Data/ByteArray.lean
```diff
@@ -109,3 +109,21 @@ theorem get_extract_aux {a : ByteArray} {start stop} (h : i < (a.extract start s
 @[simp] theorem get_extract {a : ByteArray} {start stop} (h : i < (a.extract start stop).size) :
     (a.extract start stop)[i] = a[start+i]'(get_extract_aux h) := by
   simp [getElem_eq_data_getElem]
+
+/-! ### ofFn -/
+
+/--- `ofFn f` with `f : Fin n → UInt8` returns the byte array whose `i`th element is `f i`. --/
+def ofFn (f : Fin n → UInt8) : ByteArray where
+  data := .ofFn f
```
digama0, line 117: This is an inefficient implementation, although I think we want this to be a defeq, so it should be overridden using `@[csimp]`.

fgdorais: I'm glad you chimed in since I was wanted to ask you about this. I have a more efficient `csimp` version in the works but I paused it since I wasn't sure it was actually worthwhile. 

Both `ByteArray.mk` and `Array.ofFn` have efficient implementations. What ends up happening is that `Array.ofFn` builds an array boxing all the `UInt8`s, then `ByteArray.mk` (implemented in C as "lean_byte_array_mk") reads that array and unboxes all the `UInt8`'s. There's evident redundancy here but it's not clear to me that the one-pass version implemented in Lean would be that much faster in reality. (Of course, the gains in memory allocation are very clear.)

Since my primary usage is to use `ofFn` to make basic specs for some currently `opaque` defs, I don't currently have a need for an efficient version.

Anyway, I'm happy to continue working on my efficient `csimp` version, whether it ends up in this PR or a later one. I just wanted to consult the wizards before going further on this road!

digama0: Yes, that's exactly the inefficiency I'm talking about. People expect functions like this to perform as well as can be expected under the circumstances, and this doesn't meet that bar. I would prefer to not merge this until this issue is addressed.

fgdorais: Sounds good. Thanks!

fgdorais: All done!

Std/Data/ByteArray.lean
```diff
@@ -127,3 +128,18 @@ def ofFn (f : Fin n → UInt8) : ByteArray where
 
 @[simp] theorem getElem_ofFn (f : Fin n → UInt8) (i) (h : i < (ofFn f).size):
     (ofFn f)[i] = f ⟨i, size_ofFn f ▸ h⟩ := get_ofFn ..
+
+/-- More efficient implementation of `ByteArray.ofFn` -/
+def ofFnAux (f : Fin n → UInt8) : ByteArray := go 0 (mkEmpty n) where
```
digama0, line 133:
```suggestion
private def ofFnAux (f : Fin n → UInt8) : ByteArray := go 0 (mkEmpty n) where
```

fgdorais: Should I also add `@[specialize]`? I'm never sure when that makes a difference.

Std/Data/ByteArray.lean
```diff
@@ -109,3 +110,34 @@ theorem get_extract_aux {a : ByteArray} {start stop} (h : i < (a.extract start s
...
+    if h : i < n then go (i+1) (acc.push (f ⟨i, h⟩)) else acc
+termination_by _ => n - i
+
+@[simp] theorem ofFnAux_data (f : Fin n → UInt8) :
+    (ofFnAux.go f i acc).data = Array.ofFn.go f i acc.data := by
+  rw [ofFnAux.go, Array.ofFn.go]; split; rw [ofFnAux_data, push_data]; rfl
+termination_by _ => n - i
+
+@[csimp] theorem ofFn_eq_ofFnAux : @ofFn = @ofFnAux := by
+  funext n f; ext; simp [ofFnAux, Array.ofFn, ofFnAux_data, mkEmpty]
```
digama0, line 143:
```suggestion
@[csimp] theorem ofFn_eq_ofFnAux : @ofFn = @ofFnAux := by
  let rec ofFnAux_data (f : Fin n → UInt8) :
      (ofFnAux.go f i acc).data = Array.ofFn.go f i acc.data := by
    rw [ofFnAux.go, Array.ofFn.go]; split; rw [ofFnAux_data, push_data]; rfl
  termination_by _ => n - i
  funext n f; ext; simp [ofFnAux, Array.ofFn, ofFnAux_data, mkEmpty]
```

fgdorais: I should have paid attention to "or similar" :)

fgdorais: The `@_ = @_` makes this change more tedious than necessary. If you have a trick to make it work, please let me know.

fgdorais: Never mind, I got it!

Std/Data/ByteArray.lean
```diff
@@ -109,3 +111,34 @@ theorem get_extract_aux {a : ByteArray} {start stop} (h : i < (a.extract start s
...
+
+@[simp] theorem getElem_ofFn (f : Fin n → UInt8) (i) (h : i < (ofFn f).size):
+    (ofFn f)[i] = f ⟨i, size_ofFn f ▸ h⟩ := get_ofFn ..
+
+private def ofFnAux (f : Fin n → UInt8) : ByteArray := go 0 (mkEmpty n) where
+  go (i : Nat) (acc : ByteArray) : ByteArray :=
+    if h : i < n then go (i+1) (acc.push (f ⟨i, h⟩)) else acc
+termination_by _ => n - i
+
+@[csimp] theorem ofFn_eq_ofFnAux : @ofFn = @ofFnAux := by
```
digama0, line 138:
```suggestion
@[csimp] private theorem ofFn_eq_ofFnAux : @ofFn = @ofFnAux := by
```

Std/Data/ByteArray.lean
```diff
@@ -109,3 +111,34 @@ theorem get_extract_aux {a : ByteArray} {start stop} (h : i < (a.extract start s
...
+@[simp] theorem ofFn_data (f : Fin n → UInt8) : (ofFn f).data = .ofFn f := rfl
+
+@[simp] theorem size_ofFn (f : Fin n → UInt8) : (ofFn f).size = n := by
+  simp [size]
+
+@[simp] theorem get_ofFn (f : Fin n → UInt8) (i : Fin (ofFn f).size) :
+    (ofFn f).get i = f (i.cast (size_ofFn f)) := by
+  simp [get, Fin.cast]
+
+@[simp] theorem getElem_ofFn (f : Fin n → UInt8) (i) (h : i < (ofFn f).size):
```
digama0, line 130:
```suggestion
@[simp] theorem getElem_ofFn (f : Fin n → UInt8) (i) (h : i < (ofFn f).size) :
```

Ruben-VandeVelde:
The names `ofFn_data` and `size_ofFn` seem inconsistent. Was that intentional?

fgdorais:
Yes. This "inconsistency" is everywhere in `Array` and `ByteArray`.

PR #528, opened 2024-01-15
feat: correctness for `Fin.foldl` and `Fin.foldr`

fgdorais:


semorrison (approved):
LGTM

Std/Data/Fin/Basic.lean
```diff
@@ -49,6 +49,12 @@ def natAdd (n) (i : Fin m) : Fin (n + m) := ⟨n + i, Nat.add_lt_add_left i.2 _
 /-- `min n m` as an element of `Fin (m + 1)` -/
 def clamp (n m : Nat) : Fin (m + 1) := ⟨min n m, Nat.lt_succ_of_le (Nat.min_le_right ..)⟩
 
+/-- `enum n` is the array of all elements of `Fin n` in order -/
+def enum (n) : Array (Fin n) := Array.ofFn id
+
+/-- `list n` is the list of all elements of `Fin n` in order -/
+def list (n) : List (Fin n) := (enum n).data
```
digama0, line 56: not going to block the PR for this, but I think this can be done in one pass without an additional array allocation.

fgdorais: Yes, but that makes it difficult to prove anything. It might make sense to have a csimp lemma though.

digama0: It's a bit weird for this list function to be defined in terms of array functions, we should have enough to do the proof directly over lists

fgdorais: I agree. The issue is proving it _here_, this would be much easier with `import Std`...

I tried a few different approaches and this is the one that had was the least unpleasant.

Std/Data/Fin/Lemmas.lean
```diff
@@ -723,6 +725,78 @@ theorem addCases_right {m n : Nat} {motive : Fin (m + n) → Sort _} {left right
...
+
+theorem foldr_eq_foldr_list (f : Fin n → α → α) (x) : foldr n f x = (list n).foldr f x := by
+  induction n using Nat.recAux with
+  | zero => rfl
+  | succ n ih => rw [foldr_succ, ih, list_succ, List.foldr_cons, List.foldr_map]
+
+private def listAux (n) : List (Fin n) := foldr n (.::.) []
+
+@[csimp] private theorem list_eq_listAux : @list = @listAux := by
+  funext n; rw [listAux, foldr_eq_foldr_list]; generalize list n = xs; cases xs <;> simp
```
joehendrix, line 798: What's the rational for this `csimp`?

fgdorais: Mario pointed out that `Fin.list` is slightly inefficient since it first allocates an array and then extracts the list. It's more efficient to skip the array, which is what `Fin.foldl`/`Fin.foldr` are actually designed for. Defining `Fin.list` correctly at the outset leads to difficult correctness proofs because of the circularity. This csimp attempts to fix the situation after the fact. There probably is a better code organization that avoids this but it's kind of like a jenga tower...

It's also questionable to have a csimp hidden in the lemmas.

fgdorais: We could drop the csimp since, for example, `List.ofFn` does the same thing: construct an array then extract the list.

fgdorais:
Some weird git issues happened here. On hold for now.

fgdorais:
Git issues fixed but history may have been mangled a bit during surgery. Let's merge this soon before it rots again.

Std/Data/Fin/Basic.lean
```diff
@@ -9,6 +9,12 @@ namespace Fin
...
```
eric-wieser, line 16: This function already exists in Mathlib as `List.finRange`.  Was the new choice of name deliberate?

fgdorais: No, I'll fix that.

fgdorais: See #899

PR #530, opened 2024-01-15
feat: add type classes for decidable < and <= relations

david-christiansen:
This means that instead of writing
```
def f [inst : LT α] [DecidableRel (@LT.lt α inst)] (args ...) := ...
```
users can write
```
def f [DecidableLT α] (args ...) := ...
```
I find this much easier to figure out, and it doesn't impose any extra work on library authors.

This came up while @nomeata and I were writing code examples for a tutorial today.

### DIFF
--- a/Std/Classes/DecidableLE.lean
+++ b/Std/Classes/DecidableLE.lean
@@ -1,0 +1,15 @@
+/--
+Asserts that the "less-than-or-equal" relation on `α` is decidable, that is, `a ≤ b` is decidable
+for all `a b : α`. See `Decidable`.
+
+It is generally not necessary to create instances of this class directly, because Lean will create
+one whenever a `DecidableRel` instance exists for the `LE` instance in question.
+-/
+class DecidableLE (α : Type u) extends LE α where
+  /-- Decide whether `x` is less than or equal to `y`. -/
+  decLE : (x y : α) → Decidable (LE.le x y)
+
+instance [DecidableLE α] {x y : α} : Decidable (x ≤ y) := DecidableLE.decLE _ _
+
+instance [inst : LE α] [DecidableRel (@LE.le α inst)] : DecidableLE α where
+  decLE := inferInstance
--- a/Std.lean
+++ b/Std.lean
@@ -1,5 +1,7 @@
 import Std.Classes.BEq
 import Std.Classes.Cast
+import Std.Classes.DecidableLE
+import Std.Classes.DecidableLT
 import Std.Classes.Dvd
 import Std.Classes.LawfulMonad
 import Std.Classes.Order
--- a/Std/Classes/DecidableLT.lean
+++ b/Std/Classes/DecidableLT.lean
@@ -1,0 +1,15 @@
+/--
+Asserts that the "less-than" relation on `α` is decidable, that is, `a < b` is decidable for all
+`a b : α`. See `Decidable`.
+
+It is generally not necessary to create instances of this class directly, because Lean will create
+one whenever a `DecidableRel` instance exists for the `LT` instance in question.
+-/
+class DecidableLT (α : Type u) extends LT α where
+  /-- Decide whether `x` is less than `y`. -/
+  decLT : (x y : α) → Decidable (LT.lt x y)
+
+instance [DecidableLT α] {x y : α} : Decidable (x < y) := DecidableLT.decLT _ _
+
+instance [inst : LT α] [DecidableRel (@LT.lt α inst)] : DecidableLT α where
+  decLT := inferInstance
--- a/test/decLE.lean
+++ b/test/decLE.lean
@@ -1,0 +1,69 @@
+import Std.Classes.DecidableLT
+import Std.Classes.DecidableLE
+import Std.Tactic.GuardMsgs
+import Std.Tactic.Omega
+
+inductive Three where
+  | one | two | three
+
+inductive ThreeLT : (a b : Three) → Prop where
+  | one : ThreeLT .one any
+  | two : ThreeLT .two .three
+
+instance : LT Three where
+  lt := ThreeLT
+
+instance : DecidableRel (@LT.lt Three _) := fun a b => by
+  cases a <;> cases b <;>
+  first
+    | apply isTrue ; constructor
+    | apply isFalse ; intro h ; cases h
+
+abbrev ThreeLE (a b : Three) : Prop := ThreeLT a b ∨ a = b
+
+instance : LE Three where
+  le := ThreeLE
+
+instance : DecidableRel (@LE.le Three _) := fun a b => by
+    cases a <;> cases b <;>
+  first
+    | apply isTrue
+      first
+        | apply Or.inr
+          rfl
+        | apply Or.inl
+          constructor
+    | apply isFalse
+      intro h
+      cases h with | _ h =>
+      cases h
+
+def insertSorted [DecidableLT α] (arr : Array α) (i : Fin arr.size) : Array α :=
+  match i with
+  | ⟨0, _⟩ => arr
+  | ⟨i' + 1, _⟩ =>
+    have : i' < arr.size := by omega
+    if arr[i'] < arr[i] then
+      arr
+    else
+      insertSorted
+        (arr.swap ⟨i', by assumption⟩ i)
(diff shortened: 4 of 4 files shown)


david-christiansen:
The name is chosen by analogy from `DecidableEq`

Std/Classes/DecidableLE.lean
```diff
@@ -0,0 +1,15 @@
...
+one whenever a `DecidableRel` instance exists for the `LE` instance in question.
+-/
+class DecidableLE (α : Type u) extends LE α where
+  /-- Decide whether `x` is less than or equal to `y`. -/
+  decLE : (x y : α) → Decidable (LE.le x y)
+
+instance [DecidableLE α] {x y : α} : Decidable (x ≤ y) := DecidableLE.decLE _ _
+
+instance [inst : LE α] [DecidableRel (@LE.le α inst)] : DecidableLE α where
+  decLE := inferInstance
```
nomeata, line 15: I hope this will not make type class inference run into circles, or be significantly slower.

digama0: I would like sign-off from mathlib maintainers for this

david-christiansen:
I think an alternative version that required explicit instances would also serve the goals here, at the cost of a bit more implementation work.

Std/Classes/DecidableLE.lean
fgdorais: Would an `abbrev` work better?

fgdorais: Does
```lean
abbrev DecidableLE (α) {inst : LE α} := DecidableRel inst.le
```
work at all?

If you're proposing bundling `LE` and `DecidableRel` then I would favor something more basic like
```lean
class BLE (α) where ble : α → α → Bool
```
I have a vague feeling this has been debated abundantly elsewhere...

fgdorais: Ah! This is tied to the long awaited `Decidable` overhaul. Since `Decidable` is not a structure (it's a rare `class inductive`) it can't easily be mixed in with other classes.

david-christiansen: RE `ble` - I think the important thing here is that people who are new to Lean can do basic operations like comparisons without understanding _all_ the subtleties of implicit arguments and dependent types that are implied by `[inst : LE α] [DecidableRel (@LE.le _ inst)]`.

Would `ble`/`blt` solve this? I would think that it would imply using some other version of `<`, just as `==` is the Boolean notion of equality. But I don't think users will have an easy time guessing non-standard comparison syntax.

fgdorais: I think the core issue is that we can't use the `class abbrev` mechanism with `Decidable` since `Decidable` is a class inductive and not a class structure. This could be fixed by [lean4#2038](https://github.com/leanprover/lean4/pull/2038).

The circular dependency is probably going to slow down class inference for `LE`. Here are three less risky options:

1. A convenience abbreviation:
```lean
abbrev DecidableLE (α) [inst : LE α] := DecidableRel inst.le

def order [LE α] [DecidableLE α] : α × α → α × α
  | (x, y) => if x ≤ y then (x, y) else (y, x)
```

2. A non-circular version:
```lean
class DecidableLE (α) extends LE α where
  instDecidable (x y : α) : Decidable (x ≤ y)
attribute [instance] DecidableLE.instDecidable

def order [DecidableLE α] : α × α → α × α
  | (x, y) => if x ≤ y then (x, y) else (y, x)
```

3. A boolean version with no new notation:
```lean
class BLE (α) where ble : α → α → Bool
export BLE (ble)
instance [BLE α] : LE α where le x y := ble x y 
instance [BLE α] (x y : α) : Decidable (x ≤ y) := inferInstanceAs (Decidable (_ = true))

def order [BLE α] : α × α → α × α
  | (x, y) => if x ≤ y then (x, y) else (y, x)
```

fgdorais: While we can't use `class abbrev` here, I think the result would be the same as this.
```lean
class DecidableLE (α : Type _) extends LE α where
  [instDecidable (x y : α) : Decidable (x ≤ y)]
attribute [instance] DecidableLE.mk DecidableLE.instDecidable
```

eric-wieser:
I worry that this is going to just become a trap for new users, who write `[PartialOrder X] [DecidableLE X]` and now end up with two unrelated ≤ operators.

Std/Classes/DecidableLE.lean
david-christiansen: Number 1 here imposes less, but still oddly large, syntactic overhead on people writing what should be basic programs. I can at least imagine how I'd write a good error message for that one!

Number 2 seems good - should I just update this PR to it? It'd be a bit annoying to write the instances, but not the end of the world. We're speculating about performance here, so it would be nice to compare e.g. the Mathlib build times on top of each option to find out rather than making educated guesses. Do we have a better way to benchmark std4's type classes?

The `BLE` one seems potentially problematic in the same ways that the split between `DecidableEq` and `BEq` are today. We could easily end up with two separate LE predicates for a type, with different behavior obtained only by replacing `BLE` with the pair of `LE` and `DecidableRel` in the signature. This strikes me as a footgun.

What's the best way to run real experiments here in `std4`? I think this goal is important but I'm by no means wedded to a particular implementation.

david-christiansen:
`PartialOrder` is not in `std4`, nor in core Lean. It seems important to me that users using only `Std` can write `if x < y then ... else ...` without first understanding dependent types. But fixing this new user papercut should also not make Mathlib worse!

Please help me understand the issue better - why is this worse than if the user had written `[PartialOrder X] [LE X]`, which could easily happen today?

digama0:
> Please help me understand the issue better - why is this worse than if the user had written `[PartialOrder X] [LE X]`, which could easily happen today?

It's not, really. But I think people are more likely to reach for `DecidableLE` as a way to say "LE is decidable" rather than "I have a LE and it is decidable", and in the example, the correct solution is `[PartialOrder X] [DecidableRel (@LT.lt X _)]` which reintroduces exactly the problem we're trying to solve here. This is what I would call a pitfall: the natural option is the wrong one.

Maybe an alternative which solves this issue is to have `DecidableLE A` just mean the mixin part, meaning that you have to write `[LE A] [DecidableLE A]` if you just want a decidable LE and nothing more, but this composes better with alternative sources for the LE as in `[PartialOrder A] [DecidableLE A]`.

joehendrix:
GHC has an extension [ConstraintKinds](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/constraint_kind.html) that if implemented in Lean, could address this.

It lets one wrote constraint synonyms:
```haskell
type Stringy a = (Read a, Show a)
foo :: Stringy a => a -> (String, String -> a)
foo x = (show x, read)
```

I'm not sure on the syntax, but it could be something like:

```lean
class DecidableLT α := (LT α, DecidableRel (α := α) (· < ·))

def f [DecidableLT α] (x y : α) : Bool := decide (x < y)
```

Std/Classes/DecidableLE.lean
joehendrix: I like the idea of solving this at the language level rather than library level.

There are a lot of classes related to binary ordering operations already and this feels a bit like adding one more to solve a particular problem.  I expect some of the Mathlib ones will flow down to Std eventually for theorems about ordered data structures.

The syntax for expressing the decidability of a binary operation isn't great though, so I'm not opposed to `DecidableLT α` as an abbreviation for `DecidableRel (α := α) (· < ·)` (or with the extra `inferInstance`) if that works out as a language feature.

david-christiansen:
Just an update from my side - I've been busy with some other things, but I plan to run some Mathlib builds with a few of the options we've discussed here, then report back on the impact on build times.

Std/Classes/DecidableLE.lean
eric-wieser: I agree that `abbrev DecidableLT (α : Type u) := DecidableRel (α := α) (· < ·)` would be a welcome addition; after all, we already have something analogous for `DecidableEq`.

fgdorais:
Heartbeat check: is this PR still active?

PR #531, opened 2024-01-15
feat: add `Std.Data.List.SplitOnList`

chabulhwi:
* Move a theorem from `Mathlib.Data.List.Basic` to Std.
* Add theorems about `List.isPrefixOf` and `List.isSuffixOf`.
* Add `Std.Data.List.SplitOnList`.

`List.splitOnList` is a specification for `String.splitOn`. We need it to prove `String.splitOn_of_valid`, which is on the to-do list in `Std.Data.String.Lemmas`.

Co-authored-by: Mario Carneiro <di.gama@gmail.com>
Co-authored-by: Scott Morrison <scott.morrison@gmail.com>

---

See https://github.com/leanprover/std4/pull/495.

### DIFF
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -139,6 +139,8 @@ theorem drop_eq_get_cons : ∀ {n} {l : List α} (h), drop n l = get l ⟨n, h�
 @[simp] theorem isEmpty_nil : ([] : List α).isEmpty = true := rfl
 @[simp] theorem isEmpty_cons : (x :: xs : List α).isEmpty = false := rfl
 
+theorem isEmpty_iff_eq_nil {l : List α} : l.isEmpty ↔ l = [] := by cases l <;> simp [isEmpty]
+
 /-! ### append -/
 
 theorem append_eq_append : List.append l₁ l₂ = l₁ ++ l₂ := rfl
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -2116,9 +2116,8 @@ theorem length_le_of_isPrefixOf [BEq α] {l₁ l₂ : List α} (h : l₁.isPrefi
   | [],    _     => simp
   | _::_,  []    => simp [isPrefixOf] at h
   | a::as, b::bs =>
-    simp [isPrefixOf] at h
-    simp [Nat.succ_le_succ_iff]
-    exact length_le_of_isPrefixOf h.2
+    simp only [isPrefixOf, Bool.and_eq_true] at h
+    simpa [Nat.succ_le_succ_iff] using length_le_of_isPrefixOf h.2
 
 theorem length_le_of_isSuffixOf [BEq α] {l₁ l₂ : List α} (h : l₁.isSuffixOf l₂) :
     l₁.length ≤ l₂.length := by
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -2108,6 +2108,22 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
       refine ⟨fun h => h₀ _ (mem_of_mem_drop h) rfl, ?_⟩
       exact disjoint_take_drop h₁ (Nat.le_of_succ_le_succ h)
 
+/-! ### isPrefixOf and isSuffixOf -/
+
+theorem length_le_of_isPrefixOf [BEq α] {l₁ l₂ : List α} (h : l₁.isPrefixOf l₂) :
+    l₁.length ≤ l₂.length := by
+  match l₁, l₂ with
+  | [],    _     => simp
+  | _::_,  []    => simp [isPrefixOf] at h
+  | a::as, b::bs =>
+    simp [isPrefixOf] at h
+    simp [Nat.succ_le_succ_iff]
+    exact length_le_of_isPrefixOf h.2
+
+theorem length_le_of_isSuffixOf [BEq α] {l₁ l₂ : List α} (h : l₁.isSuffixOf l₂) :
+    l₁.length ≤ l₂.length := by
+  simpa [length_reverse] using length_le_of_isPrefixOf h
+
 /-! ### takeWhile and dropWhile -/
 
 @[simp] theorem takeWhile_append_dropWhile (p : α → Bool) :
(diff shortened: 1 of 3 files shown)


Std/Data/List/Lemmas.lean
```diff
@@ -2185,6 +2187,22 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+
+theorem length_le_of_isPrefixOf [BEq α] {l₁ l₂ : List α} (h : l₁.isPrefixOf l₂) :
+    l₁.length ≤ l₂.length := by
+  match l₁, l₂ with
+  | [],    _     => simp
+  | _::_,  []    => simp [isPrefixOf] at h
+  | a::as, b::bs =>
+    simp [isPrefixOf] at h
+    simp [Nat.succ_le_succ_iff]
+    exact length_le_of_isPrefixOf h.2
```
semorrison, line 2200: Non-terminal `simp`. Can you use `simpa`? Or `simp only`?

Std/Data/List/SplitOnList.lean
```diff
@@ -0,0 +1,72 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Bulhwi Cha, Mario Carneiro
+-/
+import Std.Data.Nat.Lemmas
+import Std.Data.List.Lemmas
+
+namespace List
+
+/-- Returns `(l₁, l₂)` for the first split `l = l₁ ++ l₂` such that `P l₂` returns true. -/
+def splitOnceP (P : List α → Bool) (l : List α) : Option (List α × List α) :=
```
semorrison, line 12: If I only read the name and signature (and not the doc-string), I would be surprised by the behaviour here; I would have guessed "`P l₁` returns true".

I appreciate that this is motivated by `splitOnList`, but could we either have a more information name, or make this private? (or defined via a `where` clause in `splitOnList`?)

Std/Data/List/SplitOnList.lean
```diff
@@ -0,0 +1,72 @@
...
+theorem eq_append_of_splitOnceP (P : List α → Bool) (l : List α) :
+    ∀ l₁ l₂, splitOnceP P l = some (l₁, l₂) → l = l₁ ++ l₂ := by
+  induction l with
+  | nil => simp [splitOnceP]
+  | cons a l ih =>
+    simp [splitOnceP]
+    if h : P (a :: l) then
+      simp [h]
+    else
+      simp [h]
```
semorrison, line 29: Non-terminal simp.

Std/Data/List/SplitOnList.lean
```diff
@@ -0,0 +1,72 @@
...
+      | none => simp
+      | some (lf, rt) => simp; apply ih; assumption
+
+/--
+Split a list at every occurrence of a separator list. The separators are not in the result.
+```
+[1, 1, 2, 3, 2, 4, 4].splitOnList [2, 3] = [[1, 1], [2, 4, 4]]
+```
+-/
+def splitOnList [BEq α] (l sep : List α) : List (List α) :=
```
semorrison, line 54: Could I ask what the application of this is? It doesn't seem like a particularly common function to need!

semorrison:
awaiting-author

Std/Data/List/SplitOnList.lean
digama0: This function is derived from specification of `String.splitOn`, see #495

semorrison: Got it, thanks. That makes sense.

Std/Data/List/Lemmas.lean
chabulhwi: I used `simpa` to avoid a non-terminal `simp`.

Std/Data/List/SplitOnList.lean
chabulhwi: I replaced it with a `simp only`.

Std/Data/List/SplitOnList.lean
chabulhwi: Do you think `splitOnceRightP` is a good name?

Std/Data/List/SplitOnList.lean
chabulhwi: We need it to prove `String.splitOn_of_valid`, which is on the to-do list in `Std.Data.String.Lemmas`.

chabulhwi:
awaiting-review

Std/Data/List/SplitOnList.lean
semorrison: Sounds better to me.

Std/Data/List/SplitOnList.lean
```diff
@@ -0,0 +1,70 @@
+/-
+Copyright (c) 2024 Bulhwi Cha, Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Bulhwi Cha, Mario Carneiro
+-/
+import Std.Data.Nat.Lemmas
+import Std.Data.List.Lemmas
+
```
semorrison, line 8: Could we have a module doc-string? I think it would be help to explain that the motivation is for specifying the behaviour of String.splitOn, even before that is actually implemented.

Std/Data/List/SplitOnList.lean
```diff
@@ -0,0 +1,70 @@
...
+    match e : splitOnceP sep.isPrefixOf l with
+    | none => [l]
+    | some (l₁, l₂) =>
+      have : length l₂ - sep.length < length l := by
+        rw [eq_append_of_splitOnceP (sep.isPrefixOf) l l₁ l₂ e, length_append]
+        calc
+          length l₂ - length sep < length l₂ := by
+            apply Nat.sub_lt_self
+            · apply length_pos.mpr; simp [← isEmpty_iff_eq_nil, h]
+            · apply length_le_of_isPrefixOf (P_of_splitOnceP sep.isPrefixOf l l₁ l₂ e)
```
semorrison, line 65:
```suggestion
          length l₂ - length sep < length l₂ :=
            Nat.sub_lt_self (by simp [length_pos, ← isEmpty_iff_eq_nil, h])
             (length_le_of_isPrefixOf (P_of_splitOnceP sep.isPrefixOf l l₁ l₂ e))
```

semorrison (approved):
Otherwise, LGTM.

Std/Data/List/SplitOnList.lean
chabulhwi: Thanks!

Std/Data/List/SplitOnList.lean
chabulhwi: I added it!

PR #532, opened 2024-01-15
chore: Add empty collection instance to BinomialHeap

joehendrix:


Std/Data/BinomialHeap/Basic.lean
```diff
@@ -502,6 +502,7 @@ variable {α : Type u} {le : α → α → Bool}
 /-- `O(1)`. Make a new empty binomial heap. -/
 @[inline] def empty : BinomialHeap α le := mkBinomialHeap α le
 
+instance : EmptyCollection (BinomialHeap α le) := ⟨.empty⟩
 instance : Inhabited (BinomialHeap α le) := ⟨.empty⟩
```
fgdorais, line 507: Would it make sense to have an `Inhabited` instance derived from `EmptyCollection` higher up? It could always be overridden in the extremely unusual case where that's not the right choice for `default`.

semorrison: It does seem reasonable, but not here. Please either open an issue or a PR if you're interested in doing either!

PR #533, opened 2024-01-16
chore: reduce imports

semorrison:
Reduce imports.


semorrison:
@digama0, do you know why `lake exe shake Std` (run from Mathlib) doesn't report this?

digama0:
Uh oh, there was a bug which after fixing is generating loads more shake reports...

semorrison:
Good news! :-)

digama0:
I pushed the results of the shake report to this branch. I'll make a PR separately to mathlib with the noshake.json changes for std, such that running `lake exe shake Std` works as expected. (This is a bit of a layering violation, we should probably upstream `shake` if we want to keep this working.)

joehendrix (approved):

PR #534, opened 2024-01-16
chore: split Std.Data.Int.Lemmas

semorrison:
This PR is prompted by trying to reduce imports of `omega`, but does not yet make progress in that direction.

We split `Std.Data.Int.Lemmas` into two files, adding `Std.Data.Int.Order`.

No statements have changed, but there has been lots of shuffling to try to reunite lost friends, and put things into sections.

I did change the proof of `Int.mul_eq_zero` completely, making it simpler and not requiring any order lemmas.

There are further sensible splits to `Std.Data.Int.Order`, which I am inclined to do in a follow-up PR:
* move the material about order and multiplication to a separate file (`OrderedRing?`)
* move the material about `natAbs` and `toNat` to a separate file (possible also `sign`)
These are two independent "tails" of the current file.


digama0:
Note, this is a breaking change because imports of `Std.Data.Int.Lemmas` no longer pick up lemmas from `Std.Data.Int.Order`. It makes me wonder whether having `Lemmas` files are a good idea in the first place, because they tend to encourage a large file and to some extent overlap in purpose with `Std.Data.Int`. With the version in this PR, I'm no longer clear on what exactly the semantic meaning of `import Std.Data.Int.Lemmas` is supposed to be.

semorrison:
I agree that the current interpretation of `Lemma` is an invitation to a bloated disorganised files that grow without bound. :-)

I quite like Mathlib's `Def` and `Basic` split, followed by growing collection of other files for subtopics. With the addition of Std's folder-files, there would be at least some protection against changes that move lemmas between `Basic` and subtopic files: you can revert you import to the folder-file, and then reshake.

digama0:
I also like the `Def` and `Basic` split, but this is alpha-equivalent to the issue being discussed here. What is the upper bound on `Foo.Basic` files in mathlib? I think there is no clear upper bound, and this has historically lead to a lot of import bloat because actually basic things are put next to much less basic things which are nevertheless topically related. With `Defs` we have a clear place to draw the line: things with computational content (although even here we can't always accommodate everything, if there are definitions that depend on a big proof like in `Std.Data.RBMap.WF`). For lemmas I don't really know when to say that basic stuff ends here.

semorrison:
It feels unlikely that there will be a general answer here, other the hard or soft caps on LoC. I think we should encourage as much as possible that files are organised into sections with headings (and ideally doc-string introductions to each section!).

I think this split is good: `Order` is a natural subtopic, and it cleanly divides the file in two, and then generally splits and better than not splits!

But I'm happy to put them back together: this PR also does a fair amount of shuffling lemmas back into their thematic groups, so would still be an improvement if we just re-concatenated the files.

semorrison:
In either case, there are now later PRs depending on this one, so a decision would be nice. :-)

joehendrix:
I think I jumped the gun on `will-merge-sooon`.  I like reorganizing things out of very large files.

I'm wondering if we should work towards a hard requirement that tactics only need things in an `Init` namespace.

Wrt to the lemmas files, I'd prefer we just make sure the `Std.Data.Int` level modules contain everything we thing people should normally use.  I think the `Std.Data.Int.Basic` module is reasonable though.

I'm in favor of merging this and punting on the longer-term organization though in this PR though.

digama0:
> I'm wondering if we should work towards a hard requirement that tactics only need things in an `Init` namespace.

I agree with this. Currently the only exceptions I am aware of are in `Std.Logic`; I did a pass that did roughly this not too long ago. The nice thing about `Init` namespace is that it doesn't have the "grow without bound" issue because it has a clear inclusion criterion: only things which are transitive dependencies of some user-facing tactic, definition or theorem defined in Std should be in an `Init` folder. (The original design here also includes `Foo.Init.Init.Lemmas` for lemmas that are needed by `Foo.Init.Basic`, but so far we haven't needed that many alternations yet.)

PR #535, opened 2024-01-16
Add specialized associative List folds

joehendrix:
This uses the associativity, commutativity and identity classes recently added to Lean core to define folds with simplification lemmas that can move intermediate results outside the fold for better use in tactics.

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -258,7 +258,7 @@ where
 instance {α : Type _} : Std.Associative (α := List α) List.append where
   assoc := append_assoc
 
-instance {α : Type _} : Std.LawfulIdentity (α := List α) List.append nil where
+instance {α : Type _} : Std.LawfulIdentity (α := List α) (· ++ ·) [] where
   left_id := nil_append
   right_id := append_nil
 
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -298,8 +298,7 @@ Product of elements in a list.
 
 `List.prod [a, b, c] = ((1 * a) * b) * c`
 -/
-def prod {α} [Mul α] [One α] : List α → α := foldl (· * ·) 1
-
+def prod {α} [Mul α] [OfNat α 1] : List α → α := foldl (· * ·) 1
 
 /--
 `prodMap f l` returns `prod (map f l)` without creating an intermediate list.
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -280,7 +280,9 @@ def foldMap {α : Type _} {β : Type _} (op : β → β → β) {o : β} [Std.Le
   l.foldl (init := o) (fun b a => op b (f a))
 
 /--
-Sum of elements in list.
+Sum of elements in a list.
+
+`List.sum [a, b, c] = ((0 + a) + b) + c`
 -/
 def sum [Add α] [OfNat α 0] (l : List α) : α :=
   l.foldl (init := 0) (· + ·)
@@ -292,10 +294,12 @@ def sumMap {α : Type u} {β : Type v} [Add β] [OfNat β 0] (f : α → β) (l 
   l.foldl (init := 0) (· + f ·)
 
 /--
-Product of elements in list.
+Product of elements in a list.
+
+`List.prod [a, b, c] = ((1 * a) * b) * c`
 -/
-def prod [Mul α] [OfNat α 1] (l : List α) : α :=
-  l.foldl (init := 1) (· * ·)
+def prod {α} [Mul α] [One α] : List α → α := foldl (· * ·) 1
+
 
 /--
 `prodMap f l` returns `prod (map f l)` without creating an intermediate list.
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -255,7 +255,7 @@ where
 
 /-! ## Algebraic instances -/
 
-instance {α : Type _} : Std.Associative (α := List α) List.append where
+instance {α : Type _} : Std.Associative (α := List α) (· ++ ·) where
   assoc := append_assoc
 
 instance {α : Type _} : Std.LawfulIdentity (α := List α) (· ++ ·) [] where
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -253,6 +253,44 @@ where
     | _::_ => by simp [intercalateTR.go, go]
     simp [intersperse, go]
 
+/-! ## Algebraic instances -/
+
+instance {α : Type _} : Std.Associative (α := List α) List.append where
+  assoc := append_assoc
+
+instance {α : Type _} : Std.LawfulIdentity (α := List α) List.append nil where
+  left_id := nil_append
+  right_id := append_nil
+
+/-! ## Generic fold and sum operations. -/
+
+/--
+Applies a fold operation to join elements together.
+-/
+@[nolint unusedArguments] -- Std.LeftIdentity is needed for inference only.
+def fold {α : Type _} (op : α → α → α) {o : α} [Std.LeftIdentity op o] (l : List α) : α :=
+  l.foldl (init := o) op
+
+/--
+`foldMap op f l` is equivalent to `fold op (map f l)`.
+-/
+@[nolint unusedArguments] -- Std.LeftIdentity is needed for inference only.
+def foldMap {α : Type _} {β : Type _} (op : β → β → β) {o : β} [Std.LeftIdentity op o] (f : α → β)
+    (l : List α) : β :=
+  l.foldl (init := o) (fun b a => op b (f a))
+
+/--
+Returns sum of elements in list.
+-/
+def sum [Add α] [OfNat α 0] (l : List α) : α :=
+  l.foldl (init := 0) (· + ·)
+
+/--
+sumMap f l returns the sum of `sum (map f l)` without creating an intermediate list.
+-/
+def sumMap {α : Type u} {β : Type v} [Add β] [OfNat β 0] (f : α → β) (l : List α) : β :=
+  l.foldl (init := 0) (· + f ·)
+
 /-! ## New definitions -/
 
 /--
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -280,16 +280,28 @@ def foldMap {α : Type _} {β : Type _} (op : β → β → β) {o : β} [Std.Le
   l.foldl (init := o) (fun b a => op b (f a))
 
 /--
-Returns sum of elements in list.
+Sum of elements in list.
 -/
 def sum [Add α] [OfNat α 0] (l : List α) : α :=
   l.foldl (init := 0) (· + ·)
 
 /--
-sumMap f l returns the sum of `sum (map f l)` without creating an intermediate list.
+`sumMap f l` returns `sum (map f l)` without creating an intermediate list.
 -/
 def sumMap {α : Type u} {β : Type v} [Add β] [OfNat β 0] (f : α → β) (l : List α) : β :=
   l.foldl (init := 0) (· + f ·)
+
+/--
+Product of elements in list.
+-/
+def prod [Mul α] [OfNat α 1] (l : List α) : α :=
+  l.foldl (init := 1) (· * ·)
+
+/--
+`prodMap f l` returns `prod (map f l)` without creating an intermediate list.
+-/
+def prodMap {α : Type u} {β : Type v} [Mul β] [OfNat β 1] (f : α → β) (l : List α) : β :=
+  l.foldl (init := 1) (· * f ·)
 
 /-! ## New definitions -/
 
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -375,26 +375,22 @@ theorem foldr_assoc (op : α → α → α) [h : Associative op] (a b : α) (l :
   | nil => simp
   | cons c l ind => simp [ind, h.assoc]
 
-@[simp]
-theorem fold_nil (op : α → α → α) [LeftIdentity op o] : fold op (@nil α) = o := rfl
+@[simp] theorem fold_nil (op : α → α → α) [LeftIdentity op o] : fold op (@nil α) = o := rfl
 
-@[simp]
-theorem fold_cons (op : α → α → α) [ha : Associative op] [hl : LawfulIdentity op o]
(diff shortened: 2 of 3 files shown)


Std/Data/Nat/Lemmas.lean
```diff
@@ -133,6 +131,26 @@ theorem recDiagOn_succ_succ {motive : Nat → Nat → Sort _} (zero_zero : motiv
...
+
+instance : Std.IsAssociative (α := Nat) (· * ·) where
+  assoc := Nat.mul_assoc
+
+instance : Std.IsCommutative (α := Nat) (· * ·) where
+  comm := Nat.mul_comm
+
+instance : Std.IsCommId (α := Nat) (· * ·) 1 where
+  right_id := Nat.mul_one
+
```
fgdorais, line 153: These could be in Basic? I'm sincerely asking, this is debatable.

Std/Data/List/Lemmas.lean
```diff
@@ -181,6 +181,9 @@ theorem mem_iff_append {a : α} {l : List α} : a ∈ l ↔ ∃ s t : List α, l
 
 /-! ### map -/
 
+@[simp]
+theorem map_eq_map (f : α → β) (l : List α) : f <$> l = map f l := rfl
```
fgdorais, line 185: I don't think there is a style guidance about this but if it can all fit in one line then I tend to prefer not breaking a line after `@[simp]`. This applies elsewhere but it's just nitpick.

Std/Classes/Algebra.lean
```diff
@@ -0,0 +1,76 @@
...
+  assoc := IsAssociative.assoc
+
+/--
+A binary operation with an associated identity element that can be inferred
+through class inference.
+
+This intentionally does not have associated lemmas.  For lemmas, see @IsLeftId@,
+@IsRightId@ and @IsId .
+ -/
+class HasId {α : Type u} (op : α → α → α) (o : outParam α) : Prop where
```
fgdorais, line 50:
```suggestion
class HasId {α : Type u} (op : α → α → α) (o : outParam α) : Prop
```

Std/Classes/Algebra.lean
```diff
@@ -0,0 +1,76 @@
...
+These classes index properties by the operation rather than the type as
+in Haskell.
+-/
+
+namespace Std
+
+/-- A symmetric operation. -/
+class IsSymmOp {α : Type u} {β : Sort v} (op : α → α → β) : Prop where
+  /-- Symmetric operation -/
+  symm_op : ∀ a b, op a b = op b a
```
fgdorais, line 22: Where is this used other than the one derived instance below? I think this could stay in Mathlib until needed.

joehendrix: It's not.  I added it because it was next to the other declarations and we have a `Symm` tactic.

digama0:
Note that these classes in mathlib are deprecated and slated for removal; they were in lean 3 core and rarely used. I think we should add these on an as-needed basis.

Std/Data/Nat/Lemmas.lean
digama0: I would be inclined not to unless it appears in a definition. I don't think it does in this case, because list folds don't require any properties on the operation (unlike finset folds)

Std/Data/List/Lemmas.lean
joehendrix, line 185:
```suggestion
@[simp] theorem map_eq_map (f : α → β) (l : List α) : f <$> l = map f l := rfl
```

Std/Data/Nat/Lemmas.lean
joehendrix: `List.fold` does have a typeclass dependency on `HasId` in this implementation to find the identity (there aren't any properties though in that class -- it's essentially a way to associate an identity value to a function.

I think that's a slight argument for putting them in `Basic` so that the typeclass instance is available (via `IsCommId`).  I kept it here though as they are almost entirely for verification though.

joehendrix:
I'm open to other approaches then reviving these classes, but it did seem like a way to get the monoid properties I wanted for writing fold/sum lemmas.

`IsSymmOp` isn't strictly needed but it seemed to naturally go with the other properties, and the associated Symm tactic is already in Std.

fgdorais:
Is the only reason not to use the Lean core classes that these classes are `Prop`?

joehendrix:
The associativity and commutative ones don't have a good reason.  Perhaps I should use the Lean core versions.  The Identity classes use an out parameter for the identity/neutral element.

fgdorais:
Maybe it would be better to add the missing `outParam` in core? Perhaps also make these classes `Prop` to avoid diamonds? This kind of PR to core can backfire in interesting ways but that would give more guidance what to do in Std.

Another option is to move `ac_rfl` to Std. That way we could more freely turn it into something more useful.

Std/Classes/Algebra.lean
fgdorais: I experimented with deleting `ac_rfl` from core and nothing seems to break: https://github.com/fgdorais/lean4/tree/ac_rfl

joehendrix:
I'm looking into both changing the core axioms and/or moving/removing ac_rfl.

If ac_rfl is removed from core, then technically the classes could live in Std.  However, I think it's good for tactics that don't use Std (potentially so they can migrate to Std) to have a common classes for this.

joehendrix:
I'll revisit this when leanprover/lean4#3195 is merged.

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+/--
+`foldMap op f l` is equivalent to `fold op (map f l)`.
+-/
+@[nolint unusedArguments] -- HasId needed for inference only.
+def foldMap {α : Type _} {β : Type _} (op : β → β → β) {o : β} [Std.LeftIdentity op o] (f : α → β)
+    (l:List α) : β :=
+  l.foldl (init := o) (fun b a => op b (f a))
+
+/--
+Returns sum of elements in list addinging init in.
```
semorrison, line 284:
```suggestion
Returns sum of elements in list.
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+-/
+def sum [Add α] [OfNat α 0] (l:List α) : α :=
```
semorrison, line 286:
```suggestion
def sum [Add α] [OfNat α 0] (l : List α) : α :=
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+  l.foldl (init := 0) (· + ·)
+
+/--
+sumMap f l returns the sum of `sum (map f l)` without creating an intermediate list.
+-/
+def sumMap {α : Type u} {β : Type v} [Add β] [OfNat β 0] (f : α → β) (l:List α) : β :=
```
semorrison, line 292:
```suggestion
def sumMap {α : Type u} {β : Type v} [Add β] [OfNat β 0] (f : α → β) (l : List α) : β :=
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+  assoc := append_assoc
+
+instance {α : Type _} : Std.LawfulIdentity (α := List α) List.append nil where
+  left_id := nil_append
+  right_id := append_nil
+
+/-! ## Generic fold and sum operations. -/
+
+/--
+Applies a fold operation to join elements togethrr.
```
fgdorais, line 269:
```suggestion
Applies a fold operation to join elements together.
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+-/
+@[nolint unusedArguments] -- HasId needed for inference only.
```
fgdorais, line 271:
```suggestion
@[nolint unusedArguments] -- Std.LeftIdentity is needed for inference only.
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+def fold {α : Type _} (op : α → α → α) {o : α} [Std.LeftIdentity op o] (l:List α) : α :=
+  l.foldl (init := o) op
+
+/--
+`foldMap op f l` is equivalent to `fold op (map f l)`.
+-/
+@[nolint unusedArguments] -- HasId needed for inference only.
```
fgdorais, line 278:
```suggestion
@[nolint unusedArguments] -- Std.LeftIdentity is needed for inference only.
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+  left_id := nil_append
+  right_id := append_nil
+
+/-! ## Generic fold and sum operations. -/
+
+/--
+Applies a fold operation to join elements togethrr.
+-/
+@[nolint unusedArguments] -- HasId needed for inference only.
+def fold {α : Type _} (op : α → α → α) {o : α} [Std.LeftIdentity op o] (l:List α) : α :=
```
fgdorais, line 272:
```suggestion
def fold {α : Type _} (op : α → α → α) {o : α} [Std.LeftIdentity op o] (l : List α) : α :=
```

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,44 @@ where
...
+  l.foldl (init := o) op
+
+/--
+`foldMap op f l` is equivalent to `fold op (map f l)`.
+-/
+@[nolint unusedArguments] -- HasId needed for inference only.
+def foldMap {α : Type _} {β : Type _} (op : β → β → β) {o : β} [Std.LeftIdentity op o] (f : α → β)
+    (l:List α) : β :=
```
fgdorais, line 280:
```suggestion
    (l : List α) : β :=
```

Std/Data/List/Lemmas.lean
```diff
@@ -361,8 +363,87 @@ theorem map_snd_zip :
...
+  | nil => simp
+  | cons c l ind => simp [ind, h.assoc]
+
+theorem foldr_assoc (op : α → α → α) [h : Std.Associative op] (a b : α) (l : List α) :
+    foldr op (op a b) l = op (foldr op a l) b := by
+  induction l generalizing a b with
+  | nil => simp
+  | cons c l ind => simp [ind, h.assoc]
+
+open Std
```
fgdorais, line 381: Move to top?

Std/Data/List/Lemmas.lean
```diff
@@ -361,8 +363,87 @@ theorem map_snd_zip :
...
+
+@[simp]
+theorem fold_nil (op : α → α → α) [LeftIdentity op o] : fold op (@nil α) = o := rfl
+
+@[simp]
+theorem fold_cons (op : α → α → α) [ha : Associative op]  [hl : LawfulIdentity op o]
```
fgdorais, line 387:
```suggestion
theorem fold_cons (op : α → α → α) [ha : Associative op] [hl : LawfulIdentity op o]
```

Std/Data/List/Lemmas.lean
```diff
@@ -361,8 +363,87 @@ theorem map_snd_zip :
...
+    (a : α) (l : List α) : fold op (a :: l) = op a (fold op l) := by
+  unfold fold
+  simp [foldl_cons]
+  conv => lhs; rw [hl.left_id, ←hl.right_id a]
+  simp [foldl_assoc, hl.left_id]
+
+@[simp]
+theorem fold_append (op : α → α → α) [ha : Associative op]  [hl : LawfulIdentity op o]
```
fgdorais, line 395:
```suggestion
theorem fold_append (op : α → α → α) [ha : Associative op] [hl : LawfulIdentity op o]
```

Std/Data/List/Lemmas.lean
```diff
@@ -361,8 +363,87 @@ theorem map_snd_zip :
...
+  induction l generalizing init with
+  | nil => rfl
+  | cons a l ind => simp [ind]
+
+@[simp]
+theorem sum_nil [Add α] [OfNat α 0] : sum (@nil α) = 0 := rfl
+
+@[simp]
+theorem sum_cons [Add α] [h : OfNat α 0]
+    [ha : Associative (α := α) (· + ·)] [hl : Std.LawfulIdentity (α := α) (· + ·) 0]
```
fgdorais, line 424:
```suggestion
    [ha : Associative (α := α) (· + ·)] [hl : LawfulIdentity (α := α) (· + ·) 0]
```

Std/Data/List/Lemmas.lean
```diff
@@ -361,8 +363,87 @@ theorem map_snd_zip :
...
+    (a : α) (l : List α) : sum (a :: l) = a + sum l := by
+  unfold sum
+  simp [foldl_cons]
+  conv => lhs; rw [hl.left_id, ←hl.right_id a]
+  simp [foldl_assoc, hl.left_id]
+
+@[simp]
+theorem sum_append [Add α] [OfNat α 0]
+    [ha : Std.Associative (α := α) (· + ·)]  [hl : Std.LawfulIdentity (α := α) (· + ·) 0]
```
fgdorais, line 433:
```suggestion
    [ha : Associative (α := α) (· + ·)]  [hl : LawfulIdentity (α := α) (· + ·) 0]
```

fgdorais (changes requested):

semorrison:
I've changed the base of the PR to `main`, as it is now on `v4.6.0-rc1` which includes the upstream changes. I've also set the `lean-toolchain` to `v4.6.0-rc1` to resolve the conflict.

Std/Data/List/Basic.lean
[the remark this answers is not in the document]
joehendrix: I've added these.

Std/Data/List/Basic.lean
```diff
@@ -254,6 +254,59 @@ where
     | _::_ => by simp [intercalateTR.go, go]
     simp [intersperse, go]
 
+/-! ## Algebraic instances -/
+
+instance {α : Type _} : Std.Associative (α := List α) List.append where
+  assoc := append_assoc
+
+instance {α : Type _} : Std.LawfulIdentity (α := List α) List.append nil where
```
fgdorais, line 262: I think this should be
```suggestion
instance {α : Type _} : Std.LawfulIdentity (α := List α) (· ++ ·) [] where
```
so that it works with `Append` notation.

fgdorais (approved):

fgdorais:
All the prereqs are good now, so this could be merged soon.

fgdorais:
Looks like the only issue left is the `to_additive` problem in Mathlib. We need an expert to deal with that...

fgdorais:
This PR was marked `will-merge-soon` since Februrary 16, I removed that tag on July 20. Presumably something went wrong along the way... What is the current status of this PR?

PR #538, opened 2024-01-17
feat: lemmas about Int.gcd

semorrison:
We move the definition of `Nat.lcm` from `Std.Data.Nat.Gcd` to `Std.Data.Nat.Basic`, and prove a few lemmas about `Int.gcd`, which are just thin wrappers around the corresponding lemams for `Nat.gcd`, dealing with `natAbs` as necessary.

- [ ] depends on #537

### DIFF
--- a/Std/Data/Int/Gcd.lean
+++ b/Std/Data/Int/Gcd.lean
@@ -23,11 +23,11 @@ theorem gcd_dvd_right {a b : Int} : (gcd a b : Int) ∣ b := by
   rw [← Int.ofNat_dvd] at this
   exact Int.dvd_trans this natAbs_dvd_self
 
-@[simp] theorem gcd_one_left {a : Int} : gcd 1 a = 1 := by simp [gcd]
-@[simp] theorem gcd_one_right {a : Int} : gcd a 1 = 1 := by simp [gcd]
+@[simp] theorem one_gcd {a : Int} : gcd 1 a = 1 := by simp [gcd]
+@[simp] theorem gcd_one {a : Int} : gcd a 1 = 1 := by simp [gcd]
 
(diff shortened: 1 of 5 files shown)


Std/Data/Int/Gcd.lean
```diff
@@ -0,0 +1,45 @@
...
+theorem gcd_dvd_right {a b : Int} : (gcd a b : Int) ∣ b := by
+  have := Nat.gcd_dvd_right a.natAbs b.natAbs
+  rw [← Int.ofNat_dvd] at this
+  exact Int.dvd_trans this natAbs_dvd_self
+
+@[simp] theorem gcd_one_left {a : Int} : gcd 1 a = 1 := by simp [gcd]
+@[simp] theorem gcd_one_right {a : Int} : gcd a 1 = 1 := by simp [gcd]
+
+@[simp] theorem gcd_neg_left {a b : Int} : gcd a (-b) = gcd a b := by simp [gcd]
+@[simp] theorem gcd_neg_right {a b : Int} : gcd a (-b) = gcd a b := by simp [gcd]
```
fgdorais, line 30: This is one of those edge cases where an exception could make sense, but these should use infix names.
```suggestion
@[simp] theorem one_gcd {a : Int} : gcd 1 a = 1 := by simp [gcd]
@[simp] theorem gcd_one {a : Int} : gcd a 1 = 1 := by simp [gcd]

@[simp] theorem neg_gcd {a b : Int} : gcd a (-b) = gcd a b := by simp [gcd]
@[simp] theorem gcd_neg {a b : Int} : gcd a (-b) = gcd a b := by simp [gcd]
```

fgdorais (approved):
Approved modulo comment.

PR #540, opened 2024-01-17
fix: make exfalso actually do what it claims to

eric-wieser:
Before this change,
```lean
example {A} : False → A := by exfalso
```
would work, but the docstring ("converts a goal `⊢ tgt` into `⊢ False`") says it should not.


semorrison:
Could you please add some tests exercising the intended behaviour?

semorrison (approved):

PR #542, opened 2024-01-18
feat: restore lemmas about order and Int.ediv

semorrison:
These lemmas had been commented out after the change in meaning of `/` (the `div` vs `ediv` change).

I had to restore to preliminary lemmas that seem to have gone missing. The proofs of these are note particularly golfed. Please feel free to merge, golf, or ask me to golf as you see fit!

- [x] depends on: #534


joehendrix:
@semorrison This got dropped with all the migrations.  Should it be resolved here or in core?

Std/Data/Int/DivMod.lean
```diff
@@ -1056,71 +1082,71 @@ theorem bmod_natAbs_plus_one (x : Int) (w : 1 < x.natAbs) : bmod x (x.natAbs + 1
...
-theorem div_eq_div_of_mul_eq_mul {a b c d : Int}
+theorem ediv_eq_ediv_of_mul_eq_mul {a b c d : Int}
     (H2 : d ∣ c) (H3 : b ≠ 0) (H4 : d ≠ 0) (H5 : a * d = b * c) : a / b = c / d :=
-  Int.div_eq_of_eq_mul_right H3 <| by
-    rw [← Int.mul_div_assoc _ H2] <;> exact (Int.div_eq_of_eq_mul_left H4 H5.symm).symm
+  Int.ediv_eq_of_eq_mul_right H3 <| by
+    rw [← Int.mul_ediv_assoc _ H2]; exact (Int.ediv_eq_of_eq_mul_left H4 H5.symm).symm
+
+/-!
+### The following lemmas have been commented out here for a while, and need restoration.
```
fgdorais, line 1140: Maybe use `proof_wanted`?

fgdorais (approved):

Std/Data/Int/DivMod.lean
semorrison: Maybe later. :-)

semorrison:
I think this is still orthogonal to our changes on `nightly-testing`, so I'm merging before it's too late.

PR #543, opened 2024-01-18
feat: cooper resolution

kim-em:
This doesn't need review for now. This is a statement of "Cooper resolution" as used in Jovanović and de Moura's "Cutting to the chase". This may eventually end up in an alternative to `omega`.

- [x] depends on #542


kim-em:
This has been replaced by https://github.com/leanprover/lean4/pull/6862

PR #544, opened 2024-01-18
feat: `HashMap.empty_find?` + helper lemmas

technosentience:
Unsure how this interacts with #279, but it's marked as `proof_wanted` so why not.

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -30,8 +30,8 @@ theorem getElem?_neg [GetElem Cont Idx Elem Dom]
 
 @[simp] theorem mkArray_data (n : Nat) (v : α) : (mkArray n v).data = List.replicate n v := rfl
 
-@[simp] theorem mkArray_getElem (n : Nat) (v : α) (h : i < n) :
-    (mkArray n v)[i]'(by simp [h]) = v := by simp [Array.getElem_eq_data_get]
+@[simp] theorem getElem_mkArray (n : Nat) (v : α) (h : i < (mkArray n v).size) :
+    (mkArray n v)[i]'h = v := by simp [Array.getElem_eq_data_get]
 
 namespace Array
 
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -30,6 +30,9 @@ theorem getElem?_neg [GetElem Cont Idx Elem Dom]
 
 @[simp] theorem mkArray_data (n : Nat) (v : α) : (mkArray n v).data = List.replicate n v := rfl
 
+@[simp] theorem mkArray_getElem (n : Nat) (v : α) (h : i < n) :
+    (mkArray n v)[i]'(by simp [h]) = v := by simp [Array.getElem_eq_data_get]
+
 namespace Array
 
 attribute [simp] isEmpty uget
--- a/Std/Data/HashMap/Lemmas.lean
+++ b/Std/Data/HashMap/Lemmas.lean
@@ -4,12 +4,15 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Scott Morrison
 -/
 import Std.Data.HashMap.Basic
+import Std.Data.Array.Lemmas
 import Std.Util.ProofWanted
 
 namespace Std.HashMap
 
-@[simp] proof_wanted empty_find? [BEq α] [Hashable α] {a : α} :
-    (∅ : HashMap α β).find? a = none
+@[simp] theorem empty_buckets [BEq α] [Hashable α] : (∅ : HashMap α β).1.buckets.1 = mkArray 8 AssocList.nil := rfl
+
+@[simp] theorem empty_find? [BEq α] [Hashable α] {a : α} :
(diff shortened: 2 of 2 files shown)


technosentience:
awaiting-review

Std/Data/HashMap/Lemmas.lean
```diff
@@ -4,12 +4,15 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
 -/
 import Std.Data.HashMap.Basic
+import Std.Data.Array.Lemmas
 import Std.Util.ProofWanted
 
 namespace Std.HashMap
 
-@[simp] proof_wanted empty_find? [BEq α] [Hashable α] {a : α} :
-    (∅ : HashMap α β).find? a = none
+@[simp] theorem empty_buckets [BEq α] [Hashable α] : (∅ : HashMap α β).1.buckets.1 = mkArray 8 AssocList.nil := rfl
```
digama0, line 12: This should be in the `HashMap.Imp` namespace as it is talking about implementation details.

technosentience: Fixed + split in two.

Std/Data/HashMap/Lemmas.lean
```diff
@@ -4,12 +4,21 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
 
-@[simp] proof_wanted empty_find? [BEq α] [Hashable α] {a : α} :
-    (∅ : HashMap α β).find? a = none
+namespace Imp
+
+@[simp] theorem empty_buckets : (empty : Imp α β).buckets = ⟨mkArray 8 AssocList.nil, by simp⟩ := rfl
+
+end Imp
+
+@[simp] theorem empty_imp [BEq α] [Hashable α] : (∅ : HashMap α β).1 = Imp.empty := rfl
```
joehendrix, line 18: It looks like `empty_imp` should be put in the `Imp` namespace similar to `empty_buckets`.

technosentience: Done, does the new name work?

Std/Data/Array/Lemmas.lean
```diff
@@ -30,6 +30,9 @@ theorem getElem?_neg [GetElem Cont Idx Elem Dom]
 
 @[simp] theorem mkArray_data (n : Nat) (v : α) : (mkArray n v).data = List.replicate n v := rfl
 
+@[simp] theorem mkArray_getElem (n : Nat) (v : α) (h : i < n) :
+    (mkArray n v)[i]'(by simp [h]) = v := by simp [Array.getElem_eq_data_get]
```
joehendrix, line 34:
```suggestion
@[simp] theorem getElem_mkArray (n : Nat) (v : α) (h : i < (mkArray n v).size) :
    (mkArray n v)[i]'h = v := by simp [Array.getElem_eq_data_get]
```

technosentience:
awaiting-review

digama0 (approved):

Std/Data/HashMap/Lemmas.lean
```diff
@@ -4,12 +4,21 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+import Std.Data.Array.Lemmas
 import Std.Util.ProofWanted
 
 namespace Std.HashMap
 
-@[simp] proof_wanted empty_find? [BEq α] [Hashable α] {a : α} :
-    (∅ : HashMap α β).find? a = none
+namespace Imp
+
+@[simp] theorem empty_buckets : (empty : Imp α β).buckets = ⟨mkArray 8 AssocList.nil, by simp⟩ := rfl
```
timotree3, line 14: Do we really want to be hard-coding 8 here? It seems like either this should be a constant `Imp.emptyCapacity` or we should avoid stating it. If we want to avoid stating this lemma entirely, the `simp` invocation in `empty_find?` could be fixed be adding `[Imp.empty', Imp.empty, Imp.Buckets.mk]`.

Std/Data/HashMap/Lemmas.lean
```diff
@@ -4,12 +4,21 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+@[simp] theorem empty_val [BEq α] [Hashable α] : (∅ : HashMap α β).val = Imp.empty := rfl
+
+end Imp
+
+@[simp] theorem empty_find? [BEq α] [Hashable α] {a : α} :
+    (∅ : HashMap α β).find? a = none := by simp [find?, Imp.find?, Imp.mkIdx, USize.modn_lt]
```
timotree3, line 21:
```suggestion
    (∅ : HashMap α β).find? a = none := by simp [find?, Imp.find?]
```
The goal is closed without those two lemmas on my machine

technosentience: Fixed, this was due to my previous definition of `getElem_mkArray`

Std/Data/HashMap/Lemmas.lean
joehendrix: That's a good catch.  I think the approach of adding additional defs to unfold is good.

Alternatively, `empty_buckets` should work over any user-defined `capacity` value passed into `Imp.empty`.

PR #546, opened 2024-01-19
feat: MVarId.let, the let-binding analogue of MVarId.note

semorrison:


digama0:
> inlining the `Option.getDM`.

I thought we upstreamed this already?

semorrison:
> > inlining the `Option.getDM`.
> 
> I thought we upstreamed this already?

Indeed, it has been upstreamed, but I'd thought previously that we couldn't import `Std.Data.Option.Basic` from `Std.Lean.Meta.Basic` without introducing a bad cycle. 

However this was wrong, the import seems harmless, so I've now added the import and both `let` and `note` can use it now.

digama0:
> Indeed, it has been upstreamed, but I'd thought previously that we couldn't import `Std.Data.Option.Basic` from `Std.Lean.Meta.Basic` without introducing a bad cycle.

If this is an issue, we can move it to `Std.Data.Option.Init.Basic`

semorrison:
> > Indeed, it has been upstreamed, but I'd thought previously that we couldn't import `Std.Data.Option.Basic` from `Std.Lean.Meta.Basic` without introducing a bad cycle.
> 
> If this is an issue, we can move it to `Std.Data.Option.Init.Basic`

It seems fine as is.

PR #547, opened 2024-01-19
chore: don't call symmSaturate repeatedly in solve_by_elim

semorrison:
The `symm` config option for `solve_by_elim` results in calling `symmSaturate`, to add a hypothesis `h_symm` for every hypothesis with an applicable `@[symm]` lemma.

This was being run every time we wanted to apply a lemma, rather than once at the beginning! I can't think of any reason this would be necessary, and it seems horribly inefficient.


semorrison:
Hmm, there's a problem in Mathlib.

PR #548, opened 2024-01-19
feat: let_projs tactic, recursively adding projections of hypotheses

semorrison:
`let_projs` adds let bindings for all projections of local hypotheses, recursively.

It's probably rarely useful interactively, but may be useful as a preprocessing step for tactics such as `solve_by_elim`.

### DIFF
--- a/Std/Tactic/LetProjs.lean
+++ b/Std/Tactic/LetProjs.lean
@@ -1,0 +1,105 @@
+/-
+Copyright (c) 2023 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Scott Morrison
+-/
+import Std.Lean.Meta.Basic
+import Std.Lean.Expr
+import Lean.Elab.Tactic.ElabTerm
+
+/-!
+# `let_projs`
+
+`let_projs` adds let bindings for all projections of local hypotheses, recursively.
+
+It should not be used interactively,
+but may be useful as a preprocessing step for tactics such as `solve_by_elim`.
+-/
+
+open Lean Meta
+
+/--
+Analogue of `Lean.getStructureFields`,
+but return an empty array rather than panicking if `structName` does not refer to a structure.
+-/
+def Lean.getStructureFields! (env : Environment) (structName : Name) : Array Name :=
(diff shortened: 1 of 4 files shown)


digama0:
I'm inclined to not surface this as a user-facing tactic at all, and just have a `MetaM` tactic for it. This introduces more API surface area which we have to avoid breaking later, without a compelling use case.

semorrison:
Good idea. I mostly wrote the syntactical frontend so I could test it. I've moved the syntax into the test file, and added a note that it is not useful otherwise.

Std/Tactic/LetProjs.lean
```diff
@@ -0,0 +1,82 @@
...
+/--
+Builds all the projections of an expression,
+returning an array of pairs consisting of the projection name
+and the projection applied to the original expression.
+-/
+def allProjs (e : Expr) : MetaM (Array (Name × Expr)) := do
+  let (c, _) := (← inferType e).getAppFnArgs
+  let env ← getEnv
+  unless isStructure env c do return #[]
+  (getStructureFields! env c).filterMapM fun f => return (f, ← mkProjection e f)
```
nomeata, line 42: `getStructureFields!" already returns `#[]` if it’s not a structure, so maybe
```suggestion
  (getStructureFields! env c).filterMapM fun f => return (f, ← mkProjection e f)
```
and maybe don’t name it with `!`?

nomeata:
Is there no worry that eagerly adding all projections is going to blow up in some cases, with large or deeply nested structures? Or is this only going to be used in places where the user expects to wait a bit longer (interactive `exact?` maybe)? Or is this simply a normal approach?

Std/Tactic/LetProjs.lean
semorrison: I've just inlined it.

semorrison:
> Is there no worry that eagerly adding all projections is going to blow up in some cases, with large or deeply nested structures? Or is this only going to be used in places where the user expects to wait a bit longer (interactive `exact?` maybe)? Or is this simply a normal approach?

@joehendrix's suggestion is that when other tactics are using this, if it is a user-called tactic it's okay to do something potentially expensive like this, but tactics called by other tactics should not.

PR #549, opened 2024-01-19
feat: allow solve_by_elim to use projections of hypotheses

semorrison:
This powers up `solve_by_elim` by allowing it to use projections of hypotheses. It does so by saturating the context with let-bindings for projections (recursively).

I'm proposing that we turn this on by default, but there is a flag `solve_by_elim (config := {letProjs := false})` to disable it.

As a simple example, this allows
```
example (x : P ∧ Q) : P := by
  solve_by_elim
```
which  previously would fail.

It also powers up `exact?` (example from Mathlib):
```
example (p : ℕ) [Fact p.Prime] : 1 < p := by exact? says exact Nat.Prime.one_lt Fact.out
```

- [ ] depends on #547
- [ ] depends on #548

### DIFF
--- a/Std.lean
+++ b/Std.lean
@@ -101,6 +101,7 @@ import Std.Tactic.HaveI
 import Std.Tactic.Instances
 import Std.Tactic.LabelAttr
 import Std.Tactic.LeftRight
+import Std.Tactic.LetProjs
 import Std.Tactic.LibrarySearch
 import Std.Tactic.Lint
 import Std.Tactic.Lint.Basic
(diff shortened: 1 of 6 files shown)


semorrison:
I will check this against Mathlib.

semorrison:
`apply_rules` should probably turn this off, as it is non-terminal.

semorrison:
> `apply_rules` should probably turn this off, as it is non-terminal.

Done.

semorrison:
> I will check this against Mathlib.

Mathlib reveals a nasty problem, which is that somehow we are now generating proofs with additional, unused, universe variables in them!

semorrison:
Checking what this does to Mathlib in https://github.com/leanprover-community/mathlib4/pull/10266.

PR #553, opened 2024-01-21
rfc: drop non-specific keys in `exact?`

semorrison:
This is a discrepancy between the implementations of `std_exact?` and `exact?`: Mathlib's version does not index lemmas whose `DiscrTree.Key` would be `[*]` or `[Eq, *, *, *]`.

Typically the results indexed by these keys are way off-topic, and it is not helpful to report them to the user in `apply?`. Of course in `exact?`, where we report at most one result, if there's time available we might as well try them.

This PR is not complete as is: it just restores the Mathlib behaviour in the non-lazy branch of the code. Some combination of Joe and I would also have to make the corresponding changes in the lazy branch.

Perhaps the best solution, however, is that we index all such lemmas, and instead filter them out at the end:
* `exact?` would just use them
* `apply?` would only report them if it managed to use at least one local hypothesis?
* maybe a configuration option, or `apply?!`, to have it report them all?


semorrison:
This has been done upstream now.

PR #556, opened 2024-01-23
feat: OrderedAssocList

kim-em:
Ordered wrapper around `AssocList`, and basic functions `find?`, `insert?`, `filterMapVal`, and `merge`. An extensionality lemma `(∀ a, l₁.find? a = l₂.find? a) → l₁ = l₂`.

I will later use this to implement sparse coefficients in `omega`.

- [ ] depends on #1953


joehendrix:
I'm assuming this is useful for kernel execution efficiency.  Is that the case?  Could you test performance against RBMap?

Std/Data/OrderedAssocList.lean
```diff
@@ -871,11 +874,13 @@ Add two `OrderedAssocList`s, taking the value from one list if the other value i
 def add [Add β] (l₁ : OrderedAssocList cmp β) (l₂ : OrderedAssocList cmp β) :
     OrderedAssocList cmp β :=
   merge (fun _ => addOption) l₁ l₂
-where addOption : Option β → Option β → Option β
-  | some x, some y => some (x + y)
-  | some x, none => some x
-  | none, some y => some y
-  | none, none => none
+where
+    /-- Add two values, treating missing values as `0`. -/
```
digama0, line 878:
```suggestion
  /-- Add two values, treating missing values as `0`. -/
```

semorrison:
> I'm assuming this is useful for kernel execution efficiency. Is that the case? Could you test performance against RBMap?

No, actually, there's no need to run any of this in the kernel. I want this because I think it will be easier to write the theorems about algebraic operations on finitely supported functions `Nat \to Int` that I need in this representation than in `RBMap`.

digama0:
In that case I'm a bit dubious of adding a new data structure like this to std, since it has neither good kernel performance nor good runtime performance.

semorrison:
Hmm, okay. I will see how proofs go on top of `RBMap` + a wrapper asserting there are no zero values.

digama0:
> Hmm, okay. I will see how proofs go on top of `RBMap` + a wrapper asserting there are no zero values.

What are you trying to do more specifically? Why do you need proofs in this case? If this is an `omega` data structure I would expect proofs to not be necessary.

semorrison:
> > Hmm, okay. I will see how proofs go on top of `RBMap` + a wrapper asserting there are no zero values.
> 
> What are you trying to do more specifically? Why do you need proofs in this case? If this is an `omega` data structure I would expect proofs to not be necessary.

This is very slightly inaccurate (because of subsequent laziness taking some shortcuts), but the functions and theorems I need on the representation of coefficients are given in https://github.com/leanprover/std4/blob/main/Std/Tactic/Omega/Coeffs/IntList.lean

Currently all these functions are implemented on `List Int`, i.e. a dense representation of coefficients, only trimming trailing zeros as needed to make the theorems true!

As you can see the intention is that I should be able to swap out any representation satisfying that API.

This is not particularly urgent, as right now now one is actually calling `omega` with more than ~10 variables... But Leo assures me that that someone will want to eventually.

digama0:
So why not just use `AssocList` there? Or `Array (Nat x Int)` for that matter.

digama0:
> > So why not just use `AssocList` there? Or `Array (Nat x Int)` for that matter.
> 
> Because then the merge operation (which is actually run, not just proved about) is `O(n1 * n2)`, and I wanted to avoid that.

I mean, you can still implement a `O(n1 + n2)` merge operation on `AssocList` which works provided the inputs are sorted.

semorrison:
> > > So why not just use `AssocList` there? Or `Array (Nat x Int)` for that matter.
> > 
> > 
> > Because then the merge operation (which is actually run, not just proved about) is `O(n1 * n2)`, and I wanted to avoid that.
> 
> I mean, you can still implement a `O(n1 + n2)` merge operation on `AssocList` which works provided the inputs are sorted. That is, you are working with ordered `AssocList`s, just not as a distinct type.

That's what this PR does! `AssocList.orderedMerge`. It seems really cumbersome to then state all the theorems about it (e.g. how it relates to `find?`, associativity, etc) if you forbid yourself from mentioning the bundled type that wraps up the witness of ordered-ness.

digama0:
Is it? I would expect it to just be a simple predicate, e.g.
```
theorem orderedMerge_ordered : Ordered l1 -> Ordered l2 -> Ordered (l1.orderedMerge l2) := ...
```

semorrison:
> Is it?

I've started refactoring, and agree this is better. Thanks for talking me around. :-)

Std/Data/AssocList.lean
```diff
@@ -3,7 +3,7 @@ Copyright (c) 2019 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura, Mario Carneiro
 -/
-import Std.Data.List.Basic
+import Std.Data.List.Lemmas
```
fgdorais, line 6: I'm not a fan of breaking the Basic/Lemmas split this way. This suggests that it's time to split `AssocList` into Basic/Lemmas.

fgdorais:
I think all my concerns have been addressed. Just one additional remark: the design of `AssocList` so far is that defs are specced using `toList`, correctness and other lemmas are then proved using the `toList` translation and the `List` library. I think this makes good sense and should be considered in the revisions.

semorrison:
> the design of `AssocList` so far is that defs are specced using `toList`, correctness and other lemmas are then proved using the `toList` translation and the `List` library.

This is still WIP, I will probably eventually remove the bundled `OrderedAssocList` entirely when I get back to this PR.

However, I am not intending to use `toList` for specification: the great property of an ordered assoc list is that `find?` suffices for specifications, and this is intentionally different from what we have to do for AssocList.

kim-em:
I've cleanup this up. I decided to keep the bundled structure because the whole point is the nice extensionality lemma available there.

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1018 @@
...
+  | .nil => none
+  | .cons a _ _ => some a
+
+@[simp] theorem headKey?_nil : headKey? (.nil : AssocList α β) = none := rfl
+@[simp] theorem headKey?_cons : headKey? (.cons a b t) = some a := rfl
+
+/--
+The condition that an element is less than the first key of an `AssocList`, or that list is empty.
+-/
+abbrev ltHeadKey? (cmp : α → α → Ordering) (a : α) (t : AssocList α β) : Prop :=
```
fgdorais, line 61: Makes more sense as `Bool` valued. Something about the `?` makes me unhappy here but it's hard to think of a more accurate alternative.

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1018 @@
...
+/--
+The predicate that the keys of an `AssocList` are
+in strictly increasing order according to the comparator `cmp`.
+-/
+def keysOrdered (cmp : α → α → Ordering) : AssocList α β → Prop
+  | .nil => True
+  | .cons _ _ .nil => True
+  | .cons a _ (.cons x y t) => cmp a x = .lt ∧ keysOrdered cmp (.cons x y t)
+
+instance instKeysOrderedDecidablePred : DecidablePred (keysOrdered cmp : AssocList α β → Prop) := by
```
fgdorais, line 77: Won't be necessary if `keysOrdered` returns `Bool`.

fgdorais (commented):
I only made it part way but I don't know when I will come back to it :/

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1018 @@
...
+
+@[simp] theorem ltHeadKey?_nil {cmp : α → α → Ordering} :
+    ltHeadKey? cmp a (.nil : AssocList α β) = True := rfl
+@[simp] theorem ltHeadKey?_cons : ltHeadKey? cmp a (.cons x y t) = (cmp a x = .lt) := rfl
+
+/--
+The predicate that the keys of an `AssocList` are
+in strictly increasing order according to the comparator `cmp`.
+-/
+def keysOrdered (cmp : α → α → Ordering) : AssocList α β → Prop
```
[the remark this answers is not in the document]
kim-em: I'm skeptical. What is the pay-off? It's kind of annoying to adapt to this.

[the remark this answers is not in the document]
fgdorais: True, it's a major refactor. My concern is that the decision procedure for the current version is not necessarily as efficient. Can we have both? Also, `keysOrdered` should be `KeysOrdered` if it's a proposition.

Have you considered using `ltHeadKey?` here to save some cases?
```lean
def KeysOrdered (cmp : α → α → Ordering) : AssocList α β → Bool
  | .nil => True
  | .cons a _ t => ltHeadKey? cmp a t ∧ KeysOrdered cmp t
```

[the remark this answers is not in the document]
kim-em: `keysOrdered` is not meant to ever "run". It's not a decision procedure.

Batteries/Data/OrderedAssocList.lean
kim-em: Again, I disagree. This is an invariant of the data structures, we will *never* "run" this on an input.

Batteries/Data/OrderedAssocList.lean
kim-em: I've removed it, as it isn't necessary at all.

Batteries/Data/OrderedAssocList.lean
fgdorais: So it really is a `Prop`: rename to `LtHeadKey`.

kim-em: Already renamed to `LHHeadKey?`.

fgdorais: Sorry! I wasn't clear. I meant to remove the question mark. Consider `IsLtHeadKey` if you prefer.

kim-em: But the `?` is part of the name being referred to.

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+/--
+The predicate that the keys of an `AssocList` are
+in strictly increasing order according to the comparator `cmp`.
+-/
+def KeysOrdered (cmp : α → α → Ordering) : AssocList α β → Prop
+  | .nil => True
+  | .cons _ _ .nil => True
+  | .cons a _ (.cons x y t) => cmp a x = .lt ∧ KeysOrdered cmp (.cons x y t)
+
+@[simp] theorem KeysOrdered_cons_nil : KeysOrdered cmp (.cons a b nil) := trivial
```
fgdorais:
```suggestion
@[simp] theorem keysOrdered_cons_nil : KeysOrdered cmp (.cons a b nil) := trivial
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+  dsimp [LTHeadKey?]
+  split
+  · simp
+  · exact TransCmp.lt_of_le_of_lt h
+
+/--
+A relation on two `AssocList`s, asserting that the head key of the first list is at most
+the head key of the second list, or that the second list is empty.
+-/
+abbrev headKey?_le_headKey?
```
fgdorais: It's a Prop so the name should be uppercase CamelCase. The literal suggestion would be `HeadKey?LEHeadKey?` but that is ugly, how about the simple:
```suggestion
abbrev LEHead
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+  revert p w
+  match headKey? s, headKey? t with
+  | some a, some b =>
+    intro p w
+    simp [TransCmp.lt_of_lt_of_le w p]
+  | some a, none => intros; trivial
+  | none, some b => simp
+  | none, none => intros; trivial
+
+theorem KeysOrdered_cons {cmp : α → α → Ordering}
```
fgdorais:
```suggestion
theorem keysOrdered_cons {cmp : α → α → Ordering}
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+    (w : LTHeadKey? cmp a t) (h : KeysOrdered cmp t) :
+    KeysOrdered cmp (.cons a b t) := by
+  match t with
+  | .nil => trivial
+  | .cons x y s => exact ⟨w, h⟩
+
+theorem find?_eq_none_of_LTHeadKey? {cmp : α → α → Ordering} [TransCmp cmp] [BEq α] [LawfulBEq α]
```
fgdorais:
```suggestion
theorem find?_eq_none_of_ltHeadKey? {cmp : α → α → Ordering} [TransCmp cmp] [BEq α] [LawfulBEq α]
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+theorem headKey?_orderedInsert_or (cmp) (l : AssocList α β) (a) (b) :
+    headKey? (orderedInsert cmp l a b) = some a ∨
+      headKey? (orderedInsert cmp l a b) = headKey? l := by
+  rw [headKey?_orderedInsert]
+  match l with
+  | .nil => left; rfl
+  | .cons x y s => dsimp; cases cmp a x <;> simp
+
+mutual
+theorem orderedInsert_KeysOrdered_aux [LawfulEqCmp cmp] [OrientedCmp cmp] {x y t}
```
fgdorais:
```suggestion
theorem orderedInsert_keysOrdered_aux [LawfulEqCmp cmp] [OrientedCmp cmp] {x y t}
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+        · dsimp [LTHeadKey?]
+          rcases headKey?_orderedInsert_or cmp t a b with (p|p)
+          · rw [p]
+            exact OrientedCmp.lt_of_gt w
+          · rw [p]
+            exact ltHeadKey?_of_keysOrdered_cons h
+        · apply orderedInsert_KeysOrdered
+          exact h.tail
+
+theorem orderedInsert_KeysOrdered [LawfulEqCmp cmp] [OrientedCmp cmp] (h : KeysOrdered cmp l) :
```
fgdorais:
```suggestion
theorem keysOrdered_orderedInsert [LawfulEqCmp cmp] [OrientedCmp cmp] (h : KeysOrdered cmp l) :
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+  match l with
+  | .nil => simp [headKey?_le_headKey?]
+  | .cons x y t =>
+    simp [filterMapVal]
+    match f x y with
+    | none =>
+      exact headKey?_le_headKey?_cons h (headKey?_le_headKey?_filterMapVal h.tail)
+    | some _ => simp [ReflCmp.compare_self]
+
+theorem filterMapVal_KeysOrdered [TransCmp cmp] (h : KeysOrdered cmp l) :
```
fgdorais:
```suggestion
theorem keysOrdered_filterMapVal [TransCmp cmp] (h : KeysOrdered cmp l) :
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+      · exact ltHeadKey?_orderedMerge (ltHeadKey?_of_cons h₁ w₁) (ltHeadKey?_of_cons h₂ w₂)
+          w₁.tail w₂.tail
+    | .gt =>
+      dsimp
+      split
+      · exact h₂
+      · exact ltHeadKey?_orderedMerge h₁ (ltHeadKey?_of_cons h₂ w₂) w₁ w₂.tail
+
+unseal orderedMerge in
+theorem orderedMerge_KeysOrdered [LawfulEqCmp cmp] [TransCmp cmp]
```
fgdorais:
```suggestion
theorem keysOrdered_orderedMerge [LawfulEqCmp cmp] [TransCmp cmp]
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+
+/--
+An `OrderedAssocList` is an `AssocList` with the additional invariant that
+the keys are in strictly increasing order according to some specified comparator function.
+-/
+structure OrderedAssocList {α : Type u} (cmp : α → α → Ordering) (β : Type v) where
+  /-- The underlying `AssocList` of an `OrderedAssocList`. -/
+  list : AssocList α β
+  /-- The invariant that the keys are in strictly increasing order according to `cmp`. -/
+  KeysOrdered : list.KeysOrdered cmp
```
fgdorais:
```suggestion
  keysOrdered : list.KeysOrdered cmp
```

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+
+end AssocList
+
+/--
+An `OrderedAssocList` is an `AssocList` with the additional invariant that
+the keys are in strictly increasing order according to some specified comparator function.
+-/
+structure OrderedAssocList {α : Type u} (cmp : α → α → Ordering) (β : Type v) where
+  /-- The underlying `AssocList` of an `OrderedAssocList`. -/
+  list : AssocList α β
```
fgdorais:
```suggestion
  toAssocList : AssocList α β
```
Or extend `AssocList`?

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1025 @@
...
+
+/-- The head key of a tail is either `none`, or greater than the original head key. -/
+theorem ltheadKey?_tail (h : AssocList.KeysOrdered cmp (.cons a b t)) :
+    LTHeadKey? a ⟨t, h.tail⟩ := by
+  dsimp [LTHeadKey?]
+  match t with
+  | .nil => trivial
+  | .cons _ _ _ => exact h.1
+
+theorem getElem?_eq_none_of_LTHeadKey? (l : OrderedAssocList cmp β) (w : LTHeadKey? x l) :
```
fgdorais:
```suggestion
theorem getElem?_eq_none_of_ltHeadKey? (l : OrderedAssocList cmp β) (w : LTHeadKey? x l) :
```

fgdorais (commented):
Naming conventions. Consider dot notation in some places, if only to avoid awkward lowercasing of type names.

Batteries/Data/OrderedAssocList.lean
kim-em: Changed my mind, you're right.

Std/Data/AssocList.lean
kim-em: Done.

kim-em:
> I think all my concerns have been addressed. Just one additional remark: the design of `AssocList` so far is that defs are specced using `toList`, correctness and other lemmas are then proved using the `toList` translation and the `List` library. I think this makes good sense and should be considered in the revisions.

Mostly done, to the extend the corresponding operations exist on `List`.

Batteries/Data/OrderedAssocList.lean
kim-em: I went with LEHeadKey

Batteries/Data/OrderedAssocList.lean
kim-em: I went with `keysOrdered_orderedInsert_aux`.

Batteries/Data/OrderedAssocList.lean
kim-em: `AssocList` is an inductive, so we can't.

Batteries/Data/OrderedAssocList.lean
```diff
@@ -0,0 +1,1099 @@
...
+/-- The head key of an `AssocList`, or `none` if the list is empty. -/
+def headKey? (l : AssocList α β) : Option α :=
+  match l with
+  | .nil => none
+  | .cons a _ _ => some a
+
+@[simp] theorem headKey?_nil : headKey? (.nil : AssocList α β) = none := rfl
+@[simp] theorem headKey?_cons : headKey? (.cons a b t) = some a := rfl
+
+theorem headKey?_eq_head? (l : AssocList α β) : l.headKey? = l.toList.head?.map Prod.fst := by
```
fgdorais:
```suggestion
theorem headKey?_eq_map_fst_head?_toList (l : AssocList α β) : l.headKey? = l.toList.head?.map Prod.fst := by
```

PR #558, opened 2024-01-24
feat: lemmas for BitVec

semorrison:
A user was having trouble with:

```
import Std

open Std BitVec

example (v : Std.BitVec n)  : 0#0 ++ v = BitVec.cast (by omega) v := sorry
```

With these changes:
* add `@[ext]` to `BitVec.eq_of_getLsb_eq`
* add `@[simp]` to a few existing lemmas
* prove additional theorems about `getLsb` applied to other operations

the proof becomes 
```
example (v : Std.BitVec n)  : 0#0 ++ v = BitVec.cast (by omega) v := by ext; simp
```

### DIFF
--- a/Std/Data/BitVec/Basic.lean
+++ b/Std/Data/BitVec/Basic.lean
@@ -334,7 +334,13 @@ As a numeric operation, this is equivalent to `a / 2^s`, rounding down.
 
 SMT-Lib name: `bvlshr` except this operator uses a `Nat` shift value.
 -/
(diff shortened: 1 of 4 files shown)


joehendrix:
It'd be good to have the example in tests.  Maybe in `test/bitvec.lean`?  It'd be good to test the expressive power and normal forms of the simp rules.

joehendrix:
@semorrison I've added a bunch of lemmas too.  I plan to merge this soon unless you find other changes you'd like.

PR #563, opened 2024-01-25
feat: proofs of bitblasted not and bitblasted negation

mhk119:


Std/Data/BitVec/Lemmas.lean
```diff
@@ -155,6 +157,41 @@ theorem truncate_succ (x : BitVec w) :
...
+theorem sub_cast {x y : BitVec w} : (x - y).toNat = (x.toNat + (2^w - y.toNat)) % 2^w := rfl
+
+@[simp]
+theorem neg_cast {x : BitVec w} : (-x).toNat = (2^w- x.toNat) % 2^w := by
+  rw [← Nat.zero_add (2^w - x.toNat)]; rfl
+
+@[simp]
+theorem xor_cast {x y: BitVec w} : (x ^^^ y).toNat = x.toNat ^^^ y.toNat:= rfl
+
+/-! ### ring properties -/
```
fgdorais, line 174: Let's try to be a bit less mathy in Std. Besides, since there are no mul lemmas (yet), "ring" is a bit too ambitious for now. Split these lemmas in sections for `add`, `sub`, etc.

Std/Data/BitVec/Lemmas.lean
```diff
@@ -155,6 +157,41 @@ theorem truncate_succ (x : BitVec w) :
...
+
+theorem eq_sub_of_add_eq {a b c : BitVec w} (h : a + b = c) : a = c - b := by
```
fgdorais, line 176:
```suggestion
theorem eq_sub_of_add_eq (a b c : BitVec w) (h : a + b = c) : a = c - b := by
```
Also move to sub section.

Std/Data/BitVec/Lemmas.lean
```diff
@@ -155,6 +157,41 @@ theorem truncate_succ (x : BitVec w) :
...
+  apply BitVec.eq_of_toNat_eq; simp [Nat.add_comm]
+
+theorem add_sub_assoc {x y z : BitVec w} : x + y - z = x + (y - z) := by
+  apply BitVec.eq_of_toNat_eq; simp [Nat.add_sub_assoc, Nat.add_assoc]
+
+theorem add_right_neg {x : BitVec w} : x + -x = 0 := by
+  apply BitVec.eq_of_toNat_eq
+  simp [← Nat.add_sub_assoc (Nat.le_of_lt x.isLt), Nat.add_sub_self_left]
+
+theorem zero_sub {x : BitVec w} : 0#w - x = -x := rfl
```
fgdorais, line 193:
```suggestion
theorem zero_sub (x : BitVec w) : 0#w - x = -x := rfl
```
Also move to sub section.

Std/Data/BitVec/Lemmas.lean
```diff
@@ -155,6 +157,41 @@ theorem truncate_succ (x : BitVec w) :
     have j_lt : j.val < i := Nat.lt_of_le_of_ne (Nat.le_of_succ_le_succ j.isLt) j_eq
     simp [j_eq, j_lt]
 
+/-! ### cast -/
+
+@[simp]
+theorem add_cast {x y : BitVec w} :  (x + y).toNat = (x.toNat + y.toNat) % 2^w := rfl
```
fgdorais, line 163:
```suggestion
@[simp]
theorem toNat_add (x y : BitVec w) :  (x + y).toNat = (x.toNat + y.toNat) % 2^w := rfl
```

fgdorais: Same for the remaining `_cast` theorems.

Std/Data/BitVec/Bitblast.lean
```diff
@@ -157,3 +158,58 @@ theorem adc_correct (x y : BitVec w) (c : Bool) :
...
+
+/-- Bitwise 2's complement implemented via `iunfoldr`. -/
+def bit_neg (x : BitVec w) : BitVec w := (adc (bit_not x) (BitVec.ofNat w 1) false).snd
+
+-- @[simp]
+-- theorem not_cast {x : BitVec w} : (~~~x).toNat = (2^w - 1 -x.toNat)%(2^w) := by
+--   simp only [Complement.complement, BitVec.not]
+--   simp only [xor_cast, toNat_ofFin]
+--   rw [Nat.shiftLeft_eq, Nat.one_mul]
+--   sorry
```
fgdorais, line 176: Delete.

Std/Data/BitVec/Bitblast.lean
```diff
@@ -157,3 +158,58 @@ theorem adc_correct (x y : BitVec w) (c : Bool) :
...
   simp [adc_correct]
+
+/-! ### Negation -/
+
+/-- Bitwise 1's complement implemented via `iunfoldr`. -/
+def bit_not (x : BitVec w) : BitVec w :=
+  ((iunfoldr fun (i : Fin w) c => (c, !(x.getLsb i))) ()).snd
+
+/-- Bitwise 2's complement implemented via `iunfoldr`. -/
+def bit_neg (x : BitVec w) : BitVec w := (adc (bit_not x) (BitVec.ofNat w 1) false).snd
```
fgdorais, line 169: What is the motivation for these definitions? Is the implementation via `iunfoldr` more efficient? I don't understand why there is a need for two implementations of the same function.

Std/Data/BitVec/Bitblast.lean
```diff
@@ -157,3 +158,58 @@ theorem adc_correct (x y : BitVec w) (c : Bool) :
...
+--   rw [Nat.shiftLeft_eq, Nat.one_mul]
+--   sorry
+
+theorem bit_not_testBit (x : BitVec w) (i : Fin w) :
+  getLsb (bit_not x) i.val = !(getLsb x i.val) := by
+  simp only [bit_not]
+  apply iunfoldr_getLsb (fun _ => ()) i (by simp)
+
+
+theorem bit_not_eq_not' (x : BitVec w) : bit_not x + x  = -1:= by
```
fgdorais, line 184:
```suggestion
theorem bit_not_add_self (x : BitVec w) : bit_not x + x  = -1 := by
```

fgdorais (changes requested):

Std/Data/BitVec/Lemmas.lean
fgdorais: This suggestion was a mistake, the original was fine. Sorry. I was probably on auto-pilot :/ This should still be in a section for `sub`.

Std/Data/BitVec/Bitblast.lean
mhk119: The idea is to implement a bit-blasted version of each `BitVec` operation and show it is equivalent to the original definition. The advantage of this is that SMT solvers use bitblasting to convert such operations into a boolean formula which can be sent to a SAT solver. Our aim is to develop a tactic that integrates SMT solvers with Lean in this project: [lean-smt](https://github.com/ufmg-smite/lean-smt).

fgdorais: I understand. It's unusual to have defs that aren't meant to be run. Would it be sufficient to have a theorem that says `-x =  (adc (bit_not x) (BitVec.ofNat w 1) false).snd`? It's fine if not, so long as it's clear that these are alternate implementations not meant to be used in code.

semorrison:
This PR will need to be migrated to a PR to the Lean repo, as the BitVec material has moved in the meantime.

PR #564, opened 2024-01-25
chore: remove spurious bool lemmas; add dual lemmas for List.any/all

joehendrix:
This adds a few lemmas to better simplify `t = false` expressions and removes a couple lemmas almost immediately following core lemmas.

### DIFF
--- a/Std/Data/Bool.lean
+++ b/Std/Data/Bool.lean
@@ -42,9 +42,13 @@ theorem false_ne_true : false ≠ true := Bool.noConfusion
 
 theorem eq_false_or_eq_true : (b : Bool) → b = true ∨ b = false := by decide
 
-theorem eq_false_iff : {b : Bool} → b = false ↔ b ≠ true := by decide
+-- N.B. The replacement lemma is symmetric to the original so it is
+-- not a direct replacement.
+@[deprecated Bool.not_eq_true]
+theorem eq_false_iff {b : Bool} : b = false ↔ b ≠ true := Iff.of_eq (Bool.not_eq_true b).symm
 
-theorem ne_false_iff : {b : Bool} → b ≠ false ↔ b = true := by decide
+@[deprecated Bool.not_eq_false]
+theorem ne_false_iff {b : Bool} : b ≠ false ↔ b = true := Iff.of_eq (Bool.not_eq_false b)
 
 /-! ### and -/
 
@@ -72,8 +76,13 @@ theorem and_xor_distrib_right : ∀ (x y z : Bool), (xor x y && z) = xor (x && z
 /-- De Morgan's law for boolean and -/
 theorem not_and : ∀ (x y : Bool), (!(x && y)) = (!x || !y) := by decide
 
+@[deprecated and_eq_true]
 theorem and_eq_true_iff : ∀ (x y : Bool), (x && y) = true ↔ x = true ∧ y = true := by decide
 
+@[simp] theorem and_eq_false (a b : Bool) : ((a && b) = false) = (a = false ∨ b = false) := by
+  cases a <;> cases b <;> decide
+
+@[deprecated and_eq_false]
 theorem and_eq_false_iff : ∀ (x y : Bool), (x && y) = false ↔ x = false ∨ y = false := by decide
 
 /-! ### or -/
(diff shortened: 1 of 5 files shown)


Std/Data/Bool.lean
```diff
@@ -72,8 +76,13 @@ theorem and_xor_distrib_right : ∀ (x y z : Bool), (xor x y && z) = xor (x && z
 /-- De Morgan's law for boolean and -/
 theorem not_and : ∀ (x y : Bool), (!(x && y)) = (!x || !y) := by decide
 
+@[deprecated and_eq_true]
 theorem and_eq_true_iff : ∀ (x y : Bool), (x && y) = true ↔ x = true ∧ y = true := by decide
```
fgdorais, line 80: Here and elsewhere, the `Iff` versions are useful since they gives access to the two implications via `.mp` and `.mpr`. This doesn't work with the `Eq` version. 

That said, they would be more useful if the variables were marked implicit, as they would with the two implications.

Std/Data/Bool.lean
```diff
@@ -42,9 +42,13 @@ theorem false_ne_true : false ≠ true := Bool.noConfusion
 
 theorem eq_false_or_eq_true : (b : Bool) → b = true ∨ b = false := by decide
 
-theorem eq_false_iff : {b : Bool} → b = false ↔ b ≠ true := by decide
+-- N.B. The replacement lemma is symmetric to the original so it is
+-- not a direct replacement.
+@[deprecated Bool.not_eq_true]
+theorem eq_false_iff {b : Bool} : b = false ↔ b ≠ true := Iff.of_eq (Bool.not_eq_true b).symm
```
fgdorais, line 48: The direction here is important because the coercion from `Bool` to `Prop` is `_ = true`. So this direction is the useful one to convert a `Bool` expression to an equivalent `Prop`.

Std/Data/Bool.lean
```diff
@@ -97,8 +106,13 @@ theorem or_and_distrib_right : ∀ (x y z : Bool), ((x && y) || z) = ((x || z) &
 /-- De Morgan's law for boolean or -/
 theorem not_or : ∀ (x y : Bool), (!(x || y)) = (!x && !y) := by decide
 
+@[simp] theorem or_eq_false (a b : Bool) : ((a || b) = false) = (a = false ∧ b = false) := by
+  cases a <;> cases b <;> decide
```
fgdorais, line 110: Match surrounding style.
```suggestion
@[simp] theorem or_eq_false : ∀ (x y : Bool). ((x || y) = false) = (x = false ∧ y = false) := by decide
```

Std/Data/Bool.lean
```diff
@@ -72,8 +76,13 @@ theorem and_xor_distrib_right : ∀ (x y z : Bool), (xor x y && z) = xor (x && z
 /-- De Morgan's law for boolean and -/
 theorem not_and : ∀ (x y : Bool), (!(x && y)) = (!x || !y) := by decide
 
+@[deprecated and_eq_true]
 theorem and_eq_true_iff : ∀ (x y : Bool), (x && y) = true ↔ x = true ∧ y = true := by decide
 
+@[simp] theorem and_eq_false (a b : Bool) : ((a && b) = false) = (a = false ∨ b = false) := by
+  cases a <;> cases b <;> decide
```
fgdorais, line 83: Math surrounding style.
```suggestion
@[simp] theorem and_eq_false : ∀ (x y : Bool) : ((x && y) = false) = (x = false ∨ y = false) := by decide
```

fgdorais (changes requested):

Std/Data/Bool.lean
joehendrix: That helps me understand the conventions here.  Thanks

joehendrix: This is on pause for a bit while I work through a comprehensive look at `Bool`/`Prop`, but I noticed that `Eq` also give access to the implications via `Eq.mp` and `Eq.mpr`.

I think I want to systematically use one (probably `Eq`), but that's not settled just yet.

joehendrix:
Obsolete after upstreaming Bool and Nat lemmas

PR #566, opened 2024-01-27
fix: use getTransparency in librarySearch SolveByElim.Config

dwrensha:


dwrensha:
[zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60exact.3F.60.20failure.3A.20.22maximum.20recursion.20depth.20has.20been.20reached.22/near/417649319)

PR #568, opened 2024-01-29
chore: remove Std.Data.List import from Omega

joehendrix:
This adds lemmas to Std.Data.List from Omega dependencies so that Omega can be used in List lemmas.

I changed a few proofs as part of this to reduce dependencies and include fewer lemmas in Init.

### DIFF
--- a/Std/Data/List/Init/Basic.lean
+++ b/Std/Data/List/Init/Basic.lean
@@ -10,7 +10,7 @@ namespace List
 Version of `List.zipWith` that continues to the end of both lists, passing `none` to one argument
 once the shorter list has run out.
 -/
--- TODO We should add a tail-recursive version as we do for other `zip` functions above.
+-- TODO We should add a tail-recursive version as we do for other `zip` functions.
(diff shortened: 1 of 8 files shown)


Std/Data/List/Init/Basic.lean
```diff
@@ -0,0 +1,28 @@
...
+Authors: Leonardo de Moura
+-/
+
+namespace List
+
+/--
+Version of `List.zipWith` that continues to the end of both lists, passing `none` to one argument
+once the shorter list has run out.
+-/
+-- TODO We should add a tail-recursive version as we do for other `zip` functions above.
```
digama0, line 13: they aren't above anymore

semorrison:
```suggestion
-- TODO We should add a tail-recursive version as we do for other `zip` functions.
```

semorrison (commented):
Looks good to me.

There are lots of theorems in `Std.Tactic.Omega.MinNatAbs` that are not actually used (they are leftovers from the verified version), but I don't think them removing them actually reduces how much we need to move to `Init`, in any case.

PR #569, opened 2024-01-30
feat: `List.insertP` and lemmas

fgdorais:
`List.insertP` is a generalized, tail-recursive version of the insert step in insertion-sort. Sample use:
```lean
def insertionSort [LT α] [DecidableRel (α:=α) (· < ·)] (l : List α) := 
  l.foldl (fun s a => s.insertP (a < ·) a) [] -- fully tail recursive and stable!
```
It will be useful to streamline and optimize `ac_rfl` once moved to Std.

PS: Perhaps someone has a better name?


Std/Data/List/Basic.lean
```diff
@@ -1676,4 +1676,14 @@ See `isSubperm_iff` for a characterization in terms of `List.Subperm`.
 -/
 def isSubperm [BEq α] (l₁ l₂ : List α) : Bool := ∀ x ∈ l₁, count x l₁ ≤ count x l₂
 
-end List
+/--
+`O(|l|)`. Inserts `a` in `l` right before the first element such that `p` fails, or at the end of
+the list if `p` never fails on `l`.
```
digama0, line 1681: Before reading the docs, my initial guess was that the polarity of `p` would be flipped, and it inserts `a` before the first element for which `p` is true. This seems more consistent with `findP` et al.

joehendrix (changes requested):
I don't have a specific objection to `insertP`, but we are working on triaging what parts of Std can and should be moved to Lean core (with further polishing) to provide a good new user experience.

As such I'd like to hold off introducing new operations unless they enhance existing capabilities or are needed for high-priority tactics.

fgdorais:
The `ac_rfl` move has been canceled so there is no immediate need for this. It's still a useful addition so I'm reluctant to close but closing might be better than rotting.

Batteries/Data/List/Lemmas.lean
```diff
@@ -562,10 +562,36 @@ theorem indexOf_mem_indexesOf [BEq α] [LawfulBEq α] {xs : List α} (m : x ∈
...
+    insertP p a (b :: l) = b :: insertP p a l := by
+  simp only [insertP, insertP.loop, cond, h]; exact insertP_loop ..
+
+@[simp] theorem length_insertP (p : α → Bool) (a l) : (insertP p a l).length = l.length + 1 := by
+  induction l with simp [insertP, insertP.loop, cond]
+  | cons _ _ ih => split <;> simp [insertP_loop, ih]
+
+@[simp] theorem mem_insertP (p : α → Bool) (a l) : a ∈ insertP p a l := by
+  induction l with simp [insertP, insertP.loop, cond]
+  | cons _ _ ih => split <;> simp [insertP_loop, ih]
```
digama0, line 588:
```suggestion
  | cons _ _ ih => split <;> simp [insertP_loop, ih]

```

PR #573, opened 2024-01-31
feat: move `ac_rfl` to Std

fgdorais:
- [ ]  Depends on [lean4#3223](https://github.com/leanprover/lean4/pull/3223)

See issue [lean4#3198](https://github.com/leanprover/lean4/issues/3198) for background.

This is intended to be a "clean move" of `ac_rfl` from Lean core to Std. Changes are very minor: some namespace changes, style changes to pass linting, etc. There are no significant changes in functionality nor in implementation. The intent is to provide continuity where Std's `ac_rfl` should be as effortless as possible for `ac_rfl` users to adopt.

There is room for improvements, some are already planned, but these are best left to future PRs.


semorrison:
The diff will hopefully be more useful if you change the base of the PR to `bump/v4.6.0`.

fgdorais:
Move canceled upstream.

PR #575, opened 2024-02-01
feat: omega handles min, max, if

semorrison:
```
example {a : Nat} (_ : a < min a b) : False := by omega 
example {a : Nat} (_ : max a b < b) : False := by omega 

example {a b : Nat} (_ : a = 7) (_ : b = 3) : min a b = 3 := by omega

example {a b : Nat} (_ : a + b = 9) : (min a b) % 2 + (max a b) % 2 = 1 := by omega

example {a : Int} (_ : a < if a ≤ b then a else b) : False := by omega
example {a b : Int} : (if a < b then a else b - 1) ≤ b := by omega
```

* Upon encountering `min a b`, add the facts `min a b <= a` and `min a b <= b` (similarly for max)
* Upon encountering `natAbs a`, add the facts `a <= natAbs a` and `-a <= natAbs a`
* If `splitMinMax := true`, allow the case split on `min a b = if a <= b then a else b` (similarly for max) as needed
* If `splitNatAbs := true`, allow the case split on `natAbs a = if 0 <= a then a else -a`

By default `splitMinMax := true`, as in interactive mode it may as well try.

When used in automation we will probably just set `disjunctions := false`, which disables all case splits.

### DIFF
--- a/Std/Tactic/Omega/Config.lean
+++ b/Std/Tactic/Omega/Config.lean
@@ -32,6 +32,11 @@ structure OmegaConfig where
   `0 ≤ a ∧ Int.natAbs a = a ∨ a < 0 ∧ Int.natAbs a = - a` for later splitting.
   -/
   splitNatAbs : Bool := true
+  /--
+  Whenever `min a b` or `max a b` is found, rewrite in terms of the definition
+  `if a ≤ b ...`, for later case splitting.
+  -/
+  splitMinMax : Bool := true
 
 /--
 Allow elaboration of `OmegaConfig` arguments to tactics.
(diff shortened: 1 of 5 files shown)


joehendrix:
My inclination would be to add the bounds hypotheses (e.g., `min a b <= a`) automatically, but default not adding disjunctions.   My own usage is typically around array bounds, and so inequalities are good enough for most use cases.

semorrison:
@joehendrix, I've redesigned, to just add the inequalities at first. Splitting the disjunction is behind an option `splitMinMax`, although for interactive use I've still defaulted this to `true`. There's minimal performance impact if they are not needed.

joehendrix (approved):

PR #578, opened 2024-02-04
fix: omega uses local values

semorrison:
Per @joehendrix's [request](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/.60omega.60.20did.20not.20find.20a.20contradiction/near/419179638).

### DIFF
--- a/test/omega/test.lean
+++ b/test/omega/test.lean
@@ -367,3 +367,9 @@ example {a b : Nat} (_ : a + b = 9) : (min a b) % 2 + (max a b) % 2 = 1 := by
 
 example {a : Int} (_ : a < if a ≤ b then a else b) : False := by omega
 example {a b : Int} : (if a < b then a else b - 1) ≤ b := by omega
+
+-- Check that we use local values.
+example (i j : Nat) (p : i ≥ j) : True := by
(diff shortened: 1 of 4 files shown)


test/omega/test.lean
```diff
@@ -367,3 +367,9 @@ example {a b : Nat} (_ : a + b = 9) : (min a b) % 2 + (max a b) % 2 = 1 := by
 
 example {a : Int} (_ : a < if a ≤ b then a else b) : False := by omega
 example {a b : Int} : (if a < b then a else b - 1) ≤ b := by omega
+
+-- Check that we use local values.
+example (i j : Nat) (p : i ≥ j) : True := by
+  let l := j - 1
+  have _ : i ≥ l := by omega
+  trivial
```
kmill, line 375: Could you add another test that shows that it can handle more indirection? Maybe something like
```
example (i j : Nat) (p : i ≥ j) : True := by
  let l := j - 1
  let k := l
  have _ : i ≥ k := by omega
  trivial
```

semorrison: Done!

joehendrix (approved):

PR #579, opened 2024-02-04
feat: `List.merge` and lemmas

fgdorais:
`List.merge` the merge step from merge sort (without any ordering assumptions). This will be useful to streamline and optimize `ac_rfl`.

Note that `List.merge` exists in Mathlib but the Mathlib implementation is not tail recursive and it uses a decidable relation instead of a `Bool` relation. A Mathlib adaptation PR is in progress: https://github.com/leanprover-community/mathlib4/tree/fgdorais-patch-list-merge


semorrison (approved):

fgdorais:
The `ac_rfl` move has been canceled so there is no immediate need for this. It's still a useful addition so I'm reluctant to close but closing might be better than rotting.

PR #581, opened 2024-02-04
feat: simprocs for `BitVec` literals

leodemoura:
Depends on https://github.com/leanprover/lean4/pull/3252

This PR temporarily adds nolint exceptions for the simprocs introduced here. The doc-strings all actually exist. Unfortunately the `simproc` macro rules drop the doc-strings at present. This is fixed in leanprover/lean4#3259, so after we switch to the next release candidate we can just run `runLinter --update` to clear them all out again.


semorrison:
I'll merge conflicts and adjust the toolchain once the nightly is out.

PR #583, opened 2024-02-05
feat: Introduce #check_simp command and usage.

joehendrix:
This does some basic simp tests for arrays as well

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -8,7 +8,6 @@ import Std.Data.Nat.Lemmas
 import Std.Data.List.Lemmas
 import Std.Data.Array.Basic
 import Std.Tactic.HaveI
-import Std.Tactic.Omega
 import Std.Tactic.Simpa
 import Std.Util.ProofWanted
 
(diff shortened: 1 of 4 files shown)


Std/Data/Array/Lemmas.lean
```diff
@@ -131,6 +148,22 @@ theorem get_set (a : Array α) (i : Fin a.size) (j : Nat) (hj : j < a.size) (v :
...
+
+/--
+This lemma simplifies a normal form from `get!`
+-/
+@[simp] theorem getD_get?_setD (a : Array α) (i : Nat) (v d : α) :
+  Option.getD (setD a i v)[i]? d = if i < a.size then v else d := by
+  if h : i < a.size then
+    simp [setD, h, getElem?, get_set]
+  else
+    have p : i ≥ a.size := by omega
```
fgdorais, line 164:
```suggestion
    have p : i ≥ a.size := Nat.le_of_not_gt h
```

Std/Data/Array/Lemmas.lean
```diff
@@ -8,6 +8,7 @@ import Std.Data.Nat.Lemmas
 import Std.Data.List.Lemmas
 import Std.Data.Array.Basic
 import Std.Tactic.HaveI
+import Std.Tactic.Omega
```
fgdorais, line 11: No longer necessary.

fgdorais (approved):

PR #584, opened 2024-02-05
feat: reduce assumptions from `size_swap!`

nomeata:
this is possible after the fix in
https://github.com/leanprover/lean4/pull/3197 and will hopefully help
with array bounds checking.


nomeata:
awaiting-review

joehendrix (approved):

PR #586, opened 2024-02-06
chore: update exact? tests

semorrison:
These tests were not being run in CI (per #585), so had broken.


Ruben-VandeVelde:
Note that 453176a is empty; presumably these changes already landed.

PR #591, opened 2024-02-06
feat: omega using Fin.isLt

semorrison:
```
example (i : Fin 7) : (i : Nat) < 8 := by omega
```

Noticed this was missing in #590.


joehendrix (approved):

semorrison:
(Discussed with Joe, merging.)

PR #594, opened 2024-02-07
chore: make Std.Tactic.Basic an import only

semorrison:
Currently Std.Tactic.Basic serves two purposes:
* to set up some small tactics used everywhere
* to collect the imports of a bunch of medium sized tactics used everywhere

To make is easier for me to understand the import dependencies in Std, I'd like to split these two purposes:

* the small tactics are now in Std.Tactic.Init
* Std.Tactic.Basic is now an import only file


joehendrix (approved):

Std/Tactic/Basic.lean
```diff
@@ -1,165 +1,15 @@
...
-/-- The `conv` tactic `equals` claims that the currently focused subexpression is equal
- to the given expression, and proves this claim using the given tactic.
-```
-example (P : (Nat → Nat) → Prop) : P (fun n => n - n) := by
-  conv in (_ - _) => equals 0 =>
-    -- current goal: ⊢ n - n = 0
-    apply Nat.sub_self
-  -- current goal: P (fun n => 0)
-```
+This is an import only file.
```
digama0, line 14: Some of the style linters use a regex-based identification of import-only files, so I would recommend not putting any module docs in them. Use regular comments if needed.

PR #596, opened 2024-02-07
feat: more BitVec lemmas

semorrison:
We complete some missing API about the not operation on BitVec.

Along the way we get to generalize facts like `testBit (2^n-1) i = decide (i < n)` to `testBit (2^n - (x + 1)) i = (decide (i < n) && ! testBit x i)` for all `x < 2^n`.

Later we should have theorems for `testBit (pred x) i` and indeed addition/subtraction.


Std/Data/Nat/Lemmas.lean
```diff
@@ -1145,6 +1145,21 @@ protected theorem mul_pow (a b n : Nat) : (a * b) ^ n = a ^ n * b ^ n := by
   | zero => rw [Nat.pow_zero, Nat.pow_zero, Nat.pow_zero, Nat.mul_one]
   | succ _ ih => rw [Nat.pow_succ, Nat.pow_succ, Nat.pow_succ, Nat.mul_mul_mul_comm, ih]
 
+protected theorem one_lt_two_pow (h : n ≠ 0) : 1 < 2 ^ n :=
+  match n, h with
+  | n+1, _ => by
+    rw [Nat.pow_succ', ← Nat.one_mul 1]
+    exact Nat.mul_lt_mul_of_lt_of_le' (by decide) (Nat.pow_two_pos n) (by decide)
```
joehendrix, line 1152: How about make this a simp rule?  `1 < 2 ^ n ↔ 0 < n`?  Then `one_lt_two_pow_succ` probably won't beneeded?

Std/Data/BitVec/Lemmas.lean
```diff
@@ -205,11 +237,35 @@ protected theorem extractLsb_ofNat (x n : Nat) (hi lo : Nat) :
...
+    rw [Nat.succ_eq_add_one] at h
+    rw [← h]
+    rw [Nat.testBit_two_pow_sub_succ (toNat_lt _)]
+    · cases w : decide (i < v)
+      · simp at w
+        simp [w]
+        rw [Nat.testBit_lt_two_pow]
+        calc BitVec.toNat x < 2 ^ v := toNat_lt _
+          _ ≤ 2 ^ i := Nat.pow_le_pow_of_le_right Nat.zero_lt_two w
+      · simp
```
joehendrix, line 262: I'm glad you corrected the statement; shows the danger of proof_wanted.

I'd be tempted to write the right hand side as `(~~~x).toNat = (2^v - 1) - x.toNat` just because there are `Nat` theorems abort `2^n - 1` since it's `n` ones in binary.

Std/Data/BitVec/Lemmas.lean
```diff
@@ -35,10 +35,18 @@ theorem testBit_toNat (x : BitVec w) : x.toNat.testBit i = x.getLsb i := rfl
...
   have p : 2^w ≤ 2^i := Nat.pow_le_pow_of_le_right (by omega) ge
   omega
 
+theorem getLsb_eq_decide_and_getLsb {x : BitVec w} {i : Nat} :
+    getLsb x i = (decide (i < w) && getLsb x i) := by
+  if h : i < w then
+    simp [h]
+  else
+    have h' := Nat.ge_of_not_lt h
+    simp [decide_False, h, h']
```
joehendrix, line 48: This is an usual style.  I'd instead expect something like `getLsb x i -> i < w` or the contra-positive.  Could you comment on the choice?

Std/Data/BitVec/Lemmas.lean
semorrison: Changed the statement (but only minimally the proof).

Std/Data/Nat/Lemmas.lean
semorrison: I added the simp lemma, and removed one_lt_two_pow_succ

Std/Data/BitVec/Lemmas.lean
semorrison: Yeah, it is weird. The trick is where I use it in
```
@[simp] theorem getLsb_not {x : BitVec v} : (~~~x).getLsb i = (decide (i < v) && ! x.getLsb i) := by
  simp only [not_def, getLsb_xor, getLsb_allOnes]
  rw [getLsb_eq_decide_and_getLsb]
  -- case bash!
  cases decide (i < v) <;> simp
```
The point is that this introduces all the facts you want into the boolean expression, which can then be handled by boolean reasoning (in this case implemented via `cases` and `simp` because we don't have a sat solver).

Of course I could do it the tradional way with an implication, but it is extra work mixing the propositional and boolean approaches.

Std/Data/Nat/Bitwise.lean
```diff
@@ -254,24 +264,38 @@ theorem testBit_two_pow_add_gt {i j : Nat} (j_lt_i : j < i) (x : Nat) :
       else
         simp [i_lt_j]
 
-private theorem testBit_one_zero : testBit 1 0 = true := by trivial
+@[simp] theorem testBit_one_zero : testBit 1 0 = true := by trivial
```
tobiasgrosser, line 267:
```suggestion
theorem testBit_one_zero : testBit 1 0 = true := by trivial
```

This will calm down the linter without breaking the proofs.

Std/Data/BitVec/Lemmas.lean
joehendrix: How about we make this private for now?  I think the right approach is to have some form of programable [forward chaining](https://www.cs.utexas.edu/users/moore/acl2/manuals/current/manual/index-seo.php/ACL2____FORWARD-CHAINING), but I don't know of current Lean tactics that do that.

digama0: I think this would be less objectionable if there were separate theorems saying `getLsb x i -> i < w` and `(a -> b) -> b && a = a` and `getLsb_not` just composed them.

semorrison: I think I'll do all these:
* split it into two steps
* make the lemma private
* leave a note at the point of use that a tactic is missing.

semorrison: I found a much better proof anyway.

joehendrix (approved):

tobiasgrosser:
Nice. Thank you!

PR #601, opened 2024-02-07
chore: retroactive review of BitVec.Bitblast

semorrison:
Attempting to get up to speed on this file, initial comments. Please feel free to accept or reject piecewise via emoji, and I'll update. :-)

### DIFF
--- a/Std/Data/BitVec/Bitblast.lean
+++ b/Std/Data/BitVec/Bitblast.lean
@@ -18,11 +18,8 @@ The module is named for the bit-blasting operation in an SMT solver that convert
 expressions into expressions about individual bits in each vector.
 
 ## Main results
-* `x + y : BitVec w` is equivalent to `adc x y false`.
+* `x + y : BitVec w` is `(adc x y false).2`.
 
-## Future work
-All other operations are to be PR'ed later and are already proved in
-https://github.com/mhk119/lean-smt/blob/bitvec/Smt/Data/BitVec.lean.
 -/
 
 open Nat Bool
@@ -78,7 +75,7 @@ private theorem mod_two_pow_lt (x i : Nat) : x % 2 ^ i < 2^i := Nat.mod_lt _ (Na
 /-! ### Addition -/
 
 /-- carry w x y c returns true if the `w` carry bit is true when computing `x + y + c`. -/
-def carry (w x y : Nat) (c : Bool) : Bool :=  decide (x % 2^w + y % 2^w + c.toNat ≥ 2^w)
+def carry (w x y : Nat) (c : Bool) : Bool := decide (x % 2^w + y % 2^w + c.toNat ≥ 2^w)
 
 @[simp] theorem carry_zero : carry 0 x y c = c := by
   cases c <;> simp [carry, mod_one]
@@ -104,7 +101,7 @@ theorem carry_succ (w x y : Nat) (c : Bool) : carry (succ w) x y c =
   generalize testBit x w = xh
   generalize testBit y w = yh
   have sum_bnd : x%2^w + (y%2^w + c.toNat) < 2*2^w := by
-          simp [Nat.mul_comm 2 _, ←Nat.pow_succ ]
+          simp only [← Nat.pow_succ']
           exact adc_overflow_limit x y w c
   simp only [Nat.pow_succ]
   cases xh <;> cases yh <;> cases Decidable.em (x%2^w + (y%2^w + toNat c) ≥ 2 ^ w) with | _ pred =>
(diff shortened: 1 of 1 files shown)


Std/Data/BitVec/Bitblast.lean
```diff
@@ -18,11 +18,8 @@ The module is named for the bit-blasting operation in an SMT solver that convert
 expressions into expressions about individual bits in each vector.
 
 ## Main results
-* `x + y : BitVec w` is equivalent to `adc x y false`.
+* `x + y : BitVec w` is `(adc x y false).2`.
 
-## Future work
-All other operations are to be PR'ed later and are already proved in
-https://github.com/mhk119/lean-smt/blob/bitvec/Smt/Data/BitVec.lean.
```
semorrison, line 25: As far as I can see, that file doesn't contain anything analogous to the contents of this file? Perhaps the comment is an orphan from an earlier period when BitVec first arrived in Std?

Std/Data/BitVec/Bitblast.lean
```diff
@@ -18,11 +18,8 @@ The module is named for the bit-blasting operation in an SMT solver that convert
 expressions into expressions about individual bits in each vector.
 
 ## Main results
-* `x + y : BitVec w` is equivalent to `adc x y false`.
+* `x + y : BitVec w` is `(adc x y false).2`.
```
semorrison, line 21: I'm still not thrilled by this file, but not putting my finger on it.

My main complaint is that it proves that something I already knew about is equal to something new.

(Does that make sense? The new thing needs to be introduced and motivated first, and then its interesting to prove old and new things are equal. Sorry if that is a bit vague.)

Std/Data/BitVec/Bitblast.lean
```diff
@@ -99,20 +96,21 @@ theorem adc_overflow_limit (x y i : Nat) (c : Bool) : x % 2^i + (y % 2^i + c.toN
   exact Nat.mod_lt _ (Nat.pow_two_pos i)
 
 theorem carry_succ (w x y : Nat) (c : Bool) : carry (succ w) x y c =
+    --
     decide ((x.testBit w).toNat + (y.testBit w).toNat + (carry w x y c).toNat ≥ 2) := by
```
semorrison, line 100: I'm sad that this isn't in terms of the at-least-two boolean operator `x && y || x && c || y && c`.

Should I be sad?

Maybe this is what I was saying above about `adc` being something new that I don't care about:

I would like the beginning of the file to define
`Nat.carry`, `Bool.atLeastTwo`, `Bool.xorThree`, and `BitVec.adc`.

Then I would like to have the theorems explaining how they relate via `BitVec.toNat` and `BitVec.getLsb`.

Finally (possibly in a different file!) I would like `+ = adc`.

Std/Data/BitVec/Bitblast.lean
alexkeizer: I can understand your point. In some sense, `adc` is an alternative definition of the addition operation we already have. But, having `adc` allows us to prove things about addition from the perspective as a bit-by-bit transformation with one bit of state.

This was helpful, for example, when proving `x + ~~~x = -1#w`. The argument here is that, actually, the carry bit is always false, and the proof went as follows:
```lean
rw [add_as_adc, adc, iunfoldr_replace (fun _ => false) (-1#w)]
  · rfl
  · simp [adcb]
```

This doesn't necessarily mean we should keep exposing `adc`. Probably we should find the fundamental theorems about `add` that would allow us to make this argument, and expose only those. What comes to mind is:
* Some theorem about `(concat xs x) + (concat ys y) = concat (xs + ys + _) _`
* An analogue of `iunfoldr_replace`, essentially a coinduction principle for addition as an arbitrary-width bitvector transformation

Std/Data/BitVec/Bitblast.lean
mhk119: Sorry, meant to link this [file](https://github.com/mhk119/lean-smt/blob/bitvec/Smt/Data/Bitwise.lean).

semorrison: Thanks.

Std/Data/BitVec/Bitblast.lean
semorrison: Thanks for the explanations.

semorrison:
I've made more use of `omega` in `carry_succ`, and also removed the `x.toNat + y.toNat + z.toNat >= 2` in favour of `x && y || x && z || y && z`, which further shortens the proofs.

joehendrix (approved):

PR #604, opened 2024-02-08
chore: remove duplicated delayedAssign functions

semorrison:


semorrison:
I'm disinclined to add deprecation warning on these:
1. metaprogramming stuff comes with less guarantee
2. it would involve rejiggering imports to allow `@[deprecated] alias` here.

If it's important we could add the aliases in a later file.

digama0:
I'm okay with removing it without deprecation as long as you double check it's not used in aesop or mathlib

PR #607, opened 2024-02-09
chore: upstream Fin.coe_sub_iff_le from Mathlib

semorrison:


Std/Data/Fin/Lemmas.lean
```diff
@@ -752,6 +754,30 @@ theorem addCases_right {m n : Nat} {motive : Fin (m + n) → Sort _} {left right
...
+
+theorem coe_sub_iff_lt {a b : Fin n} : (↑(a - b) : Nat) = n + a - b ↔ a < b := by
+  rw [sub_def, lt_def]
+  dsimp only
+  if h : n ≤ a + (n - b) then
+    rw [Nat.mod_eq_sub_of_lt_two_mul h]
+    all_goals omega
+  else
+    rw [Nat.mod_eq_of_lt]
+    all_goals omega
```
joehendrix, line 779: The above two lemmas seem very specific.  What are they for?

semorrison: These are the remaining NLSym dependency on Mathlib.Data.Fin.Basic.

joehendrix (approved):

PR #610, opened 2024-02-09
chore: split Int/Order and Int/DivMod

semorrison:
The criterion for this split (of `Int.Order` into `Int.Order` and `Int.Init.Order` and simlarly for `Int.DivMod`) is solely "is it needed for omega".

I'm still on the fence about whether this is a good idea. As a consequence we are allowed to import `omega` in `Int.Order` itself (and I did it, to double check it was allowed now), and a big fraction of the lemmas there could be `by omega` now.

I also moved `Int.Lemmas` into `Int.Init.Lemmas` for consistency: it is pretty much *all* needed by `omega`, and it is needed by `Int.Init.Order` and `Int.Init.DivMod`.

`Int.Order` was about 1000 lines: 400 moved into `Init` (so 600 are newly accessible to `omega`). `Int.DivMod` was about 1100 lines, 350 moved into `Init` (and so 750 are newly accessible, although `omega` is relatively less useful here).

I also commented out quite a few unused proofs in `omega`, that I guess are remnants from the verified algorithm. I'll triage them later for interest.


Std/Tactic/Omega/IntList.lean
```diff
@@ -35,9 +35,9 @@ theorem get_of_length_le {xs : IntList} (h : xs.length ≤ i) : xs.get i = 0 :=
   rw [get, List.get?_eq_none.mpr h]
   rfl
 
-theorem lt_length_of_get_nonzero {xs : IntList} (h : xs.get i ≠ 0) : i < xs.length := by
-  revert h
-  simpa using mt get_of_length_le
+-- theorem lt_length_of_get_nonzero {xs : IntList} (h : xs.get i ≠ 0) : i < xs.length := by
+--   revert h
+--   simpa using mt get_of_length_le
```
joehendrix, line 40: What is the intent with the commented out code?  Could it be deleted?

semorrison: Yes: last para of the PR description. I found lots of proofs in `Tactic/Omega/` are just not used. I commented them out so they didn't cause spurious dependencies, but I didn't want to take the time right now to decide if anything is worth saving.

semorrison: But I think I'll just delete it.

PR #611, opened 2024-02-09
feat: Remove `ext` attribute on `Rat`

adomani:
Following [this discussion](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Compute.20roots.20of.20polynomials/near/420643573), I would like to remove the `ext` attribute from `Rat`.

This is my first Std PR: I hope that I am doing things correctly!

### DIFF
--- a/Std/Data/Rat/Lemmas.lean
+++ b/Std/Data/Rat/Lemmas.lean
@@ -11,6 +11,9 @@ import Std.Tactic.SeqFocus
 
 namespace Rat
 
+@[simp] theorem mk_den_one {r : Int} :
+    ⟨r, 1, Nat.one_ne_zero, (Nat.coprime_one_right _)⟩ = (r : Rat) := rfl
+
 @[simp] theorem zero_num : (0 : Rat).num = 0 := rfl
 @[simp] theorem zero_den : (0 : Rat).den = 1 := rfl
 @[simp] theorem one_num : (1 : Rat).num = 1 := rfl
@@ -218,7 +221,7 @@ theorem divInt_add_divInt (n₁ n₂ : Int) {d₁ d₂} (z₁ : d₁ ≠ 0) (z�
 @[simp] theorem neg_den (a : Rat) : (-a).den = a.den := rfl
 
 theorem neg_normalize (n d z) : -normalize n d z = normalize (-n) d z := by
-  simp [normalize, maybeNormalize_eq]; ext <;> simp [Int.neg_div]
+  simp [normalize]; rfl
 
 theorem neg_mkRat (n d) : -mkRat n d = mkRat (-n) d := by
   if z : d = 0 then simp [z]; rfl else simp [← normalize_eq_mkRat z, neg_normalize]
@@ -344,17 +347,16 @@ theorem intCast_one : ((1 : Int) : Rat) = (1 : Rat) := rfl
 @[simp, norm_cast] theorem intCast_add (a b : Int) :
     ((a + b : Int) : Rat) = (a : Rat) + (b : Rat) := by
   rw [add_def]
-  ext <;> simp [normalize_eq]
+  simp [normalize_eq]
 
-@[simp, norm_cast] theorem intCast_neg (a : Int) : ((-a : Int) : Rat) = -(a : Rat) := by
-  ext <;> simp [normalize_eq]
+@[simp, norm_cast] theorem intCast_neg (a : Int) : ((-a : Int) : Rat) = -(a : Rat) := rfl
 
 @[simp, norm_cast] theorem intCast_sub (a b : Int) :
     ((a - b : Int) : Rat) = (a : Rat) - (b : Rat) := by
   rw [sub_def]
-  ext <;> simp [normalize_eq]
+  simp [normalize_eq]
 
 @[simp, norm_cast] theorem intCast_mul (a b : Int) :
     ((a * b : Int) : Rat) = (a : Rat) * (b : Rat) := by
   rw [mul_def]
-  ext <;> simp [normalize_eq]
+  simp [normalize_eq]


adomani:
awaiting-review

fgdorais:
Could you summarize the Zulip discussion here?

adomani:
The discussion was essentially that `ext` digging into `ℚ` is like `ext` digging into `ℂ` and `ℂ`'s `ext` was removed.

In the specific case, calling `ext` on an equality of polynomials with rational coefficients entered the polynomials through the equality of their coefficients and then broke up the (rational) coefficients into their numerators and denominators.  This leads you astray more often than it is helpful.

Std/Data/Rat/Lemmas.lean
```diff
@@ -16,6 +16,8 @@ namespace Rat
 @[simp] theorem one_num : (1 : Rat).num = 1 := rfl
 @[simp] theorem one_den : (1 : Rat).den = 1 := rfl
 
+@[simp] theorem mk'_one {r : Int} : mk' r 1 Nat.one_ne_zero (Nat.coprime_one_right _) = r := rfl
```
fgdorais, line 19: Other lemmas don't use the prime. It's still an odd name. Can anyone think of something better?
```suggestion
@[simp] theorem mk_one {r : Int} : ⟨r, 1, Nat.one_ne_zero, Nat.coprime_one_right _⟩ = r := rfl
```

fgdorais (approved):
Approved after minor change.

Std/Data/Rat/Lemmas.lean
adomani: Would `mk_den_one` be a better name?

adomani:
In the process of fixing `Mathlib`, `Rat.ext` was useful exactly once, to prove [NNRat.ext_num_den](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Rat/NNRat.html#NNRat.ext_num_den).

Thus, I propose to add also this lemma, but not tagging it with `ext`.


```lean
theorem Rat.ext (p q : ℚ) (hn : p.num = q.num) (hd : p.den = q.den) : p = q := by
  cases p; cases q;  subst hn hd; rfl
```

If this is reasonable, then I will add it to this PR.

Also, the proof of `Rat.toNNRat_mul` had `ext <;> simp ...` and is `ext; simp ...` after the PR.

I am happy to do a pre-emptive PR to Mathlib to make explicit use of the `Rat.ext` lemma, rather than relying on the `ext` tactic, if this is the standard approach.

fgdorais:
Yes, please add the manual `Rat.ext` lemma (`p` and `q` should be implicit) along with a comment explaining why this is not marked `@[ext]`.

Std/Data/Rat/Lemmas.lean
fgdorais: This needs `(r : Rat)` on the rhs for the anonymous constructor to resolve the type.

digama0:
Does it work to use `local attribute [ext] Rat` in the Lemmas file?

adomani:
@digama0, `local attribute` complains:
```
unexpected token 'attribute'; expected 'binder_predicate', 'builtin_simproc', 'elab', 'elab_rules', 'infix', 'infixl', 'infixr', 'instance', 'macro', 'macro_rules', 'notation', 'postfix', 'prefix', 'simproc', 'syntax' or 'unif_hint'
```

Std/Data/Rat/Lemmas.lean
```diff
@@ -16,6 +16,9 @@ namespace Rat
 @[simp] theorem one_num : (1 : Rat).num = 1 := rfl
 @[simp] theorem one_den : (1 : Rat).den = 1 := rfl
 
+@[simp] theorem mk_den_one {r : Int} :
+    (⟨r, 1, Nat.one_ne_zero, Nat.coprime_one_right _⟩ : Rat) = r := rfl
```
fgdorais, line 20: Coercion should be on the rhs.
```suggestion
@[simp] theorem mk_den_one (r : Int) : 
  ⟨r, 1, Nat.one_ne_zero, Nat.coprime_one_right _⟩ = (r : Rat) := rfl
```

digama0:
> @digama0, `local attribute` complains:
> 
> ```
> unexpected token 'attribute'; expected 'binder_predicate', 'builtin_simproc', 'elab', 'elab_rules', 'infix', 'infixl', 'infixr', 'instance', 'macro', 'macro_rules', 'notation', 'postfix', 'prefix', 'simproc', 'syntax' or 'unif_hint'
> ```

sorry, the syntax is `attribute [local ext]` instead of `local attribute [ext]`

Std/Data/Rat/Lemmas.lean
adomani: I accepted all but the indent, since I think that 4 spaces is standard for pre-proof.

adomani:
@digama0 I'll try it, but tomorrow, although I'm sure that it will work, since you suggested it!

I imagine that this means to revert all the remaining changes, right?

Std/Data/Rat/Basic.lean
```diff
@@ -27,6 +27,10 @@ denominator is positive and the numerator and denominator are coprime.
   reduced : num.natAbs.Coprime den := by decide
   deriving DecidableEq
 
+-- `Rat` is not tagged with the `ext` attribute, since this is more often than not undesirable
+theorem Rat.ext {p q : Rat} (hn : p.num = q.num) (hd : p.den = q.den) : p = q := by
+  cases p; cases q; subst hn hd; rfl
+
```
fgdorais, line 33: Irresistible golfing...
```suggestion
theorem Rat.ext : {p q : Rat} → p.num = q.num → p.den = q.den → p = q
  | ⟨_,_,_,_⟩, ⟨_,_,_,_⟩, rfl, rfl => rfl
```

adomani:
@digama0 and @fgdorais -- I reverted all changes and made `ext` a local attribute!

The extra `simp` lemma that I had added seemed to make proof nicer, in my opinion: can you say why the current ones are better?

digama0:
No I did mean to use `attribute [local ext] Rat` here in Std, because this has the side effect of creating `Rat.ext` (and marking it `[local ext]` for the file, which might be useful since this file wants to prove facts about Rat). Downstream, mathlib can use `attribute [local ext] Rat.ext` to use it in `ext` proofs.

fgdorais:
But the generated `Rat.ext` doesn't have the implicit p and q so it is less suitable for manual use. This is better:
```lean
@[local ext] theorem Rat.ext : {p q : Rat} → p.num = q.num → p.den = q.den → p = q
  | ⟨_,_,_,_⟩, ⟨_,_,_,_⟩, rfl, rfl => rfl
```
The changes Damiano had made were all simplifications of the original and ultimately showed that this ext lemma is not necessary even in this file.

digama0:
in that case we should fix `ext`. I don't like the precedent this sets, that `@[ext]` on structures should be avoided even when it is producing the correct extensionality lemma. This seems like it's working around either an issue in `ext` or a misuse of it.

fgdorais:
Yes, it's an issue in `ext`. This workaround also shows up in `Std.Tactic.Ext` itself.

fgdorais:
I still don't want to lose the simpler proofs that Damiano had proposed.

adomani:
Can I suggest merging the current minimalistic state of the PR and I can produce the rest of the changes in a separate PR?

adomani:
Also, as far as I understand, the reason there was no need to use `ext` in the earlier version was the new `simp` lemma that took care of closing the missing goals: I initially did not introduce the manual `Rat.ext` lemma, I simply removed it.

fgdorais:
In addition to the sub-optimal binders, there is another bug in `ext` where marking a structure with `@[local ext]` generates the `.ext` and `.ext_iff` lemmas and incorrectly marks `.ext` with `[ext]` instead of `[local ext]`.

fgdorais:
I don't want this PR to depend on fixing `ext`. Since the `@[local ext]` attribute currently does the same thing as just `@[ext]` and this is undesirable, the workaround
```lean
@[local ext] theorem Rat.ext : {p q : Rat} → p.num = q.num → p.den = q.den → p = q
  | ⟨_,_,_,_⟩, ⟨_,_,_,_⟩, rfl, rfl => rfl
```
at the top of the lemma file is the best option.

In the future, this should be replaced by `attribute [local ext] Rat` at the top of the file.

Std/Data/Rat/Basic.lean
fgdorais: Rendered irrelevant.

Std/Data/Rat/Lemmas.lean
```diff
@@ -11,6 +11,8 @@ import Std.Tactic.SeqFocus
 
 namespace Rat
 
+attribute [local ext] Rat
```
fgdorais, line 14: This currently doesn't work as intended. See #618.

fgdorais (changes requested):

adomani:
@fgdorais I agree with you: in a file that imports another one containing `attribute [local ext] Rat` the `ext` tactic works on `Rat`s.

@digama0 I can revert this PR to its original form, where I had simply removed the `ext` attribute, introduced `mk_den_one` and fix the `ext` proofs in `Std`, if you think that you would merge it.  Otherwise, I am also happy to wait for the fix to the `local ext`, since the workaround for avoiding `ext` on `Rat` is simply to add a stopper `ext x : whatever`.

PR #612, opened 2024-02-10
chore: rename Classical.and_not

semorrison:
We just added `Classical.and_not`, but it is causing too many disambiguation errors with `_root_.and_not` in Mathlib, so I am giving it a longer name.


semorrison:
Apologies, I am just going to merge this myself, as I really want to get Mathlib's nightly-testing working again, and don't want to have to do all the `_root_.not_and` disambiguation in Mathlib needlessly.

Please ping me and I'm happy to do another name change later if it is needed.

digama0:
Can mathlib just re-export the theorem instead of defining an alias?

PR #615, opened 2024-02-10
chore: adaptations for nightly-2024-02-10

semorrison:
This is a PR to bump/v4.7.0. It is larger than usual, deleting a number of files which have been moved to the Lean repository in the last two days.


Std/Tactic/NoMatch.lean
```diff
@@ -1,65 +0,0 @@
...
-impossible pattern. The `match x with.` and `intro.` tactics do the same thing but in tactic mode.
--/
-namespace Std.Tactic
-open Lean Elab Term Parser.Term
-
-/--
-The syntax `match x with.` is a variant of `nomatch x` which supports pattern matching on multiple
-discriminants, like regular `match`, and simply has no alternatives in the match.
--/
-syntax:lead (name := noMatch) "match " matchDiscr,* " with" "." : term
```
digama0, line 26: We can't remove this file, it needs to be a deprecation

Std/Tactic/NoMatch.lean
```diff
@@ -1,65 +0,0 @@
...
-@[inherit_doc noFun] macro tk:"λ" "." : term => `(fun%$tk .)
-
-@[inherit_doc noMatch] macro "match " discrs:matchDiscr,* " with" "." : tactic =>
-  `(tactic| exact match $discrs,* with.)
-
-/--
-The tactic `intro.` is shorthand for `exact fun.`: it introduces the assumptions, then performs an
-empty pattern match, closing the goal if the introduced pattern is impossible.
--/
-macro "intro" "." : tactic => `(tactic| exact fun.)
```
digama0, line 65: One thing I don't like about `nofun` replacing `intro.` is that now we don't have keyword equivalence between the 0-ary and n-ary versions. Although this leads us down the path of renaming `intro` to `fun`...

Std/Data/RBMap/Alter.lean
```diff
@@ -183,16 +183,16 @@ protected theorem Balanced.del {path : Path α}
     | red, ⟨_, h⟩ | black, ⟨_, _, h⟩ => exact h.setBlack
   | @redL _ n _ _ hb hp ih => match c', n, ht with
     | red, _, _ => cases hc rfl rfl
-    | black, _, ⟨_, rfl, ha⟩ => exact ih ((hb.balLeft ha).of_false (fun.)) (fun.)
+    | black, _, ⟨_, rfl, ha⟩ => exact ih ((hb.balLeft ha).of_false (nofun)) (nofun)
```
digama0, line 186:
```suggestion
    | black, _, ⟨_, rfl, ha⟩ => exact ih ((hb.balLeft ha).of_false nofun) nofun
```
Probably find/replace this, there are a few dozen occurrences

Std/Data/RBMap/Lemmas.lean
```diff
@@ -328,7 +328,7 @@ strictly greater than the cut (so there is no exact match, and nothing closer to
 theorem Ordered.lowerBound?_least [@TransCmp α cmp] [IsCut cmp cut] (ht : Ordered cmp t)
     (H : t.lowerBound? cut none = some x) (hy : y ∈ t)
     (xy : cmp x y = .lt) (hx : cut x = .gt) : cut y = .lt :=
-  ht.lowerBound?_least_lb (by exact fun.) H hy hx xy
+  ht.lowerBound?_least_lb (by exact nofun) H hy hx xy
```
digama0, line 331:
```suggestion
  ht.lowerBound?_least_lb (by nofun) H hy hx xy
```

Std/Tactic/NoMatch.lean
semorrison: Okay, I've added this back. It felt like an extra hoop to jump through to add the warnings manually, if there's a better method for deprecating syntax I'm happy to try again.

digama0: Well ideally there would be something to copy paste from, but we don't deprecate syntax very often, this is the first time it has happened in std. (Lean core likes to just move fast and break things here, but std has a deprecation policy.)

PR #616, opened 2024-02-11
chore: add @[simp] to imp_false

semorrison:
Adds `@[simp]` to `theorem imp_false : (a → False) ↔ ¬a`.

I think this is a good simp lemma, and Mathlib adds it immediately in any case.

(I've deleted a unused proof that broke as a result: it was an evil proof with a nonterminal simp, and it is being deleted in another open PR in any case.)

### DIFF
--- a/Std/Logic.lean
+++ b/Std/Logic.lean
@@ -126,7 +126,7 @@ theorem true_imp_iff (α : Prop) : (True → α) ↔ α := ⟨fun h => h trivial
 
 @[simp] theorem imp_self : (a → a) ↔ True := iff_true_intro id
 
-theorem imp_false : (a → False) ↔ ¬a := Iff.rfl
+@[simp] theorem imp_false : (a → False) ↔ ¬a := Iff.rfl
 
 theorem imp.swap : (a → b → c) ↔ (b → a → c) := ⟨flip, flip⟩
 
--- a/Std/Tactic/Omega/IntList.lean
+++ b/Std/Tactic/Omega/IntList.lean
@@ -506,28 +506,7 @@ def trim (xs : IntList) : IntList := xs.trim?.getD xs
 
 theorem trim?_eq_some {xs : IntList} (w : xs.trim? = some t) : xs.trim = t := by simp [trim, w]
 
-theorem trim_spec {xs : IntList} : xs.trim = (xs.reverse.dropWhile (· == 0)).reverse := by
-  dsimp [trim, trim?]
-  split <;> rename_i h h'
-  · simp_all
-  · simp_all [List.dropWhile_cons]
-  · simp_all [List.dropWhile_cons]
-    rw [if_neg h, ← h', List.reverse_reverse]
-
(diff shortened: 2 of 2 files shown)


joehendrix:
One issue is that `imp_false` is not confluent with `and_imp`:

```lean
@[simp] theorem and_imp : (a ∧ b → c) ↔ (a → b → c) :=
  ⟨fun h ha hb => h ⟨ha, hb⟩, fun h ⟨ha, hb⟩ => h ha hb⟩
```

If you consider the term `a ∧ b → False` then `and_imp` followed by `imp_false` yields `a → ¬b`, but if you apply `imp_false` first, then we get `¬(a ∧ b)` (which I think is superior).  Potentially, `and_imp` could be lower priority so `imp_false` is triggered first.  I'd like to do this in the context of the Boolean changes in #593 (or something similar).

digama0:
I think we want to prefer `a → ¬b` there, we should already have a simp lemma rewriting `¬(a ∧ b)` to `a → ¬b` if `and_imp` doesn't already apply.

semorrison:
Yes, it does seem that `and_imp` and `¬(a ∧ b)` to `a → ¬b` should both be simps.

joehendrix:
This was one of the annoyances that I found with importing Mathlib in any projects I had.  I don't know why one would want to break a flattened conjunction like ¬(... ∧ ... ∧ ... ∧ ..) into an implication based on the right-most term that happened to match.

I think confluence is more important than happening to occasionally trigger an implication.

digama0:
what do you mean by "based on the right-most term that happened to match"? Note that simplifying `¬(a ∧ b)` to `a → ¬b` results in a confluent system together with `and_imp` and `imp_false`.

joehendrix:
I've thought about this a bit, and I'm still not a fan, but think compatibility with Mathlib is the way to go for now.

The main thing that would have been good for my earlier experience is how to normalize the implication chains that come out of this.  With an And, it's a matter of running simp with associativity, commutativity and a left commutative rule.  With implication, it's a little less clear what the rules are to rewrite modulo, the consequent has a somewhat unique position of having an explicit rather than implicit negation.

digama0:
> The main thing that would have been good for my earlier experience is how to normalize the implication chains that come out of this. With an And, it's a matter of running simp with associativity, commutativity and a left commutative rule. With implication, it's a little less clear what the rules are to rewrite modulo, the consequent has a somewhat unique position of having an explicit rather than implicit negation.

Well, when rewriting everything into implications associativity is built in and commutativity isn't necessary since the last thing in the implication chain is special / not subject to reassociation, so the only thing you need is the left commutativity rule `imp.swap`. It's true that adding negation into the mix (using `imp_false` as a simp lemma rather than the reverse) complicates matters, and in that case you also need to add commutativity of the last two conjuncts which is `imp_not_comm`.

joehendrix:
`imp_false` is now simp upstream.

PR #617, opened 2024-02-11
feat: add `Rat.mk_den_one` simp lemma

adomani:
This PR extracts the new lemma `mk_den_one`, previously introduced in #611 and uses it to golf some proofs.

### DIFF
--- a/Std/Data/Rat/Basic.lean
+++ b/Std/Data/Rat/Basic.lean
@@ -13,6 +13,7 @@ import Std.Tactic.Ext
 Rational numbers, implemented as a pair of integers `num / den` such that the
 denominator is positive and the numerator and denominator are coprime.
 -/
+-- `Rat` is not tagged with the `ext` attribute, since this is more often than not undesirable
 structure Rat where
   /-- Constructs a rational number from components.
   We rename the constructor to `mk'` to avoid a clash with the smart constructor. -/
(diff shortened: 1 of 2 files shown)


Std/Data/Rat/Basic.lean
```diff
@@ -13,7 +13,8 @@ import Std.Tactic.Ext
 Rational numbers, implemented as a pair of integers `num / den` such that the
 denominator is positive and the numerator and denominator are coprime.
 -/
-@[ext] structure Rat where
+-- `Rat` is not tagged with the `ext` attribute, since this is more often than not undesirable
+structure Rat where
```
fgdorais, line 17: Restore to original state to avoid dependency on #611 .
```suggestion
@[ext] structure Rat where
```

Std/Data/Rat/Lemmas.lean
```diff
@@ -11,6 +11,11 @@ import Std.Tactic.SeqFocus
 
 namespace Rat
 
+attribute [local ext] Rat
```
fgdorais, line 14: Remove to avoid dependency on #611.

Std/Data/Rat/Lemmas.lean
```diff
@@ -11,6 +11,11 @@ import Std.Tactic.SeqFocus
...
+
+@[simp] theorem mk_den_one {r : Int} :
```
fgdorais, line 16: Excellent simp lemma! The name is still unusual but there are probably no uses outside simp, so it doesn't matter much.

fgdorais (changes requested):
This can be made independent of #611.

adomani:
@fgdorais thanks for the suggestions!  I made this PR independent on the one with `Rat.ext`.

fgdorais (approved):

joehendrix (approved):
The seems sensible to me.

PR #628, opened 2024-02-13
chore: upstream pow_dvd_pow_iff_pow_le_pow

semorrison:


Std/Data/Nat/Lemmas.lean
```diff
@@ -573,6 +573,15 @@ theorem sub_lt_succ (a b) : a - b < succ a := lt_succ_of_le (sub_le a b)
 theorem sub_one_sub_lt (h : i < n) : n - 1 - i < n := by
   rw [Nat.sub_right_comm]; exact Nat.sub_one_lt_of_le (Nat.sub_pos_of_lt h) (Nat.sub_le ..)
 
+protected theorem exists_eq_add_of_le (h : m ≤ n) : ∃ k : Nat, n = m + k :=
```
fgdorais, line 576: This is basically `Nat.le.dest` but I very much prefer `Nat.exists_eq_add_of_le`. Should `Nat.le.dest` be deprecated?

semorrison: Could we pursue that separately?

joehendrix (approved):

PR #637, opened 2024-02-13
feat: lemmas about `BitVec.getLsb` and bitwise ops (and/or/xor)

alexkeizer:
Some more lemmas that rewrite `toFin` applied to a bitwise operation (`and`/`or`/`xor`)


alexkeizer:
awaiting-review

PR #638, opened 2024-02-13
chore: lemma for BitVec.ofFin to .ofNat

joehendrix:
I think readability is improved if we make this a simp rule, but it is currently too easy to create simp loops with it on.

### DIFF
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -15,6 +15,12 @@ import Std.Util.ProofWanted
 
 namespace Std.BitVec
 
+/--
+This normalized a bitvec using `ofFin` to `ofNat`.
+-/
+theorem ofFin_eq_ofNat : @BitVec.ofFin w (Fin.mk x lt) = BitVec.ofNat w x := by
+  simp only [BitVec.ofNat, Fin.ofNat', lt, Nat.mod_eq_of_lt]
+
 /-- Prove equality of bitvectors in terms of nat operations. -/
 theorem eq_of_toNat_eq {n} : ∀ {i j : BitVec n}, i.toNat = j.toNat → i = j
   | ⟨_, _⟩, ⟨_, _⟩, rfl => rfl
--- a/test/bitvec.lean
+++ b/test/bitvec.lean
@@ -114,3 +114,8 @@ def testMatch8 (i : BitVec 32) :=
 #guard toString 5#13 = "0x0005#13"
 #guard toString 5#12 = "0x005#12"
 #guard toString 5#13 = "0x0005#13"
+
+-- Simp
+
+example (n w : Nat) (p : n < 2^w) : { toFin := { val := n, isLt := p } : BitVec w} = .ofNat w n := by
+  simp only [ofFin_eq_ofNat]


alexkeizer:
Could we make it a simp-proc instead, to avoid the loops? Are simp-procs a thing yet in Std's version of Lean?

joehendrix:
Could you help me understand with an example?

What's a theorem that user's want to prove about bitvectors that is easy to express with a `ofFin` normal form, but not an `ofNat?

alexkeizer:
Consider `toNat_ofFin` vs `toNat_ofNat`: the latter introduces a `% 2 ^ w` which the former can leave out because of the staticly known `Fin`-bound. Concretely:

```lean
example (x : Nat) (h : x < 2^w) : BitVec.toNat (ofFin ⟨x, h⟩) = z := by
  simp
  -- ⊢ x = z
  sorry

@[simp] theorem ofFin_eq_ofNat : @BitVec.ofFin w (Fin.mk x lt) = BitVec.ofNat w x := by
  simp only [BitVec.ofNat, Fin.ofNat', lt, Nat.mod_eq_of_lt]

example (x : Nat) (h : x < 2^w) : BitVec.toNat (ofFin ⟨x, h⟩) = z := by
  simp
  -- ⊢ x % 2 ^ w = z
  rw [Nat.mod_eq_of_lt h]
  -- ⊢ x = z
  sorry
```

By rewriting the `ofFin` into the `ofNat`, our goal state becomes `x % 2 ^ w = z` rather than just `x = z`. Admittedly, in this example it's easy enough to get rid of the modulus with `Nat.mod_eq_of_lt`, but this requires the user to manually supply the `x < 2^w` proof. `simp` is not able to automatically apply this lemma:
```lean
example (x : Nat) (h : x < 2^w) : BitVec.toNat (ofFin ⟨x, h⟩) = z := by
  simp [Nat.mod_eq_of_lt]
  -- ⊢ x % 2 ^ w = z (just like before)
```

joehendrix:
Thanks for the example.  I think we need to ensure the normal form is very easy since I think on industrial problems there may be many of them.  Readability and avoiding dependent types for the `w` are my primary reason for preferring something that de-elaborates into the `x#w` syntax.

We do some ability to automatically know bounds.  One can get that through using `toFin` instead of `toNat`.  Perhaps though omega offer a user-extensible way of generating additional constraints from terms appearing in the context.

alexkeizer:
Could we solve this by adding a special notation for `BitVec.ofFin` that is easier to read.
Possibly following the example of `GetElem` notation, where the proof of the index being in-bounds is part of the expression, but hidden by the delaborator. To illustrate:
```lean
example : xs[1]'(by decide) = xs[1] := by
  -- ⊢ xs[1] = xs[1]
  rfl
```

This way, we gain readability, but don't lose the `Fin`-structure and the rewrites it gives us, at the downside of now having two notations that do slightly different things being possibly a bit confusing.

joehendrix:
I'm going punt on the simp discussion for this PR.

I see your point.  I'm still skeptical that we want to optimize normal forms to carry proof information with them long term, but I can see the utility of having proofs available too given the current level of automation has (forward chaining could be a way to provide it as needed).  I think we need some case studies and more experimentation to know is best for users.

PR #641, opened 2024-02-13
refactor: improve consistency of BitVec lemmas

alexkeizer:
Ensure all lemmas about `toNat`/`toFin` applied to an operation take their arguments explicitly.

It was pointed out in #636 that the new `toFin_` lemmas had inconsistent argument implicitness, which brought to light also the inconsistency of existing related `toNat_` lemmas. Both are addressed with the current PR.


alexkeizer:
awaiting review

alexkeizer:
awaiting-review

PR #643, opened 2024-02-14
chore: Drop Std.Data.Nat.Lemmas from Omega dependencies.

joehendrix:
This moves a bunch of lemmas around to remove the remaining lemmas in Std.Data.Nat.Lemmas from Omega's dependencies.

It may lose some organization, but at least the remaining lemmas could potentially use Omega.


semorrison:
@joehendrix, this is set, I think incorrectly, as a PR to `bump/v4.7.0`. I'm going to switch it back to `main` now.

PR #645, opened 2024-02-14
feat: ∀(x : BitVec w.succ), msb x = 1 iff x >= 2^w

bollu:
This characterizes when the `msb` is 1 in terms of an arithmetic inequality.

### DIFF
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -117,6 +117,22 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
   else
     simp [h]
 
+
+/-! ### msb -/
+
+theorem msb_decide (x : BitVec (Nat.succ w)) : BitVec.msb x = decide (x.toNat ≥ 2^w) := by
+  simp only [BitVec.msb, getMsb, Nat.zero_lt_succ, decide_True, getLsb, Nat.testBit, Nat.succ_sub_succ_eq_sub, 
+    Nat.sub_zero, Nat.and_one_is_mod, Bool.true_and, ge_iff_le]
+  rw [Nat.shiftRight_eq_div_pow]
+  rcases (Nat.lt_or_ge (BitVec.toNat x) (2 ^ w)) with h | h
+  · simp [Nat.div_eq_of_lt h, h]
+  · simp only [h, decide_True, bne_iff_ne, ne_eq]
+    rw [Nat.div_eq_sub_div (Nat.two_pow_pos _) (by omega), Nat.div_eq_of_lt]
+    · decide
+    · have hxmax : BitVec.toNat x < 2^(Nat.succ w) := x.isLt
+      rw [Nat.pow_succ, Nat.mul_two] at hxmax
+      omega
+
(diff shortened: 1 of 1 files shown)


bollu:
awaiting-review

Std/Data/BitVec/Lemmas.lean
```diff
@@ -117,6 +117,23 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
   else
     simp [h]
 
+
+/-! ### msb -/
+
+theorem msb_decide (x : BitVec (Nat.succ w)) : BitVec.msb x = decide (x.toNat ≥ 2^w) := by
```
alexkeizer, line 123:
```suggestion
theorem msb_eq_decide (x : BitVec (Nat.succ w)) : BitVec.msb x = decide (2 ^ w ≤ x.toNat) := by
```
Even though this is not a simp-lemma, it makes sense to use the simp-normal form (which prefers `<=` over `>=`). Also, the name should have `eq`

bollu:
awaiting-author

alexkeizer:
wip

Std/Data/BitVec/Lemmas.lean
```diff
@@ -117,6 +117,21 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
   else
     simp [h]
 
+
+/-! ### msb -/
+
+theorem msb_eq_decide (x : BitVec (Nat.succ w)) : BitVec.msb x = decide (2 ^ w ≤ x.toNat) := by
```
semorrison, line 123:
```suggestion
theorem msb_eq (x : BitVec (Nat.succ w)) : BitVec.msb x = decide (2 ^ w ≤ x.toNat) := by
```
I think the naming convention is [slightly unclear here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Naming.20convention/near/421544382), but my preference is that the name stops at `eq` if the RHS is obvious and unambiguous. So this could just be `msb_eq`.

semorrison: I guess, however, in this case we also want `x.msb = x.getMsb 0`, and possibly also `x.msb = x.getLsb (Fin.last w)`, so on second thoughts keep the name as is.

What do you think of adding the other two lemmas as well (and possibly also the corresponding ones for `x.lsb`)? Fine if not now.

semorrison:
Eventually we will need `simp` normal forms here. This doesn't need to be decided here, as you're not proposing to write `@[simp]`. I *think* my preference is for `x.msb ~> x.getLsb (Fin.last _)` and separately `x.getLsb (Fin.last _) ~> decide (...)`.

Std/Data/BitVec/Lemmas.lean
semorrison: @bollu, I think you may have committed before seeing my second message. :-)

bollu: @semorrison Yes, I did. I shall keep the name with the `msb_eq_decide ` on this one, and then add the other two as a follow up PR.

bollu:
awaiting-review

PR #646, opened 2024-02-14
refactor: consistently use variable `w` for the width of a bitvector

alexkeizer:
Currently, lemmas about bitvectors inconsistently use `n, m, v` and `w` for the width.
I've (arbitrarily) picked `w` as the blessed variable, and changed all lemmas to use this, only using `v` when a lemma involves two bitvectors of differing widths.
I don't think the exact choice of variable matters much, but it does matter that we pick one and consistently use it.

This changes only the name of variables, it should not break any other code.

Additionally, I've dropped `{w}` implicit arguments in favor of a `variable {w : Nat}` at the start of the file.
This has one functional change: the type of `getLsb_cons` used to be `(b : Bool) -> {w : Nat} -> ...`, but is now `{w : Nat} -> (b : Bool)`. Since this only changes the order of an implicit argument, the impact of this change should be very small.

This is still a work-in-progress. I've opened the PR already as a draft to see if I can get a leaff report to confirm there are no functional changes besides the one described above

### DIFF
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -15,8 +15,11 @@ import Std.Util.ProofWanted
 
 namespace Std.BitVec
 
+-- We use variables `w`, `v` and `u` (in order of preference) to represent the width of a bitvector
+variable {w v u : Nat}
+
 /-- Prove equality of bitvectors in terms of nat operations. -/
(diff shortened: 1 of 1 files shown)


alexkeizer:
WIP

semorrison:
If there's still useful material here, it will need to be moved to the Lean repo, and BitVec/Lemmas has been upstreamed.

PR #647, opened 2024-02-14
feat: Add a simp lemma `Nat.cast x = BitVec.ofNat _ x`

alexkeizer:
We already have a simp lemma that rewrites `OfNat.ofNat` into `BitVec.ofNat`.
This PR extends that by rewriting `Nat.cast` into `BitVec.ofNat` as well, making `BitVec.ofNat` the canonical simp-normal form for all ways to turn a `Nat` into a `BitVec`.


alexkeizer:
awaiting-review

joehendrix (approved):

Issue #618, opened 2024-02-11
Bugs in `ext` tactic


fgdorais:
These bugs were noticed in the discussion of #611.

The `.ext` and `.ext_iff` lemmas generated by applying the `@[ext]` attribute to a structure has sub-optimal binders. For example:
```lean
@[ext] structure S (α β) where (a : α) (b : β)
/-
S.ext.{u_1, u_2} {α : Sort u_1} {β : Sort u_2} (x y : S α β) (a : x.a = y.a) (b : x.b = y.b) : x = y
-/
#check S.ext
```
One would expect `x y` to be implicit parameters.

The `@[local ext]`/`@[scoped ext]` variants don't work correctly when applied to structures. 
* The generated `.ext` lemma is marked plain `@[ext]` instead of `@[local ext]`/`@[scoped ext]`.
* The `.ext` and `.ext_iff` are unconditionally generated, which fails if the lemmas were already generated by an unrelated `@[local ext]` in a a dependency, for example.

fgdorais:
After a bit of scouting, a better implementation should make use of these tools from core [`Lean.Meta.CongrTheorems`](https://github.com/leanprover/lean4/blob/4718af54744ec297d51cd5ec40c3ec26c4623ead/src/Lean/Meta/CongrTheorems.lean). 

I'm willing to do this but since my Lean time is limited it could take a long time.

fgdorais:
Migrated to [lean4#3643](https://github.com/leanprover/lean4/issues/3643)

PR #651, opened 2024-02-15
Rat: remove `ext` attribute

adomani:
Following [this discussion](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Compute.20roots.20of.20polynomials/near/420643573)*, I would like to remove the `ext` attribute from `Rat` (and manually add the `Rat.ext` lemma).

* The discussion was essentially that `ext` digging into ℚ is like `ext` digging into ℂ and ℂ's `ext` was removed.

  In the specific case, calling `ext` on an equality of polynomials with rational coefficients entered the polynomials through the equality of their coefficients (expected and desirable) and then broke up the (rational) coefficients into their numerators and denominators (undesirable). This leads you astray more often than it is helpful.

This uncovered that `local ext` does not really make the attribute local -- #618.  However, with the refactor introduced in #617 that is now merged, there is no need for `Std` to use the `ext` attribute for `Rat`.  Still, according to a local build, `Mathlib` may need the `Rat.ext` lemma in one file.

### DIFF
--- a/Std/Data/Rat/Basic.lean
+++ b/Std/Data/Rat/Basic.lean
@@ -13,7 +13,8 @@ import Std.Tactic.Ext
 Rational numbers, implemented as a pair of integers `num / den` such that the
 denominator is positive and the numerator and denominator are coprime.
 -/
-@[ext] structure Rat where
+-- `Rat` is not tagged with the `ext` attribute, since this is more often than not undesirable
+structure Rat where
   /-- Constructs a rational number from components.
   We rename the constructor to `mk'` to avoid a clash with the smart constructor. -/
   mk' ::
--- a/Std/Data/Rat/Basic.lean
+++ b/Std/Data/Rat/Basic.lean
@@ -28,6 +28,9 @@ structure Rat where
   reduced : num.natAbs.Coprime den := by decide
   deriving DecidableEq
 
+theorem Rat.ext : {p q : Rat} → p.num = q.num → p.den = q.den → p = q
+  | ⟨_,_,_,_⟩, ⟨_,_,_,_⟩, rfl, rfl => rfl
+
 instance : Inhabited Rat := ⟨{ num := 0 }⟩
 
 instance : ToString Rat where
--- a/Std/Data/Rat/Basic.lean
+++ b/Std/Data/Rat/Basic.lean
@@ -28,9 +28,6 @@ structure Rat where
   reduced : num.natAbs.Coprime den := by decide
   deriving DecidableEq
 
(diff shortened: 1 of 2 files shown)


adomani:
@digama0 @fgdorais 

Dear Both,

I had messed up #611, due to my inexperience with PRs from forks.  This led to the other PR getting closed, when I synchronized `Std/main` with my fork.  I reopened the PR here, on a branch that is no longer called `main`.

adomani:
awaiting-review

adomani:
You are right, I think that it is good to have the `Rat.ext` lemma -- I just added it back.

I was not sure whether adding the lemma now was preferred to waiting for a fix of the `local ext` bug.

fgdorais (approved):

fgdorais:
I'm afraid `ext` needs more than a quick patch. We can't count on it getting fixed soon. Maybe add a comment suggesting to replace with `attribute [local ext] Rat` once #618 is fixed.

adomani:
`Rat.ext` is not actually used in `Std`, so the comment might simply say to remove the lemma: there is no need to replace it with a `local ext`.  What may need the `ext`-lemma are downstream projects, notably Mathlib seems to have used it once.

I can add the comment, but tomorrow: I'm off to sleep now!

Std/Data/Rat/Basic.lean
```diff
@@ -27,6 +28,9 @@ denominator is positive and the numerator and denominator are coprime.
   reduced : num.natAbs.Coprime den := by decide
   deriving DecidableEq
 
+theorem Rat.ext : {p q : Rat} → p.num = q.num → p.den = q.den → p = q
+  | ⟨_,_,_,_⟩, ⟨_,_,_,_⟩, rfl, rfl => rfl
+
```
fgdorais, line 33: This should be in Rat.Lemmas.

It should also have `@[local ext]`. Applying the local attribute to an existing lemma still works. It's only the part that automatically generates the lemmas for a structure which is broken.

joehendrix: I don't know if a local ext matters much.  I'm inclined to upstream this.  @adomani if you want to move it, that's great, but I can also just pull it in and change it in a small PR.

adomani: @joehendrix you mean to move the lemma to `Rat.lemmas`?  I can do that right away, if that is what you wanted!

joehendrix: @adomani Thanks, moving to Rat.Lemmas si what I meant.  Once that's moved, I'll merge.

adomani: Great, thanks: I just pushed the move!

PR #652, opened 2024-02-15
chore: adaptations for nightly-2024-02-15

semorrison:


joehendrix (approved):

digama0:
For someone who has not been watching the progress of the lean4 PRs, what are the changes? It is very difficult to review this since these files have been moved to core with unknown modifications.

digama0 (changes requested):
For someone who has not been watching the progress of the lean4 PRs, what are the changes? It is very difficult to review this since these files have been moved to core with unknown modifications.

Std/Tactic/Omega/Logic.lean
```diff
@@ -5,15 +5,14 @@ Authors: Scott Morrison
...
-import Std.Logic
 
 /-!
 # Specializations of basic logic lemmas
 
 These are useful for `omega` while constructing proofs, but not considered generally useful
 so are hidden in the `Std.Tactic.Omega` namespace.
 
-If you find yourself needing them elsewhere, please move them first to `Std.Logic`.
+If you find yourself needing them elsewhere, please move them first to another file.
```
fgdorais, line 15: I think the right message here is that `Std.Omega.Logic` should not be an import outside the implementation of `omega`. Users can still import `Std.Omega` as a whole and these lemmas come for free. It's unreasonable to expect users to inspect every lemma the use to see whether it comes from this file.

Std/Lean/Position.lean
```diff
@@ -4,8 +4,8 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
 import Lean.Syntax
-import Lean.Data.Lsp.Utf16
 import Lean.Meta.Tactic.TryThis
+import Lean.Data.Lsp.Utf16
```
fgdorais, line 8: Is this import reordering significant?

fgdorais (changes requested):

Std/Tactic/Omega/Logic.lean
joehendrix: Agreed.  We are working on getting omega into lean itself -- which provides a path to addressing this issue by making the lemmas part of the Lean package.  However, that update is not feasible in this PR.

Std/Lean/Position.lean
joehendrix: I think that's an accidental diff.

joehendrix:
> For someone who has not been watching the progress of the lean4 PRs, what are the changes? It is very difficult to review this since these files have been moved to core with unknown modifications.

The main downstream visible changes are in module structure.  Tactics that previously required an Std module import that are migrated will be available without import.  Similarly definitions in Std.Tactic are getting moved elsewhere (Lean.Meta.Tactic).

We'll aim to provide guidance on what has changed in the next release.

Std/Tactic/Omega/Logic.lean
fgdorais: I can live with the current wording since it's not expected to last long.

fgdorais:
I agree with @digama0. Where should be guidelines for move PRs since these tend to be big and hard to review. 

One general principle should be that moves should minimize changes to the bare essentials (imports, namespaces, for example), and later PRs should be used for any other changes.

Another principle is that when moving away, all additions should be explained somehow, probably in the PR description.

Std/Lean/Position.lean
joehendrix: Turns out the import was no longer needed.

fgdorais: @joehendrix There's a bunch of other code additions that are just small changes to imports in a file. They might need a quick check to see if they are also no longer needed.

joehendrix:
The import additions are to just to keep declarations in scope.  We'll probably need to shake the imports further.  The only other move of note I see is `Bool.eq_iff_iff` from Std.Logic to Std.Data.Bool and that's largely because most of the contents of Std.Logic have been ported to Lean itself.

To the extent there are changes; it's largely happening within the Lean repo as we adapt Lean to address inconsistencies between Lean and Std (e.g, soon Lean master will use `ediv`/`emod` for Int division).

I appreciate this is really hard to review (and even just track).  We are trying to improve the consistency of new user experience across Lean, Std and Mathlib and are focused on getting tactics (including omega) and dependencies needed for tactics upstream rapidly.

Std/Data/Nat/Lemmas.lean
```diff
@@ -1266,3 +1256,6 @@ instance decidableExistsLT [h : DecidablePred p] : DecidablePred fun n => ∃ m
 instance decidableExistsLE [DecidablePred p] : DecidablePred fun n => ∃ m : Nat, m ≤ n ∧ p m :=
   fun n => decidable_of_iff (∃ m, m < n + 1 ∧ p m)
     (exists_congr fun _ => and_congr_left' Nat.lt_succ_iff)
+
+@[deprecated] protected alias lt_connex := Nat.lt_or_gt_of_ne
+@[deprecated] alias pow_two_pos := Nat.two_pow_pos -- deprecated 2024-02-09
```
fgdorais, line 1261: Not sure where these came from but I don't think the first one is necessary anymore. The second could stay for a bit since the comment makes it look recent but I have no idea where it came from.

digama0:
> To the extent there are changes; it's largely happening within the Lean repo as we adapt Lean to address inconsistencies between Lean and Std (e.g, soon Lean master will use `ediv`/`emod` for Int division).

This is my concern. Not so much the moving / import changes, but rather the little changes that happen after that in order to make the code "fit" inside the Lean package better. I would like those changes to be explicitly recorded and tracked in the relevant PRs, and/or avoided if at all possible, because as things stand I have no idea if this broke backward compatibility and what mitigations we need to apply because of it.

fgdorais:
I concur again with Mario. Moves should have a "move-point" after the basic move and before any adaptations. I don't think it is necessary for the "move-point" to compile, so long as the errors make sense and are fixed in adaptation PRs. This is complicated but I think this is necessary, I've moved several times in my life and it was always complicated: 😟 🚚 😟 🚚 😄

PR #654, opened 2024-02-16
chore: refactor Backtrack and SolveByElim to remove Nondet dependency

joehendrix:
This doesn't remove the import so an existing signature is unchanged, but it effectively replaces the Nondet and implicit MLList dependency in Backtrack and SolveByElim with an iterator.

The iterator has a narrower interface, but a simpler design.

This is a step towards upstreaming library search to Lean 4.

### DIFF
--- a/Std/Lean/Meta/Iterator.lean
+++ b/Std/Lean/Meta/Iterator.lean
@@ -32,7 +32,7 @@ def ofList (l : List α) : MetaM (Iterator α) := do
 Map and filter results of iterator and returning only those values returned
 by `f`.
 -/
-partial def filterMapM (f :  α → MetaM (Option β)) (L : Iterator α) : Iterator β :=
+partial def filterMapM (f : α → MetaM (Option β)) (L : Iterator α) : Iterator β :=
     { next := _next }
   where _next := do
     match ← L.next with
--- a/Std/Lean/Meta/Iterator.lean
+++ b/Std/Lean/Meta/Iterator.lean
@@ -33,7 +33,8 @@ Map and filter results of iterator and returning only those values returned
 by `f`.
 -/
 partial def filterMapM (f :  α → MetaM (Option β)) (L : Iterator α) : Iterator β :=
-  let rec next := do
+    { next := _next }
+  where _next := do
     match ← L.next with
     | none =>
       pure none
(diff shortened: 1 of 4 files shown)


Std/Lean/Meta/Iterator.lean
```diff
@@ -0,0 +1,68 @@
...
+      restoreState s
+      let r ← f v
+      match r with
+      | none =>
+        _next
+      | some r =>
+        pure <| some (r, ←saveState)
+
+/--
+Find the first alternative in a nondeterministic value, as a monadic value.
```
semorrison, line 51: Needs an update!

Std/Lean/Meta/Iterator.lean
```diff
@@ -0,0 +1,68 @@
...
+    | r :: l =>
+      ref.set l
+      pure <| some (r, ←saveState)
+  pure { next }
+
+/--
+Map and filter results of iterator and returning only those values returned
+by `f`.
+-/
+partial def filterMapM (f :  α → MetaM (Option β)) (L : Iterator α) : Iterator β :=
```
semorrison, line 35:
```suggestion
partial def filterMapM (f : α → MetaM (Option β)) (L : Iterator α) : Iterator β :=
```

semorrison:
We should test Mathlib before merging.

Std/Lean/Meta/Iterator.lean
```diff
@@ -0,0 +1,68 @@
...
+
+namespace Lean.Meta
+
+/--
+Provides an iterface for iterating over values that are defined in a particular
+meta state.
+-/
+structure Iterator (α : Type) where
+  /-- Function for getting next value and state pair. -/
+  next : MetaM (Option (α × Meta.SavedState))
```
alexkeizer, line 11: I imagine we might want to define a more general notion of `Iterator` at some point in the future. Should we call this one `MetaIterator`, or make it `protected`, to prevent future name-clash problems?

joehendrix: I agree.  I explored going the more general route directly to a polymorphic `BacktrackingIterator`, but it turns out to need several classes, and I think is a premature generalization at the moment.

So I just made this one protected.

joehendrix:
> We should test Mathlib before merging.

The only change needed was to import `Std.Control.Nondet.Basic` in `Mathlib.Tactic.LibrarySearch`.

PR #655, opened 2024-02-17
feat: `BitVec.toFin_not` lemma

alexkeizer:
Adds a lemma to rewrite `toFin (~~~x)`.

There is a slight inconsistency with `toNat_not`, which takes its argument implicitly.
I followed the other `toFin_` lemmas in making the argument explicit, but choose not to change `toNat_not` to keep this PR focussed.

Morally, the result of `toFin (~~~x)` should be `Fin.last _ - x.toFin`. However, this doesn't typecheck, as `Fin.last n` is of type `Fin (n + 1)`. Alternatives that do type-check are:
*  Inline the definition of `Fin.last` to get `⟨2^w - 1, Nat.sub_lt (Nat.two_pow_pos _)⟩`
* Use `Fin.cast`, as in `(Fin.last (2^w-1)).cast (_ : 2^w-1+1 = 2^w)`
* Define `Fin.last' (h : n \neq 0) : Fin n` (using one of the preceding options), and use that

I've chosen to go with the first option (inlining), since it seemed the most straightforward.

### DIFF
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -285,11 +285,10 @@ theorem not_def {x : BitVec v} : ~~~x = allOnes v ^^^ x := rfl
       · simp
 
 @[simp] theorem toFin_not (x : BitVec w) :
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
awaiting-review

PR #656, opened 2024-02-17
feat: allow writing (0 : Fin (2^n))

semorrison:


eric-wieser (commented):
For what it's worth, this already works with:
```lean
import Mathlib.Data.Fin.Basic

variable (n : ℕ)
#check (0 : Fin (2 ^ n))
```

PR #658, opened 2024-02-17
feat: simp-lemmas about `BitVec.cast`

alexkeizer:
Adds a few simple simp-lemmas about `BitVec.cast`.

Also changes the two existing lemmas (`toNat_cast` and `getLsb_cast`) to be more consistent. In particular, `getLsb_cast` now takes it's proof of width-equality explicitly, just like `toNat_cast` already does, and the variable names have been aligned.

### DIFF
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -128,10 +128,10 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
   rfl
 
 @[simp] theorem getLsb_cast (h : w = v) (x : BitVec w) : (cast h x).getLsb i = x.getLsb i := by
-  subst h ; simp
+  subst h; simp
 
 @[simp] theorem getMsb_cast (h : w = v) (x : BitVec w) : (cast h x).getMsb i = x.getMsb i := by
-  subst h ; simp
+  subst h; simp
 @[simp] theorem msb_cast (h : w = v) (x : BitVec w) : (cast h x).msb = x.msb := by
   simp [BitVec.msb]
 
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -122,21 +122,18 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
 
 /-! ### cast -/
 
-section
-variable (h : w = v) (x : BitVec w)
+@[simp] theorem toNat_cast (h : w = v) (x : BitVec w) : (cast h x).toNat = x.toNat := rfl
+@[simp] theorem toFin_cast (h : w = v) (x : BitVec w) :
+    (cast h x).toFin = x.toFin.cast (by rw [h]) :=
+  rfl
 
-@[simp] theorem toNat_cast : (cast h x).toNat = x.toNat := rfl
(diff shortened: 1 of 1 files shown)


Std/Data/BitVec/Lemmas.lean
```diff
@@ -122,11 +122,21 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
...
+variable (h : w = v) (x : BitVec w)
 
-@[simp] theorem getLsb_cast : (cast h v).getLsb i = v.getLsb i := by
-  cases h
-  simp
+@[simp] theorem toNat_cast : (cast h x).toNat = x.toNat := rfl
+@[simp] theorem toFin_cast : (cast h x).toFin = x.toFin.cast (by rw [h]) := rfl
+
+@[simp] theorem getLsb_cast : (cast h x).getLsb i = x.getLsb i := by
+  subst h ; simp
```
semorrison, line 132:
```suggestion
  subst h; simp
```
I think usually no space before `;`, but probably the style guide doesn't care. One day no one will care, because a code formatter will run automatically.

Std/Data/BitVec/Lemmas.lean
```diff
@@ -122,11 +122,21 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
...
+
+@[simp] theorem getMsb_cast : (cast h x).getMsb i = x.getMsb i := by
+  subst h ; simp
```
semorrison, line 135:
```suggestion
  subst h; simp
```

Std/Data/BitVec/Lemmas.lean
```diff
@@ -122,11 +122,21 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
 
 /-! ### cast -/
 
-@[simp] theorem toNat_cast (e : m = n) (x : BitVec m) : (cast e x).toNat = x.toNat := rfl
+section
+variable (h : w = v) (x : BitVec w)
```
semorrison, line 126: I would prefer *not* to use `variable` in a section in order to override explicitness. I know Mathlib does this, but it is hard to read and maintain. It's fine in the lemma statement to just write e.g. `theorem toNat_cast (h) : ...`

semorrison (approved):
Otherwise, LGTM.

Std/Data/BitVec/Lemmas.lean
alexkeizer: Are you sure? Other lemmas in this same file have it written with the space, which is what made me include it, too. For example:
```lean
@[simp] theorem ofNat_sub_ofNat {n} (x y : Nat) : x#n - y#n = .ofNat n (x + (2^n - y % 2^n)) := by
  apply eq_of_toNat_eq ; simp [BitVec.ofNat]

@[simp] protected theorem sub_zero (x : BitVec n) : x - (0#n) = x := by apply eq_of_toNat_eq ; simp
```

Should I change those as well, for the sake of consistency?

Std/Data/BitVec/Lemmas.lean
alexkeizer: Just `theorem toNat_cast (h) : ...` doesn't work, because Lean can't infer the second width of the equality. I just went with `theorem toNat_cast (h : w = v) (x : BitVec w) : ...` full-out everywhere

PR #660, opened 2024-02-18
feat: establish `BitVec.allOnes` as simp normal form

alexkeizer:
Adds a lemma `negOne_eq_allOnes`, ensuring that `(-1 : BitVec w)` and `-1#w` are both (transitively) simplified to `allOnes w`.

The choice of `allOnes` as the canonical form was inspired by the fact we already have a few simp-lemmas about it, such as `toNat_allOnes` but also `not_def`. The downside is that `allOnes` is itself defined as `-1`, so we can get a simp loop with `simp [allOnes]`.

### DIFF
--- a/Std/Data/BitVec/Lemmas.lean
+++ b/Std/Data/BitVec/Lemmas.lean
@@ -234,6 +234,9 @@ private theorem allOnes_def :
   else
     simp [h]
 
+@[simp] theorem negOne_eq_allOnes : -1#w = allOnes w :=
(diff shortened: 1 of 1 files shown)


alexkeizer:
awaiting-review

joehendrix:
I think this makes sense.  We may end up with a simproc that rewrites literal values and it'd be good to have this simp lemma available when deciding what to rewrite negative literals to.

PR #663, opened 2024-02-18
chore: introduce ofNatLit to reduce usage of Fin in bitvec proofs

joehendrix:
This PR is an effort to improve reasoning at the `Nat` level about bitvectors and reduce of Fin and Nat.

It slightly tightens some proofs, but is generally aimed at reducing inconsistencies between definitions at the `Nat` and `Fin` types in favor of more consistently using `Nat` operations.


tobiasgrosser:
Dear @joehendrix, it might be helpful to explain briefly in the commit message (and maybe even in code/) why reasoning using `Nat` and `ofNatLt` is preferable.

tobiasgrosser:
> chore: introduce ofNatLit to reduce usage of Fin in bitvec proofs #663

I guess you mean `ofNatLt` in the commit message.

joehendrix:
It's mostly about consistency.  If we use the same types and style in definitions and proofs, there is less surprise when unfolding or otherwise using definitions.  We use some `Nat` based operations that haven't been extended to `Fin` such as the bitwise operations, and I don't want to pay the overhead of introducing a `Fin` version of every Bitvector operation.  
 So this basically means `Nat` is preferred.

One argument potentially in favor of `Fin` is that we could reuse results proven there, but that doesn't really seem to be the case so far.

A second argument is that we want to simplify expression to use more canonical forms and we currently can pretty-print those operations better using `ofNat` than `ofFin`.  We could define the notations using `ofFin` of course though, but that's additional operators that will show up in expressions.

joehendrix:
I think it'd be good to have a writeup of normal forms and how to reason about Bitvectors, but I don't want to hold up PRs for that if there aren't objections to this code.

alexkeizer:
> One argument potentially in favor of `Fin` is that we could reuse results proven there, but that doesn't really seem to be the case so far.

I'd say this is mostly a symptom of a lacking `Fin` API in std. For now, we've proved most things directly in terms of `Nat`  because we don't have the equivalent `Fin`-result yet, and proving the `Fin` version first add overhead.
However, I would argue that some of these result are ones we should have for `Fin` at some point, and it would be nice if, when we do, the `BitVec` proof is becomes a trivial collorary.

For example, consider `add_not_self : x + ~~~x = allOnes w` (which currently we prove via the "bitblasted" `adc`). Morally, this is the same as
```lean
/-- Analogue of `Fin.last` given `n ≠ 0` -/
def Fin.last' (h : n ≠ 0) : Fin n :=
  ⟨n - 1, Nat.sub_lt (Nat.zero_lt_of_ne_zero h) Nat.zero_lt_one⟩

theorem add_rev_eq_last' (i : Fin n) : i + i.rev = Fin.last' _
```
Which I'd argue makes for an interesting statement about `Fin`s on its own merit

> A second argument is that we want to simplify expression to use more canonical forms and we currently can pretty-print those operations better using `ofNat` than `ofFin`. We could define the notations using `ofFin` of course though, but that's additional operators that will show up in expressions.

I agree a good, simple notation for creating a bitvector with known-bound is beneficial, but what is wrong with `ofFin`?
That is, what goes wrong if we define the notation as
```lean
macro_rules | `($i#'$p) => `(BitVec.ofFin <| Fin.mk $i $p)
```
With the unexpander:
```lean
/-- Unexpander for bit vector literals. -/
@[app_unexpander BitVec.ofFin] def unexpandBitVecOfNatLt : Lean.PrettyPrinter.Unexpander
  | `($_ { val := $i, isLt := $p }) => `($i#'$p)
  | s => throw ()
```

joehendrix:
I don't think coupling `BitVec` so closely to `Fin` is good practice.  `Fin` has some of the operations that `BitVec` has, but not all are natural (such as most-significant bit-based indexing), and it's primarily used in my experience in array indexing.

This PR is also showing that using `Fin` wasn't buying much in terms of simplifying the code.  One could conceptually hide the `Fin.mk` behind syntax (just as we hide `BitVec.ofNat` and `BitVec.ofNatLt`), but I don't see the point.  If you are specifying an ISA or something like LLVM, you probably dont need a close `Fin` connection.  Crypto-might benefit a bit more, but even then they may have custom types for GF(2^n) or prime fields rather than use `BitVec` or Fin directly.

There's also the whole class of SAT tactics that work very nicely with `BitVec`, but don't have naturally `Fin` analogues.

In short, I think the benefits of decoupling and simple user types outweigh the potential code savings.  The consistency with `Fin` could be addresses through tooling eventually, but I think is low priority.

Std/Data/BitVec/Basic.lean
```diff
@@ -46,6 +46,11 @@ this is truncation of the high bits when downcasting and zero-extension when upc
 protected def ofNat (n : Nat) (i : Nat) : BitVec n where
   toFin := Fin.ofNat' i (Nat.two_pow_pos n)
 
+/-- The `BitVec` with value `i mod 2^n`. Treated as an operation on bitvectors,
+this is truncation of the high bits when downcasting and zero-extension when upcasting. -/
+protected def ofNatLt {n : Nat} (i : Nat) (p : i < 2^n) : BitVec n where
+  toFin := ⟨i, p⟩
```
alexkeizer, line 52: The docstring is a bit of, the value is not `i mod 2 ^ n`, it's just `i` because we know that `i < 2^n`. There is no truncation involved.
```suggestion
/-- The `BitVec` with value `i`, given a proof that `i < 2^n`. -/
protected def ofNatLt {n : Nat} (i : Nat) (p : i < 2^n) : BitVec n where
  toFin := ⟨i, p⟩
```

Std/Data/BitVec/Basic.lean
```diff
@@ -387,9 +415,9 @@ def rotateRight (x : BitVec w) (n : Nat) : BitVec w := x >>> n ||| x <<< (w - n)
 A version of `zeroExtend` that requires a proof, but is a noop.
 -/
 def zeroExtend' {n w : Nat} (le : n ≤ w) (x : BitVec n)  : BitVec w :=
-  ⟨x.toNat, by
+  .ofNatLt (x.toNat) (by
     apply Nat.lt_of_lt_of_le x.isLt
-    exact Nat.pow_le_pow_of_le_right (by trivial) le⟩
+    exact Nat.pow_le_pow_of_le_right (by trivial) le)
```
alexkeizer, line 420: Should we use the notation here, rather than `ofNatLt`?

Std/Data/BitVec/Lemmas.lean
```diff
@@ -108,6 +108,10 @@ theorem ofBool_eq_iff_eq : ∀(b b' : Bool), BitVec.ofBool b = BitVec.ofBool b'
   getLsb (x#n) i = (i < n && x.testBit i) := by
   simp [getLsb, BitVec.ofNat, Fin.val_ofNat']
 
+@[simp] theorem getLsb_ofNatLt {n : Nat} (x : Nat) (lt : x < 2^n) (i : Nat) :
+  getLsb (.ofNatLt x lt) i = x.testBit i := by
+  simp [getLsb, BitVec.ofNatLt]
```
alexkeizer, line 113:
```suggestion
@[simp] theorem getLsb_ofNatLt {n : Nat} (x : Nat) (lt : x < 2^n) (i : Nat) :
    getLsb (.ofNatLt x lt) i = x.testBit i := by
  simp [getLsb, BitVec.ofNatLt]
```
fornatting

alexkeizer (commented):
Fair enough. You make a good point that the current situation, where some operations are defined in terms of `Fin` and others in terms of `Nat` is needlessly confusing.
The code duplication for the definitions is minimal, so I agree the benefit of consistency outweigh the cost.

When it comes to lemmas: so long as we have the equivalence between the bitvector operations and the `Fin` operations (i.e., the various `toFin_add`, `toFin_sub`, etc. lemmas), then it'll be easy enough to reuse results if we want to.

Should we add an `ofFin_eq_ofNatLt : ofFin i = ofNatLt i.val i.isLt` simp-lemma as well, to cement the fact that the `_#'_` notation is the preferred way to write bitvector values?

Std/Data/BitVec/Basic.lean
```diff
@@ -87,6 +92,18 @@ attribute [match_pattern] BitVec.ofNat
   | `($(_) $n $i) => `($i#$n)
   | _ => throw ()
 
+/-- Extended notation for bit vector literals. `i#'lt` is a shorthand for `BitVec.ofNatLt i lt`. -/
+scoped syntax:max term:max noWs "#'" noWs term:max : term
+macro_rules | `($i#'$p) => `(BitVec.ofNatLt $i $p)
+
+/- Support for `i#n` notation in patterns.  -/
+attribute [match_pattern] BitVec.ofNatLt
```
fgdorais, line 100: Shouldn't that be on the def above?

PR #669, opened 2024-02-23
docs: fix documentation build instructions

zdb999:
The README documentation build instructions are missing the `-R` flag that needs to be passed to `lake`. This generates configuration files that are read by [doc-gen4](https://github.com/leanprover/doc-gen4).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,10 @@ You can generate `std4`'s documentation with
 
 ```text
 # if you're generating documentation for the first time
-> lake -Kdoc=on update
+> lake -R -Kdoc=on update
 ...
 # actually generate the documentation
-> lake -Kdoc=on build Std:docs
+> lake -R -Kdoc=on build Std:docs
 ...
 > ls build/doc/index.html
 build/doc/index.html
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ You can generate `std4`'s documentation with
 # actually generate the documentation
 > lake -R -Kdoc=on build Std:docs
 ...
-> ls build/doc/index.html
+> open build/doc/index.html
 build/doc/index.html
 ```
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ You can generate `std4`'s documentation with
 # actually generate the documentation
 > lake -R -Kdoc=on build Std:docs
 ...
-> open build/doc/index.html
+> ls build/doc/index.html
 build/doc/index.html
 ```
 


zdb999:
awaiting-review

README.md
```diff
@@ -33,12 +33,12 @@ You can generate `std4`'s documentation with
...
 # if you're generating documentation for the first time
-> lake -Kdoc=on update
+> lake -R -Kdoc=on update
 ...
 # actually generate the documentation
-> lake -Kdoc=on build Std:docs
+> lake -R -Kdoc=on build Std:docs
 ...
-> ls build/doc/index.html
+> open build/doc/index.html
```
fgdorais, line 41: `ls` is right here given the expected output. I believe this is only to demonstrate the file exists. The `open` command doesn't work on many platforms.
```suggestion
> ls build/doc/index.html
```

README.md
```diff
@@ -33,12 +33,12 @@ You can generate `std4`'s documentation with
 
 ```text
 # if you're generating documentation for the first time
-> lake -Kdoc=on update
+> lake -R -Kdoc=on update
 ...
 # actually generate the documentation
-> lake -Kdoc=on build Std:docs
+> lake -R -Kdoc=on build Std:docs
```
fgdorais, line 39: Why is `-R` needed? Please include that in the PR description.

fgdorais (changes requested):
Please explain why `-R` is needed in the PR description.

README.md
zdb999: Ok, I'll revert this

README.md
zdb999: The `-R` flag generates configuration files that seems to be needed by [doc-gen4](https://github.com/leanprover/doc-gen4), as documented on the README. Without the flag, the docs didn't build for me. I also found [this interaction](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/how.20to.20generate.20html.20documentation.20for.20new.20lean.20file/near/419851869) on Zulip, which seems to suggest the `-R` flag is required. Adding it fixed my issue.

fgdorais (approved):

digama0:
I don't think we can recommend this without also reminding people to `lake build -R` afterward to clear the flag when they are done. It's not as simple as just putting `-R` on this command line.

zdb999:
Would the following addendum work @fgdorais ?

After generating the docs, run `lake build -R` to reset the configuration.

fgdorais:
Please add it to this PR. If necessary, we may recommend specific edits but mostly for polishing at this point.

joehendrix (approved):

PR #672, opened 2024-02-24
feat: add print structure command

fgdorais:
A handy `#print` command for structures.


fgdorais:
Still WIP but comments and suggestions welcome!

digama0:
This should go in core, replacing the existing result of `#print` for structures.

fgdorais:
@digama0 I finally got around to posting an RFC [lean4#3644](https://github.com/leanprover/lean4/issues/3644).

PR #673, opened 2024-02-25
chore: add test for simultaneous 'import Lean' and 'import Std'

semorrison:


.github/workflows/build.yml
```diff
@@ -67,4 +67,4 @@ jobs:
       - name: Don't 'import Lean', use precise imports
         if: always()
         run: |
-          ! (find . -name "*.lean" -type f -print0 | xargs -0 grep -E -n '^import Lean$')
+          ! (find . -name "*.lean" ! -path "./test/import_lean.lean" -type f -print0 | xargs -0 grep -E -n '^import Lean$')
```
digama0, line 70: Maybe this should just be restricted to the src (I mean `Std`) directory?

PR #687, opened 2024-03-07
feat: add Decidable instance for List.Forall₂

Seppel3210:


Seppel3210:
awaiting-review

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,23 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
+/--
+Decision procedure for `Forall₂ R l₁ l₂`
+-/
+def Forall₂.dec {R : α → β → Prop} [∀ a b, Decidable (R a b)] : ∀ l₁ l₂, Decidable (Forall₂ R l₁ l₂)
+  | [],    []    => isTrue .nil
+  | a::as, b::bs =>
+    if h₁ : R a b then
+      match dec as bs with
+      | isTrue h_all => isTrue (.cons h₁ h_all)
+      | isFalse h_all => isFalse fun (.cons _ h_all') => absurd h_all' h_all
```
digama0, line 835: Verify this is tail recursive. It doesn't look like it would be, unless the compiler is abnormally smart. An alternative strategy that will be tail recursive is to call a boolean function instead and prove that the function is true iff the Forall2 statement holds.

Seppel3210: I think optimizing C compilers and LLVM should be able to pull that off, but I implemented it anyways

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,40 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
 
 attribute [simp] Forall₂.nil
 
+/--
+Decision procedure for `Forall₂ R l₁ l₂`
+-/
+def Forall₂.dec (R : α → β → Prop) [∀ a b, Decidable (R a b)] : List α → List β → Bool
```
fgdorais, line 829: I think this would be more useful as a plain `Bool` function:
```lean
def all₂ (r : α → β → Bool) : List α → List β → Bool
```

eric-wieser: Do we need this function at all? I would guess you can just directly implement the Decidable instance without any intermediate declarations

fgdorais: Kind of. It's harder to make the decidable instance tail recursive.

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,40 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
+      simp [dec, h]
+      intro h'
+      have .cons head _ := h'
+      exact h head
+  | _::_, [] | [], _::_ => by
+    simp [dec]
+    intro h
+    nomatch h
+
+instance {R : α → β → Prop} [∀ a b, Decidable (R a b)] : ∀ l₁ l₂, Decidable (Forall₂ R l₁ l₂) :=
```
eric-wieser, line 857:
```suggestion
instance {R : α → β → Prop}  [DecidableRel R] : DecidableRel (Forall₂ R) :=
```

eric-wieser: Oh nevermind

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,40 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
+-/
+def Forall₂.dec (R : α → β → Prop) [∀ a b, Decidable (R a b)] : List α → List β → Bool
+  | [], [] => true
+  | a::as, b::bs =>
+    if R a b then
+      dec R as bs
+    else false
+  | _, _ => false
+
+theorem Forall₂.dec_iff {R : a → b → Prop} [∀ a b, Decidable (R a b)] :
```
eric-wieser, line 837:
```suggestion
theorem Forall₂.dec_iff {R : α → β → Prop} [∀ a b, Decidable (R a b)] :
```

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,40 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
+    ∀ l₁ l₂, Forall₂.dec R l₁ l₂ ↔ Forall₂ R l₁ l₂
+  | [], [] => by simp [dec]
+  | a::as, b::bs => by
+    if h : R a b then
+      simp [dec, h, dec_iff]
+      constructor <;> intro h'
+      exact Forall₂.cons h h'
+      have .cons _ tail := h'
+      exact tail
```
eric-wieser, line 846: I think it would make sense to upstream `forall₂_cons` from mathlib in this PR, which can be golfed to
```
theorem forall₂_cons {R : α → β → Prop} {a b as bs} :
    Forall₂ R (a :: as) (b :: bs) ↔ R a b ∧ Forall₂ R as bs :=
  ⟨fun | .cons h tail => ⟨h, tail⟩, fun ⟨head, tail⟩ => Forall₂.cons head tail⟩
```

Std/Data/List/Basic.lean
Seppel3210: doesn't work because a and b are different types

Std/Data/List/Basic.lean
Seppel3210: yea, I changed it to guarantee that it's tail recursive

Std/Data/List/Basic.lean
Seppel3210: oh, oops

Std/Data/List/Basic.lean
Seppel3210: should it maybe be called `Forall₂.cons_iff` or your name? Like, should it be under the `Forall₂` namespace or not

Seppel3210: I guess if that's already used I'll go with that

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,36 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
 
 attribute [simp] Forall₂.nil
 
+theorem forall₂_cons {R : α → β → Prop} {a b as bs} :
```
eric-wieser, line 826:
```suggestion
@[simp]
theorem forall₂_cons {R : α → β → Prop} {a b as bs} :
```
to match the version in Mathlib.

Perhaps it makes sense to upstream the entire [file](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/List/Forall2.html#List.forall%E2%82%82_cons) in a follow-up PR, to avoid accidentally going out of sync.

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,36 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
+theorem all₂_iff_forall₂ {R : α → β → Prop} [∀ a b, Decidable (R a b)] :
+    ∀ l₁ l₂, all₂ R l₁ l₂ ↔ Forall₂ R l₁ l₂
+  | [], [] => by simp [all₂]
+  | a::as, b::bs => by
+    by_cases h : R a b
+      <;> simp [all₂, h, all₂_iff_forall₂, forall₂_cons]
+  | _::_, [] | [], _::_ => by
+    simp [all₂]
+    intro h
+    nomatch h
```
eric-wieser, line 851: I think `exact nofun` would work too.

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,36 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
+def all₂ (R : α → β → Prop) [∀ a b, Decidable (R a b)] : List α → List β → Bool
+  | [], [] => true
+  | a::as, b::bs =>
+    if R a b then
+      all₂ R as bs
+    else false
+  | _, _ => false
+
+theorem all₂_iff_forall₂ {R : α → β → Prop} [∀ a b, Decidable (R a b)] :
+    ∀ l₁ l₂, all₂ R l₁ l₂ ↔ Forall₂ R l₁ l₂
```
eric-wieser, line 843:
```suggestion
@[simp]
theorem all₂_eq_true {R : α → β → Prop} [∀ a b, Decidable (R a b)] :
    ∀ l₁ l₂, all₂ R l₁ l₂ ↔ Forall₂ R l₁ l₂
```
to match `List.all_eq_true`

Std/Data/List/Basic.lean
Seppel3210: oh I didn't know that macro :+1:

Std/Data/List/Basic.lean
Seppel3210: yeah, makes more sense when viewing the generated docs :+1:

Std/Data/List/Basic.lean
```diff
@@ -823,6 +823,35 @@ inductive Forall₂ (R : α → β → Prop) : List α → List β → Prop
...
 
+@[simp] theorem forall₂_cons {R : α → β → Prop} {a b l₁ l₂} :
+    Forall₂ R (a :: l₁) (b :: l₂) ↔ R a b ∧ Forall₂ R l₁ l₂ :=
+  ⟨fun | .cons h tail => ⟨h, tail⟩, fun ⟨head, tail⟩ => .cons head tail⟩
+
+/--
+Check if all elements in the first List are `R` related to the element at the same position in the
+second List.
+-/
+def all₂ (R : α → β → Prop) [∀ a b, Decidable (R a b)] : List α → List β → Bool
```
fgdorais, line 834: I believe the suggestion was this:
```lean
def all₂ (r : α → β → Bool) : List α → List β → Bool
```
Please avoid amending reviewed commits, that makes it hard to track how review comments were addressed.

Seppel3210: Ah ok, I wasn't aware. And I guess I kinda misread the suggestion 😅

fgdorais (approved):

joehendrix (approved):

PR #688, opened 2024-03-07
feat: add more theorems about `Array.swap`'s basic properties

somombo:
Here is a description of each theorem being added. Given some array:
1. If you swap element at position i with that at j, then element at j in the new array (after swap) will be the element that was at i in the old array (before the swap)
2. If you swap element at position i with that at j, then element at i in the new array will be the element that was at j in the old array 
3. If you swap element at position i with that at j, then all elements that are neither at positions i nor j remain unchanged from what they were before the swap
4. If you swap element at position i with that at j, get a new array, then in this new array if you swap element at position i with that at j again, what you get is the original array you had before any of the swaps
5. If you swap element at position i with that at j, what you get is the same as if you instead had swapped element at position j with that at i

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -844,3 +844,59 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
 /-! ### erase -/
 
 @[simp] proof_wanted erase_data [BEq α] {l : Array α} {a : α} : (l.erase a).data = l.data.erase a
+
+/-! ### swap -/
+
+@[simp] theorem get_swap_right (a : Array α) {i j : Fin a.size} : (a.swap i j)[j.val] = a[i] :=
+  by simp only [swap, fin_cast_val, get_eq_getElem, getElem_set_eq, getElem_fin]
+
+@[simp] theorem get_swap_left (a : Array α) {i j : Fin a.size} : (a.swap i j)[i.val] = a[j] :=
+  if he : ((Array.size_set _ _ _).symm ▸ j).val = i.val then by
+    simp only [←he, fin_cast_val, get_swap_right, getElem_fin]
+  else by
+    apply Eq.trans
+    · apply Array.get_set_ne
+      · simp only [size_set, Fin.is_lt]
+      · assumption
+    · simp [get_set_ne]
+
+@[simp] theorem get_swap_of_ne (a : Array α) {i j : Fin a.size} (hp : p < a.size)
+    (hi : p ≠ i) (hj : p ≠ j) : (a.swap i j)[p]'(a.size_swap .. |>.symm ▸ hp) = a[p] := by
+  apply Eq.trans
+  · have : ((a.size_set i (a.get j)).symm ▸ j).val = j.val := by simp only [fin_cast_val]
+    apply Array.get_set_ne
+    · simp only [this]
+      apply Ne.symm
+      · assumption
+  · apply Array.get_set_ne
+    · apply Ne.symm
+      · assumption
+
+theorem get_swap (a : Array α) (i j : Fin a.size) (k : Nat) (hk: k < a.size) :
+    (a.swap i j)[k]'(by simp_all) = if k = i then a[j] else if k = j then a[i]  else a[k] := by
+  split
+  · simp_all only [get_swap_left]
+  · split <;> simp_all
+
+theorem get_swap' (a : Array α) (i j : Fin a.size) (k : Nat) (hk' : k < (a.swap i j).size) :
+    (a.swap i j)[k] = if k = i then a[j] else if k = j then a[i] else a[k]'(by simp_all) := by
+  apply get_swap
+
+@[simp] theorem swap_swap (a : Array α) {i j : Fin a.size} :
+    (a.swap i j).swap ⟨i.1, (a.size_swap ..).symm ▸i.2⟩ ⟨j.1, (a.size_swap ..).symm ▸j.2⟩ = a := by
+  apply ext
+  · simp only [size_swap]
+  · intros
+    simp only [get_swap']
+    split
+    · simp_all
+    · split <;> simp_all
+
+theorem swap_comm (a : Array α) {i j : Fin a.size} : a.swap i j = a.swap j i := by
+  apply ext
+  · simp only [size_swap]
+  · intros
+    simp only [get_swap']
+    split
+    · split <;> simp_all
+    · split <;> simp_all


somombo:
awaiting-review

somombo:
WIP

somombo:
awaiting-review

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
 /-! ### erase -/
 
 @[simp] proof_wanted erase_data [BEq α] {l : Array α} {a : α} : (l.erase a).data = l.data.erase a
+
+/-! ### swap -/
+
+@[simp] theorem swap_def_jeqi {a: Array α} {i j : Nat}
+    (_ : i < a.size := by trivial) (_ : j < a.size := by trivial)
```
semorrison, line 731: Why are you writing `by trivial` here? Doesn't seem like a useful autoparam.

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+    (_ : j < (a.swap ⟨i, ‹_›⟩ ⟨j, ‹_›⟩).size := by trivial) : (a.swap ⟨i, _⟩ ⟨j, _⟩)[j] = a[i] :=
```
semorrison, line 732: This shouldn't be a separate argument, it follows from the previous one.

semorrison: Lemma should be called `get_swap_right`, I think.

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+  suffices a'[i.val]'(by simp [a, a']) = vj from by simp [a'] at this; assumption
+  if he : j.val = i.val then
+    by simp [←he, vi, vj, j, a', fin_cast_val]
+  else
+    suffices a[i.val]'(by simp [a]) = vj from (a.get_set_ne j vi _ he).symm ▸ this
+    by simp [get_set_ne, a]
+
+@[simp] theorem swap_def_else {a: Array α} {i j : Fin a.size}
+    (_ : p < a.size := by trivial) (_ : p < (a.swap i j).size := by trivial)
+    (_ : i.val ≠ p) (_ : j.val ≠ p) : (a.swap i j)[p] = a[p] :=
```
semorrison, line 763: Maybe `p \ne i` is better?

Call this `get_swap_of_ne`?

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+  let i : Fin a.size := ⟨i, _⟩
```
semorrison, line 764: Does this line have any effect?

I haven't read your proofs, but they seem longer than necessary.

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+  have : j.val=j'.val := by simp [j', fin_cast_val]
+  have h1 : a'[p.val]'(by simp [a', a_]) = a_[p.val]'(by simp [a_]) :=
+    a_.get_set_ne (j') vi _ (this ▸ hj )
+  have h2 : a_[p.val]'(by simp [a_]) = a[p.val] := a.get_set_ne i vj _ hi
+
+  trans h1 h2
+
+
+theorem swap_p {a: Array α} {i j p : Fin a.size} :
+    (a.swap i j)[p] = a[if i.val=p.val then j else if j.val=p.val then i else p] :=
```
semorrison, line 788: You can just write `i = p`, no need for `i.val = p.val`.

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+
+  have : j.val=j'.val := by simp [j', fin_cast_val]
+  have h1 : a'[p.val]'(by simp [a', a_]) = a_[p.val]'(by simp [a_]) :=
+    a_.get_set_ne (j') vi _ (this ▸ hj )
+  have h2 : a_[p.val]'(by simp [a_]) = a[p.val] := a.get_set_ne i vj _ hi
+
+  trans h1 h2
+
+
+theorem swap_p {a: Array α} {i j p : Fin a.size} :
```
semorrison, line 787: Should be called `get_swap_eq_get_if`.

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+  else if hj : j.val=p.val then by
+    simp [hi]
+    simp [←hj]
+    apply swap_def_jeqi
+  else by
+    simp [hi]
+    simp [hj]
+    apply swap_def_else <;> assumption
+
+theorem swap_p' {a: Array α} {i j p : Fin a.size} :
```
semorrison, line 801: `get_swap_if`

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,182 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+  else if hj : j.val=p.val then by
+    simp [hi]
+    simp [←hj]
+    apply swap_def_ieqj
+  else by
+    simp [hi]
+    simp [hj]
+    apply swap_def_else <;> assumption
+
+@[simp] theorem swap_swap_id {a: Array α} {i j : Fin a.size} :
```
semorrison, line 815:
```suggestion
@[simp] theorem swap_swap {a: Array α} {i j : Fin a.size} :
```

somombo (commented):
I will change the names of the functions as suggested and try to shorten the proofs

Std/Data/Array/Lemmas.lean
somombo: @semorrison Could I have further guidance on how best to state this `get_swap_of_ne` theorem ? In particular, how should an end-user of this theorem furnish a proof that `p` is within bounds of `(a.swap i j)`:

#### Option 1 (as above):
as above (without the autoparam) user could explicitly pass it as an argument.
```lean
example {a: Array α} {i j : Fin a.size} 
    (hp : p < a.size) (hp' : p < (a.swap i j).size) (_ : p ≠ i) (_ : p ≠ j) : 
    (a.swap i j)[p] = a[p] := 
  sorry
```

this is one more thing to type at the "invocation" site


#### Option 2:
Alternatively, the proof could be automatically derived by simplification on `_ : p < a.size` :
```lean
example {a: Array α} {i j : Fin a.size} 
    (hp : p < a.size) (hi : p ≠ i) (hj : p ≠ j) : 
    (a.swap i j)[p]'(by simp_all) = a[p] := 
  sorry
```

#### Option 3
If the `get_elem_tactic_trivial` tactic already included such a simplification, then the theorem can be stated in the most ideal and concise manner.. 
```
section
local macro_rules | `(tactic| get_elem_tactic_trivial) => `(tactic| simp_all) 
example {a: Array α} {i j : Fin a.size} 
    (hp : p < a.size) (_ : p ≠ i) (_ : p ≠ j) : 
    (a.swap i j)[p] = a[p] := 
  sorry
end 
```
But this it begs the question of whether such an augmentation to that macro should be done globally (where the macro itself is originally defined).

somombo:
WIP

Std/Data/Array/Lemmas.lean
somombo: no problem. .. often times when I am using theorems like this these minor hypothesis can be distracting, so I made them option assumption they have already been proven somewhere before the theorem's use. This is just a personal preference so I will go ahead and remove these optional params

Std/Data/Array/Lemmas.lean
somombo: will provide `Fin` params instad of `Nat` for simplicity

Std/Data/Array/Lemmas.lean
somombo: sorry about this good catch, let me look over the proofs more carefully .. I wrote them a long time ago. Will try and shorten them.

Std/Data/Array/Lemmas.lean
somombo: @semorrison I have re-written the proofs of `swap_swap` and `swap_comm` from scratch to make them shorter and easier to follow. In particular they now nolonger rely on these two theorems `swap_p` and `swap_p'` ( aka.. `get_swap_eq_get_if`/`get_swap_if`) whose only purpose in my orignial draft was as prerequisite lemmas for `swap_swap` and `swap_comm`.. Since I now prove the essential theorems `swap_swap`/`swap_comm` without `swap_p` /`swap_p'` (and done so with a more concise shorter proof), I have simply removed  `swap_p` and `swap_p'` from the scope of this PR

somombo:
awaiting-review

somombo (commented):
I'm on the fence about whether `get_swap_right`, `get_swap_left` and `get_swap_of_ne` and be annotated `@[simp]`.. May be too aggressive for some users .. In my personal usage it has worked out just fine though

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,105 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+      (Array.get_set_ne _ _ _ _ he).symm ▸ (show _[i.val] = _ by simp [get_set_ne])
+
+@[simp] theorem get_swap_of_ne (a : Array α) {i j : Fin a.size} (hp : p < a.size)
+    (hi : p ≠ i) (hj : p ≠ j) : (a.swap i j)[p]'(a.size_swap .. |>.symm ▸ hp) = a[p] :=
+  let vi := a.get i
+  let vj := a.get j
+
+  let j' := a.size_set i vj ▸ j
+  let a_ := a.set i vj
+  let a' := a_.set j' vi
```
semorrison, line 749: These are rather unhelpful names.

Std/Data/Array/Lemmas.lean
```diff
@@ -724,3 +724,105 @@ theorem all_eq_true_iff_forall_mem {l : Array α} : l.all p ↔ ∀ x, x ∈ l 
...
+    ∀ p : Nat,
+      (_ : p < ((a.swap i j).swap (hss ▸ i) (hss ▸ j) ).size) →
+      (_ : p < a.size) →
+      ((a.swap i j).swap (hss ▸ i) (hss ▸ j))[p]  = a[p]
+  from Array.ext ((a.swap i j).swap (hss ▸ i) (hss ▸ j)) a hs.symm this
+
+  fun p _ hp =>
+    have hp' : p < (a.swap i j).size := hss ▸ hp
+
+    if hi : p=i.val then
```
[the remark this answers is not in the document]
semorrison: and elsewhere.

Std/Data/Array/Lemmas.lean
somombo: I find term based proofs easier to read, but no problem,  will replace with tactic based proofs which wont have any auxiliary variables such as these.

somombo:
WIP

somombo (commented):
I have converted all the added proofs to tactic based proofs to make it easier for you @semorrison. Let me know what you think.

Std/Data/Array/Lemmas.lean
```diff
@@ -844,3 +848,61 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
...
+  split
+  · simp_all
+  · split <;> simp_all
+
+theorem get_swap' (a : Array α) (i j : Fin a.size) (k : Nat) (hk' : k < (a.swap i j).size) :
+    (a.swap i j)[k] = if k = i then a[j] else if k = j then a[i] else a[k]'(by simp_all) := by
+  apply get_swap
+
+@[simp] theorem swap_swap (a : Array α) {i j : Fin a.size} :
+    (a.swap i j).swap ⟨i.1, (a.size_swap ..).symm ▸ i.2⟩ ⟨j.1, (a.size_swap ..).symm ▸ j.2⟩ = a := by
```
somombo, line 892: I use `⟨i.1, (a.size_swap ..).symm ▸ i.2⟩` instead of `Fin.cast  (a.size_swap ..).symm i` because the former in definitionally equivalent but I found that it makes my life easier in actual usage (I dont have an additional trivial theorem proving that they are equal if I an in a context that expects the former)

somombo:
awaiting-review

Std/Data/Array/Lemmas.lean
semorrison: No problem. If `simp` can still cope that I think this is fine.

Std/Data/Array/Lemmas.lean
```diff
@@ -51,6 +51,10 @@ theorem not_mem_nil (a : α) : ¬ a ∈ #[] := nofun
 
 /-- # get lemmas -/
 
+theorem ext_get {a b : Array α} (h1 : a.size = b.size)
+    (h2 : ∀ i : Fin a.size, a[i] = b[i]) : a = b :=
+  Array.ext a b h1 (fun i h _ => h2 ⟨i, h⟩)
+
```
semorrison, line 57: Can you just drop this and use `apply Array.ext` directly below?

Std/Data/Array/Lemmas.lean
```diff
@@ -844,3 +848,61 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
...
+      · simp
+      · assumption
+    · simp [get_set_ne]
+
+@[simp] theorem get_swap_of_ne (a : Array α) {i j : Fin a.size} (hp : p < a.size)
+    (hi : p ≠ i) (hj : p ≠ j) : (a.swap i j)[p]'(a.size_swap .. |>.symm ▸ hp) = a[p] := by
+  apply Eq.trans
+  · have : ((a.size_set i (a.get j)).symm ▸ j).val = j.val := by simp only [fin_cast_val]
+    apply Array.get_set_ne
+    · simp [this]
```
semorrison, line 874: Use `simp?` to turn this non-terminal simp into a `simp only`.

Std/Data/Array/Lemmas.lean
```diff
@@ -844,3 +848,61 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
...
+
+theorem get_swap' (a : Array α) (i j : Fin a.size) (k : Nat) (hk' : k < (a.swap i j).size) :
+    (a.swap i j)[k] = if k = i then a[j] else if k = j then a[i] else a[k]'(by simp_all) := by
+  apply get_swap
+
+@[simp] theorem swap_swap (a : Array α) {i j : Fin a.size} :
+    (a.swap i j).swap ⟨i.1, (a.size_swap ..).symm ▸ i.2⟩ ⟨j.1, (a.size_swap ..).symm ▸ j.2⟩ = a := by
+  apply ext_get
+  · intros
+    simp [get_swap']
```
semorrison, line 895: And here.

Std/Data/Array/Lemmas.lean
```diff
@@ -844,3 +848,61 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
...
+    split
+    · simp_all
+    · split <;> simp_all
+  · simp
+
+theorem swap_comm (a : Array α) {i j : Fin a.size} : a.swap i j = a.swap j i := by
+  apply ext_get
+  · intro k ; have _ := k.isLt
+    simp [get_swap']
```
semorrison, line 904: And here.

somombo:
WIP

Std/Data/Array/Lemmas.lean
somombo: bummer, i kind of liked how much simpler this one was stated. But I get that it appears there is no huge practical benefit

Std/Data/Array/Lemmas.lean
somombo: done

somombo:
@semorrison awaiting-review

somombo:
awaiting-review

Std/Data/Array/Lemmas.lean
```diff
@@ -844,3 +844,61 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
 /-! ### erase -/
 
 @[simp] proof_wanted erase_data [BEq α] {l : Array α} {a : α} : (l.erase a).data = l.data.erase a
+
+/-! ### swap -/
+
+@[simp] theorem get_swap_right (a : Array α) {i j : Fin a.size} :
+    (a.swap i j)[j.val]'(a.size_swap .. |>.symm ▸ j.isLt) = a[i] :=
```
semorrison, line 851:
```suggestion
    (a.swap i j)[j.val] = a[i] :=
```

Std/Data/Array/Lemmas.lean
```diff
@@ -844,3 +844,61 @@ instance [DecidableEq α] (a : α) (as : Array α) : Decidable (a ∈ as) :=
...
+  by simp only [swap, fin_cast_val, get_eq_getElem, getElem_set_eq, getElem_fin]
+
+@[simp] theorem get_swap_left (a : Array α) {i j : Fin a.size} :
+    (a.swap i j)[i.val]'(a.size_swap .. |>.symm ▸ i.isLt) = a[j] :=
```
semorrison, line 855:
```suggestion
    (a.swap i j)[i.val] = a[j] :=
```

Std/Data/Array/Lemmas.lean
semorrison: Unneeded.

semorrison, line 851: It seems unnatural to state this using `j.val` instead of just `j`.

@joehendrix or @digama0, do you know what the right approach is here? The author needs this version using `j.val` below, but the plain `j` version might also be useful. Is there a better alternative to duplicating the lemma?

somombo: oh this, sorry this is an artifact of having additionally tested against v4.6.1, forgot to remove it.

somombo: This statement now says:
```lean
@[simp] theorem get_swap_right (a : Array α) {i j : Fin a.size} : (a.swap i j)[j.val] = a[i] := sorry -- #mwe
```
(Removed `'(a.size_swap .. |>.symm ▸ j.isLt)` as @semorrison suggested)

somombo: Part of the question is whether this theorem (and `get_swap_left`) should even be annotated with `@[simp]`. To be clear, the only reason I opted it use `j.val` instead of  `j` (as was originally intended) is because when I have `j` and tag this as a `@[simp]`, everything builds fine but I get the following Std4 linting error (When I remove the `@[simp]` tag this linting error also disappears)

```
-- Found 1 error in 3922 declarations (plus 16441 automatically generated ones) in Std with 13 linters

/- The `simpNF` linter reports:
SOME SIMP LEMMAS ARE NOT IN SIMP-NORMAL FORM.
see note [simp-normal form] for tips how to debug this.
https://leanprover-community.github.io/mathlib_docs/notes.html#simp-normal%20form -/
-- Std.Data.Array.Lemmas
./././Std/Data/Array/Lemmas.lean:850:9: error: Array.get_swap_right.{u_1} Left-hand side simplifies from
  (Array.swap a i j)[j]
to
  (Array.swap a i j)[↑j]
using
  simp only [@getElem_fin, @Array.data_length, @Fin.is_lt]
Try to change the left-hand side to the simplified term!
```

somombo: It seems to me that the statement that has the `j.val` option is the most robust as is shown in the following #mwe:

```lean
import Std.Data.Array.Lemmas

@[simp] theorem Array.get_swap_right (a : Array α) {i j : Fin a.size} : (a.swap i j)[j.val] = a[i] := sorry -- if not already defined

example (a : Array α) (i j : Fin a.size) : sorry :=
  have _ : (a.swap i j)[j] = a[i] := by simp -- `j` version works
  have _ : (a.swap i j)[j] = a[i] := a.get_swap_right -- `j` version works

  sorry
```

semorrison: The comments both say `j` version works --- is that what you intended? That would give the opposite conclusion.

semorrison: The simpNF linter is usually right: it is telling you what the LHS should be, to avoid non-confluence.

somombo: perhaps a comment can be added that says that this theorem can also be used to prove `|- (a.swap i j)[j] = a[i]`

somombo:
lgtm

PR #701, opened 2024-03-19
chore: removed unnecessary array indexing proofs

semorrison:
Inference of array indexing proofs has improved. This PR removes those that no longer need to be given explicitly.

I *think* this is an improvement but would like a second opinion.

### DIFF
--- a/Std/Data/Array/Lemmas.lean
+++ b/Std/Data/Array/Lemmas.lean
@@ -10,8 +10,6 @@ import Std.Data.Array.Basic
 import Std.Tactic.SeqFocus
 import Std.Util.ProofWanted
 
-local macro_rules | `($x[$i]'$h) => `(getElem $x $i $h)
-
 @[simp] theorem getElem_fin [GetElem Cont Nat Elem Dom] (a : Cont) (i : Fin n) (h : Dom a i) :
     a[i] = a[i.1] := rfl
 
@@ -30,7 +28,7 @@ theorem getElem?_neg [GetElem Cont Idx Elem Dom]
 @[simp] theorem mkArray_data (n : Nat) (v : α) : (mkArray n v).data = List.replicate n v := rfl
 
 @[simp] theorem getElem_mkArray (n : Nat) (v : α) (h : i < (mkArray n v).size) :
-    (mkArray n v)[i]'h = v := by simp [Array.getElem_eq_data_get]
+    (mkArray n v)[i] = v := by simp [Array.getElem_eq_data_get]
 
 namespace Array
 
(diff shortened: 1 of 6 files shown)


digama0:
Hm, too bad we don't have !bench here. Did you notice any performance degradation with this?

Std/Data/Array/Lemmas.lean
```diff
@@ -108,7 +106,7 @@ theorem get?_push {a : Array α} : (a.push x)[i]? = if i = a.size then some x el
 @[simp] theorem data_set (a : Array α) (i v) : (a.set i v).data = a.data.set i.1 v := rfl
 
 theorem get_set_eq (a : Array α) (i : Fin a.size) (v : α) :
-    (a.set i v)[i.1]'(by simp [i.2]) = v := by
+    (a.set i v)[i.1] = v := by
   simp only [set, getElem_eq_data_get, List.get_set_eq]
```
joehendrix, line 110: It feels like `get_set_eq` should replace the `(i : Fin a.size) with `i : Nat` and a separate proof.  The current proof seems overly specific.

digama0: This may not work with rewriting if the argument to `set` is not a `Fin.mk` application.

semorrison:
> Hm, too bad we don't have !bench here. Did you notice any performance degradation with this?

Before:
```
% lake build && hyperfine "rm -f .lake/build/lib/Std/Data/Array/Lemmas.olean && lake build Std.Data.Array.Lemmas"
Benchmark 1: rm -f .lake/build/lib/Std/Data/Array/Lemmas.olean && lake build Std.Data.Array.Lemmas
  Time (mean ± σ):      3.555 s ±  0.017 s    [User: 3.305 s, System: 0.395 s]
  Range (min … max):    3.528 s …  3.582 s    10 runs
```
After
```
% lake build && hyperfine "rm -f .lake/build/lib/Std/Data/Array/Lemmas.olean && lake build Std.Data.Array.Lemmas"
Benchmark 1: rm -f .lake/build/lib/Std/Data/Array/Lemmas.olean && lake build Std.Data.Array.Lemmas
  Time (mean ± σ):      3.760 s ±  0.029 s    [User: 3.497 s, System: 0.398 s]
  Range (min … max):    3.712 s …  3.805 s    10 runs
```

So there is a measurable slow-down. I'm inclined to say it's worth it, still?

Std/Data/Array/Lemmas.lean
joehendrix: I don't follow but don't need to.  If these changes are useful to others, I am fine with merging them.

PR #702, opened 2024-03-19
chore: remove upstreamed `rfl` macro rules

nomeata:
they are already defined in
https://github.com/leanprover/lean4/blob/3ab1c23500658a1997b342fad15f499d7c482365/src/Init/Tactics.lean#L370
and
https://github.com/leanprover/lean4/blob/3ab1c23500658a1997b342fad15f499d7c482365/src/Init/Core.lean#L797

### DIFF
--- a/Std/Tactic/Init.lean
+++ b/Std/Tactic/Init.lean
@@ -29,13 +29,6 @@ syntax (name := failIfSuccessConv) "fail_if_success " Conv.convSeq : conv
 
 attribute [tactic failIfSuccessConv] evalFailIfSuccess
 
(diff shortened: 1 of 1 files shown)


nomeata:
I am using `nightly-testing` as a base here, no sure if this is most appropriate. Maybe one of the `bump` branches is more suitable? @semorrison will surely know :-)

nomeata:
Oh, these changes are already in lean 4.7.0-rc1!
https://github.com/leanprover/lean4/commit/c138801c3ab3f350b3278b91d65312134177a781

So this should go into `master`.

PR #704, opened 2024-03-20
fix: updates for #489 against `main`

semorrison:


Std/Data/Array/Lemmas.lean
```diff
@@ -524,6 +517,13 @@ theorem mapIdx_spec (as : Array α) (f : Fin as.size → α → β)
   unfold modify modifyM Id.run
   split <;> simp
 
+theorem get_modify {arr : Array α} {x i} (h : i < arr.size) :
```
digama0, line 520: what's this move about?

semorrison: This lemma needs `size_modify`, which used to be in `Init` but now is in this file, so I've moved `get_modify` immediately below `size_modify` (in the `modify` section).

PR #712, opened 2024-03-26
chore: remove triv tactic

semorrison:
This is not used in Mathlib; all uses can just become `trivial`. Let's remove the unnecessary duplication.

https://github.com/leanprover-community/mathlib4/pull/11679

### DIFF
--- a/Std/Tactic/Init.lean
+++ b/Std/Tactic/Init.lean
@@ -92,13 +92,6 @@ example (h : ∀ x : Nat, x = x → True) : True := by
 elab "eapply " e:term : tactic =>
   evalApplyLikeTactic (·.apply (cfg := {newGoals := .nonDependentOnly})) e
 
-/--
-Tries to solve the goal using a canonical proof of `True`, or the `rfl` tactic.
-Unlike `trivial` or `trivial'`, does not use the `contradiction` tactic.
--/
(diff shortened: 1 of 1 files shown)


Std/Tactic/Init.lean
```diff
@@ -92,13 +92,6 @@ example (h : ∀ x : Nat, x = x → True) : True := by
 elab "eapply " e:term : tactic =>
   evalApplyLikeTactic (·.apply (cfg := {newGoals := .nonDependentOnly})) e
 
-/--
-Tries to solve the goal using a canonical proof of `True`, or the `rfl` tactic.
-Unlike `trivial` or `trivial'`, does not use the `contradiction` tactic.
--/
-macro (name := triv) "triv" : tactic =>
```
digama0, line 99:
```suggestion
elab (name := triv) "triv" : tactic => throwError "`triv` has been removed; use `trivial` instead"
```

digama0: This is a fairly strong kind of deprecation, but it should be fine given that it doesn't have many existing uses. (Less strong approaches would be to make it do `trivial` anyway, give a warning, and/or come with a code action performing the replacement.) We should make it easier to do one-line try-this style deprecation for tactics.

digama0:
approved pending merge of the mathlib4 PR.

PR #714, opened 2024-03-26
chore: monadic implementation of classical

semorrison:
Refactor of `classical` so there is a `MetaM` / `TacticM` level interface.

(This I think will allow Mathlib's tauto to use this, instead of `classical!`, thereby removing the last use of `classical!`. It's a positive change regardless.)

### DIFF
--- a/Std/Tactic/Classical.lean
+++ b/Std/Tactic/Classical.lean
@@ -28,8 +28,8 @@ macro (name := classical!) "classical!" : tactic =>
 `classical t` runs `t` in a scope where `Classical.propDecidable` is a low priority
 local instance.
 -/
-def classical [Monad m] [MonadEnv m] [MonadFinally m] [MonadLiftT MetaM m] (t : m Unit) :
-    m Unit := do
+def classical [Monad m] [MonadEnv m] [MonadFinally m] [MonadLiftT MetaM m] (t : m α) :
+    m α := do
   modifyEnv Meta.instanceExtension.pushScope
   Meta.addInstance ``Classical.propDecidable .local 10
   try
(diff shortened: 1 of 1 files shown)


Std/Tactic/Classical.lean
```diff
@@ -24,6 +24,19 @@ Consider using `classical` instead if you want to use the decidable instance whe
 macro (name := classical!) "classical!" : tactic =>
   `(tactic| have em := Classical.propDecidable)
 
+/--
+`classical t` runs `t` in a scope where `Classical.propDecidable` is a low priority
+local instance.
+-/
+def classical [Monad m] [MonadEnv m] [MonadFinally m] [MonadLiftT MetaM m] (t : m Unit) :
+    m Unit := do
```
digama0, line 32:
```suggestion
def classical [Monad m] [MonadEnv m] [MonadFinally m] [MonadLiftT MetaM m] (t : m α) :
    m α := do
```

Std/Tactic/Classical.lean
```diff
@@ -24,6 +24,19 @@ Consider using `classical` instead if you want to use the decidable instance whe
...
   `(tactic| have em := Classical.propDecidable)
 
+/--
+`classical t` runs `t` in a scope where `Classical.propDecidable` is a low priority
+local instance.
+-/
+def classical [Monad m] [MonadEnv m] [MonadFinally m] [MonadLiftT MetaM m] (t : m α) :
+    m α := do
+  modifyEnv Meta.instanceExtension.pushScope
+  Meta.addInstance ``Classical.propDecidable .local 10
```
fgdorais, line 34: Maybe the priority should be an optional parameter of some kind?

semorrison: Let's not over-engineer.

fgdorais (approved):

PR #716, opened 2024-03-29
chore: use BEq instead of DecidableEq in List additional ops

joehendrix:
I've aimed to more consistently use BEq instead of DecidableEq where that is feasible and rely on a LawfulBEq for proofs where needed.

### DIFF
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -98,9 +98,7 @@ theorem elem_iff [BEq α] [LawfulBEq α] {a : α} {as : List α} :
     elem a as = true ↔ a ∈ as := ⟨mem_of_elem_eq_true, elem_eq_true_of_mem⟩
 
 @[simp] theorem elem_eq_mem [BEq α] [LawfulBEq α] (a : α) (as : List α) :
-    elem a as = decide (a ∈ as) := by
-    apply Bool.eq_iff_iff.mpr
-    simp only [decide_eq_true_eq, elem_iff]
+    elem a as = decide (a ∈ as) := by rw [Bool.eq_iff_iff, elem_iff, decide_eq_true_iff]
 
 theorem mem_of_ne_of_mem {a y : α} {l : List α} (h₁ : a ≠ y) (h₂ : a ∈ y :: l) : a ∈ l :=
   Or.elim (mem_cons.mp h₂) (absurd · h₁) (·)
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -619,7 +619,8 @@ theorem Sublist.eq_of_length_le (s : l₁ <+ l₂) (h : length l₂ ≤ length l
     | refl => apply Sublist.refl
     | step => simp [*, replicate, Sublist.cons]
 
-theorem isSublist_iff_sublist [BEq α] [LawfulBEq α] {l₁ l₂ : List α} : l₁.isSublist l₂ ↔ l₁ <+ l₂ := by
+theorem isSublist_iff_sublist [BEq α] [LawfulBEq α] {l₁ l₂ : List α} :
+    l₁.isSublist l₂ ↔ l₁ <+ l₂ := by
   cases l₁ <;> cases l₂ <;> simp [isSublist]
   case cons.cons hd₁ tl₁ hd₂ tl₂ =>
     if h_eq : hd₁ = hd₂ then
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -94,8 +94,13 @@ theorem append_of_mem {a : α} {l : List α} : a ∈ l → ∃ s t : List α, l 
   | .head l => ⟨[], l, rfl⟩
   | .tail b h => let ⟨s, t, h'⟩ := append_of_mem h; ⟨b::s, t, by rw [h', cons_append]⟩
 
-@[simp] theorem elem_iff [b : BEq α] [LawfulBEq α] {a : α} {as : List α} :
-    @elem _ b a as ↔ a ∈ as := ⟨mem_of_elem_eq_true, elem_eq_true_of_mem⟩
+theorem elem_iff [BEq α] [LawfulBEq α] {a : α} {as : List α} :
+    elem a as = true ↔ a ∈ as := ⟨mem_of_elem_eq_true, elem_eq_true_of_mem⟩
+
+@[simp] theorem elem_eq_mem [BEq α] [LawfulBEq α] (a : α) (as : List α) :
+    elem a as = decide (a ∈ as) := by
+    apply Bool.eq_iff_iff.mpr
+    simp only [decide_eq_true_eq, elem_iff]
 
 theorem mem_of_ne_of_mem {a y : α} {l : List α} (h₁ : a ≠ y) (h₂ : a ∈ y :: l) : a ∈ l :=
   Or.elim (mem_cons.mp h₂) (absurd · h₁) (·)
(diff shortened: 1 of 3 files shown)


Std/Data/List/Lemmas.lean
```diff
@@ -2002,25 +2008,25 @@ theorem foldr_hom (f : β₁ → β₂) (g₁ : α → β₁ → β₁) (g₂ :
...
     x ∈ l₁ ∪ l₂ ↔ x ∈ l₁ ∨ x ∈ l₂ := by induction l₁ <;> simp [*, or_assoc]
 
 end union
 
 /-! ### inter -/
 
-theorem inter_def [DecidableEq α] (l₁ l₂ : List α)  : l₁ ∩ l₂ = filter (· ∈ l₂) l₁ := rfl
+theorem inter_def [BEq α] (l₁ l₂ : List α)  : l₁ ∩ l₂ = filter (elem · l₂) l₁ := rfl
 
-@[simp] theorem mem_inter_iff {_ : DecidableEq α} {x : α} {l₁ l₂ : List α} :
```
fgdorais, line 2023: I'm wondering why the decidable instance was plain implicit instead of instance implicit. There's usually a good reason for that.

Std/Data/List/Lemmas.lean
```diff
@@ -2002,25 +2008,25 @@ theorem foldr_hom (f : β₁ → β₂) (g₁ : α → β₁ → β₁) (g₂ :
...
 
-theorem union_def [DecidableEq α] (l₁ l₂ : List α)  : l₁ ∪ l₂ = foldr .insert l₂ l₁ := rfl
+theorem union_def [BEq α] (l₁ l₂ : List α)  : l₁ ∪ l₂ = foldr .insert l₂ l₁ := rfl
 
 @[simp] theorem nil_union (l : List α) : nil ∪ l = l := by simp [List.union_def, foldr]
 
 @[simp] theorem cons_union (a : α) (l₁ l₂ : List α) :
     (a :: l₁) ∪ l₂ = (l₁ ∪ l₂).insert a := by simp [List.union_def, foldr]
 
-@[simp] theorem mem_union_iff {_ : DecidableEq α} {x : α} {l₁ l₂ : List α} :
```
fgdorais, line 2014: I'm wondering why the decidable instance was plain implicit instead of instance implicit. There's usually a good reason for that.

fgdorais (changes requested):
It's preferable to avoid `section` and `variable` in Std so that theorems show all their hypotheses. This seems like a good opportunity to cleanup the vestigial sections for `insert`, `union` and `diff` in Std.Data.List.Lemmas.

Std/Data/List/Perm.lean
```diff
@@ -476,12 +476,13 @@ theorem Perm.diff_right {l₁ l₂ : List α} (t : List α) (h : l₁ ~ l₂) :
...
-    · simp [elem_eq_true_of_mem (h.subset (mem_of_elem_eq_true hx))]
+    · simp only [Perm.mem_iff h] at hx
+      simp [hx]
       exact ih (h.erase _)
-    · have : ¬elem x l₂ = true := fun contra =>
-        hx <| elem_eq_true_of_mem <| h.symm.subset <| mem_of_elem_eq_true contra
-      simp [this]
+    · simp only [Perm.mem_iff h] at hx
+      simp [hx]
       exact ih h
```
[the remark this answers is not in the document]
joehendrix: I simplified this some, but I wanted to avoid much after `<;>` since it makes following proofs a bit harder in my opinion.

Std/Data/List/Perm.lean
```diff
@@ -504,9 +505,7 @@ theorem Subperm.diff_right {l₁ l₂ : List α} (h : l₁ <+~ l₂) (t : List 
   | nil => simp only [List.diff]; exact h
   | cons x t ih =>
     simp only [List.diff]; split <;> rename_i hx1
-    · have : elem x l₂ = true := by
-        apply elem_eq_true_of_mem
-        apply h.subset (mem_of_elem_eq_true hx1)
+    · have : x ∈ l₂ := h.subset (mem_of_elem_eq_true hx1)
```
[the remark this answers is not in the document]
joehendrix: Thanks.  Applied

Std/Data/List/Lemmas.lean
joehendrix: I can't be sure because there is no comment, but there was an issues with synthesizing inferred type classes that was addressed in [lean4#3507](https://github.com/leanprover/lean4/pull/3507).

Std/Data/List/Lemmas.lean
joehendrix: Same as comment below

joehendrix:
I'm not convinced migrating away from section and variable is an improvement in this context, and it feels orthogonal to the changes.

fgdorais:
It is orthogonal and I'm not going to insist. The idea is to make it clear at a glance what a theorem actually depends on: a theorem in a `variable [BEq α] [LawfulBEq α]` might only depend on `BEq` or even neither.

fgdorais (approved):

Std/Data/List/Lemmas.lean
```diff
@@ -2066,8 +2070,8 @@ theorem forIn_eq_bindList [Monad m] [LawfulMonad m]
 /-! ### diff -/
 
 section Diff
--- TODO: theorems about `BEq`
```
digama0, line 2069: This doesn't eliminate the TODO, which is about having theorems also for the lawless case. I suggest moving the `[LawfulBEq A]` assumption onto the theorems that need it. (And as François says, I think we should remove both variables and put them all on the statements.)

joehendrix: I think that change is orthogonal and the TODO seems like it should be an issue.  Addressing the case without `LawfulBEq` seems very low priority to me, and not something that needs an explicit TODO in the code.

digama0: "low priority issue" is synonymous with "TODO in the code" to me.

fgdorais: It would be nice to have more low-priority issues for people looking for ways to contribute. This is a good candidate.

joehendrix: Is it though?  What specific lemmas are desired here?  Do you want ones for relationships that satisfy `LawfulBEq.rfl` so are reflexive, but not equalities?

If one isn't even willing to precisely specify the problem, then maybe it's just not important.

fgdorais: I don't understand. Isn't specifying the problem the key part of filing an issue?

digama0: > If one isn't even willing to precisely specify the problem, then maybe it's just not important.

That's what "low priority" means. That does not mean you should remove TODOs just because they aren't important.

digama0: > Is it though? What specific lemmas are desired here? Do you want ones for relationships that satisfy `LawfulBEq.rfl` so are reflexive, but not equalities?

To answer this question directly: the TODO is saying "evaluate, for each below lemma, whether it holds under a weaker (preexisting) lawfulness assumption or none at all, and weaken the hypotheses appropriately".

joehendrix: I removed the TODO since I don't think it's worth further investigation and if it were, it would be better as issue so people could discuss.  Code shouldn't be littered with TODO comments just because one doesn't want to file an issue.

These lemmas also cover the core type of `List` and thus are good candidates for upstreaming.

semorrison:
Unfortunately there is some Mathlib fallout from this one.

PR #718, opened 2024-03-31
chore: golf proof

chabulhwi:


chabulhwi:
awaiting-review

Std/Data/List/Lemmas.lean
```diff
@@ -646,7 +646,7 @@ theorem head!_of_head? [Inhabited α] : ∀ {l : List α}, head? l = some a →
   | _a::_l, rfl => rfl
 
 theorem head?_eq_head : ∀ l h, @head? α l = some (head l h)
-  | [], h => nomatch h rfl
+  | [], h => nomatch h
   | _::_, _ => rfl
```
fgdorais, line 650: If I'm not mistaken, the first case is not necessary:
```suggestion
  | _::_, _ => rfl
```

chabulhwi: I applied this suggestion.

fgdorais (approved):

PR #725, opened 2024-04-08
doc: add examples to `Std.Data.HashMap.Init` docs

### LINKED ISSUE #469: Usage examples for HashMap.lean
Currently the HashMap docs have no code examples. We should have examples similar to those from `Std.Data.List.Basic`:
````lean
/--
Split a list at every occurrence of a separator element. The separators are not in the result.
```
[1, 1, 2, 3, 2, 4, 4].splitOn 2 = [[1, 1], [3], [4, 4]]
```
-/
@[inline] def splitOn [BEq α] (a : α) (as : List α) : List (List α) := as.splitOnP (· == a)
````

### PR DESCRIPTION
**Pull request by austinletson**
Create usage examples in `Std.Data.HashMap.Init` similar to `Std.Data.List.Basic`

Closes #469

### DIFF
--- a/Std/Data/HashMap/Basic.lean
+++ b/Std/Data/HashMap/Basic.lean
@@ -262,26 +262,52 @@ instance [BEq α] [Hashable α] : Inhabited (HashMap α β) where
 
 instance [BEq α] [Hashable α] : EmptyCollection (HashMap α β) := ⟨mkHashMap⟩
 
-/-- Make a new empty hash map. -/
+/--
+Make a new empty hash map
+```
+(empty : Std.HashMap Int Int).toList = []
+```
+-/
 @[inline] def empty [BEq α] [Hashable α] : HashMap α β := mkHashMap
 
 variable {_ : BEq α} {_ : Hashable α}
 
-/-- The number of elements in the hash map. -/
+/--
+The number of elements in the hash map.
+```
+(ofList [("one", 1), ("two", 2)]).size = 2
+```
+-/
 @[inline] def size (self : HashMap α β) : Nat := self.1.size
 
-/-- Is the map empty? -/
+/--
+Is the map empty?
+```
+(empty : Std.HashMap Int Int).isEmpty = true
+(ofList [("one", 1), ("two", 2)]).isEmpty = false
+```
+-/
 @[inline] def isEmpty (self : HashMap α β) : Bool := self.size = 0
 
 /--
 Inserts key-value pair `a, b` into the map.
 If an element equal to `a` is already in the map, it is replaced by `b`.
+```
+def hashMap := ofList [("one", 1), ("two", 2)]
+hashMap.insert "three" 3 = ofList [("one", 1), ("two", 2), ("three", 3)]
+hashMap.insert "two" 0 = ofList [("one", 1), ("two", 0)]
+```
 -/
 def insert (self : HashMap α β) (a : α) (b : β) : HashMap α β := ⟨self.1.insert a b, self.2.insert⟩
 
 /--
 Similar to `insert`, but also returns a boolean flag indicating whether an existing entry has been
 replaced with `a => b`.
+```
+def hashMap := ofList [("one", 1), ("two", 2)]
+(hashMap.insert' "three" 3).snd = false
+(hashMap.insert' "two" 0).snd = true
+```
 -/
 @[inline] def insert' (m : HashMap α β) (a : α) (b : β) : HashMap α β × Bool :=
   let old := m.size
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Std/Data/HashMap/Basic.lean [hunk 1]
```diff
@@ -344,13 +448,35 @@ instance : GetElem (HashMap α β) α (Option β) fun _ _ => True where
...
+-/
 @[inline] def forM [Monad m] (f : α → β → m PUnit) (self : HashMap α β) : m PUnit := self.1.forM f
 
-/-- Converts the map into a list of key-value pairs. -/
+/--
+Converts the map into a list of key-value pairs.
+```
+TODO: confirm this acceptable
+def hashMap := ofList [("one", 1), ("two", 2)]
+hashMap.toList = [("one", 1), ("two", 2)] ∨ hashMap.toList = [("two", 2), ("one", 1)]
```
austinletson, line 468: Is it acceptable to use `∨` to demonstrate the order of the list isn't deterministic?

Std/Data/HashMap/Basic.lean
```diff
@@ -291,51 +317,129 @@ replaced with `a => b`.
...
+hashMap.erase "three" = ofList [("one", 1), ("two", 2)]
+```
 -/
 @[inline] def erase (self : HashMap α β) (a : α) : HashMap α β := ⟨self.1.erase a, self.2.erase⟩
 
 /--
 Performs an in-place edit of the value, ensuring that the value is used linearly.
 The function `f` is passed the original key of the entry, along with the value in the map.
+```
+(ofList [("one", 1), ("two", 2)]).modify "one" (fun _ v => v + 1) = ofList [("one", 2), ("two", 2)]
```
austinletson, line 332: Given that there is no instance of `BEQ` for `HashMap` this an acceptable way to write this example? Would it be better to use `=>` instead of `=`?

Std/Data/HashMap/Basic.lean
digama0: How about just asserting that it's a permutation of `[("one", 1), ("two", 2)]`?

austinletson: That is much more clear. I have updated this example (and similarly the `toArray` example) to `(ofList [("one", 1), ("two", 2)]).toList ∈ List.permutations [("one", 1), ("two", 2)]`

austinletson:
awaiting-review

Std/Data/HashMap/Basic.lean
digama0: (maybe we'll make that actual syntax at some point, it would be nice to have a way to create concrete hashmaps that way)

I agree that it would be best if the LHS at least was actual code, because even if it's pseudocode now we still want to test it (occasionally these doctests reveal issues in either the doc or the function itself).

Std/Data/HashMap/Basic.lean
digama0: List.Perm has a syntax `~` which is available if you `open List`, so you could also write `(ofList [("one", 1), ("two", 2)]).toList ~ [("one", 1), ("two", 2)]`

austinletson: Ah, nice. Updated `toList` and `toArray` examples with the `~` syntax

austinletson:
Marked as ready for review as my initial questions were answered. Open to any general feedback

Std/Data/HashMap/Basic.lean
```diff
@@ -262,26 +262,52 @@ instance [BEq α] [Hashable α] : Inhabited (HashMap α β) where
...
 -/
 def insert (self : HashMap α β) (a : α) (b : β) : HashMap α β := ⟨self.1.insert a b, self.2.insert⟩
 
 /--
 Similar to `insert`, but also returns a boolean flag indicating whether an existing entry has been
 replaced with `a => b`.
+```
+def hashMap := ofList [("one", 1), ("two", 2)]
+(hashMap.insert' "three" 3).snd = false
+(hashMap.insert' "two" 0).snd = true
```
digama0, line 309:
```suggestion
hashMap.insert' "three" 3 = ({"one" ↦ 1, "two" ↦ 2, "three" ↦ 3}, false)
hashMap.insert' "two" 0 = ({"one" ↦ 1, "two" ↦ 0}, true)
```

Std/Data/HashMap/Basic.lean
```diff
@@ -291,51 +317,128 @@ replaced with `a => b`.
...
+hashMap.erase "three" = {"one" ↦ 1, "two" ↦ 2}
+```
 -/
 @[inline] def erase (self : HashMap α β) (a : α) : HashMap α β := ⟨self.1.erase a, self.2.erase⟩
 
 /--
 Performs an in-place edit of the value, ensuring that the value is used linearly.
 The function `f` is passed the original key of the entry, along with the value in the map.
+```
+(ofList [("one", 1), ("two", 2)]).modify "one" (fun _ v => v + 1) = {"one" ↦ 2, "two" ↦ 2}
```
digama0, line 332:
```suggestion
(ofList [("one", 1), ("two", 2)]).modify "one" (fun _ v => v + 1) = {"one" ↦ 2, "two" ↦ 2}
(ofList [("one", 1), ("two", 2)]).modify "three" (fun _ v => v + 1) = {"one" ↦ 1, "two" ↦ 2}
```

Std/Data/HashMap/Basic.lean
```diff
@@ -291,51 +317,128 @@ replaced with `a => b`.
...
 
-/-- Given a key `a`, returns a key-value pair in the map whose key compares equal to `a`. -/
+/--
+Given a key `a`, returns a key-value pair in the map whose key compares equal to `a`.
+```
+def hashMap := ofList [("one", 1), ("two", 2)]
+hashMap.findEntry? "one" = some ("one", 1)
+hashMap.findEntry? "three" = none
+```
+-/
```
digama0, line 345: There is a subtlety in this function that is only demonstrated if the key type has a BEq instance which is weaker than `=`: the returned key is not always the same as the input key, it is only `==` to it. Unfortunately I don't think there are any builtin types which have this behavior. Are there examples of this in the RBMap or List docs? I think the example might have been a pair where the equality is defined to only look at the first component.

austinletson: There is no example in [the doc for the `findEntry?` function for RBMap](https://leanprover-community.github.io/mathlib4_docs/Std/Data/RBMap/Basic.html#Std.RBMap.findEntry%3F) and I don't see a `findEntry?` for List. Would it be in another function?

There are some functions in `RBMap` like [this](https://leanprover-community.github.io/mathlib4_docs/Std/Data/RBMap/Basic.html#Std.RBMap.instBEqRBMap) one which mentions the compatibility of the ordering `cmp` and `=`.

Std/Data/HashMap/Basic.lean
```diff
@@ -262,26 +262,52 @@ instance [BEq α] [Hashable α] : Inhabited (HashMap α β) where
 
 instance [BEq α] [Hashable α] : EmptyCollection (HashMap α β) := ⟨mkHashMap⟩
 
-/-- Make a new empty hash map. -/
+/--
+Make a new empty hash map
```
digama0, line 266:
```suggestion
Make a new empty hash map.
```

Std/Data/HashMap/Basic.lean
austinletson: `↦` was causing the [Check for forbidden character ↦](https://github.com/leanprover/std4/blob/main/.github/workflows/build.yml#L41) CI step to fail.

Replaced `↦` with `=>` to align with #252

Std/Data/HashMap/Basic.lean
```diff
@@ -291,51 +317,128 @@ replaced with `a => b`.
...
+```
+(ofList [("one", 1), ("two", 2)]).modify "one" (fun _ v => v + 1) = {"one" ↦ 2, "two" ↦ 2}
+```
 -/
 def modify (self : HashMap α β) (a : α) (f : α → β → β) : HashMap α β :=
   ⟨self.1.modify a f, self.2.modify⟩
 
-/-- Given a key `a`, returns a key-value pair in the map whose key compares equal to `a`. -/
+/--
+Given a key `a`, returns a key-value pair in the map whose key compares equal to `a`.
```
digama0, line 340:
```suggestion
Given a key `a`, returns a key-value pair in the map whose key compares equal to `a`.
Note that the returned key may not be identical to the input, if `==` ignores some part
of the value.
```

Std/Data/HashMap/Basic.lean
digama0: I can't find the example I was thinking of. In the absence of good demonstrative examples we can just have a line of explanatory text.

PR #728, opened 2024-04-11
feat: `#print instances` command

digama0:
This implements the command `#print instances Foo`, which is similar to `#synth` but shows all matching instances instead of just the first one. (It also only performs one step of instance synthesis instead of iterating.) Instances are shown in priority order (from highest to lowest priority).


digama0:
duplicate of #148

Issue #565, opened 2024-01-26
Add docker / gitpod integrations


alexjbest:
Mathlib has `.docker` and `.devcontainer` folders that (for one thing) make it easy to quickly load in to a fresh install of lean + mathlib (eg with gitpod.io). I think adding these to Std would be helpful

semorrison:
PR welcome! :-)

Note that we're hoping to distribute Std directly with Lean itself (i.e. no `require std` required for projects that use it), but I think these integrations would remain just as useful even if we can pull that off.

alexjbest:
Ok, I'll make a PR when I get a chance!

semorrison:
Partially resolved by #732, but perhaps someone should still copy over Mathlib's `.devcontainer/*`, deleting the line about `lake exe cache get`.

fgdorais:
Anything left to do here?

PR #732, opened 2024-04-12
chore: add .gitpod.yml and .docker/gitpod/Dockerfile

austinletson:
Add docker files and .gitpod.yml to enable use of Gitpod with std4. Both of these are copied from mathlib with modifications for std4

I have opened std4 in Gitpod and tested,
- building std4
- running tests with make
- building and view docs

related to #565

### DIFF
--- a/.docker/gitpod/Dockerfile
+++ b/.docker/gitpod/Dockerfile
@@ -7,7 +7,7 @@ FROM ubuntu:jammy
 
 USER root
 
-RUN apt-get update && apt-get install sudo git curl git bash-completion python3 python3-requests gcc make -y && apt-get clean
+RUN apt-get update && apt-get install sudo git curl bash-completion python3-requests gcc make -y && apt-get clean
 
 RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod \
     # passwordless sudo for users in the 'sudo' group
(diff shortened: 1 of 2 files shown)


austinletson:
awaiting-review

.docker/gitpod/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# This is the Dockerfile for leanprover/std4
+# This file is mostly copied from [mathlib4](https://github.com/leanprover-community/mathlib4/blob/master/.docker/gitpod/Dockerfile)
+
+# gitpod doesn't support multiple FROM statements, (or rather, you can't copy from one to another)
+# so we just install everything in one go
+FROM ubuntu:jammy
+
+USER root
+
+RUN apt-get update && apt-get install sudo git curl git bash-completion python3 python3-requests gcc make -y && apt-get clean
```
semorrison, line 10: I don't think we should need any python3 anymore. (Ideally we'd also remove this on the mathlib side).

There's also a repeated `git` here.

Could you test without?

austinletson: I have removed `git` and `python3` and ran everything again successfully. 

I didn't remove `python3-requests`. We still need that right? If you remove it you can't run `python3 -m http.server` to expose the documentation.

austinletson:
awaiting-review

.docker/gitpod/Dockerfile
semorrison: I guess --- I've never done that. I guess it doesn't hurt too much having python bundled in.

austinletson: Alright, my thinking was that all the commands in the README.md should work out of the box.

I just launched Gitpod successfully from https://github.com/leanprover/std4 and built std4. Thanks!

PR #734, opened 2024-04-13
feat: `#show_unused` command

digama0:
`#show_unused decl1 decl2 ..` will highlight every theorem or definition in the current file not involved in the definition of declarations `decl1`, `decl2`, etc. The result is shown both in the message on  `#show_unused`, as well as on the declarations themselves.

This is requested periodically on Zulip. The most recent related mention I can find is [`#delete_unused_by`](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.23delete_unused_by/near/430912325) which is a slight variation which removes the theorems instead of highlighting them.

### DIFF
--- a/Std/Tactic/ShowUnused.lean
+++ b/Std/Tactic/ShowUnused.lean
@@ -12,7 +12,7 @@ import Std.Lean.Delaborator
 
 `#show_unused decl1 decl2 ..` will highlight every theorem or definition in the current file
 not involved in the definition of declarations `decl1`, `decl2`, etc. The result is shown
-both in the message on  `#show_unused`, as well as on the declarations themselves.
+both in the message on `#show_unused`, as well as on the declarations themselves.
 -/
 
 namespace Std.Tactic.ShowUnused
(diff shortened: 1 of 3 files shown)


Std/Tactic/ShowUnused.lean
```diff
@@ -0,0 +1,73 @@
...
+import Lean.Util.FoldConsts
+import Lean.Linter.UnusedVariables
+import Std.Lean.Delaborator
+
+/-!
+# The `#show_unused` command
+
+`#show_unused decl1 decl2 ..` will highlight every theorem or definition in the current file
+not involved in the definition of declarations `decl1`, `decl2`, etc. The result is shown
+both in the message on  `#show_unused`, as well as on the declarations themselves.
```
semorrison, line 15:
```suggestion
both in the message on `#show_unused`, as well as on the declarations themselves.
```

Std/Tactic/ShowUnused.lean
```diff
@@ -0,0 +1,73 @@
...
+    | some (ConstantInfo.quotInfo _)   => pure ()
+    | some (ConstantInfo.ctorInfo v)   => visitExpr v.type
+    | some (ConstantInfo.recInfo v)    => visitExpr v.type
+    | some (ConstantInfo.inductInfo v) => visitExpr v.type *> v.ctors.forM visit
+    | none                             => pure ()
+
+/--
+`#show_unused decl1 decl2 ..` will highlight every theorem or definition in the current file
+not involved in the definition of declarations `decl1`, `decl2`, etc. The result is shown
+both in the message on  `#show_unused`, as well as on the declarations themselves.
```
semorrison, line 40:
```suggestion
both in the message on `#show_unused`, as well as on the declarations themselves.
```

PR #739, opened 2024-04-13
feat: RBMap lemmas

digama0:
- [x] Depends on: #735
- [x] Depends on: #737
- [x] Depends on: #738

This adds RBMap lemmas corresponding to all of the RBSet lemmas.


Std/Data/RBMap/Lemmas.lean
```diff
@@ -933,3 +934,126 @@ theorem find?_insert [@TransCmp α cmp] (t : RBSet α cmp) (v v') :
...
+  le_lt_trans h₁ h₂ := TransCmp.lt_le_trans h₂ h₁
+  le_gt_trans h₁ := Decidable.not_imp_not.1 (TransCmp.le_trans · h₁)
+  exact h := (TransCmp.cmp_congr_left h).symm
+
+theorem findEntry?_some [@TransCmp α cmp] {t : RBMap α β cmp} :
+    t.findEntry? x = some y ↔ y ∈ toList t ∧ cmp x y.1 = .eq := RBSet.findP?_some
+
+theorem find?_some [@TransCmp α cmp] {t : RBMap α β cmp} :
+    t.find? x = some v ↔ ∃ y, (y, v) ∈ toList t ∧ cmp x y = .eq := by
+  simp [find?, findEntry?_some]; constructor
```
fgdorais, line 1010:
```suggestion
  simp only [find?, findEntry?_some, Option.map_eq_some']; constructor
```

Std/Data/RBMap/Lemmas.lean
```diff
@@ -6,6 +6,7 @@ Authors: Mario Carneiro
 import Std.Data.RBMap.Alter
 import Std.Data.Nat.Lemmas
 import Std.Data.List.Lemmas
+import Std.Tactic.Lint
```
fgdorais, line 9: This import is not actually used.

Std/Data/RBMap/Lemmas.lean
```diff
@@ -6,6 +6,7 @@ Authors: Mario Carneiro
 import Std.Data.RBMap.Alter
 import Std.Data.Nat.Lemmas
 import Std.Data.List.Lemmas
```
fgdorais, line 8:
```suggestion
import Batteries.Data.RBMap.Alter
import Batteries.Data.Nat.Lemmas
import Batteries.Data.List.Lemmas
```

fgdorais (approved):
Just a few trivial fixes. Otherwise looks good!

Std/Data/RBMap/Lemmas.lean
digama0:
```suggestion
```

PR #744, opened 2024-04-16
chore: move RBNode depth lemmas out

digama0:
These lemmas are separate from the rest and need not be pulled in by the main file.


semorrison:
Merged main.

PR #749, opened 2024-04-17
feat: string length lemmas

tjf801:
Figured these belonged in here, especially `length_append`, since `List.length_append` exists already

### DIFF
--- a/Std/Data/String/Lemmas.lean
+++ b/Std/Data/String/Lemmas.lean
@@ -33,6 +33,16 @@ theorem ext_iff {s₁ s₂ : String} : s₁ = s₂ ↔ s₁.data = s₂.data := 
   rw [push, mk_length, List.length_append, List.length_singleton, Nat.succ.injEq]
   rfl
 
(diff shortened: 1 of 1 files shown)


tjf801:
awaiting-review

Std/Data/String/Lemmas.lean
```diff
@@ -33,6 +33,16 @@ theorem ext_iff {s₁ s₂ : String} : s₁ = s₂ ↔ s₁.data = s₂.data :=
   rw [push, mk_length, List.length_append, List.length_singleton, Nat.succ.injEq]
   rfl
 
+@[simp] theorem length_pushn (c : Char) (n : Nat) : (pushn s c n).length = s.length + n := by
+  induction n with
+  | zero => simp [pushn, Nat.repeat]
+  | succ k h =>
+    simp [pushn, Nat.repeat, Nat.add_succ]
+    exact h
```
digama0, line 41:
```suggestion
  unfold pushn; induction n <;> simp [Nat.repeat, Nat.add_succ, *]
```

tjf801: Yeah that's way better, I wasn't really trying to golf that proof in the first place

PR #750, opened 2024-04-18
feat: `Rat.toFloat`

digama0:
This adds an implementation of exact (IEEE conforming) rounding from `Rat` to `Float` rounding nearest ties to even. Note that unlike most other lean functions (but like most float functions), division by zero results in `Float.inf` or `Float.nan`, not `0`.


test/float.lean
```diff
@@ -41,3 +41,19 @@ import Std.Lean.Float
 #guard 1e1000.toStringFull == "inf"
 #guard (-1e1000).toStringFull == "-inf"
 #guard (-0/0:Float).toStringFull == "NaN"
+
+#guard Nat.divFloat 1 0 == Float.inf
+#guard Nat.divFloat 50 0 == Float.inf
```
eric-wieser, line 46: Can you add some tests for `Int.divFloat` and `-Float.inf` too?

PR #751, opened 2024-04-19
chore: remove unneeded nolint

semorrison:


semorrison:
I guess it is needed, sad.

PR #755, opened 2024-04-19
refactor: move `Function.id_def` from mathlib

chabulhwi:
* Move the theorem from `Mathlib.Logic.Function.Basic` to Std.
* We need this theorem to prove `String.splitOn_of_valid`.


chabulhwi:
awaiting-review

digama0 (approved):

PR #763, opened 2024-04-23
feat: merging functions on List + mergeSort

digama0:
Complement to #762. Adds List version of all the definitions in `Data.Array.Merge`. Also adds `List.mergeSort` (upstreamed from mathlib), as a replacement for `Array.qsort` on lists (because `qsort` cannot easily be expressed as is).

### DIFF
--- a/Std/Data/List/Basic.lean
+++ b/Std/Data/List/Basic.lean
@@ -1620,13 +1620,124 @@ See `isSubperm_iff` for a characterization in terms of `List.Subperm`.
 def isSubperm [BEq α] (l₁ l₂ : List α) : Bool := ∀ x ∈ l₁, count x l₁ ≤ count x l₂
 
 /--
-`O(|l| + |r|)`. Merge two lists using `s` as a switch.
+`O(|xs| + |ys|)`. Merge lists `xs` and `ys`. If the lists are sorted according to `lt`, then the
+result is sorted as well. If two (or more) elements are equal according to `lt`, they are preserved.
 -/
-def merge (s : α → α → Bool) (l r : List α) : List α :=
-  loop l r []
+def merge (lt : α → α → Bool) : (xs ys : List α) → List α
+  | [], xs
+  | xs, [] => xs
+  | x :: xs, y :: ys =>
+    bif lt x y then x :: merge lt xs (y :: ys) else y :: merge lt (x :: xs) ys
+
+/-- Tail recursive version of `merge`. -/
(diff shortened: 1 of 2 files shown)


Std/Data/List/Basic.lean
```diff
@@ -1620,13 +1620,124 @@ See `isSubperm_iff` for a characterization in terms of `List.Subperm`.
...
+    · exact go _ (x::xs) ys
+  simp [mergeTR, go]
+
+/--
+`O(|xs| + |ys|)`. Merge lists `xs` and `ys`, which must be sorted according to `compare` and must
+not contain duplicates. Equal elements are merged using `merge`. If `merge` respects the order
+(i.e. for all `x`, `y`, `y'`, `z`, if `x < y < z` and `x < y' < z` then `x < merge y y' < z`)
+then the resulting list is again sorted.
+-/
+def mergeDedupWith [Ord α] (merge : α → α → α) : (xs ys : List α) → List α
```
fgdorais, line 1657: The `merge` parameter is a bit confusing. How about `dedup`, `join` or `meld`?

Std/Data/List/Basic.lean
```diff
@@ -1620,13 +1620,124 @@ See `isSubperm_iff` for a characterization in terms of `List.Subperm`.
...
+/--
+`O(|xs|)`. Deduplicate a sorted list. The list must be sorted with to an order which agrees with
+`==`, i.e. whenever `x == y` then `compare x y == .eq`.
+-/
+def dedupSorted [BEq α] (xs : List α) : List α :=
+  xs.mergeAdjacentDups fun x _ => x
+
+namespace MergeSort
+
+/-- `O(|l|)`. Split `l` into two lists of approximately equal length.
```
fgdorais, line 1706:
```suggestion
/-- `O(|l|)`. Split alternating elements of list `l` into two lists. The two lists will have equal length if `|l|` is even, otherwise the first list will be one element longer than the second.
```

fgdorais (changes requested):

Std/Data/List/Basic.lean
digama0: `combine`?

digama0: or just `f`

fgdorais: Yes, the issue is that `merge` clashes with a definition in this file. I'm happy with `f`, `combine` or whatever you decide.

PR #764, opened 2024-04-24
chore: adaptations for leanprover/lean4#3756

negiizhao:


negiizhao:
There are some other changes made to the latest Lean4. It seems difficult to test my PR.

digama0:
I think this has the wrong merge-base. There are currently a ton of bad commits in the PR.

negiizhao:
It's my first time making a PR like this and I'm not familiar with it. I will make a new one.

PR #767, opened 2024-04-27
feat(Nat/Gcd): drop an unneeded assumption

urkud:
Lemma `Nat.exists_coprime` does not need to assume `gcd m n > 0`.


urkud:
awaiting-review

PR #770, opened 2024-04-28
Add `String.length_join` and `List.length_join`

tjf801:


tjf801:
awaiting-review

Std/Data/List/Lemmas.lean
```diff
@@ -365,6 +365,9 @@ theorem exists_of_mem_join : a ∈ join L → ∃ l, l ∈ L ∧ a ∈ l := mem_
 
 theorem mem_join_of_mem (lL : l ∈ L) (al : a ∈ l) : a ∈ join L := mem_join.2 ⟨l, lL, al⟩
 
+theorem length_join (l : List (List α)) : (join l).length = Nat.sum (l.map length) := by
+  induction l <;> simp [*]
```
semorrison, line 369:
```suggestion
theorem length_join (l : List (List α)) : (join l).length = (l.map length).sum := by
  induction l <;> simp [*]
```

Std/Data/String/Lemmas.lean
```diff
@@ -499,6 +499,11 @@ theorem join_eq (ss : List String) : join ss = ⟨(ss.map data).join⟩ := go ss
 @[simp] theorem data_join (ss : List String) : (join ss).data = (ss.map data).join := by
   rw [join_eq]
 
+@[simp] theorem comp_length_data : List.length ∘ String.data = String.length := rfl
```
semorrison, line 502: I think this is more useful applied to an argument. Otherwise it is very unlikely to fire in practice.

Std/Data/String/Lemmas.lean
```diff
@@ -499,6 +499,11 @@ theorem join_eq (ss : List String) : join ss = ⟨(ss.map data).join⟩ := go ss
...
+
+theorem length_join (l : List String) : (String.join l).length = Nat.sum (l.map String.length) := by
```
semorrison, line 504:
```suggestion
theorem length_join (l : List String) : (String.join l).length = (l.map String.length).sum := by
```

semorrison:
Otherwise, great, thank you!

Std/Data/String/Lemmas.lean
tjf801: That's fair, I originally had it as a composition just so it can be used in `length_join` underneath, but I guess using `funext` is a pretty easy way to do that. Thanks!

tjf801:
awaiting-review

semorrison:
Oops, I merged this, but it caused `master` to fail. Very sorry, but would you reopen against current `main`, fix the problem, and then ping me?

tjf801:
Yup! No problem. Seems to be that somehow `Std.Data.Nat` got removed through a few indirect layers of imports since the code was tested, so I just explicitly added it back into `Std.Data.List.Lemmas`.

tjf801:
@semorrison opened the new PR under #779

PR #771, opened 2024-04-30
chore: add `Union` and `SDiff` instances for `Std.RBSet`

thelissimus:
There are `union` and `sdiff` function implementations for `Std.RBSet`. But there are no `Union` and `SDiff` class instances, which I find inconsistent.

Is there any reason behind this? I tried asking in Zulip ([thread](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Union.20and.20SDiff.20instances.20for.20Std.2ERBSet)). Got no answers so far, prohibiting having these instances. So, I am making this PR.


thelissimus:
awaiting-review

PR #772, opened 2024-04-30
chore: adaptations for nightly-2024-04-30

semorrison:
Trying to get all the bump branches up to date in anticipation of v4.8.0-rc1.


Std/Data/Array/Lemmas.lean
```diff
@@ -5,7 +5,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro, Gabriel Ebner
 -/
 import Std.Data.List.Lemmas
-import Std.Data.Array.Init.Lemmas
```
digama0, line 8: Maybe these files should be left in place but empty? It's annoying to have to recreate them and the header stuff every time.

test/congr.lean
```diff
@@ -33,19 +33,24 @@ example {α β : Type _} {F : _ → β} {f g : { f : α → β // f = f }}
   guard_target = type_of% h
   exact h
 
-private opaque List.sum : List Nat → Nat
+-- Adaptation note: the next two examples have always failed if `List.ext` was in scope,
+-- but until nightly-2024-04-24 (when `List.ext` was upstreamed), it wasn't in scope.
```
digama0, line 37: Is it possible to `attribute [-ext] List.ext` here? Alternatively, use a similar but different example.

Std/Data/Array/Lemmas.lean
semorrison: I'd prefer to hope to not need to recreate them? They don't seem inevitable!

test/congr.lean
semorrison: Done.

Std/Data/Array/Lemmas.lean
semorrison: I've now added them.

digama0 (approved):

PR #779, opened 2024-05-03
Add String.length_join and List.length_join

tjf801:
Quick fix for #770.

### DIFF
--- a/Std/Data/List/Lemmas.lean
+++ b/Std/Data/List/Lemmas.lean
@@ -5,6 +5,7 @@ Authors: Parikshit Khanna, Jeremy Avigad, Leonardo de Moura, Floris van Doorn, M
 -/
 import Std.Control.ForInStep.Lemmas
 import Std.Data.List.Basic
+import Std.Data.Nat
 import Std.Tactic.Init
 import Std.Tactic.Alias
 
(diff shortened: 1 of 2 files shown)


tjf801:
awaiting-review

Std/Data/List/Lemmas.lean
```diff
@@ -5,6 +5,7 @@ Authors: Parikshit Khanna, Jeremy Avigad, Leonardo de Moura, Floris van Doorn, M
 -/
 import Std.Control.ForInStep.Lemmas
 import Std.Data.List.Basic
+import Std.Data.Nat
```
digama0, line 8: This is a cross cutting import, and also not a precise import. The needed theorem(s) should move to `Std.Data.Nat.Init.Lemmas` and that should be imported here.

urkud:
Note that Mathlib has https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/BigOperators/List/Basic.html#List.length_join stated in terms of a generic `List.sum` instead of `Nat.sum`.

Std/Data/List/Lemmas.lean
semorrison: Is it possible to move the needed lemmas up into an `Init` folder, to reduce the imports needed here?

semorrison:
Given that mathlib already has a `List.length_join`, we'll need to have a PR for mathlib that uses this branch of Std ready to go before this can be merged.

PR #780, opened 2024-05-03
chore: rename to batteries

semorrison:
I think this thoroughly renames. It adds deprecations for everything in `Data`, leaves the `Std.Linter` and `Std.CodeAction` namespaces as they are, but does not attempt provide deprecations for `Tactic`.

I've changed URLs and repo references too, so when we are ready to merge this, the merge should be followed immediately by the transfer.


Batteries/Tactic/Basic.lean
```diff
@@ -0,0 +1,7 @@
+import Lean.Elab.Tactic.ElabTerm
+import Batteries.Linter
+import Batteries.Tactic.Init
+import Batteries.Tactic.SeqFocus
+import Batteries.Util.ProofWanted
+
+-- This is an import Batteries
```
digama0, line 7: is this copypasta?

Batteries/Tactic/Lint/Basic.lean
```diff
@@ -85,28 +86,29 @@ initialize stdLinterExt :
...
-Defines the `@[std_linter]` attribute for adding a linter to the default set.
-The form `@[std_linter disabled]` will not add the linter to the default set,
+Defines the `@[batteries_linter]` attribute for adding a linter to the default set.
+The form `@[batteries_linter disabled]` will not add the linter to the default set,
 but it will be shown by `#list_linters` and can be selected by the `#lint` command.
 
 Linters are named using their declaration names, without the namespace. These must be distinct.
 -/
-syntax (name := std_linter) "std_linter" &" disabled"? : attr
+syntax (name := batteries_linter) "batteries_linter" &" disabled"? : attr
```
digama0, line 95:
```suggestion
syntax (name := env_linter) "env_linter" &" disabled"? : attr
```
The name `std_linter` was never very good to begin with. We could do this separately but it would be nice if people didn't have to rename all the downstream linters twice...

test.lean
```diff
@@ -1,9 +1,9 @@
-import Std
+import Batteries
```
digama0, line 1: This file should not exist at all. It looks like a mistakenly pushed local test file

test/import_lean.lean
```diff
@@ -1,6 +1,6 @@
 import Lean
-import Std
+import Batteries
 
 /-!
-This file ensures that we can import all of `Lean` and `Std` without name conflicts.
+This file ensures that we can import Batteries` without name conflicts.
```
digama0, line 5:
```suggestion
This file ensures that we can import all of `Lean` and `Batteries` without name conflicts.
```

test/lint_empty.lean
```diff
@@ -16,7 +16,7 @@ def id (x : α) := x
 
 open Lean Parser Elab in
 /--
-Test that the linters imported from Std all work when run on a declaration in an empty environment.
+Test that the linters import Batteries all work when run on a declaration in an empty environment.
```
digama0, line 19:
```suggestion
Test that the linters imported from Batteries all work when run on a declaration in an empty environment.
```

Batteries/Tactic/Basic.lean
semorrison: Oh dear, it appears to be a sed gone bad. :-(

semorrison, line 7:
```suggestion
-- This is an import only file for common tactics used throughout Batteries
```

README.md
```diff
@@ -23,28 +23,28 @@ Additionally, please make sure that you're using the version of Lean that the cu
...
-* To build `std4` run `lake build`. To build and run all tests, run `make`.
-* If you added a new file, run the command `scripts/updateStd.sh` to update the
+* To build `batteries` run `lake build`. To build and run all tests, run `make`.
+* If you added a new file, run the command `scripts/updateBatteries.sh` to update the
   imports.
 
 # Documentation
 
-You can generate `std4`'s documentation with
+You can generate `batteries`'s documentation with
```
grunweg, line 32: Should this be "`batteries`'" instead? (This might also apply in other places.)

scripts/nolints.json
```diff
@@ -1,46 +1 @@
-[["docBlame", "Std.BitVec.reduceAbs"],
```
grunweg, line 1: I wouldn't have expected to see this changes: is this a bug or were these entries stale before?

grunweg (commented):
Two small questions, not critical.

scripts/nolints.json
digama0: the entries were stale. (I think they were all upstreamed before fixing though, and core doesn't track docBlame because it has a huge number of violators to begin with. :( )

README.md
semorrison: Thanks, looks like the only one.

digama0 (approved):

PR #781, opened 2024-05-04
chore(Data/List): drop `removeNth`

urkud:
Use `eraseIdx` instead.


urkud:
Should I add some deprecated aliases?
awaiting-review

digama0:
Yes, for all names that disappear you should have a deprecated alias for it

urkud:
If I use `@[deprecated] alias`, then the aliases will be about `eraseIdx`, not `removeNth`. Should I manually restate lemmas for `removeNth`?

digama0:
That's okay, the two will be defeq anyway and it's better to get the code action.

urkud:
Done

urkud:
awaiting-review

digama0 (approved):

PR #743, opened 2024-04-16
feat: add String.splitOn_of_valid

chabulhwi:
This is the third version of https://github.com/leanprover/std4/pull/495, and it also includes https://github.com/leanprover/std4/pull/531.

refactor: move `Function.id_def` from mathlib

* Move the theorem from `Mathlib.Logic.Function.Basic` to Std.

refactor: move theorems about lists from mathlib

* Move theorems about lists from `Mathlib.Data.List.Basic` to Std.
* I removed the `nolint` attribute from `List.modifyHead_modifyHead`.

refactor: move `List.cons_prefix_iff` from mathlib

* Move the theorem from `Mathlib.Data.List.Infix` to Std.

feat: add lemmas about lists

feat: add `Std.Data.List.SplitOnList`

* `List.splitOnceP` returns `(l₁, l₂)` for the first split `l = l₁ ++ l₂` such that `P l₂` returns true.
* `List.splitOnList` splits a list at every occurrence of a separator list. The separators are not in the result.
* `List.splitOnListAux` is an auxiliary definition for proving `String.splitOnAux_of_valid`.

Co-authored-by: Mario Carneiro <di.gama@gmail.com>
Co-authored-by: Kim Morrison <scott.morrison@gmail.com>

feat: add lemmas about `Nat.add`

feat: add `String.splitOn_of_valid`

* Add `String.splitOnAux_of_valid` and `String.splitOn_of_valid`.

---

When the latest Lean toolchain has the fixed versions of `String.splitOnAux` and `String.splitOn`, remove the section `DELETE_THIS_LATER` in `Std.Data.String.Lemmas` and replace `splitOnAux'` and `splitOn'` with `splitOnAux` and `splitOn`, respectively.

See https://github.com/leanprover/lean4/pull/3832.

PR #782, opened 2024-05-06
feat: Add some theorems for `String.substrEq` and `String.isPrefixOf`

tjf801:


tjf801:
awaiting-review

Std/Data/String/Lemmas.lean
```diff
@@ -60,6 +60,9 @@ instance : Std.BEqOrd String := .compareOfLessAndEq String.lt_irrefl
 
 @[simp] theorem data_append (s t : String) : (s ++ t).1 = s.1 ++ t.1 := rfl
 
+@[simp] theorem cons_append (c : Char) (cs : List Char) (t : String) :
```
semorrison, line 63: I'm skeptical this should be a `@[simp]` lemma.

Std/Data/String/Lemmas.lean
```diff
@@ -779,8 +826,66 @@ theorem mapAux_of_valid (f : Char → Char) : ∀ l r, mapAux f ⟨utf8Len l⟩
...
+  apply And.intro
+  · rfl
+  · have : stp.byteIdx - (off.byteIdx + csize (get s off)) < stp.byteIdx - off.byteIdx := by
+      simpa [Nat.sub_add_eq] using Nat.sub_lt (Nat.sub_pos_of_lt offb_lt_stpb) (csize_pos _)
+    exact substrEq_loop_self_append (valid_next h_off off_lt_end) h_stp
+termination_by stp.1 - off.1
+
+/-! ### isPrefixOf -/
+
+@[nolint unusedHavesSuffices, simp] -- false positive from unfolding substrEq.loop
```
semorrison, line 863:
```suggestion
-- unusedHavesSuffices lint false positive from unfolding substrEq.loop
-- This will be fixed by https://github.com/leanprover/lean4/pull/4143
@[nolint unusedHavesSuffices, simp] 
```

semorrison (approved):
Otherwise LGTM.

semorrison:
Please merge `main` and resolve the conflict.

tjf801:
awaiting-review

Std/Data/String/Lemmas.lean
semorrison: You don't seem to have removed this (or explained why you think it should stay).

Batteries/Data/String/Lemmas.lean
```diff
@@ -58,7 +58,8 @@ instance : Batteries.BEqOrd String := .compareOfLessAndEq String.lt_irrefl
 
 @[simp] theorem data_push (s : String) (c : Char) : (s.push c).1 = s.1 ++ [c] := rfl
 
-@[simp] theorem data_append (s t : String) : (s ++ t).1 = s.1 ++ t.1 := rfl
```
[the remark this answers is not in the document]
tjf801: Oh my god, I completely screwed that up. I meant to try to remove `cons_append` (since I oftentimes end up having lemmas that aren't necessary later on), and after seeing the tests pass, I just assumed that was all right, even though it was completely the wrong lemma... Big failure to read on my part, and just reverted that change

tjf801:
awaiting-review

chabulhwi:
@chabulhwi [said](https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/ten.20unproved.20string.20theorems/near/433490590):

> By the way, I proved that [`List.IsPrefix`](https://github.com/leanprover/std4/pull/743/files#diff-4236f31530e7cfa1ed2974fdb2f8c4d8fbac8cce221f2661b74913ffd6a86bffR2415-R2430) is equivalent to `List.isPrefixOf`, and [`List.IsSuffix`](https://github.com/leanprover/std4/pull/743/files#diff-4236f31530e7cfa1ed2974fdb2f8c4d8fbac8cce221f2661b74913ffd6a86bffR2432-R2448) is equivalent to `List.isSuffixOf`. It'll be nice if Lean can automatically generate lemmas for `isPrefixOf` and `isSuffixOf` whenever I prove those for `IsPrefix` and `IsSuffix`. Do you think this is a good idea? If so, how do we do that?

@digama0 [said](https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/ten.20unproved.20string.20theorems/near/433490838):

> the idea is to have all the lemmas about `IsPrefix`, and have a simp lemma saying `isPrefixOf l1 l2 <-> IsPrefix l1 l2`, and then you never have to worry about `isPrefixOf` again

I'm pretty sure `String.IsPrefix` and `String.isPrefixOf` are equivalent, as `List.IsPrefix` and `List.isPrefixOf` do. I suggest you replace your theorems about `isPrefixOf` with the `IsPrefix` counterparts.

tjf801:
I'm pretty sure that `String.IsPrefix` is in `Mathlib`, so all of that would require upstreaming that definition from there. (On a related note, if `String` and `List` eventually both have `IsPrefix` in `Batteries`, would it be a good idea to turn the `<+:` operator, which currently is an alias for `List.isPrefixOf` into a typeclass? But that's a question for another time.)

However, in the time that this PR has been opened, I used these theorems to actually prove the equivalence between `String.isPrefixOf` and `List.isPrefixOf`, and at least the way I did it, it was non-trivial and uses most (if not all?) of these theorems about `isPrefixOf` specifically as dependencies. At this point in time, it's probably better to just close this PR and do one with all of these changes. 

Also, I'm sorry! I didn't realize those two lines served a specific purpose. Won't make that mistake again 🫡

PR #784, opened 2024-05-07
refactor: structurally recursive List.ofFn

nomeata:
This used to be defined via `Array.ofFn`
but `Array.ofFn.go` is defined by well-founded recursion (slow to reduce)
and used `Array.push` (quadratic complexity on lists). Since mathlib relies on
reducing `List.ofFn`, use a structurally recursive definition here.

### DIFF
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -736,15 +736,28 @@ def sigmaTR {σ : α → Type _} (l₁ : List α) (l₂ : ∀ a, List (σ a)) : 
 ofFn f = [f 0, f 1, ... , f (n - 1)]
 ```
 -/
-def ofFn {n} (f : Fin n → α) : List α := go n (Nat.le_refl _) where
-  /-- Auxiliary for `List.ofFn`. -/
-  -- This used to be defined via `Array.ofFn`
-  -- but `Array.ofFn.go` is defined by well-founded recursion (slow to reduce)
-  -- and used `Array.push` (quadratic complexity on lists). Since mathlib relies on
-  -- reducing `List.ofFn`, use a structurally recursive definition here.
-  go : (i : Nat) → (h : i ≤ n) → List α
-  | 0, _ => []
-  | i+1, h => f ⟨n - (i+1), by omega⟩ :: go i (by omega)
+def ofFn {n} (f : Fin n → α) : List α := go n 0 rfl where
+  /-- Auxiliary for `List.ofFn`. `ofFn.go f i j _ = [f j, ..., f (n - 1)]`. -/
+  -- This used to be defined via `Array.ofFn` but mathlib relies on reducing it,
+  -- so we use a structurally recursive definition here.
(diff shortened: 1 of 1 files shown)


nomeata:
Adaption PR at https://github.com/leanprover-community/mathlib4/compare/joachim/batteries-pr-testing-784

nomeata:
Mathlib adoption branch is green.

Oddly, no perf benefit on mathlib. I guess unfolding wf recursive functions isn’t too bad after all.
http://speed.lean-fro.org/mathlib4/compare/9a73b242-225f-4397-bdb2-6dde26984f6e/to/e10aac6e-c174-493a-98b6-e9d1c4223e8a

But in light of the upcoming https://github.com/leanprover/lean4/pull/4061 probably still worth doing?

Batteries/Data/List/Basic.lean
```diff
@@ -736,7 +736,15 @@ def sigmaTR {σ : α → Type _} (l₁ : List α) (l₂ : ∀ a, List (σ a)) :
 ofFn f = [f 0, f 1, ... , f (n - 1)]
 ```
 -/
-def ofFn {n} (f : Fin n → α) : List α := (Array.ofFn f).data
+def ofFn {n} (f : Fin n → α) : List α := go n (Nat.le_refl _) where
+  /-- Auxillary for `List.ofFn`. -/
```
semorrison, line 740:
```suggestion
  /-- Auxiliary for `List.ofFn`. -/
```
I'm glad to discover that I'm not the only person who can not spell this word. :-)

Batteries/Data/List/Basic.lean
```diff
@@ -736,7 +736,15 @@ def sigmaTR {σ : α → Type _} (l₁ : List α) (l₂ : ∀ a, List (σ a)) :
...
+  -- This used to be defined via `Array.ofFn`
+  -- but `Array.ofFn.go` is defined by well-founded recursion (slow to reduce)
+  -- and used `Array.push` (quadratic complexity on lists). Since mathlib relies on
```
digama0, line 743: I don't know what is meant by this: `Array.push` is not quadratic complexity on arrays, and it's an array function.

digama0:
Pushed a variation which uses a structural recursive version for the defeqs and a tail-rec version for performance.

Batteries/Data/List/Basic.lean
nomeata: I assumes that the original definition, `List.ofFn = Array.ofFn` would in terms of definitional equality use the logical definition of the Array functions, which work on Arrays as if they are structures with a fields `.data` which is a list, and thus `Array.push` is actually `List.concat`. Am I mistaken here?

nomeata:
> Pushed a variation which uses a structural recursive version for the defeqs and a tail-rec version for performance.

Thanks! I wager that the tail-recursive variant could even be the only definition and we’d still get the defeqs that we had with the Array variant, but didn’t test this.

nomeata:
Can this be merged now? (I'd like to have this off my plate.)

PR #787, opened 2024-05-09
feat: use 'lake test'

semorrison:


lakefile.lean
```diff
@@ -20,3 +20,7 @@ lean_exe runLinter where
 
 meta if get_config? doc |>.isSome then
 require «doc-gen4» from git "https://github.com/leanprover/doc-gen4" @ "main"
+
+@[test_runner]
+lean_exe test where
+  srcDir := "scripts"
```
semorrison, line 26: I choose to do this with an executable, rather than a script, so that it is toml compatible. (And also, because I struggled to find examples of scripts: @nomeata has pointed me in the right direction since.)

urkud: What's the pointer? Does this config add `test` executable to any project that depends on `batteries`?

semorrison: It does in fact add `lake exe test` to everything downstream. That hadn't occurred to me, but maybe it is not a problem? I'm not sure if `lake test` works out of the box downstream however. I'll try!

semorrison: `lake exe test` is inherited in Mathlib (and appears to work fine!), but `lake test` is not inherited. I'm actually not sure it's possible to do so, pointing to the Batteries executable!

@tydeu?

austinletson: If `lake test` was inherited in Mathlib (or another downstream dependency  of `batteries`) would it conflict with a  `@[test_runner]` declaration in Mathlib's `lakefile.lean`?

tydeu: The executable is inherited (and that does not conflict with `@[test_runner]`). However, there is no way to use a upstream executable for a test runner. That seems like it could be a good feature, though.

GNUmakefile
```diff
@@ -1,22 +1,12 @@
...
-
-test: $(TESTS)
-
-%.run: build
-	lake env lean $(@:.run=.lean)
+test:
+	lake test
 
 lint: build
 	./.lake/build/bin/runLinter
```
tydeu, line 12: Separate from this PR, should this be using `lake exe runLinter`?

lakefile.lean
urkud: Does it mean that downstream can no longer call their test runner `test`, because two `lake exe test`s will conflict? If yes, then it should be called `stdTest`.

@semorrison said:
> (I don't understand the "What's the pointer?" part of your question.)

This was about

> (And also, because I struggled to find examples of scripts: @nomeata has pointed me in the right direction since.)

tydeu: @urkud No multiple packages can have executables with the same name, the root package's executable is preferred in cases of ambiguity. Uses can also manually disambiguate via e.g., `batteries/test`.

semorrison: > This was about
> 
> > (And also, because I struggled to find examples of scripts: @nomeata has pointed me in the right direction since.)

@urkud, unfortunately the example is in an FRO private repo!

GNUmakefile
```diff
@@ -1,22 +1,12 @@
...
 
 build:
 	lake build
 
-# Find all .lean files in test and subdirectories, replace .lean with .run
-TESTS := $(shell find test -type f -name '*.lean' | sed 's/\.lean$$/.run/')
-
-test: $(TESTS)
-
-%.run: build
```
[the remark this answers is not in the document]
semorrison: I've updated the PR to run lake build from the new test runner.

GNUmakefile
semorrison: #791

lakefile.lean
semorrison: > The executable is inherited (and that does not conflict with `@[test_runner]`). However, there is no way to use a upstream executable for a test runner. That seems like it could be a good feature, though.

https://github.com/leanprover/lean4/issues/4121

PR #791, opened 2024-05-10
chore: update `make lint` to use `lake`

semorrison:


GNUmakefile
```diff
@@ -19,4 +19,4 @@ test: $(TESTS)
 	lake env lean $(@:.run=.lean)
 
 lint: build
```
tydeu, line 21:
```suggestion
lint:
```

Lake will do the build. ;)

PR #794, opened 2024-05-12
chore: adaptations for nightly-2024-05-11

semorrison:


Batteries/Data/Fin/Lemmas.lean
```diff
@@ -54,34 +56,39 @@ theorem foldl_loop (f : α → Fin (n+1) → α) (x) (h : m < n+1) :
...
 theorem foldl_eq_foldl_list (f : α → Fin n → α) (x) : foldl n f x = (list n).foldl f x := by
   induction n generalizing x with
-  | zero => rfl
+  | zero => rw [foldl_zero, list_zero, List.foldl_nil]
   | succ n ih => rw [foldl_succ, ih, list_succ, List.foldl_cons, List.foldl_map]
 
 /-! ### foldr -/
 
-theorem foldr_loop_zero (f : Fin n → α → α) (x) : foldr.loop n f ⟨0, Nat.zero_le _⟩ x = x := rfl
+theorem foldr_loop_zero (f : Fin n → α → α) (x) : foldr.loop n f ⟨0, Nat.zero_le _⟩ x = x :=
```
digama0, line 71: Can these be `unseal` theorems? This will make them applicable by `dsimp` (although I'm not sure if this causes problems).

semorrison: I'm sorry, I don't understand the question.

digama0: my proposal is to do
```suggestion
unseal foldr.loop in
theorem foldr_loop_zero (f : Fin n → α → α) (x) : foldr.loop n f ⟨0, Nat.zero_le _⟩ x = x := rfl
```
thereby allowing `dsimp` to use the theorem

semorrison: @nomeata, I presume you'll be happy with this, so I'll go ahead and make this change. Please ping me if you disagree.

nomeata: I have no opinion here

PR #793, opened 2024-05-11
feat: statically sized Vector type

Shreyas4991:
This PR adds a static vector type to batteries akin to `Mathlib.Data.Vector` with a key difference: It is based on `Array` instead of `List` and is designed for efficient operations with while offering a static size . ~~The PR attempts to make the API of `Batteries.Data.Vector` isomorphic to `Mathlib.Data.Vector` to the maximum possible extent.~~

EDIT: This PR also extends all the Array API to Vector
For discussion see [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/RFC.3A.20Arrays.20with.20fixed.20length/near/438122610)

EDIT:  After consideration of the magnitude of the task, I limit this PR itself to porting non-monadic functions from Init.Data.Array.Basic and some surrounding API, as well as the extension theorem. I will create further PRs  for adapting monadic functions, relevant theorems, and elements of the Array API already present in Batteries and Init

Depends on
- [x] #801 
- [x] #877  : Later Vector.map_empty needs to be changed to use the lemma introduced there.

PR #800, opened 2024-05-16
Array changes corresponding to changes in Vector PR.

Shreyas4991:
This PR adds functions and theorems to the array API that were found necessary in #793 or useful from it.#793 

- [ ] Depends on PR #793

### DIFF
--- a/Batteries.lean
+++ b/Batteries.lean
@@ -37,6 +37,7 @@ import Batteries.Data.String
 import Batteries.Data.Sum
 import Batteries.Data.UInt
 import Batteries.Data.UnionFind
+import Batteries.Data.Vector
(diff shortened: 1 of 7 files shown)


fgdorais:
This PR doesn't actually depend on #793. The second commit is completely independent and the first commit shouldn't be there at all.

Shreyas4991:
awaiting-review

Shreyas4991:
> This PR doesn't actually depend on #793. The second commit is completely independent and the first commit shouldn't be there at all.

Yeah you're right. There is no dependency on 793 for these functions. How do I change this?

fgdorais:
Git cherry picking would be the trick but since there is no history here, might as well close this PR and make a new one.

PR #802, opened 2024-05-16
fix: unused variables

nomeata:
these were missed by the linter, which is probably a bug that
is fixed or made fire less often with
https://github.com/leanprover/lean4/pull/4192


nomeata:
awaiting-review

nomeata:
Merging this may mean less work with nightly-testing later :-)

PR #803, opened 2024-05-16
feat: safe Nat indexed Array functions using get_elem_tactic

Shreyas4991:
Adds safe Nat indexed variants of Array functions: `getN`, `setN`, `swapN`, `swapAtN`, and `eraseIdxN`.

### DIFF
--- a/Batteries/Data/Array/Basic.lean
+++ b/Batteries/Data/Array/Basic.lean
@@ -162,14 +162,6 @@ should prove the index bound.
 -/
 
 /--
-`getN a i` gets the `iᵗʰ` element of `a` if valid.
-By default, the proof of validity `h : i < a.size` is supplied by the `get_elem_tactic`.
-If this fails, a proof must be supplied at the function call.
--/
-abbrev getN (a : Array α) (i : Nat) (h : i < a.size := by get_elem_tactic) : α :=
-  Array.get a ⟨i, h⟩
-
-/--
 `setN a i h x` sets an element in a vector using a Nat index which is provably valid.
 A proof by `get_elem_tactic` is provided as a default argument for `h`.
 This will perform the update destructively provided that `a` has a reference count of 1 when called.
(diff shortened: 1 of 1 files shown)


Shreyas4991:
awaiting-review

Batteries/Data/Array/Basic.lean
```diff
@@ -153,6 +153,58 @@ Unsafe implementation of `attachWith`, taking advantage of the fact that the rep
...
+should prove the index bound.
+-/
+
+/--
+`getN a i` gets the `iᵗʰ` element of `a` if valid.
+By default, the proof of validity `h : i < a.size` is supplied by the `get_elem_tactic`.
+If this fails, a proof must be supplied at the function call.
+-/
+abbrev getN (a : Array α) (i : Nat) (h : i < a.size := by get_elem_tactic) : α :=
+  Array.get a ⟨i, h⟩
```
dupuisf, line 170: For this particular one, is there a scenario where one would use `as.getN i` instead of just `as[i]`?

Shreyas4991: It's to be sure that `h` is being used and not some variable potentially captured from the environment (variable declarations).

dupuisf: I don't understand: if `h` is generated by `get_elem_tactic` then it's the same thing as `as[i]`, and if it's provided explicitly, then it's the same as `as[i]'h`, no?

Shreyas4991: How does `a[i]` generate a proof of validity?

Shreyas4991: Sorry edited that one and lost the old message. But yeah crashes happen in both `as[i]` and `as.getN i` for invalid indices. Only `as.get` seems safe.

Shreyas4991: Should this be deleted?

fgdorais: I think so. Sometimes it makes sense to have a non-notation-based version for special cases but this doesn't strike me as one.

Shreyas4991: Done.

fgdorais: Same idea applies to #793, I think.

Shreyas4991: Done.

dupuisf (approved):

fgdorais (approved):

PR #805, opened 2024-05-21
chore: update `build.yml` to use `lean-action`

austinletson:
Replace the `install elan`,  `build batteries`, `test batteries`, and `lint batteries` steps of the Build job with `lean-action`.


austinletson:
@semorrison what should we verify here before merging?

Based on the build output, `lean-action` seems to be doing the same steps as the step-by-step ci. I plan to take a closer look tomorrow morning with fresh eyes.

Additionally, I am happy to perform any `batteries` specific testing on this PR. What do you think?

austinletson:
[Here](https://github.com/leanprover-community/batteries/actions/runs/9190636187/job/25275288278?pr=805) is an example of a build failure.

austinletson:
[Here](https://github.com/leanprover-community/batteries/actions/runs/9190708090/job/25275529374?pr=805) is an example of a test failure.

.github/workflows/build.yml
```diff
@@ -14,26 +14,13 @@ jobs:
...
-        run: lake build -Kwerror
-
-      - name: test batteries
-        if: steps.build.outcome == 'success'
-        run: lake test
-
-      - name: lint batteries
-        if: steps.build.outcome == 'success'
-        run: make lint
+      - id: lean-action
```
semorrison, line 19:
```suggestion
      - id: lean-action
      - name: build, test, and lint batteries
```

PR #807, opened 2024-05-23
feat: size lemma for `Array.set!`

fgdorais:


Batteries/Data/Array/Lemmas.lean
```diff
@@ -104,6 +104,9 @@ theorem get?_push {a : Array α} : (a.push x)[i]? = if i = a.size then some x el
 
 @[simp] theorem data_set (a : Array α) (i v) : (a.set i v).data = a.data.set i.1 v := rfl
 
+theorem size_set! (a : Array α) (i v) : (a.set! i v).size = a.size := by
```
eric-wieser, line 107:
```suggestion
@[simp] theorem size_set! (a : Array α) (i v) : (a.set! i v).size = a.size := by
```
to match `size_setD`

eric-wieser (approved):

Batteries/Data/Array/Lemmas.lean
fgdorais: The simp linter whines about this one.

fgdorais:
@eric-wieser I just realized there is an ambiguity in your last comment. I read it to mean that the `simp` was not necessary but you might have meant that the lemma wasn't necessary?

I do have use cases where this lemma is useful but simp is more awkward, often in `'(...)` arguments to `getElem`. For example, I would like to use `(xs.set! ...)[i]'(xs.size_get! .. ▸ ...)` rather than a `by simp` proof in some cases.

PR #809, opened 2024-05-23
feat: Add `String.isPrefixOf` theorems

tjf801:
This PR is a redo of #782, but adds the following:

- Auxiliary theorems used to prove the theorems below
  - These were already in #782
- Proof of equivalence between `String.isPrefixOf s t` and `∃ t, s = p ++ t`
  - (This is analogous to how `List.IsPrefix` is defined.)
- Proof of equivalence between `String.isPrefixOf s t` and `List.isPrefixOf s.data t.data`


tjf801:
awaiting-review

PR #813, opened 2024-05-27
feat: dependent array type

fgdorais:
Implementation of dependent array type. [Discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/RFC.3A.20.60Object.60.20and.20.60ObjectArray.60.20types/near/440896963)


Batteries/Data/DArray/Basic.lean
```diff
@@ -0,0 +1,157 @@
...
+
+Lean's array API does not directly support dependent arrays. Each `DArray n α` is internally stored
+as an `Array NonScalar` with length `n`. This is sound since Lean's array implementation does not
+record nor use the type of the items stored in the array. So it is safe to use `UnsafeCast` to
+convert array items to the appropriate type when necessary.
+-/
+
+/-- `DArray` is a heterogenous array where the type of each item depends on the index. -/
+-- TODO: Use a structure once [lean4#2292](https://github.com/leanprover/lean4/pull/2292) is fixed.
+inductive DArray (n) (α : Fin n → Type _) where
```
eric-wieser, line 30: Could you use a structure with a `dummy : Unit` field to avoid the issue?

fgdorais: Long story short: no, nothing gets around the builtin shortcut other than the hard coded escape list at the moment.

PR #828, opened 2024-06-07
feat: strong recursor for lists

llllvvuu:
Differs from strong induction on length in that the inductive hypothesis is weakened to only need be true for suffixes. `caseStrongRecOn` is modeled after `caseStrongInductionOn` for `Nat`, `PNat`, `Finset`, `Multiset`.

Includes minimal API for `IsStrictSuffix` to derive it for e.g. recursion on `m :: n :: l`.


llllvvuu:
Actually, it turns out that this doesn't introduce the `IsStrictSuffix` hypothesis into the context at the inductive step. Oops!

PR #829, opened 2024-06-07
feat: `measureRec` for strong induction use cases

llllvvuu:
This is from @nomeata's Zulip comment: https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Contributing.20a.20strong.20induction.20for.20lists/near/442796959. It can subsume, for example, mathlib's `Finset.strongInductionOn` and `Multiset.strongInductionOn`.

In situations where it doesn't make sense to extract a goal into a top-level `theorem`/`def`, `induction _ using measureRec` will be more convenient than `let rec` -> `termination_by` -> `decreasing_by` -> `exact`.

### DIFF
--- a/Batteries/Data/Nat/Basic.lean
+++ b/Batteries/Data/Nat/Basic.lean
@@ -29,6 +29,15 @@ protected def strongRecOn (t : Nat) {motive : Nat → Sort _}
   (ind : ∀ n, (∀ m, m < n → motive m) → motive n) : motive t := Nat.strongRec ind t
 
 /--
+  Strong recursor via a `Nat`-valued measure
+-/
+@[elab_as_elim]
+def strongRecMeasure (f : α → Nat) {motive : α → Sort _}
+    (ind : ∀ x, (∀ y, f y < f x → motive y) → motive x) (x : α) : motive x :=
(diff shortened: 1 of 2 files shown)


fgdorais:
This is neat! I wonder though if this belongs in `Batteries.Data.Nat.Basic` or in `Batteries.WF`. Both options appear to have equal pros and cons to me.

llllvvuu:
> This is neat! I wonder though if this belongs in `Batteries.Data.Nat.Basic` or in `Batteries.WF`. Both options appear to have equal pros and cons to me.

Interesting idea! I just made a commit experimenting with `Batteries.WF`; if that turns out to be no good I can switch it back.

fgdorais (approved):
I'm approving and I will approve too if this needs to go back into Nat. I really don't have a very strong opinion either way other than both options should be considered. As far as I can tell, both options have equivalent pros and cons.

digama0:
I think the Nat version makes more sense.

llllvvuu:
> I think the Nat version makes more sense.

sure, changed it bakc.

llllvvuu:
awaiting-review

PR #832, opened 2024-06-09
refactor: avoid relying on rfl's behavior on ground terms

nomeata:
`rfl` for ground terms will reduce at transparency `.all`, even if the current transparency is `.default`. Maybe it's prudent to not rely on this corner case, and have more explicit proofs.

### DIFF
--- a/Batteries/Data/Fin/Lemmas.lean
+++ b/Batteries/Data/Fin/Lemmas.lean
@@ -41,14 +41,14 @@ theorem list_succ (n) : list (n+1) = 0 :: (list n).map Fin.succ := by
 theorem list_succ_last (n) : list (n+1) = (list n).map castSucc ++ [last n] := by
   rw [list_succ]
   induction n with
-  | zero => rfl
+  | zero => simp [last]
   | succ n ih =>
     rw [list_succ, List.map_cons castSucc, ih]
     simp [Function.comp_def, succ_castSucc]
 
 theorem list_reverse (n) : (list n).reverse = (list n).map rev := by
   induction n with
-  | zero => rfl
+  | zero => simp [last]
   | succ n ih =>
     conv => lhs; rw [list_succ_last]
     conv => rhs; rw [list_succ]


fgdorais:
I understand where this is coming from but I'm skeptical of `simp` as a replacement since that is a much more complex tactic. Maybe we need an enhanced `rfl`-like tactict that doesn't draw in all the simp lemmas?

nomeata:
Yes, this PR is a mit premature, I should first figure out if this rfl behavior is intended or accidental.

Using simp isn't unreasonable in the sense that the other half of that proof is also rewriting. But I am more hesitant to rely on DefEqs of irreducible definitions than others :-)

Marking as draft for now.

fgdorais:
I don't think these two changes are problematic by themselves. I'm mostly questioning the general strategy of replacing `rfl` by `simp`, especially now that defs by wf are irreducible by default. It seems to me that `rfl` is the wrong tool for reasons you mention but `simp` is also the wrong tool for different reasons. These are the tools we have and the "right tool" appears to be missing, probably because there was not much need before now.

To be more specific about my (currently purely theoretical) vision is for an `eqn` tactic that attempts to prove an equality using "defining equations". In most cases, these "defining equations" would simply be the actual definition (i.e. what `rfl` does) but in the case of irreducible defs, these would be more like specifications which have more to do with the intended meaning than the actual implementation.

Note that the equations used by `eqn` are not always `simp` lemmas. In fact, some would be really bad `simp` lemmas. But `eqn` is cannot be used to simplify a goal, so that is not a problem. In a sense, `eqn` would be orthogonal to `simp`, using overlapping but ultimately completely different set of rules.

Anyway, this discussion should probably be elsewhere.

nomeata:
I’m working on https://github.com/leanprover/lean4/pull/3772 again, so I expect his will be needed soon. I’d wager that `simp [last]` is reasonable style, at least with the idioms and tactics with have.

fgdorais:
I still don't think `simp` is the right choice here since it doesn't have the same behavior as `rfl` at all. How about having `rfl [foo, bar]` that locally changes the reducibility of `foo` and `bar` before `rfl`?

fgdorais (approved):
These two changes are fine but I would still like a better solution for the general case.

PR #835, opened 2024-06-12
chore: upstream most LazyList results from Mathlib

eric-wieser:
Either these belong in Batteries, or `LazyList` should be deprecated in favor of `MLList`


semorrison:
Do we know of any users of `LazyList`?

fgdorais:
I would like to see a `LawfulMonad (LazyList a)` before adding any more stuff here.

eric-wieser:
> I would like to see an equivalent to `LawfulMonad LazyList`

This instance already exists in mathlib, but somehow the `induction` tactic used in it's proof fails to infer the motive when used in Batteries. 

> before adding any more stuff here.

There is pressure to remove these theorems and definitions from mathlib; so it would be preferable if they could be upstreamed verbatim, and cleaned up later. 

Of course, that all predicates on the assumption that LazyList has a future at all.

fgdorais:
It takes a while to track down the history of these moved Mathlib files. This is basically straight from Mathlib3, where this lazy list type had VM support. This is of course moot in Lean 4. In fact, this particular usage of `Thunk` is perhaps inefficient in Lean 4. Some experiments for lazy lists can be found in Lean 4 tests, e.g. this [playground file](https://github.com/leanprover/lean4/blob/285a3130782e9ec91c7a7e126f1e691a70f9e756/tests/playground/lazylist.lean). In all cases, the implementation is different than that here. The `delayed` constructor is used to avoid uses of `Thunk` to avoid performance issues.

In conclusion, I think that in order to survive in Batteries, `LazyList` needs an updated implementation that is more efficient and that new implementation should be benched and profiled to ensure proper behavior. I'm not against the idea of moving here to unburden Mathlib until that update happens, but we're just cumulating idle work by doing so.

fgdorais:
I split the `Thunk.ext` as #842 so it doesn't get lost. Please advise for the correct copyright and author lines.

eric-wieser:
> I split the `Thunk.ext` as #842 so it doesn't get lost. Please advise for the correct copyright and author lines.

Given that you rewrote the whole proof and the result is trivial, I think it's fine to give yourself attribution

semorrison:
Can we make progress on this? I would like to get `LazyList` out of Mathlib.

Options:
1. Merge this as is, and then we can remove `LazyList` from Mathlib.
2. Someone says they are actively working on a new implementation of `LazyList`, with a timeline, and we wait for that.
3. If this PR stalls here, we just deprecate the material in Mathlib (as there is no evidence anyone is using it).

My preference is 1., and then 3.

semorrison:
I've changed my mind, and my preference is now 3.

@digama0, @fgdorais, would it be okay if I proceeded on that? I can add a link to this PR when deprecating.

digama0:
Let's deprecate it and point people to `MLList` for now. It's not a great replacement, but anyone who complains about it will at least prove the existence statement of lazylist users.

eric-wieser:
Surely if we are going to deprecate, we should deprecate `LazyList` in batteries, not (just) the material in mathlib?

digama0:
My impression was that `LazyList` currently doesn't exist at all in batteries, so there is nothing to deprecate... but I see now that this is not the case. Do we have MLList analogues for all the new definitions? Perhaps we should merge this and then deprecate the whole file.

PR #838, opened 2024-06-12
fix(linter): "(invalid MessageData.lazy, missing context)"

llllvvuu:
Relevant if https://github.com/leanprover/lean4/issues/4432 is wontfix

https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/v4.2E9.2E0-rc1.20discussion/near/443441359


semorrison:
I think we will fix it upstream, but we can also merge this workaround in the meantime, so I can get `nightly-testing` working again.

PR #839, opened 2024-06-12
chore(Tactic/Lint/Simp): make simpNF linter dsimp-aware

jcommelin:


Batteries/Tactic/Lint/Simp.lean
```diff
@@ -112,11 +112,18 @@ https://leanprover-community.github.io/mathlib_docs/notes.html#simp-normal%20for
     unless ← isSimpTheorem declName do return none
     let ctx := { ← Simp.Context.mkDefault with config.decide := false }
     checkAllSimpTheoremInfos (← getConstInfo declName).type fun {lhs, rhs, isConditional, ..} => do
+      let isRfl ← isRflTheorem declName
       let ({ expr := lhs', proof? := prf1, .. }, prf1Stats) ←
-        decorateError "simplify fails on left-hand side:" <| simp lhs ctx
+        decorateError "simplify fails on left-hand side:" <|
+          if !isRfl then simp lhs ctx else do
+            let (e, s) ← dsimp lhs ctx
+            return (Simp.Result.mk e .none .true, s)
```
semorrison, line 120:
```suggestion
          if !isRfl then 
            simp lhs ctx
          else do
            let (e, s) ← dsimp lhs ctx
            return (Simp.Result.mk e .none .true, s)
```

semorrison: and below.

YaelDillies (commented):
Can you update the error message as well? It currently reads `"simp can prove this:`.

Batteries/Tactic/Lint/Simp.lean
jcommelin: Done

jcommelin:
I updated the error message.

joneugster:
> Can you update the error message as well? It currently reads `"simp can prove this:`.

(and also in `formatLemmas` where it says `simp only`: [here](https://github.com/leanprover-community/batteries/blob/a7bf196220562dbe18e3ad5bfcee7cc20789e0e9/Batteries/Tactic/Lint/Simp.lean#L103))

YaelDillies (approved):
Great, thanks! LGTM

joneugster (approved):

PR #842, opened 2024-06-13
feat: ext lemma for `Thunk`

fgdorais:
Split from #835, upstreamed from Mathlib

### DIFF
--- a/Batteries/Data/Thunk.lean
+++ b/Batteries/Data/Thunk.lean
@@ -1,0 +1,7 @@
+
+namespace Thunk
+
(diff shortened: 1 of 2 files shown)


Batteries/Data/Thunk.lean
```diff
@@ -0,0 +1,7 @@
+
```
digama0, line 1: needs copyright header

fgdorais: I asked @eric-wieser what I should put there. There seems to be no claim to this. I'll put my name for now and we can fix it later.

PR #843, opened 2024-06-14
chore: adaptations for leanprover/lean4#4400, List look-up normal form

semorrison:
This shouldn't be merged until leanprover/lean4#4400 lands in a nightly.


fgdorais (approved):
I've looked at every line and the changes are correct. I approve assuming everything works out downstream.

PR #849, opened 2024-06-17
feat: add `BinaryHeap`

fgdorais:
Moved from Mathlib with minimal adaptations.


semorrison (approved):
LGTM

fgdorais:
I moved to the Batteries namespace. That can be reverted if necessary.

This can be merged as-is right now. Soon, I will update some parts to use `Vector` which will hopefully simplify adding the missing correctness proofs. But that can wait in the interest of decluttering Mathlib.

PR #851, opened 2024-06-17
feat: add `ByteSubarray`

fgdorais:
Alternative to #836. While this is not quite a drop-in replacement for users of Mathlib's `ByteSlice`, it offers the same functionality while using the same basic structure as core's `Subarray` type.

This fixes the issues I raised in #836 but I'm also happy with compromises. For example, I personally think `ByteSlice` is a better name than `ByteSubarray`. I'm also a fan of the offset/length approach but this seems to much of a divergence from `Subarray`.

### DIFF
--- a/Batteries.lean
+++ b/Batteries.lean
@@ -19,6 +19,7 @@ import Batteries.Data.BinomialHeap
 import Batteries.Data.BitVec
 import Batteries.Data.Bool
 import Batteries.Data.ByteArray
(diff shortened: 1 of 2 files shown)


semorrison:
I have no particular opinion here (as I don't know of any use cases yet), but I will happily deprecate Mathlib's `ByteArray` in favour of this when it is merged.

PR #852, opened 2024-06-18
chore: robust proof in HashMap.WF.filterMap

semorrison:


semorrison:
I needed this to fix some problems on `nightly-testing`.

PR #853, opened 2024-06-20
feat: add  simps and `toNat` lemmas for `UIntX` types

fgdorais:
Reported by @eric-wieser on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Fintype.20Char.3F/near/445317208).

Some lemmas are about `val` instead of `toNat` so they may need update if core changes the implementation of `UIntX` types.


eric-wieser (approved):
Thanks! Ir would be nice to have `ofNat` lemmas for `UInt64.toUInt32` etc too, but that doesn't have to be in this PR.

fgdorais:
> would be nice to have `ofNat` lemmas for `UInt64.toUInt32` etc too, but that doesn't have to be in this PR.

These are not as trivial, so another PR is best to avoid overloading this one.

PR #854, opened 2024-06-20
feat: order lemmas for `UIntX` types

fgdorais:


eric-wieser (approved):
It would be good to have the `val` versions of these lemmas too.

Batteries/Data/UInt.lean
```diff
@@ -42,7 +42,7 @@ theorem UInt8.max_def (x y : UInt8) : max x y = if x ≤ y then y else x := rfl
 
 theorem UInt8.min_def (x y : UInt8) : min x y = if x ≤ y then x else y := rfl
 
-theorem UInt8.toNat_max (x y : UInt8) : (max x y).toNat = max x.toNat y.toNat := by
+theorem UInt8.max_toNat (x y : UInt8) : (max x y).toNat = max x.toNat y.toNat := by
```
eric-wieser, line 45: I think the old name was correct here, this is `toNat (max x y)`?

fgdorais: Yes, I agree. But UInt uses the reverse convention everywhere so there appears to be a local convention.

fgdorais: There's a few of these (e.g. `Array.data`). They're probably all legacy names but they're pretty solidly rooted.

fgdorais: I un-reverted. Looks like it's only `zero_toNat`, `div_toNat`, `mod_toNat` that need fixing in core, so the roots are not that deep.

fgdorais: Fixing naming convention in core: [lean4#4514](https://github.com/leanprover/lean4/pull/4514).

PR #856, opened 2024-06-22
chore: adaptations for nightly-2024-06-22

semorrison:


Batteries/Lean/PersistentHashMap.lean
```diff
@@ -10,15 +10,6 @@ namespace Lean.PersistentHashMap
...
 variable [BEq α] [Hashable α]
 
-/--
-Similar to `insert`, but also returns a Boolean flag indicating whether an
-existing entry has been replaced with `a => b`.
--/
-def insert' (m : PersistentHashMap α β) (a : α) (b : β) : PersistentHashMap α β × Bool :=
-  let oldSize := m.size
-  let m := m.insert a b
-  (m, m.size == oldSize)
```
semorrison, line 20: `PersistentHashMap` no longer has a `size` method, as it was previously buggy, and doesn't seem to actually be used.

This function was added for `aesop`, but is not actually used.

PR #857, opened 2024-06-23
chore: deal with nonterminal simps in String/Lemmas

semorrison:
This file was full on non-terminal simps, and inevitable they started failing and getting in my way when attempting to change @[simp] lemmas in core.

So I've mostly removed the non-terminal simps here.

There are many cases where `simp?` is giving wrong answers because it is not correctly disambiguating namespaces. I left FIXMEs for all of these rather than doing these by hand.


chabulhwi:
@semorrison https://github.com/leanprover-community/batteries/pull/864 removes all nonterminal `simp`s from the lemmas that had the `FIXME` comment.

PR #861, opened 2024-06-25
chore: `unreachableTactic` -- replace a `Std` namespace with `Batteries` and add an `inherit_doc`.

adomani:


adomani:
awaiting-review

PR #869, opened 2024-07-01
chore: remove 'meta if' from lakefile

semorrison:
As far as I understand, no human or automation ever runs doc-gen on `Batteries`. Instead documentation is generated as part of Mathlib's documentation.

If that is the case, I would like to remove the `doc-gen` dependency from the `lakefile`.

(This is on the path to moving to a `lakefile.toml`, and eventually possibly deprecating `meta if` in `lakefile.lean`.)

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -18,9 +18,6 @@ lean_exe runLinter where
   srcDir := "scripts"
   supportInterpreter := true
 
-meta if get_config? doc |>.isSome then
-require «doc-gen4» from git "https://github.com/leanprover/doc-gen4" @ "main"
-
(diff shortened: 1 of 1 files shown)


digama0 (approved):

digama0:
> (This is on the path to moving to a lakefile.toml, and eventually possibly deprecating meta if in lakefile.lean.)

I think we should *discourage* `meta if` but not deprecate it, certainly not before getting conditional dependencies in lakefile.lean and/or .toml, but possibly even after because I'm sure there are OS dependent projects and other things that need to script their lakefiles at configure time.

PR #874, opened 2024-07-07
chore: remove `proof_wanted` in HashMap.Lemmas

TwoFX:
It would probably not be a good use of anyone's time to work on that sorry right now, and there was [some confusion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/224796-Geographic-locality/topic/New.20York.20City.2C.20USA/near/449684763) today, so I suggest removing the `proof_wanted`.


TwoFX:
awaiting-review

digama0:
how about commenting it out and put a note saying why it's not wanted?

TwoFX:
@digama0 Done.

semorrison:
Reading this without the request from Mario I would of course ask "what is the point of keeping the comment"?

But it's better than leaving it as is!

PR #877, opened 2024-07-10
feat: Array lemma for empty map added

Shreyas4991:
As the title says, this adds an API lemma `mem_empty`.

### DIFF
--- a/Batteries/Data/Array/Lemmas.lean
+++ b/Batteries/Data/Array/Lemmas.lean
@@ -129,5 +129,7 @@ theorem size_shrink (a : Array α) (n) : (a.shrink n).size = min a.size n := by
 
 /-! ### map -/
 
-unseal mapM.map in
-theorem map_empty (f : α → β) : map f empty = empty := rfl
+theorem mapM_empty [Monad m] (f : α → m β) : mapM f #[] = pure #[] := by
+  rw [mapM, mapM.map]; rfl
+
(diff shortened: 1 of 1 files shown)


Shreyas4991:
awaiting-review

Batteries/Data/Array/Lemmas.lean
```diff
@@ -126,3 +126,8 @@ theorem size_shrink_loop (a : Array α) (n) : (shrink.loop n a).size = a.size -
 theorem size_shrink (a : Array α) (n) : (a.shrink n).size = min a.size n := by
   simp [shrink, size_shrink_loop]
   omega
+
+/-! ### map -/
+
+unseal mapM.map in
+theorem map_empty (f : α → β) : map f empty = empty := rfl
```
fgdorais, line 133: Since `mapM` needs to be mentioned anyway, we might as well have the matching pair.
```suggestion
theorem mapM_empty [Monad m] (f : α → m β) : mapM f #[] = pure #[] := by
  rw [mapM, mapM.map]; rfl

@[simp] theorem map_empty (f : α → β) : map f #[] = #[] := mapM_empty ..
```

Shreyas4991: Happy to accept this. At this point, I just hope these PRs will get merged some day.

fgdorais (approved):

PR #880, opened 2024-07-11
doc: fix outdated docstring

grunweg:
The `unusedHavesSuffices` linter is run by default now: remove the part of its docstring which says otherwise.


grunweg:
awaiting-review

PR #881, opened 2024-07-12
feat: allow configuring additional linters to run

grunweg:
Extend `getChecks` to allow an additional list of linters, which is always run. This can be used downstream to e.g. enable the `docBlameThm` or `explicitVarsOfIff` linters without modifying batteries.

As part of this, we
- update fix the function docstring (which was quite outdated anyway)
- remove the `useOnly` option: the equivalent effect can be achieved by filtering the list of linters afterwards; *adding* linters cannot, hence must be performed in this function. (That said, I'm happy to add this functionality back if desired.)
- rename a variable `dflt` to the more readable `default`
- adjust the logic of the #lint command accordingly
Note that `test/lint_simpNF.lean` lies on this (the docBlame linter warns there) - so this changed is already tested.

### DIFF
--- a/Batteries/Tactic/Lint/Frontend.lean
+++ b/Batteries/Tactic/Lint/Frontend.lean
@@ -65,17 +65,23 @@ inductive LintVerbosity
   | high
   deriving Inhabited, DecidableEq, Repr
 
-/-- `getChecks slow useAlso` produces a list of linters.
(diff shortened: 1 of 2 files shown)


grunweg:
awaiting-review

grunweg:
I just pushed a commit adding back the `useOnly` option as `runOnly`.

PR #884, opened 2024-07-15
chore: deprecate LazyList

semorrison:
Per https://github.com/leanprover-community/batteries/pull/835#issuecomment-2221717519


fgdorais (approved):
Maybe add a brief explanation that this type is legacy Lean 3 and that it no longer has a special purpose in Lean 4.

PR #889, opened 2024-07-22
fix: flaky test

digama0:
This test is timing-sensitive, and has [spuriously failed](https://github.com/leanprover-community/batteries/actions/runs/10036728715/job/27735044947?pr=887) before. Increase the timings slightly to make it less easy to race.


eric-wieser (approved):
This has caught me multiple times now; though I think I'm not supposed to use my merge powers here.

PR #891, opened 2024-07-24
feat: some array lemma aliases

fgdorais:
Core uses `nil` instead of `empty` for some array lemmas and I keep getting caught by that, so I'm proposing to add aliases for the `empty` variants. I'm also proposing to add `mem_singleton` since that one comes up often enough that I'd like to avoid going through list every time.


semorrison:
I will try to fix these in Lean in not too long!

PR #897, opened 2024-07-29
feat: add `Array.Pairwise`

fgdorais:


Batteries/Data/Array/Pairwise.lean
```diff
@@ -0,0 +1,55 @@
...
+`Pairwise R as` means that all the elements of the array `as` are `R`-related to all elements with
+larger indices.
+
+`Pairwise R #[1, 2, 3] ↔ R 1 2 ∧ R 1 3 ∧ R 2 3`
+
+For example `as.Pairwise (· ≠ ·)` asserts that `as` has no duplicates, `as.Pairwise (· < ·)` asserts
+that `as` is strictly sorted and `as.Pairwise (· ≤ ·)` asserts that `as` is weakly sorted.
+-/
+abbrev Pairwise (R : α → α → Prop) (as : Array α) : Prop :=
+  ∀ (i j : Fin as.size), i < j → R as[i] as[j]
```
digama0, line 21:
```suggestion
def Pairwise (R : α → α → Prop) (as : Array α) : Prop :=
  as.data.Pairwise R
```
I think the fin definition should be a theorem.

fgdorais: I decided against that because the naive decision procedure for `Array.Pairwise` would have an unnecessary array to list overhead. I suppose a custom decidable instance could fix that.

fgdorais: Actually, a custom decidable instance would also reduce the constant in the quadratic term. So I will do that and use your preferred definition.

PR #899, opened 2024-07-30
feat: add finRange

fgdorais:
This is to correct a name clash from #528 with Mathlib pointed out by @eric-wieser 

We add `List.finRange` and `Array.finRange` (with basic lemmas). The synonyms `Fin.list` and `Fin.enum` are deprecated.

### DIFF
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -1393,3 +1393,6 @@ where
   | [], r, t => reverseAux t r
   | l, [], t => reverseAux t l
   | a::l, b::r, t => bif s a b then loop l (b::r) (a::t) else loop (a::l) r (b::t)
+
+/-- `finRange n` is the list of all elements of `Fin n` in order -/
+def finRange (n) : List (Fin n) := ofFn id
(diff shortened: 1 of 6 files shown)


eric-wieser:
Could you check that you have all the lemmas from Mathlib, and that the names and statements match?

https://github.com/leanprover-community/mathlib4/blob/df2eea5e25e8b5d262cd5ce237c6dd9335a98d22/Mathlib/Data/List/Range.lean#L87-L162

fgdorais:
It's not my aim to get all the lemmas from Mathlib, but the names for the few that are here do agree. Are there any specific lemmas from Mathlib that you think are missing?

Batteries/Data/List/Basic.lean
```diff
@@ -1393,3 +1393,6 @@ where
   | [], r, t => reverseAux t r
   | l, [], t => reverseAux t l
   | a::l, b::r, t => bif s a b then loop l (b::r) (a::t) else loop (a::l) r (b::t)
+
+/-- `finRange n` is the list of all elements of `Fin n` in order -/
```
eric-wieser, line 1397: In mathlib the docstring is
```
/-- All elements of `Fin n`, from `0` to `n-1`. The corresponding finset is `Finset.univ`. -/
```
Should the last sentence be kept, with an added "in mathlib"?

fgdorais: Good question. I'm not sure. How useful is the second sentence for Mathlib users? It seems like it is a trivial conclusion from the first sentence, isn't it?

fgdorais:
Closing for now since this will require substantial revisions after #956 gets merged.

PR #901, opened 2024-08-01
refactor: rename theorems about `Char.utf8Size`

chabulhwi:
Rename theorems whose statement includes `Char.utf8Size`.


chabulhwi:
awaiting-review

PR #903, opened 2024-08-02
chore: shake Batteries

semorrison:


Batteries/Data/List/Perm.lean
```diff
@@ -7,7 +7,6 @@ import Batteries.Tactic.Alias
 import Batteries.Data.List.Init.Attach
 import Batteries.Data.List.Pairwise
 -- Adaptation note: nightly-2024-03-18. We should be able to remove this after nightly-2024-03-19.
```
semorrison, line 9:
```suggestion
```

Batteries/Data/Array/Basic.lean
```diff
@@ -4,7 +4,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Arthur Paulino, Floris van Doorn, Jannis Limperg
 -/
 import Batteries.Data.List.Init.Attach
-import Batteries.Data.Array.Init.Lemmas
```
fgdorais, line 7: This one could stay. The import is currently empty but its purpose is to be included here (and ideally nowhere else).

Batteries/Data/List/Basic.lean
```diff
@@ -3,7 +3,6 @@ Copyright (c) 2016 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Leonardo de Moura
 -/
-import Batteries.Data.List.Init.Lemmas
```
fgdorais, line 6: This one could stay. The import is currently empty but its purpose is to be included here (and ideally nowhere else).

Batteries/Data/Array/Basic.lean
semorrison: Sure... importing empty files for which there are no active plans to add anything seems a bit of a waste of time, but no objection.

Batteries/Data/List/Basic.lean
semorrison: Done.

fgdorais (approved):

PR #904, opened 2024-08-02
chore: bump lean-action to v1

austinletson:
lean-action will now run the linter via `lake lint` instead of running it directly with `lake exe runLinter`


austinletson:
awaiting-review

fgdorais (approved):

PR #905, opened 2024-08-03
feat: getElem_tail lemmas

pimotte:
I felt like these were missing, I tried adhering to the existing standards, but happy to patch things up if I missed something:)


pimotte:
awaiting-review

Batteries/Data/List/Lemmas.lean
```diff
@@ -478,6 +478,20 @@ theorem length_eraseIdx : ∀ {l i}, i < length l → length (@eraseIdx α l i)
 
 @[simp] theorem length_tail (l : List α) : length (tail l) = length l - 1 := by cases l <;> rfl
 
+theorem succ_length_tail (l : List α) (h : 0 < length l) : (length (tail l)).succ = length l := by
+  cases l <;> simp at h ⊢
+
+theorem length_tail_le (l : List α) : length (tail l) ≤ length l := by
+  cases l <;> simp [Nat.le_succ]
+
+@[simp] theorem getElem?_tail {n} (l : List α) : l.tail[n]? = l[n + 1]? := by cases l <;> simp
```
fgdorais, line 487: Batteries uses auto-implicits so you should either have `(n)` or leave it out entirely.

Batteries/Data/List/Lemmas.lean
```diff
@@ -478,6 +478,20 @@ theorem length_eraseIdx : ∀ {l i}, i < length l → length (@eraseIdx α l i)
 
 @[simp] theorem length_tail (l : List α) : length (tail l) = length l - 1 := by cases l <;> rfl
 
+theorem succ_length_tail (l : List α) (h : 0 < length l) : (length (tail l)).succ = length l := by
+  cases l <;> simp at h ⊢
```
fgdorais, line 482: I really wish `simp [*]` or `simp_all` could just prove this. There might be a missing simp somewhere in `Nat`.

Here is a different proof but not better than yours.
```lean
theorem succ_length_tail (l : List α) (h : 0 < length l) : (length (tail l)).succ = length l := by
  cases l; contradiction; simp
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -478,6 +478,20 @@ theorem length_eraseIdx : ∀ {l i}, i < length l → length (@eraseIdx α l i)
...
+theorem length_tail_le (l : List α) : length (tail l) ≤ length l := by
+  cases l <;> simp [Nat.le_succ]
+
+@[simp] theorem getElem?_tail {n} (l : List α) : l.tail[n]? = l[n + 1]? := by cases l <;> simp
+
+@[simp] theorem getElem_tail {n} (l : List α) (h : n < l.tail.length) :
+    l.tail[n] = l[n + 1]'(Nat.lt_of_lt_of_le (succ_lt_succ h)
+    (succ_length_tail _ (Nat.lt_of_le_of_lt (zero_le n)
+    (Nat.lt_of_lt_of_le h (length_tail_le _))) ▸ Nat.le_refl l.length)) := by
+  cases l <;> simp [Nat.not_lt_zero] at h ⊢
```
fgdorais, line 493: When the `[]'()` proof part is that complicated, I like to put it as an optional argument as follows. That makes the theorem easier to read but I honestly have no evidence whether this is better or worse.
```suggestion
@[simp] theorem getElem_tail (l : List α) (h : n < l.tail.length)
    (h' : n + 1 < l.length := by
      cases l; contradiction; apply Nat.succ_lt_succ; apply Nat.lt_of_lt_of_le h; simp) :
    l.tail[n] = l[n + 1] := by
  cases l; contradiction; simp
```

fgdorais (changes requested):

Batteries/Data/List/Lemmas.lean
```diff
@@ -478,6 +478,20 @@ theorem length_eraseIdx : ∀ {l i}, i < length l → length (@eraseIdx α l i)
 
 @[simp] theorem length_tail (l : List α) : length (tail l) = length l - 1 := by cases l <;> rfl
 
+theorem succ_length_tail (l : List α) (h : 0 < length l) : (length (tail l)).succ = length l := by
+  cases l <;> simp at h ⊢
+
+theorem length_tail_le (l : List α) : length (tail l) ≤ length l := by
+  cases l <;> simp [Nat.le_succ]
```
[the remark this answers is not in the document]
pimotte: My usecase was simplifying the `[]'()` term later, so I removed it

Batteries/Data/List/Lemmas.lean
pimotte: I don't know about the missing `simp` stuff, but I do think your version shows the argument more clearly, so I adopted that one.

fgdorais: There is definitely a `simp` issue here since this minor variant works:
```lean
theorem length_tail_add_one (l : List α) (h : 1 ≤ length l) : (length (tail l)) + 1 = length l := by simp [h]
```
I'll keep investigating...

In any case, we do prefer `add_one` over `succ` since `Nat` induction now uses `_+1` instead of `succ`. So `Nat.succ` is considered low-level and most end-user theorems should try to use `_+1` instead.

pimotte: Fixed the comments. I've also looked a bit closer at the simp issue and I've pushed some stuff that might address it. I don't know if this is something that belongs in batteries, but it does seem a bit detailed for the core repo?

fgdorais: The `Nat` lemmas should go in a separate PR.

Another solution is to add `@[simp]` to `Nat.sub_one_add_one` and `Nat.sub_one_add_one_eq_of_pos` in Lean core.

pimotte: Okay, I pulled them out of this one:) I'm not great at overseeing the consequences of adding simp lemmas, do you have a preference for either solution?

fgdorais: Conditional simp lemmas are a bit tricky. When multiple such lemmas trigger on the same expression, in this case `(a - 1) + 1` in this case, they each trigger a search for the requisite hypothesis. So, fewer is better and an additional conditional simp lemma with an unlikely hypothesis is not useful. From that point of view, my solution is worse since it adds two conditional simp lemmas but then again your solution triggers on a more general expression, namely `(a - b) + b`.

Maybe there's an even better solution we're not seeing. For example, simprocs may be useful in this situation but I've delayed learning how to properly use these, so I don't know if they could actually help. Bottom line: I don't know what the right thing to do is here.

(Just in case that wasn't clear: I only recommended splitting the `Nat` lemmas because focused PRs get merged faster, not because I thought your solution was worse!)

pimotte: Gotcha:) In that case I would like to propose merging this as-is and postponing the simp-related stuff to possible future PRs.

fgdorais:
There is an unresolved "missing simp" issue here but the author and I resolved that this is not the right place to fix that problem.

Batteries/Data/List/Lemmas.lean
```diff
@@ -190,6 +190,19 @@ theorem get?_set_of_lt' (a : α) {m n} (l : List α) (h : m < length l) :
...
+theorem length_tail_add_one (l : List α) (h : 0 < length l) : (length (tail l)) + 1 = length l := by
+  simp [Nat.sub_add_cancel h]
+
+@[simp] theorem getElem?_tail (l : List α) : l.tail[n]? = l[n + 1]? := by cases l <;> simp
+
+@[simp] theorem getElem_tail (l : List α) (h : n < l.tail.length)
+    (h' : n + 1 < l.length := by
+      cases l; contradiction; apply Nat.succ_lt_succ; apply Nat.lt_of_lt_of_le h; simp) :
+    l.tail[n] = l[n + 1] := by
+  cases l; contradiction; simp
```
semorrison, line 204: This won't work as a `simp` lemma because of `h'`. You need instead:

```
@[simp] theorem getElem_tail' (l : List α) (h : n < l.tail.length) :
    l.tail[n] = l[n + 1]'(by simp at h; omega) := by
  cases l; contradiction; simp
```

It's fine to provide a separate version for writing the other way, but not at the expense of breaking the simp normal form lemma.

pimotte: I have adopted your suggestion, because I think that's what I actually wanted.

I don't quite fully understand what's going on here, though. Why does h' break `simp` here?

fgdorais:
Mathlib adaptation [#17229](https://github.com/leanprover-community/mathlib4/pull/17229)

PR #914, opened 2024-08-13
fix: naming convention

fgdorais:
Many theorems have the reverse naming convention for `Array.data` and `ByteArray.data`. This fixes many of them but there are plenty more in Lean core.

### DIFF
--- a/Batteries/Data/Array/Lemmas.lean
+++ b/Batteries/Data/Array/Lemmas.lean
@@ -11,7 +11,7 @@ import Batteries.Util.ProofWanted
 
 namespace Array
 
-theorem forIn_eq_data_forIn [Monad m]
+theorem forIn_eq_forIn_data [Monad m]
     (as : Array α) (b : β) (f : α → β → m (ForInStep β)) :
     forIn as b f = forIn as.data b f := by
   let rec loop : ∀ {i h b j}, j + i = as.size →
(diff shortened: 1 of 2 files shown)


Batteries/Data/Array/Lemmas.lean
```diff
@@ -27,10 +27,11 @@ theorem forIn_eq_data_forIn [Monad m]
       rw [loop (i := i)]; rw [← ij, Nat.succ_add]; rfl
   conv => lhs; simp only [forIn, Array.forIn]
   rw [loop (Nat.zero_add _)]; rfl
+@[deprecated (since:="2024-08-13")] alias forIn_eq_data_forIn := forIn_eq_forIn_data
```
digama0, line 30:
```suggestion
@[deprecated (since := "2024-08-13")] alias forIn_eq_data_forIn := forIn_eq_forIn_data
```
et al

Batteries/Data/ByteArray.lean
```diff
@@ -18,15 +19,18 @@ theorem getElem_eq_data_getElem (a : ByteArray) (h : i < a.size) : a[i] = a.data
 
 /-! ### empty -/
 
-@[simp] theorem mkEmpty_data (cap) : (mkEmpty cap).data = #[] := rfl
+@[simp] theorem data_mkEmpty (cap) : (mkEmpty cap).data = #[] := rfl
+@[deprecated (since:="13/08/2024")] alias mkEmpty_data := data_mkEmpty
```
digama0, line 23: inconsistent date format, I think we prefer YYYY-MM-DD

digama0 (approved):

PR #919, opened 2024-08-16
feat: faster splitAt

semorrison:
It turns out using `List` is faster than `Array` in the auxiliary functions for `splitAt` and `splitAtD`. It's about a 10% improvement (and much more for splitAtD when there is lots of padding).

I was a little surprised at first, so double checked my timings from `#time #eval ...` with a `hyperfine "lean --run ..."` benchmark.

### DIFF
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -116,13 +116,13 @@ Split a list at an index.
 splitAt 2 [a, b, c] = ([a, b], [c])
 ```
 -/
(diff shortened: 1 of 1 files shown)


digama0:
This is not actually that surprising, but I appreciate you getting hard numbers on this, I'd been hoping someone could verify this one way or another. The reason the list version performs competitively is because `List.reverse` works in-place, so it's not really wasting any allocations not involved in the final output. The array version is more compact but the array allocation is additional to the output.

PR #920, opened 2024-08-16
feat: the trailingWhitespace linter

adomani:
The `trailingWhitespace` linter emits a warning whenever a line ends with a space or a file does not end with a line break.

### DIFF
--- a/Batteries/Linter/TrailingWhitespace.lean
+++ b/Batteries/Linter/TrailingWhitespace.lean
@@ -1,0 +1,50 @@
+import Lean.Elab.Command
+import Lean.Linter.Util
+
+/-!
+#  The "trailingWhitespace" linter
+
+The "trailingWhitespace" linter emits a warning whenever a line ends with a space or a file
+does not end with a line break.
+-/
+
+open Lean Elab
+
+namespace Mathlib.Linter
+
+/--
+The "trailingWhitespace" linter emits a warning whenever a line ends with a space or a file
+does not end with a line break.
+-/
+register_option linter.trailingWhitespace : Bool := {
+  defValue := true
+  descr := "enable the trailingWhitespace linter"
+}
+
+namespace TrailingWhitespace
+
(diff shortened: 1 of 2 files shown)


digama0:
I think this should be disabled by default, or only enabled in CI. For people using vscode with the "remove trailing whitespace on save" option, it will only ever be a false positive triggering between keystrokes.

adomani:
That's true: it was not easy to get the test file to work, since I could not save it!

Would the "script-style" code below be better than a linter, then?

```lean
import Lean.Elab.Command

open Lean in
run_cmd
  for fil in ← System.FilePath.walkDir ((← IO.currentDir) / "Batteries") do
    if let some "lean" := fil.extension then
      let lines ← IO.FS.lines fil
      for l in [:lines.size] do
        let line := lines[l]!
        let tr := line.takeRightWhile (·.isWhitespace)
        if !tr.isEmpty then
          logWarning
            m!"{fil}:{l} ends with {tr.length} space{if tr.length == 1 then "" else "s"}"
      if (← IO.FS.readFile fil).back != '\n' then
        logWarning m!"{fil} does not end with a line break"
```

Batteries/Linter/TrailingWhitespace.lean
```diff
@@ -0,0 +1,48 @@
...
+open Lean Elab
+
+namespace Batteries.Linter
+
+/--
+The "trailingWhitespace" linter emits a warning whenever a line ends with a space or a file
+does not end with a line break.
+-/
+register_option linter.trailingWhitespace : Bool := {
+  defValue := true
```
adomani, line 20:
```suggestion
  defValue := false
```

adomani: As suggested by Mario.  There would still be the need to switch the option to true when building Batteries.

digama0:
We don't really have a place to put scripts like this. Ideally we could put it somewhere such that lake would make it available to downstream packages.

PR #925, opened 2024-08-17
feat: monadic defs for statically sized vectors

Shreyas4991:
This PR is the first follow up to #793. It adds all the monadic definitions for Vectors that correspond to similar definitions for Arrays in [core](https://github.com/leanprover/lean4/blob/master/src/Init/Data/Array/Basic.lean) and [Batteries](https://github.com/leanprover-community/batteries/blob/main/Batteries/Data/Array/Basic.lean) as well as defs which are immediately downstream of them. There are no plans to add theorems in this PR.

-[] Add all monadic definitions from `Init.Data.Array.Basic` and their downstream definitions for Vectors.
-[] Add all monadic definitions from `Batteries.Data.Array.Basic` and their downstream definitions for Vectors.


Shreyas4991:
WIP

Batteries/Data/Vector/Monadic.lean
```diff
@@ -0,0 +1,170 @@
...
+
+namespace Vector
+
+/--
+`modifyM v i f` applies a monadic transformation `f` to `v[i]`
+-/
+def modifyM [Monad m] (v : Vector α n) (i : Nat)
+  (f : α → m α) : m (Vector α n) := do
+    return ⟨←Array.modifyM v.toArray i f,
+      by rw[←v.size_eq]; sorry⟩
```
Shreyas4991, line 27: Any help in fixing this `sorry` is welcome. I found a lemma by the name `Array.size_modifyM` which however is not a straightforward equality.

Batteries/Data/Vector/Monadic.lean
```diff
@@ -0,0 +1,171 @@
...
+  (start := n) (stop := 0) : m β :=
+    Array.foldrM f init v.toArray start stop
+
+/--
+`mapM f v` maps a monadic function `f : α → m β` over `v` and returns
+the resultant vector in the same monad `m`.
+-/
+def mapM {α : Type u} {β : Type v} {m : Type v → Type w} [Monad m]
+  (f : α → m β) (v : Vector α n) : m (Vector β n) := do
+    return ⟨←Array.mapM f v.toArray, by rw[←v.size_eq]; sorry⟩
```
Shreyas4991, line 73: This `sorry` is also stuck for the same reason as the above [comment](https://github.com/leanprover-community/batteries/pull/925/files#r1720862336) because the corresponding Array API uses Satisfies. I am grateful for any help with these situations

Batteries/Data/Vector/Monadic.lean
```diff
@@ -0,0 +1,171 @@
...
+/--
+`mapIdx f v` maps a monadic function `f : Fin n → α → m β` that takes vector
+elements and their index, and maps them over `v`. It returns the resultant
+vector in the same monad `m`.
+-/
+@[inline]
+def mapIdxM {α : Type u} {β : Type v} {m : Type v → Type w} [Monad m]
+  (v : Vector α n) (f : Fin n → α → m β) : m (Vector β n) := do
+    return ⟨←Array.mapIdxM v.toArray (v.size_eq.symm ▸ f),
+      by rw[←v.size_eq]; sorry⟩
```
Shreyas4991, line 84: Another `sorry` that I am stuck at because the relevant API lemma uses the `Satisfies` predicate.

kim-em:
Most of these have since been implemented upstream. I'm going to close now, but if you'd like to open a PR filling any remaining gaps, that would be great. 

See https://github.com/leanprover/lean4/blob/master/tests/lean/run/list_monadic_functions.lean for a map of the current implementation status.

Shreyas4991:
@kim-em : the defs `modify`, ` modifyM`, `forRevM`, `findRevM`, `findSomeRevM`, and `findIdxM` are already in this PR and not yet added in that list.

kim-em:
Ok, if those are ready could you open a new PR containing just those?

Shreyas4991:
> Ok, if those are ready could you open a new PR containing just those?

#1107

PR #934, opened 2024-08-26
chore: robustify a proof

semorrison:


semorrison:
Ugh, this is not what I meant to commit. Trying again.

PR #936, opened 2024-08-26
chore: use emoji variant of unicode characters

joneugster:
In line with /leanprover/lean4#5015 use emoji variants of ✅️ and ❌️ in the entire Lean universe.


joneugster:
Note, this appends `\u{FE0F}` to these three emoji characters.

In particular, the following Regex helps searching for the character: `[\u{2705}\u{274C}\u{1F4A5}]\u{FE0F}\u{FE0F}`

PR #948, opened 2024-09-14
feat: add `#help note` command

edegeltje:
on [zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Where.20to.20find.20.22notes.22) it was mentioned that there used to be an easy way to find and/or read library notes, but that this feature no longer works, presumably since lean4.

This PR aims to remedy the situation by introducing the `#help note "some tag"` command, which displays all library notes marked with the tag "some tag" that are declared before the command, including in imports.


edegeltje:
awaiting review

edegeltje:
awaiting-review

fgdorais:
The Mathlib test `help_cmd.lean` will need to be updated as a result of this. Was this intended?

edegeltje:
> Was this expected?

when writing this PR i certainly did not take this in consideration, so it was not something i was expecting...
still, looking at how the failing test is written, it is not a surprise it "broke", as its success does indeed depend on what commands start with "#chec" despite that not being the goal of the test.

fgdorais:
Indeed, that is a poorly designed test. On the other hand, that test's output shows that `#check_notes` doesn't quite fit with the other `#check` commands. It fits better with `#print` commands but not perfectly. Perhaps it should be the first member of the `#show` or `#doc` commands. What do you thin?

edegeltje:
i guess it could be a `#help` command, but i think `#help` is in mathlib? i think `#show_notes` is an acceptable alternative name. What mathlib does with `#help` is their concern, they are free to reuse the code here to get the output.

fgdorais:
Actually, it makes sense to move `#help` here or even to Lean since it's useful to everyone and nothing about it is Mathlib specific. 🤔

fgdorais:
The `#help` command is now in Batteries. This PR can now be adapted as the `#help notes` subcommand.

edegeltje:
WIP

edegeltje:
awaiting-review

edegeltje:
unfortunately it seems i can't write a test to make sure the command indeed does work across imports, as i can't figure out how to import a test file in another test file, and writing a library note in the actual module for the sole purpose of testing might not be desirable

edegeltje:
particularly it is impossible to write a library note in the same file that the environment extension is defined

fgdorais:
This is what `Batteries.Test.Internal` is for! You can add `DummyLibraryNote.lean` in that folder and import it in the test file.

edegeltje:
WIP

edegeltje:
awaiting-review

fgdorais:
You need to add the dummy files to `Batteries.lean`.

This looks really good! Other `#help` subcommands support prefix search, I don't think that's essential but it would be nice.

I'll be busy for the rest of the day, so I'll come back for a review tomorrow.

edegeltje:
i added code for prefix search, but left it in comments since it depends on `List.nil_ne_mem_groupBy` (substituted with `sorry`) from mathlib... if the lemma could get upstreamed, we can enable the feature (and add appropriate tests)

fgdorais:
You can use `List.head!` instead of `List.head`.

edegeltje:
ok, it should have all the desired features now. please make sure the documentation and tests are sufficient.

Batteries/Util/LibraryNote.lean
```diff
@@ -35,6 +35,45 @@ creates a new "library note", which can then be cross-referenced using
...
+-/
+elab "#help note" name:strLit : command => do
+  let env ← getEnv
+
+  -- get the library notes from both this and imported files
+  let local_entries := (libraryNoteExt.getEntries env).reverse
+  let imported_entries := (libraryNoteExt.toEnvExtension.getState env).importedEntries
+
+  -- filter for the appropriate notes while casting to list
+  let label_prefix := name.getString.toLower
```
fgdorais, line 61: Are tags really case insensitive?

Batteries/Util/LibraryNote.lean
```diff
@@ -35,6 +35,45 @@ creates a new "library note", which can then be cross-referenced using
...
+  let imported_entries_filtered := imported_entries.flatten.toList.filterMap
+    (fun x => if label_prefix.isPrefixOf x.fst.toLower then some x else none)
+  let valid_entries := imported_entries_filtered ++ local_entries.filterMap
+    (fun x => if label_prefix.isPrefixOf x.fst.toLower then some x else none)
```
fgdorais, line 65: Style.
```suggestion
  let imported_entries_filtered := imported_entries.flatten.toList.filterMap
    fun x => if label_prefix.isPrefixOf x.fst.toLower then some x else none
  let valid_entries := imported_entries_filtered ++ local_entries.filterMap
    fun x => if label_prefix.isPrefixOf x.fst.toLower then some x else none
```

Batteries/Util/LibraryNote.lean
```diff
@@ -35,6 +35,45 @@ creates a new "library note", which can then be cross-referenced using
...
+
+  -- display results in a readable style
+  if grouped_valid_entries.isEmpty then
+    logInfo "Note not found"
+  else
+    logInfo <| "\n\n".intercalate <|
+      grouped_valid_entries.map (fun l =>
+        "library_note \"" ++ (@List.head! _ ⟨⟨"",""⟩⟩ l
+          ).fst ++ "\"\n"
+        ++ ("\n\n".intercalate <| l.map (fun e => "/--\n" ++ e.snd.trim ++ "\n-/")))
```
fgdorais, line 78: Add `deriving Inhabited` after the definition of `LibraryNoteEntry` to make this work:
```suggestion
    logInfo <| "\n\n".intercalate <|
      grouped_valid_entries.map fun l =>
        "library_note \"" ++ l.head!.fst ++ "\"\n" ++ 
        ("\n\n".intercalate <| l.map fun e => "/--\n" ++ e.snd.trim ++ "\n-/")
```

test/library_note.lean
```diff
@@ -0,0 +1,66 @@
...
+
+/--
+info: Note not found
+-/
+#guard_msgs in
+#help note "no note"
+
+/--
+info: library_note "Other"
+/--
```
fgdorais, line 12: I don't see the utility of wrapping each note in `/-- ... -/`. How about just separating with a blank line or `---`?

fgdorais (changes requested):

Batteries/Util/LibraryNote.lean
edegeltje: no, but i think it's useful to search case-insensitively. needing to remember the exact casing of a label seems like something we can prevent

test/library_note.lean
edegeltje: i think the benefit is that this is less likely to cause confusion as to where one note stops and another begins, and this mimics the way you'd find the note in the actual code.

Batteries/Util/LibraryNote.lean
edegeltje: aha... so `String.toLower` only works on basic latin characters... that's good to know. I'll think about making the a PR addressing the insensitivity.

test/library_note.lean
fgdorais: I guess my issue is that `/-- ... -/` has a specific meaning in Lean, which is not relevant here. So, for example, copy pasting the output in a Lean file would lead to multiple errors. How about `/- ... -/`? (No errors then, though that still seems visually cluttered.)

digama0: We aren't doing this for any of the other `#help` commands, I think the docs should just be indented relative to the `library_note` header.

fgdorais: I reflected on this a bit and I think `#help` subcommands should give output that is better than plain code lookup in some useful way. In many cases organization is the key difference but formatting is relevant too. I don't know what would be the best formatting for library notes. How about asking on Zulip what users would like to see for library notes?

fgdorais:
By the way, no worries about the current `breaks-mathlib` label. That will be fixed upstream once [mathlib4#17257](https://github.com/leanprover-community/mathlib4/pull/17257) gets merged.

edegeltje:
> By the way, no worries about the current `breaks-mathlib` label. That will be fixed downstream once [mathlib4#17257](https://github.com/leanprover-community/mathlib4/pull/17257) gets merged.

actually, it looks to me like the script trying to find this branch fails...
see for example how [this run](https://github.com/leanprover-community/batteries/actions/runs/11411018585/job/31754490132) fails with an error trying to get info about the PR triggering the ci, from what i can tell... so even before looking at mathlib

fgdorais:
Yes, the script is broken right now, we're doing manual testing for the time being.

test/library_note.lean
edegeltje: i asked, and the only suggestion it got was that they would like if it would display the notes with markdown. AFAIK such functionality doesn't exist yet, and i don't think creating this functionality is in scope for this PR...

fgdorais: This suggests using markdown style for bulleted lists.
```
* This is a note.

  Paragraphs within a note are separated by a blank line and indented.

* This is another note.
```

test/library_note.lean
```diff
@@ -0,0 +1,61 @@
+import Batteries.Util.LibraryNote
+import Batteries.Test.Internal.DummyLibraryNote2
+
+/--
+info: Note not found
```
digama0, line 5: this should be an error

Batteries/Util/LibraryNote.lean
```diff
@@ -35,6 +36,47 @@ creates a new "library note", which can then be cross-referenced using
...
+
+open Lean Parser in
+/--
+`#help note "foo"` searches (case-insensitively) for all library notes whose
+label starts with "foo", then displays those library notes sorted alphabetically by label,
+grouped by label.
+The command only displays the library notes that are declared in
+imported files or in the same file above the line containing the command.
+-/
+elab "#help note" name:strLit : command => do
```
fgdorais, line 58:
```suggestion
elab "#help " colGt &"note" colGt ppSpace name:strLit : command => do
```

fgdorais:
This is looking excellent and it now passes Mathlib CI!

One final thing: the command should be in Batteries.Tactic.HelpCmd along with the other `#help` commands and the module doc there should be updated.

edegeltje:
> One final thing: the command should be in Batteries.Tactic.HelpCmd along with the other `#help` commands and the module doc there should be updated.

in my opinion, it is more natural to have the tool here, in order to make sure it is available wherever library notes are...

fgdorais:
In any case, the module doc at Batteries.Tactic.HelpCmd needs to be updated.

I don't understand why you claim it's not a help command. Just like the other help commands, it's handy for users when writing code but it is not useful as code. That's why HelpCmd is not imported anywhere in Batteries and Mathlib except in the global import files. I don't see why `#help note` should be any different than the others.

fgdorais:
I'm afraid I must insist on moving to Batteries.Tactic.HelpCmd for maintainability reasons. Your proposed alternative has some issues (1) importing Batteries.Tactic.HelpCmd doesn't import all help commands and (2) importing Batteries.Util.LibraryNotes into Batteries.Util.LibraryNote would not pass shake. I see no other way around these problems.

edegeltje:
@fgdorais how about we put an alias of the command in `Batteries.Tactic.HelpCmd`? that should fix both issues, and still make sure that the functionality exists wherever you can make library notes. the "original" can then be something like `#find_note "label"`, and there can be a simple macro turning `#help note` into that...
or does that still give maintainability issues?

fgdorais:
That's too hacky for Batteries. Like I said before the reason why help commands like this one aren't available everywhere is that they are only useful for users while writing code, but of no use in the code itself. You will just have to get used to importing help commands while writing a Batteries or Mathlib PR and then deleting the import before submitting.

edegeltje:
awaiting-review

edegeltje:
i don't see the issue with having this command always available. AFAIK, it shouldn't be that impactful if these are or aren't imported.

fgdorais:
FWIW, I think Batteries.Tactic.HelpCmd would be a nice addition to Mathlib.Init.

Batteries/Tactic/HelpCmd.lean
```diff
@@ -16,12 +17,14 @@ The `#help` command can be used to list all definitions in a variety of extensib
 * `#help attr` lists attributes (used in `@[myAttr] def foo := ...`)
 * `#help cats` lists syntax categories (like `term`, `tactic`, `stx` etc)
 * `#help cat C` lists elements of syntax category C
+* `#help note "some note"` lists library notes for which "some note" is a prefix of the label
   * `#help term`, `#help tactic`, `#help conv`, `#help command`
     are shorthand for `#help cat term` etc.
   * `#help cat+ C` also shows `elab` and `macro` definitions associated to the syntaxes
```
fgdorais, line 23:
```suggestion
  * `#help term`, `#help tactic`, `#help conv`, `#help command`
    are shorthand for `#help cat term` etc.
  * `#help cat+ C` also shows `elab` and `macro` definitions associated to the syntaxes
* `#help note "some note"` lists library notes for which "some note" is a prefix of the label
```

Batteries/Tactic/HelpCmd.lean
```diff
@@ -276,6 +279,45 @@ elab_rules : command
...
+
+  -- display results in a readable style
+  if grouped_valid_entries.isEmpty then
+    logError "Note not found"
+  else
+    logInfo <| "\n\n".intercalate <|
+      grouped_valid_entries.map
+        fun l => "library_note \"" ++ l.head!.fst ++ "\"\n" ++
+          "\n\n".intercalate (l.map (·.snd.trim.makeBullet))
+    -- this could use List.head when List.ne_nil_of_mem_groupBy gets upstreamed from mathlib
```
fgdorais, line 319:
```suggestion
```

Batteries/Tactic/HelpCmd.lean
```diff
@@ -276,6 +279,45 @@ elab_rules : command
...
   | `(#help cat $[+%$more]? $cat $id:str) => elabHelpCat more cat id.getString
 
+/--
+format the string to be included in a single markdown bullet
+-/
+def _root_.String.makeBullet (s:String) := "* " ++ ("\n  ").intercalate (s.splitOn "\n")
+
+open Lean Parser Batteries.Util.LibraryNote in
+/--
+`#help note "foo"` searches (case-insensitively) for all library notes whose
```
fgdorais, line 289:
```suggestion
`#help note "foo"` searches for all library notes whose
```

fgdorais (changes requested):

fgdorais (approved):

PR #950, opened 2024-09-19
chore: simplify lakefile

kim-em:
Per discussion with Mario and Mac last night.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         name: build, test, and lint batteries
         uses: leanprover/lean-action@v1
         with:
-          build-args: '-Kwerror'
+          build-args: '---wfail'
 
       - name: Check that all files are imported
         run: lake env lean scripts/check_imports.lean
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         name: build, test, and lint batteries
         uses: leanprover/lean-action@v1
         with:
-          build-args: '---wfail'
+          build-args: '--wfail'
 
       - name: Check that all files are imported
         run: lake env lean scripts/check_imports.lean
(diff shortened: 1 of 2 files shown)


fgdorais:
Should we use TOML format?

semorrison:
Yes, but I wanted to go in two steps. I'll do that as soon as this is merged.

fgdorais:
Perhaps also remove `-Kwerror` from the build workflow.

semorrison:
> Perhaps also remove `-Kwerror` from the build workflow.

Separate PR please? Seems orthogonal to this.

fgdorais:
Does `-Kwerror` do anything without this lakefile? Sorry if I missed that somewhere upstream.

digama0:
No, I think we still want `-Kwerror`. That option is the reason why we don't need `warningAsError=true` anymore (and conditional options in general are the main blocker for moving batteries' lakefile to toml format).

fgdorais:
Ah! Found it. I think it's called `--wfail` though, I'm pretty sure `-Kwerror` won't do anything after changing the lakefile.

kim-em:
@fgdorais, I can't tell from your last message what you want done here.

fgdorais:
The `-Kwerror` in the build.yml file should be replaced by `--wfail`.

With this new lakefile `-Kwerror` does nothing: lake passes the `werror` option to the lakefile and the lakefile ignores it. To keep fail-on-warning working, the correct option is `--wfail` which tells lake itself to fail-on-warning. (I originally thought disabling fail-on-warning was part of the plan you came up with Mac and Mario, which is the root of all this confusion. I think we're all on the same page now.)

.github/workflows/build.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         name: build, test, and lint batteries
         uses: leanprover/lean-action@v1
         with:
-          build-args: '-Kwerror'
+          build-args: '---wfail'
```
fgdorais:
```suggestion
          build-args: '--wfail'
```

fgdorais (approved):

PR #951, opened 2024-09-21
chore: adapt to lean #5346

lyphyser:
qsort parameter name changed from "lt" to "f" since it now can be a <= in addition to a < and "lt" would be misleading.


kim-em:
Closing, as leanprover/lean4#5346 has stalled.

PR #953, opened 2024-09-22
feat: decidable quantifiers for `Vector`

kim-em:
- [x] depends on: #952

### DIFF
--- a/Batteries/Data/Vector/Lemmas.lean
+++ b/Batteries/Data/Vector/Lemmas.lean
@@ -46,25 +46,25 @@ protected theorem ext {a b : Vector α n} (h : (i : Nat) → (_ : i < n) → a[i
     rw [a.size_eq] at hi
     exact h i hi
 
-@[simp] theorem getElem_mk (data : Array α) (size : data.size = n) (i : Nat) (h : i < n) :
+@[simp] theorem getElem_mk {data : Array α} {size : data.size = n} {i : Nat} (h : i < n) :
     (Vector.mk data size)[i] = data[i] := rfl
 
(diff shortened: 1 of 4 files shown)


Batteries/Data/Vector/Lemmas.lean
```diff
@@ -45,3 +46,83 @@ protected theorem ext {a b : Vector α n} (h : (i : Nat) → (_ : i < n) → a[i
...
+      Vector.mk (data.push x) (by simp [size, Nat.succ_eq_add_one]) := rfl
+
+@[simp] theorem pop_mk {data : Array α} {size : data.size = n} :
+    (Vector.mk data size).pop = Vector.mk data.pop (by simp [size]) := rfl
+
+@[simp] theorem getElem_push_last {v : Vector α n} {x : α} : (v.push x)[n] = x := by
+  rcases v with ⟨data, rfl⟩
+  simp
+
+-- The `simpNF` linter incorrectly claims that this lemma can not be applied by `simp`.
```
digama0, line 64: this is a misplaced bug report

semorrison: Sorry, what do you want done with it? There's plenty more in Mathlib. :-)

It seems a bad idea to add `@[nolint simpNF]` without explanation when we know it is wrong.

digama0: I'm saying that this should not be a `nolint`, as it's working around a bug in the linter itself. It should be a bug report, and if you have to deal with the issue in the meantime, either use `nolints.json` or `@[nolint]` with some marking that makes it clear that it is not a genuine `@[nolint]` and should be removed later

fgdorais (approved):

PR #956, opened 2024-09-24
feat: compare List.ofFn and Array.ofFn

kim-em:


Batteries/Data/Array/OfFn.lean
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Array
+
+/-! ### ofFn -/
+
+@[simp]
+theorem data_ofFn (f : Fin n → α) : (ofFn f).data = List.ofFn f := by
+  ext1
+  rw [List.getElem?_eq]
+  simp [ofFnNthVal]
```
mattrobball, line 18:
```suggestion
  simp only [getElem?_eq, data_length, size_ofFn, length_ofFn, getElem_ofFn]
```

Batteries/Data/List/OfFn.lean
```diff
@@ -0,0 +1,49 @@
...
+
+@[simp]
+theorem length_ofFn (f : Fin n → α) : (ofFn f).length = n := by
+  simp only [ofFn]
+  induction n with
+  | zero => simp
+  | succ n ih => simp [Fin.foldr_succ, ih]
+
+@[simp]
+theorem getElem_ofFn (f : Fin n → α) (i : Nat) (h : i < (ofFn f).length) :
```
mattrobball, line 23: It doesn't look like this is a concern currently in the code base but should this and `getElem?_ofFn` be `protected`?

mattrobball (approved):

Batteries/Data/List/OfFn.lean
kim-em: Sure. Better done early.

mattrobball (approved):
Very nice!

PR #958, opened 2024-09-26
feat: automatically test Mathlib against Batteries PRs

kim-em:
This will only run once it is on master, so we can only test in production.

Will require a secret token installed.

Also needs an update at Mathlib to actually report back, but we could add that afterwards.


fgdorais (approved):

kim-em:
I've added the requisite tokens/secrets, documented in the usual place, so hopefully this is now ready for testing in production.

PR #968, opened 2024-09-27
fix: use `by rfl` instead of `rfl` to bypass new compiler bug

Shreyas4991:
This is a workaround for a bug reported on the [zulip thread](https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/Large.20vector.20hangs/near/473252482)

### DIFF
--- a/Batteries/Data/Vector/Basic.lean
+++ b/Batteries/Data/Vector/Basic.lean
@@ -34,7 +34,7 @@ syntax "#v[" withoutPosition(sepBy(term, ", ")) "]" : term
 
 open Lean in
 macro_rules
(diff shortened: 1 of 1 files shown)


Shreyas4991:
awaiting-review

fgdorais:
This works and it's simple but the ideal is an upstream solution, so we should wait a bit for a better solution. Otherwise, I would approve.

mattrobball:
I think `by rfl` will always be more robust than `rfl` regardless of core changes. I am inclined to merge this but want to check with @fgdorais

fgdorais:
I can't reproduce the original bug since `v4.13-rc1`. I'm not sure why the issue isn't closed upstream.

mattrobball:
I went out to 68 and the problem is gone so let's close this. 

It would be informative to ping on leanprover/lean4#5502 to see if any changes directed at this landed.

PR #969, opened 2024-09-28
feat: add `#help` commands

fgdorais:
- [X] Mathlib adaptation [#17257](https://github.com/leanprover-community/mathlib4/pull/17257)

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/348111-batteries/topic/Upstreaming.20.23help.20command.3F/near/473358619)

Co-authored-by: Mario Carneiro <di.gama@gmail.com>

### DIFF
--- a/Batteries.lean
+++ b/Batteries.lean
@@ -76,6 +76,7 @@ import Batteries.Tactic.Case
 import Batteries.Tactic.Classical
 import Batteries.Tactic.Congr
 import Batteries.Tactic.Exact
+import Batteries.Tactic.HelpCmd
 import Batteries.Tactic.Init
 import Batteries.Tactic.Instances
 import Batteries.Tactic.Lemma
--- a/Batteries/Tactic/HelpCmd.lean
+++ b/Batteries/Tactic/HelpCmd.lean
@@ -1,0 +1,319 @@
+/-
+Copyright (c) 2022 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro
+-/
+import Lean.Elab.Syntax
+import Lean.DocString
+
+/-!
+
+# The `#help` command
+
+The `#help` command can be used to list all definitions in a variety of extensible aspects of lean.
+
+* `#help option` lists options (used in `set_option myOption`)
+* `#help attr` lists attributes (used in `@[myAttr] def foo := ...`)
+* `#help cats` lists syntax categories (like `term`, `tactic`, `stx` etc)
+* `#help cat C` lists elements of syntax category C
+  * `#help term`, `#help tactic`, `#help conv`, `#help command`
+    are shorthand for `#help cat term` etc.
+  * `#help cat+ C` also shows `elab` and `macro` definitions associated to the syntaxes
+
+All forms take an optional identifier to narrow the search; for example `#help option pp` shows
+only `pp.*` options.
+
+-/
+
+namespace Mathlib.Tactic
+open Lean Meta Elab Tactic Command
+
+/--
+The command `#help option` shows all options that have been defined in the current environment.
+Each option has a format like:
+```
+option pp.all : Bool := false
+  (pretty printer) display coercions, implicit parameters, proof terms, fully qualified names,
+  universe, and disable beta reduction and notations during pretty printing
+```
+This says that `pp.all` is an option which can be set to a `Bool` value, and the default value is
+`false`. If an option has been modified from the default using e.g. `set_option pp.all true`,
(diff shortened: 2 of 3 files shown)


kim-em:
I would really like to see more thorough tests on this one, exercising most of the major functionality.

Also, the giant `match` in `getHeadTk` looks very fragile, and while this code lives in a downstream repository from where the logic that determines what needs to be on this list, it is going to break unpredictably. Tests that actually confirmed that this list was sufficient for expected behaviour would be really helpful.

fgdorais:
I added a bunch more tests. Not quite comprehensive yet, I'll have a second pass soon.

Doing so, I realized that `getHeadTk` is indeed not quite sufficient. For example, the current output of `#help command "#help"` is ironically not as helpful as it could be (`#print` is worse!).
``````
syntax "#help"... [Batteries.Tactic.«command#help_Term+____»]
  The command `#help term` shows all term syntaxes that have been defined in the current environment.
  See `#help cat` for more information.

syntax "#help"... [Batteries.Tactic.«command#help_Cat+______»]
  The command `#help cat C` shows all syntaxes that have been defined in syntax category `C` in the
  current environment.
  Each syntax has a format like:
  ```
  syntax "first"... [Parser.tactic.first]
    `first | tac | ...` runs each `tac` until one succeeds, or else fails.
  ```
  The quoted string is the leading token of the syntax, if applicable. It is followed by the full
  name of the syntax (which you can also click to go to the definition), and the documentation.
  
  * The form `#help cat C id` will show only attributes that begin with `id`.
  * The form `#help cat+ C` will also show information about any `macro`s and `elab`s
    associated to the listed syntaxes.

syntax "#help"... [Batteries.Tactic.«command#help_Command+____»]
  The command `#help command` shows all commands that have been defined in the current environment.
  See `#help cat` for more information.

syntax "#help"... [Batteries.Tactic.«command#help_AttrAttribute___»]
  The command `#help attribute` (or the short form `#help attr`) shows all attributes that have been
  defined in the current environment.
  Each attribute has a format like:
  ```
  [inline]: mark definition to always be inlined
  ```
  This says that `inline` is an attribute that can be placed on definitions like
  `@[inline] def foo := 1`. (Individual attributes may have restrictions on where they can be
  applied; see the attribute's documentation for details.) Both the attribute's `descr` field as well
  as the docstring will be displayed here.
  
  The form `#help attr id` will show only attributes that begin with `id`.

syntax "#help"... [Batteries.Tactic.«command#help_Cats___»]
  The command `#help cats` shows all syntax categories that have been defined in the
  current environment.
  Each syntax has a format like:
  ```
  category command [Lean.Parser.initFn✝]
  ```
  The name of the syntax category in this case is `command`, and `Lean.Parser.initFn✝` is the
  name of the declaration that introduced it. (It is often an anonymous declaration like this,
  but you can click to go to the definition.) It also shows the doc string if available.
  
  The form `#help cats id` will show only syntax categories that begin with `id`.

syntax "#help"... [Batteries.Tactic.«command#help_Tactic+____»]
  The command `#help tactic` shows all tactics that have been defined in the current environment.
  See `#help cat` for more information.

syntax "#help"... [Batteries.Tactic.«command#help_Conv+____»]
  The command `#help conv` shows all tactics that have been defined in the current environment.
  See `#help cat` for more information.

syntax "#help"... [Batteries.Tactic.«command#help_Option___»]
  The command `#help option` shows all options that have been defined in the current environment.
  Each option has a format like:
  ```
  option pp.all : Bool := false
    (pretty printer) display coercions, implicit parameters, proof terms, fully qualified names,
    universe, and disable beta reduction and notations during pretty printing
  ```
  This says that `pp.all` is an option which can be set to a `Bool` value, and the default value is
  `false`. If an option has been modified from the default using e.g. `set_option pp.all true`,
  it will appear as a `(currently: true)` note next to the option.
  
  The form `#help option id` will show only options that begin with `id`.  
``````
The issue is that `#help` (and `#print` and others) are command families that have more than one head token.

Things get even more complicated if we want to allow optional prefix filters to contain spaces. (I don't think that's essential for now.)

Syntax is not my forte, so I'm (unironically!) adding the `help-wanted` tag to figure out what the best thing to do with `getHeadTk`.

digama0:
`getHeadTk` is at least behaving as designed here. It's always been an issue that we have limited ability to reconstruct the syntax of a declaration. I don't think that should block this from being used such as it is. I think there are tons of ways `#help` can be improved, it was hacked together by me because we had nothing even close to it, and we still don't.

fgdorais:
OK. I think I'm done with a comprehensive test suite. I removed the `help-wanted` tag but I think I will add an issue if and when this gets merged.

PR #984, opened 2024-10-12
feat: add Mersenne Twister PRNG

fgdorais:
- [x] Depends on #985 (for testing only)

-----

TODO:
- [X] Add tests.
- [x] Improve docs.

The implementation is very general. Probably still reasonably efficient but we could link to an optimized external implementation for the standard MT19937 and MT19937-64 algorithms.


Batteries/Data/Random/MersenneTwister.lean
```diff
@@ -0,0 +1,143 @@
...
+structure State (cfg : Config) where
+  /-- Data for current state. -/
+  data : Vector (BitVec cfg.wordSize) cfg.stateSize
+  /-- Current data index. -/
+  index : Fin cfg.stateSize
+
+/-- Mersenne Twister initialization given an optional seed. -/
+@[specialize cfg] protected def Config.init (cfg : MersenneTwister.Config)
+    (seed : BitVec cfg.wordSize := cfg.initSeed) : State cfg :=
+  ⟨loop seed #[] (Nat.zero_le _), 0, cfg.zero_lt_stateSize⟩
```
digama0, line 75:
```suggestion
  ⟨loop seed (Array.withCapacity cfg.stateSize) (Nat.zero_le _), 0, cfg.zero_lt_stateSize⟩
```

Batteries/Data/Random/MersenneTwister.lean
```diff
@@ -0,0 +1,143 @@
...
+where
+  /-- Inner loop for Mersenne Twister initalization. -/
+  loop (w : BitVec cfg.wordSize) (v : Array (BitVec cfg.wordSize)) (h : v.size ≤ cfg.stateSize) :=
+    if heq : v.size = cfg.stateSize then ⟨v, heq⟩ else
+      let v := v.push w
+      let w := cfg.initMult * (w ^^^ (w >>> cfg.wordSize - 2)) + v.size
+      loop w v (by simp only [v, Array.size_push]; omega)
+
+/-- Update the state by a number of generation steps (default 1). -/
+@[specialize cfg] protected def State.update (state : State cfg) (steps := 1) : State cfg :=
```
digama0, line 85: Why does this have a loop built in when it is not used below? How about having `update1` as a separate function and using that instead in `State.next`?

fgdorais: Also, the `update` needs to be updated to run in time `O(log(steps))` instead of the current `O(steps)`. But this is not high priority on my end.

Batteries/Data/Random/MersenneTwister.lean
```diff
@@ -81,24 +81,23 @@ where
...
-  /-- Inner loop for Mersenne Twister update. -/
-  @[inline] loop (s : State cfg) (c : Nat) : State cfg :=
-    if c = 0 then s else
-      let i := s.index
-      let i' : Fin cfg.stateSize :=
-        if h : i.val+1 < cfg.stateSize then ⟨i.val+1, h⟩ else ⟨0, cfg.zero_lt_stateSize⟩
-      let y := s.data[i] &&& cfg.uMask ||| s.data[i'] &&& cfg.lMask
-      let x := s.data[i+cfg.shiftSize] ^^^ bif y[0] then y >>> 1 ^^^ cfg.xorMask else y >>> 1
-      loop ⟨s.data.set i x, i'⟩ (c-1)
+@[inline] protected def State.update (state : State cfg) (steps := 1) : State cfg :=
```
digama0, line 94: `@[inline]`? Also this could be doing structural recursion instead of well founded recursion if you use a match instead of an if statement, it should come out the same.

fgdorais: Fixed. It's kind of moot since it needs a major refactor down the road anyway.

digama0:
Needs a test file, otherwise LGTM

PR #985, opened 2024-10-13
feat: add `drop` and `take` for streams

fgdorais:


Batteries/Data/Stream.lean
```diff
@@ -0,0 +1,93 @@
...
+/-- Drop up to `n` values from the stream `s`. -/
+def drop [Stream σ α] (s : σ) : Nat → σ
+  | 0 => s
+  | n+1 =>
+    match next? s with
+    | none => s
+    | some (_, s) => drop s n
+
+/-- Read up to `n` values from the stream `s`. -/
+def take [Stream σ α] (s : σ) : Nat → List α × σ
```
mattrobball, line 18: What is the rationale for `List` in place of `Array`

mattrobball (commented):
LGTM. Could you add a comment about the choice of ordering for the extracted values in `drop`?

Batteries/Data/Stream.lean
fgdorais: Good question. I don't recall thinking about it much but it's a natural choice since `List` is a `Stream` but `Array` is not.

mattrobball: I can't think of major applications that weigh one way or the other.

PR #987, opened 2024-10-14
feat: add list function and lemmas

chabulhwi:
* Add 'leftover' lemmas I created while trying to prove
  `String.splitOn_of_valid`. See
  https://github.com/leanprover-community/batteries/pull/743.
* Add the function `List.commonPrefix` and its lemmas. It returns the
  longest common prefix of two lists.

### DIFF
--- a/Batteries/Data/List/Lemmas.lean
+++ b/Batteries/Data/List/Lemmas.lean
@@ -504,6 +504,16 @@ section
 
 variable [DecidableEq α]
 
+theorem commonPrefix_comm (l₁ l₂ : List α) : commonPrefix l₁ l₂ = commonPrefix l₂ l₁ := by
+  cases l₁ <;> cases l₂ <;> simp only [commonPrefix]
+  next a₁ l₁ a₂ l₂ =>
+  split
+  · subst a₁
+    simp only [↓reduceIte, cons.injEq, true_and]
+    apply commonPrefix_comm
+  · next h =>
+    simp [Ne.symm h]
+
 theorem commonPrefix_prefix_left (l₁ l₂ : List α) : commonPrefix l₁ l₂ <+: l₁ := by
   match l₁, l₂ with
   | [],   _  => simp [commonPrefix]
--- a/Batteries/Data/List/Lemmas.lean
+++ b/Batteries/Data/List/Lemmas.lean
@@ -546,24 +546,35 @@ theorem commonPrefix_append_append (p l₁ l₂ : List α) :
   | []   => rfl
   | a::p => simpa [commonPrefix] using commonPrefix_append_append p l₁ l₂
 
-theorem not_prefix_and_not_prefix_symm_iff_exists {l₁ l₂ : List α} :
-    ¬l₁ <+: l₂ ∧ ¬l₂ <+: l₁ ↔ ∃ c₁ c₂ pre suf₁ suf₂, c₁ ≠ c₂ ∧ l₁ = pre ++ c₁ :: suf₁ ∧
-      l₂ = pre ++ c₂ :: suf₂ := by
+theorem not_prefix_and_not_prefix_symm_iff {l₁ l₂ p t₁ t₂ : List α} (hp : commonPrefix l₁ l₂ = p)
+    (h₁ : l₁ = p ++ t₁) (h₂ : l₂ = p ++ t₂) : ¬l₁ <+: l₂ ∧ ¬l₂ <+: l₁ ↔ t₁ ≠ [] ∧ t₂ ≠ [] ∧
+      t₁.head? ≠ t₂.head? := by
   constructor <;> intro h
-  · obtain ⟨c₁, l₁, rfl⟩ := l₁.exists_cons_of_ne_nil (ne_nil_of_not_prefix h.1)
-    obtain ⟨c₂, l₂, rfl⟩ := l₂.exists_cons_of_ne_nil (ne_nil_of_not_prefix h.2)
-    simp only [cons_prefix_cons, not_and] at h
-    cases Decidable.em (c₁ = c₂)
-    · subst c₂
-      simp only [forall_const] at h
-      let ⟨c₁', c₂', pre, suf₁, suf₂, hc, heq₁, heq₂⟩ :=
-        not_prefix_and_not_prefix_symm_iff_exists.mp h
-      exact ⟨c₁', c₂', c₁::pre, suf₁, suf₂, hc, by simp [heq₁], by simp [heq₂]⟩
-    · next hc =>
-      exact ⟨c₁, c₂, [], l₁, l₂, hc, nil_append .., nil_append ..⟩
-  · let ⟨c₁, c₂, pre, suf₁, suf₂, hc, heq₁, heq₂⟩ := h
-    rw [heq₁, heq₂]
-    simp [prefix_append_right_inj, cons_prefix_cons, hc, hc.symm]
+  · obtain ⟨a₁, l₁, rfl⟩ := l₁.exists_cons_of_ne_nil (ne_nil_of_not_prefix h.1)
+    obtain ⟨a₂, l₂, rfl⟩ := l₂.exists_cons_of_ne_nil (ne_nil_of_not_prefix h.2)
+    cases Decidable.em (a₁ = a₂)
+    · subst a₂
+      simp only [← hp, commonPrefix, ↓reduceIte, cons_append, cons.injEq, true_and] at h₁ h₂
+      simp only [cons_prefix_cons, true_and] at h
+      exact (not_prefix_and_not_prefix_symm_iff rfl h₁ h₂).mp h
+    · next hne =>
+      simp only [← hp, commonPrefix, hne, ↓reduceIte, nil_append] at h₁ h₂
+      simp [← h₁, ← h₂, hne]
+  · rw [h₁, h₂]
+    let ⟨ht₁, ht₂, hhd⟩ := h
+    obtain ⟨a₁, t₁, rfl⟩ := exists_cons_of_ne_nil ht₁
+    obtain ⟨a₂, t₂, rfl⟩ := exists_cons_of_ne_nil ht₂
+    simp only [head?_cons, ne_eq, Option.some.injEq] at hhd
+    rw [← ne_eq] at hhd
+    simp [prefix_append_right_inj, not_and_of_not_left _ hhd, not_and_of_not_left _ hhd.symm]
+termination_by l₁.length
+decreasing_by
+  rename_i _₀ _l₁ _₁ _₂ _₃ _₄ _₅ _₆ _₇ _r _₈ _₉ _₁₀ _₁₁ _₁₂ _₁₃ _₁₄ _₁₅ _₁₆ _₁₇ _₁₈ _hl₁ _₁₉ _₂₀ _₂₁
+    _₂₂ _₂₃ _₂₄ _₂₅ _₂₆ _₂₇
+  clear _₀ _₁ _₂ _₃ _₄ _₅ _₆ _₇ _r _₈ _₉ _₁₀ _₁₁ _₁₂ _₁₃ _₁₄ _₁₅ _₁₆ _₁₇ _₁₈ _₁₉ _₂₀ _₂₁ _₂₂ _₂₃ _₂₄
+    _₂₅ _₂₆ _₂₇
+  subst _l₁
+  simp
 
 end
 
--- a/Batteries/Data/List/Lemmas.lean
+++ b/Batteries/Data/List/Lemmas.lean
@@ -528,17 +528,8 @@ theorem commonPrefix_prefix_left (l₁ l₂ : List α) : commonPrefix l₁ l₂ 
       simp [h]
 
 theorem commonPrefix_prefix_right (l₁ l₂ : List α) : commonPrefix l₁ l₂ <+: l₂ := by
-   match l₁, l₂ with
-  | [],   _  => simp [commonPrefix]
-  | _::_, [] => simp [commonPrefix]
-  | a₁::l₁, a₂::l₂ =>
-    simp only [commonPrefix]
-    cases Decidable.em (a₁ = a₂)
-    · next h =>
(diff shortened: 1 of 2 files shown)


chabulhwi:
awaiting-review

Batteries/Data/List/Lemmas.lean
```diff
@@ -19,6 +19,15 @@ namespace List
 
 @[deprecated (since := "2024-08-15")] alias isEmpty_iff_eq_nil := isEmpty_iff
 
+/-! ### head and tail -/
+
+theorem head_cons_tail : ∀ l h, @head α l h :: l.tail = l
```
eric-wieser, line 24: Can you write this without the `@`?

Batteries/Data/List/Lemmas.lean
```diff
@@ -482,6 +491,34 @@ theorem Sublist.erase_diff_erase_sublist {a : α} :
 
 end Diff
 
+/-! ### prefix, suffix, infix -/
+
+theorem singleton_prefix_cons (a) : [a] <+: a :: l :=
+  (prefix_cons_inj a).mpr nil_prefix
```
eric-wieser, line 497: If suitable `alias`es were added, this could be written `.cons a .nil`, which is then shorter than the lemma name.

Batteries/Data/List/Lemmas.lean
chabulhwi: In the following code, the `@` symbol was used in a similar way: https://github.com/leanprover-community/batteries/blob/5f963d5d06cc3c2d3abd0806891133137a59d7eb/Batteries/Data/List/Lemmas.lean#L22-L25.

Batteries/Data/List/Lemmas.lean
chabulhwi: Can you show me how to add those aliases? I'm not sure how to do it.

chabulhwi: I'd like to hear what other maintainers think about adding these aliases.

Batteries/Data/List/Lemmas.lean
fgdorais: Using `@` is a bit old fashioned. This would math the style of the next theorem:
```suggestion
theorem head_cons_tail : ∀ (l : List α) (hne : l ≠ []), l.head hne :: l.tail = l
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -482,6 +491,34 @@ theorem Sublist.erase_diff_erase_sublist {a : α} :
...
+
+theorem ne_nil_of_not_prefix (h : ¬l₁ <+: l₂) : l₁ ≠ [] := by
+  intro heq
+  simp [heq, nil_prefix] at h
+
+theorem not_prefix_and_not_prefix_symm_iff_exists [BEq α] [LawfulBEq α] [DecidableEq α]
+    {l₁ l₂ : List α} : ¬l₁ <+: l₂ ∧ ¬l₂ <+: l₁ ↔ ∃ c₁ c₂ pre suf₁ suf₂, c₁ ≠ c₂ ∧
+      l₁ = pre ++ c₁ :: suf₁ ∧ l₂ = pre ++ c₂ :: suf₂ := by
```
fgdorais, line 505: Since all the `∃` parts are unique, we might as well have a function that computes the pieces and then prove the specs of that function.

fgdorais: Namely:
```
def matchPrefix (l₁ l₂ : List α) : List α × List α × List α := ...

theorem match_prefix_spec : matchPrefix l₁ l₂ = (p, t₁, t₂) ↔ t₁.head? ≠ t₂.head? ∧ l₁ = p ++ t₁ ∧ l₂ = p ++ t₂ := ...
```

Batteries/Data/List/Lemmas.lean
fgdorais: These types of aliases never really caught on outside Mathlib as far as I can tell. They're convenient for sure but I'm not sure that kind of convenience is a good fit for Batteries. I would rather see more automation than these types of shortcuts as far as Batteries is concerned.

Batteries/Data/List/Lemmas.lean
eric-wieser: `t₁.head? ≠ t₂.head? ` is wrong there, it fails when both lists are equal and the tails are empty

fgdorais: Sorry, I was typing too quick. The lemma assumes that neither list is a prefix of each other. This is a better version:
```lean
theorem matchPrefix_spec : matchPrefix l₁ l₂ = (p, t₁, t₂) ↔ l₁ = l₂ ∨ t₁.head? ≠ t₂.head? ∧ l₁ = p ++ t₁ ∧ l₂ = p ++ t₂ := ...
```

chabulhwi: I renamed and restated the theorem as follows:

```lean
List.theorem not_prefix_and_not_prefix_symm_iff {l₁ l₂ p t₁ t₂ : List α} (hp : commonPrefix l₁ l₂ = p)
    (h₁ : l₁ = p ++ t₁) (h₂ : l₂ = p ++ t₂) : ¬l₁ <+: l₂ ∧ ¬l₂ <+: l₁ ↔ t₁ ≠ [] ∧ t₂ ≠ [] ∧
      t₁.head? ≠ t₂.head? := by
  sorry
```

chabulhwi (commented):
Why do I have to clean up these unwanted hypotheses?

Batteries/Data/List/Lemmas.lean
```diff
@@ -482,6 +491,93 @@ theorem Sublist.erase_diff_erase_sublist {a : α} :
...
+    obtain ⟨a₂, t₂, rfl⟩ := exists_cons_of_ne_nil ht₂
+    simp only [head?_cons, ne_eq, Option.some.injEq] at hhd
+    rw [← ne_eq] at hhd
+    simp [prefix_append_right_inj, not_and_of_not_left _ hhd, not_and_of_not_left _ hhd.symm]
+termination_by l₁.length
+decreasing_by
+  rename_i _₀ _l₁ _₁ _₂ _₃ _₄ _₅ _₆ _₇ _r _₈ _₉ _₁₀ _₁₁ _₁₂ _₁₃ _₁₄ _₁₅ _₁₆ _₁₇ _₁₈ _hl₁ _₁₉ _₂₀ _₂₁
+    _₂₂ _₂₃ _₂₄ _₂₅ _₂₆ _₂₇
+  clear _₀ _₁ _₂ _₃ _₄ _₅ _₆ _₇ _r _₈ _₉ _₁₀ _₁₁ _₁₂ _₁₃ _₁₄ _₁₅ _₁₆ _₁₇ _₁₈ _₁₉ _₂₀ _₂₁ _₂₂ _₂₃ _₂₄
+    _₂₅ _₂₆ _₂₇
```
chabulhwi, line 575: Why do I have to clean up these unwanted hypotheses?

Batteries/Data/List/Lemmas.lean
eric-wieser: Here's an argument for why we should have them; you could imagine an alternative world where `IsPrefix` is defined as
```lean
inductive IsPrefix where
| nil {l} : IsPrefix [] l
| cons (a) {l1 l2} : IsPrefix l1 l2 -> IsPrefix (a :: l1) (a :: l2)
```

The API should be agnostic to the actual implementation chosen, and so we should provide `nil` and `cons`  anyway.

Batteries/Data/List/Lemmas.lean
kmill: Suggestion:
```lean
theorem not_prefix_and_not_prefix_symm_iff {l₁ l₂ p t₁ t₂ : List α} (hp : commonPrefix l₁ l₂ = p)
    (h₁ : l₁ = p ++ t₁) (h₂ : l₂ = p ++ t₂) : ¬l₁ <+: l₂ ∧ ¬l₂ <+: l₁ ↔ t₁ ≠ [] ∧ t₂ ≠ [] ∧
      t₁.head? ≠ t₂.head? := by
  subst_vars
  rw [prefix_append_right_inj, prefix_append_right_inj]
  rw [commonPrefix_append_append, append_right_eq_self] at hp
  cases t₁ <;> cases t₂
  · simp
  · simp
  · simp
  · simp only [commonPrefix, ite_eq_right_iff, reduceCtorEq, imp_false] at hp
    simp [Ne.symm hp, hp]
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -482,6 +491,83 @@ theorem Sublist.erase_diff_erase_sublist {a : α} :
...
+  | [],   _  => simp [commonPrefix]
+  | _::_, [] => simp [commonPrefix]
+  | a₁::l₁, a₂::l₂ =>
+    simp only [commonPrefix]
+    cases Decidable.em (a₁ = a₂)
+    · next h =>
+      simp only [h, ↓reduceIte, cons_prefix_cons, true_and]
+      apply commonPrefix_prefix_left
+    · next h =>
+      simp [h]
```
kmill, line 518: I don't usually see `cases Decidable.em` in code. Instead I'd see `by_cases`:
```suggestion
    by_cases h : a₁ = a₂
    · simp only [h, ↓reduceIte, cons_prefix_cons, commonPrefix_prefix_left, and_self]
    · simp [h]
```
or `split` since there's an `if`:
```suggestion
    split
    next h =>
      simp only [h, cons_prefix_cons, true_and, commonPrefix_prefix_left]
    next h =>
      simp [h]
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -482,6 +491,83 @@ theorem Sublist.erase_diff_erase_sublist {a : α} :
...
+  | [],   _  => simp [commonPrefix]
+  | _::_, [] => simp [commonPrefix]
+  | a₁::l₁, a₂::l₂ =>
+    simp only [commonPrefix]
+    cases Decidable.em (a₁ = a₂)
+    · next h =>
+      simp only [h, ↓reduceIte, cons_prefix_cons, true_and]
+      apply commonPrefix_prefix_right
+    · next h =>
+      simp [h]
```
kmill, line 531: You can make use of previous work by adding an important `commonPrefix` lemma:
```lean
theorem commonPrefix_comm (l₁ l₂ : List α) : commonPrefix l₁ l₂ = commonPrefix l₂ l₁ := by
  induction l₁ generalizing l₂ with
  | nil => cases l₂ <;> rfl
  | cons a₁ l₂ ih =>
    cases l₂
    · rfl
    · unfold commonPrefix
      split
      · subst_vars
        simp [ih]
      next h =>
        simp [Ne.symm h]

theorem commonPrefix_prefix_right (l₁ l₂ : List α) : commonPrefix l₁ l₂ <+: l₂ := by
  rw [commonPrefix_comm]
  apply commonPrefix_prefix_left
```

Batteries/Data/List/Lemmas.lean
eric-wieser: In mathlib we'd often use `obtain rfl | hxy := Decidable.eq_or_ne x y`; perhaps that should be upstreamed to batteries?

chabulhwi: > I don't usually see `cases Decidable.em` in code. Instead I'd see `by_cases`:

It opens the `Classical` namespace, doesn't it? The `simp` tactic have already used classical theorems, anyway.

Batteries/Data/List/Lemmas.lean
chabulhwi: Thanks! I changed the proof of the theorem `List.commonPrefix_comm` a bit.

Batteries/Data/List/Lemmas.lean
chabulhwi: Thanks! I changed your proof a bit.

Batteries/Data/List/Lemmas.lean
chabulhwi: @eric-wieser The relation `List.IsPrefix` is defined in Lean's core library. Where should we put the aliases? Lean's core or Batteries?

fgdorais:
The `commonPrefix` addition is good but let's move it to a separate PR.

chabulhwi:
I'll split this pull request in two.

fgdorais:
Don't close this one! We want to keep the review history for the other parts.

Batteries/Data/List/Lemmas.lean
fgdorais: That's a fair point @eric-wieser ! However, in that alternate world, I might think that since `IsPrefix` is a `Prop`, its constructors should follow proper theorem naming conventions...

Anyway, I'm not vehemently opposed to these aliases, I just don't think they're necessary.

chabulhwi:
@fgdorais Shall I close #993?

fgdorais:
Yes.

Batteries/Data/List/Lemmas.lean
fgdorais: Should be `nil_isPrefix` and `cons_isPrefix_cons`, no?

Batteries/Data/List/Lemmas.lean
```diff
@@ -482,6 +491,15 @@ theorem Sublist.erase_diff_erase_sublist {a : α} :
...
 end Diff
 
+/-! ### prefix, suffix, infix -/
+
+theorem singleton_prefix_cons (a) : [a] <+: a :: l :=
+  (prefix_cons_inj a).mpr nil_prefix
+
+theorem ne_nil_of_not_prefix (h : ¬l₁ <+: l₂) : l₁ ≠ [] := by
+  intro heq
+  simp [heq, nil_prefix] at h
```
fgdorais, line 501:
```suggestion
theorem ne_nil_of_not_prefix : ¬l₁ <+: l₂ → l₁ ≠ [] := by
  apply mt; intro h; simp [h]
```

fgdorais (changes requested):

PR #989, opened 2024-10-15
feat: upstream DList results from Mathlib

jcommelin:
Following up on `LazyList`, we upstream results about `DList` from Mathlib to Batteries.

Along the way, we
- move DList to a subdirectory
- adjust import for new DList location
- move all defs to Defs.lean and all lemmas to Lemmas.lean
- slightly adjust docstrings

The commits attempt to be atomic, to ease review.


fgdorais:
Please rename `Defs` to `Basic`. (Admittedly, `Defs` is a more appropriate name but `Basic` is the de facto standard in Batteries.)

PR #990, opened 2024-10-15
feat: characterize `OfScientific.ofScientific` on `Rat`

eric-wieser:
With this change, `#simp Rat.ofScientific 314 false 2` in mathlib gives `3.14`.

This relates to https://github.com/leanprover/lean4/pull/5725


fgdorais (approved):

fgdorais:
Perhaps `#simp` should be upstreamed? It seems like it has broader use than just Mathlib.

PR #992, opened 2024-10-16
chore: deprecate `Expr.lambdaArity` in favor of existing upstream version

kim-em:
c.f. https://github.com/leanprover/lean4/pull/5729

No particular preference regarding naming: favoring the existing upstream name seems easiest.


fgdorais (approved):
LGTM

PR #993, opened 2024-10-16
feat: add lemmas about lists

chabulhwi:
These are 'leftover' lemmas I created while trying to prove
`String.splitOn_of_valid`. See
https://github.com/leanprover-community/batteries/pull/743.

Co-authored-by: François G. Dorais <fgdorais@gmail.com>

---

This is the second version of #987.


chabulhwi:
awaiting-review

PR #996, opened 2024-10-16
fix: print prefix bug

### LINKED ISSUE #995: `#print prefix` elaborator crashes language server downstream of Mathlib
Consider the following MWE:
```
import Mathlib

#print
```

This triggers the `#print prefix` elaborator with partial syntax, which then attempts to create a `MessageData` of all declarations in the environment of Mathlib. The effect of this is that the CPU usage of the file worker first starts increasing drastically and then eventually causes a stack overflow in `Lean.MessageData.joinSep` that crashes the file worker. Note that even if `joinSep` was tail recursive, the spike in CPU usage and attempting to print all declarations in the environment is highly undesired.

Note that this issue occurs downstream of Mathlib whenever typing a `#print <id>` declaration since the partial `#print` input while typing the full declaration triggers the `#print prefix` elaborator.

This issue was originally reported by @b-mehta.

fgdorais:
This fix makes it so that `#print prefix` does nothing when there is no identifier. Note that this doesn't actually fix the underlying issue: `#print prefix` is still horribly slow downstream of Mathlib.

Closes #995

### DIFF
--- a/Batteries/Tactic/PrintPrefix.lean
+++ b/Batteries/Tactic/PrintPrefix.lean
@@ -108,12 +108,9 @@ by setting `showTypes` to `false`:
 The complete set of flags can be seen in the documentation
 for `Lean.Elab.Command.PrintPrefixConfig`.
 -/
-elab (name := printPrefix) tk:"#print " colGt &"prefix"
+elab (name := printPrefix) tk:"#print " colGt "prefix"
     cfg:(Lean.Parser.Tactic.config)? name:(ident)? : command => liftTermElabM do
-  match name with
-  | none =>
-    logInfoAt tk "unexpected end of input; expected identifier or string literal"
-  | some name =>
+  if let some name := name then
     let opts ← elabPrintPrefixConfig (mkOptionalNode cfg)
     let mut msgs ← matchingConstants opts name.getId
     if msgs.isEmpty then
--- a/Batteries/Tactic/PrintPrefix.lean
+++ b/Batteries/Tactic/PrintPrefix.lean
@@ -108,12 +108,15 @@ by setting `showTypes` to `false`:
 The complete set of flags can be seen in the documentation
 for `Lean.Elab.Command.PrintPrefixConfig`.
 -/
-elab (name := printPrefix) "#print" tk:"prefix"
-    cfg:(Lean.Parser.Tactic.config)? name:ident : command => liftTermElabM do
-  let nameId := name.getId
-  let opts ← elabPrintPrefixConfig (mkOptionalNode cfg)
-  let mut msgs ← matchingConstants opts nameId
-  if msgs.isEmpty then
-    if let [name] ← resolveGlobalConst name then
-      msgs ← matchingConstants opts name
-  logInfoAt tk (.joinSep msgs.toList "")
+elab (name := printPrefix) tk:"#print " colGt &"prefix"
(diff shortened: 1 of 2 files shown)


digama0:
Let's just remove the `#print prefix <none>`  behavior, or replace it by `#print prefix *`. It's definitely undesirable to be printing the whole environment with a command which is transient when typing the prefix you actually want to search for.

Batteries/Tactic/PrintPrefix.lean
```diff
@@ -108,12 +108,15 @@ by setting `showTypes` to `false`:
...
-  let mut msgs ← matchingConstants opts nameId
-  if msgs.isEmpty then
-    if let [name] ← resolveGlobalConst name then
-      msgs ← matchingConstants opts name
-  logInfoAt tk (.joinSep msgs.toList "")
+elab (name := printPrefix) tk:"#print " colGt &"prefix"
+    cfg:(Lean.Parser.Tactic.config)? name:(ident)? : command => liftTermElabM do
+  match name with
+  | none =>
+    logInfoAt tk "unexpected end of input; expected identifier or string literal"
```
digama0, line 115:
```suggestion
    throwUnsupported
```
There is no need to pretend to be a parse error

Batteries/Tactic/PrintPrefix.lean
```diff
@@ -108,12 +108,15 @@ by setting `showTypes` to `false`:
...
-elab (name := printPrefix) "#print" tk:"prefix"
-    cfg:(Lean.Parser.Tactic.config)? name:ident : command => liftTermElabM do
-  let nameId := name.getId
-  let opts ← elabPrintPrefixConfig (mkOptionalNode cfg)
-  let mut msgs ← matchingConstants opts nameId
-  if msgs.isEmpty then
-    if let [name] ← resolveGlobalConst name then
-      msgs ← matchingConstants opts name
-  logInfoAt tk (.joinSep msgs.toList "")
+elab (name := printPrefix) tk:"#print " colGt &"prefix"
```
digama0, line 111:
```suggestion
elab (name := printPrefix) tk:"#print " colGt "prefix"
```
Are you sure this should be an identifier-like keyword? It is already a proper keyword, and I think this is needed to differentiate from `#print id` form.

Batteries/Tactic/PrintPrefix.lean
fgdorais: I went with the do nothing route.

fgdorais:
The `*` idea makes sense but this code needs a serious performance review, so I'm thinking we may as well defer to that time.

fgdorais:
There is currently a test that checks that `#print prefix` has empty output.

digama0:
I mean a test for bare `#print` with no argument, which was the context of the original report.

fgdorais:
Yes, the parse error makes it hard to test in this case. I don't know how to do it.

digama0:
We should put failing tests and other kinds of tests like the lean test suite on the batteries todo, but I won't insist on it here.

test/print_prefix.lean
```diff
@@ -171,3 +171,6 @@ TestInd.toCtorIdx : TestInd → Nat
 -/
 #guard_msgs in
 #print prefix TestInd
+
+-- `#print prefix` does nothing if no identifier is provided
+#print prefix
```
digama0, line 176:
```suggestion
#guard_msgs in
#print prefix
```

PR #997, opened 2024-10-17
fix: make `proof_wanted` not report unused variables

### LINKED ISSUE #938: `proof_wanted` is linted for unused arguments
We should disable the unused argument linter and the new variable inclusion mechanism linter in `proof_wanted`. Currently, we have the following behavior:
```
import Batteries.Util.ProofWanted

class Foo where

variable [Foo]

proof_wanted two_eq_two (h : 1 + 1 = 2) : 2 = 2
/-
unused variable `h`
note: this linter can be disabled with `set_option linter.unusedVariables false`

included section variable '[Foo]' is not used in 'two_eq_two', consider excluding it
-/
```

kmill:
This was requested by Terence Tao [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Is.20a.20.60definition_wanted.60.20keyword.20possible.3F/near/477325788)

Closes #938

### DIFF
--- a/Batteries/Util/ProofWanted.lean
+++ b/Batteries/Util/ProofWanted.lean
@@ -33,7 +33,10 @@ elaboration, but it's then removed from the environment.
 def elabProofWanted : CommandElab
   | `($mods:declModifiers proof_wanted $name $args* : $res) => withoutModifyingEnv do
     -- The helper axiom is used instead of `sorry` to avoid spurious warnings
-    elabCommand <| ← `(axiom helper (p : Prop) : p
-                       $mods:declModifiers
-                       theorem $name $args* : $res := helper _)
+    elabCommand <| ← `(
+      section
+      set_option linter.unusedVariables false
+      axiom helper {α : Sort _} : α
+      $mods:declModifiers theorem $name $args* : $res := helper
+      end)
   | _ => throwUnsupportedSyntax
--- a/test/proof_wanted.lean
+++ b/test/proof_wanted.lean
@@ -1,0 +1,15 @@
+import Batteries.Util.ProofWanted
+
+/-!
+No unused variable warnings.
+-/
+#guard_msgs in proof_wanted foo (x : Nat) : True
+
+/-!
+When not a proposition, rely on `theorem` command failing.
+-/
+/--
+error: type of theorem 'foo' is not a proposition
+  Nat → Nat
+-/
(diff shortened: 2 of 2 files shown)


Batteries/Util/ProofWanted.lean
```diff
@@ -33,7 +33,10 @@ elaboration, but it's then removed from the environment.
...
     -- The helper axiom is used instead of `sorry` to avoid spurious warnings
-    elabCommand <| ← `(axiom helper (p : Prop) : p
-                       $mods:declModifiers
-                       theorem $name $args* : $res := helper _)
+    elabCommand <| ← `(
+      section
+      set_option linter.unusedVariables false
+      axiom helper {α : Sort _} : α
+      $mods:declModifiers theorem $name $args* : $res := helper
+      end)
```
kmill, line 41: Another solution would be to use something like [Gabriel's `sorry`](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Is.20a.20.60definition_wanted.60.20keyword.20possible.3F/near/477287818) and capture all the fvars, but that would take making a term elaborator to replace `helper`.

Batteries/Util/ProofWanted.lean
```diff
@@ -33,7 +33,10 @@ elaboration, but it's then removed from the environment.
 def elabProofWanted : CommandElab
   | `($mods:declModifiers proof_wanted $name $args* : $res) => withoutModifyingEnv do
     -- The helper axiom is used instead of `sorry` to avoid spurious warnings
-    elabCommand <| ← `(axiom helper (p : Prop) : p
-                       $mods:declModifiers
-                       theorem $name $args* : $res := helper _)
+    elabCommand <| ← `(
+      section
+      set_option linter.unusedVariables false
+      axiom helper {α : Sort _} : α
```
fgdorais, line 39: Do we really want `Sort _` here?

fgdorais: Ah, I see this incorporates `definition_wanted`.

kmill: Yes, I found that the error messages for non-theorems was unfriendly for just Prop. The test file shows that non-Prop still gives a good error.

Batteries/Util/ProofWanted.lean
fgdorais: That would be something upstream.

PR #1019, opened 2024-10-30
fix: ignore comments in `#where`

adomani:
This modification removes trailing comments from `variable` declarations, when printing then with `#where`.

See [this Zulip discussion](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/.60.23where.60.20and.20comments) for some context.

### DIFF
--- a/Batteries/Tactic/Where.lean
+++ b/Batteries/Tactic/Where.lean
@@ -53,6 +53,16 @@ private def describeOptions (opts : Options) : CommandElabM (Option MessageData)
   else
     return MessageData.joinSep lines.toList "\n"
 
+/-- `clearTrailing stx` assumes that `stx` is a `Lean.Parser.Term.bracketedBinder` syntax node.
+It replaces the trailing `SourceInfo` with a space (` `).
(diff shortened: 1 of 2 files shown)


Batteries/Tactic/Where.lean
```diff
@@ -74,7 +84,8 @@ elab "#where" : command => do
     msg := msg.push <| "universe " ++ MessageData.joinSep levels " "
   -- Variables
   if !scope.varDecls.isEmpty then
-    msg := msg.push <| ← `(command| variable $scope.varDecls*)
+    let vars := scope.varDecls.map (⟨clearTrailing ·⟩)
```
kmill, line 87: The function `Lean.Syntax.unsetTrailing` seems to be doing what you want here, but I haven't tested it.
```suggestion
    let vars := scope.varDecls.map (⟨·.unsetTrailing⟩)
```

adomani: Ah, great!  I thought that I needed to preserve a space between the various syntax nodes, but it seems like that is already handles somewhere else.  Thanks!

adomani:
Closing, as already contained in lean4#5065.

PR #1028, opened 2024-11-07
feat: generate docs from subdirectory

fgdorais:
This uses the idea proposed by Sebastian Ullrich on [Zulip](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Improving.20the.20Windows.20user.20experience/near/479463868) in order to streamline generating `doc-gen4` documentations for batteries.


BatteriesDocs.lean
```diff
@@ -0,0 +1,19 @@
...
+
+/-! # Batteries Documentation
+
+To generate documentation for Batteries:
+
+1. Change to the `BatteriesDocs` subdirectory
+2. Run `lake build BatteriesDocs:docs`
+
+The [`doc-gen4`](https://github.com/leanprover/doc-gen4) documentation will be found in the
+`.lake/buid/doc` and `.lake/build/doc-data` subdirectories.
```
kim-em, line 17:
```suggestion
`.lake/build/doc` and `.lake/build/doc-data` subdirectories.
```

docs/lakefile.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "doc-gen4"
+git = "https://github.com/leanprover/doc-gen4"
+rev = "main"
+
+[[require]]
+name = "batteries"
+path = ".."
+
+[[lean_lib]]
+name = "BatteriesDocs"
```
digama0, line 16: Why is `BatteriesDocs` in the root instead of in the `docs` folder?

fgdorais: Yeah, this is a bit confusing. The `docs` folder is named this way just for easy integration with Github pages. It really should be the `BatteriesDocs` folder. Anyway, the reason is basically the same as why `Batteries.lean` is not in the `Batteries` folder.

fgdorais: Also `lake clean` in `docs` deletes everything in `docs`, so it's better to have the minimum possible in there.

digama0: Okay so maybe we should do that instead? Don't have a `docs/` folder at all in the repo, and treat it like a build directory. Put all the batteries docs stuff in a `BatteriesDocs` subfolder. I don't see the connection to batteries.lean being in the root, since the root of the `BatteriesDocs` project is the `BatteriesDocs/` or `docs/` folder.

fgdorais: I don't understand how that's better?

fgdorais: Wait! I think I have a bridge, give me a few minutes...

fgdorais: Is this setup better?

digama0: > I don't understand how that's better?

It means everything related to the `BatteriesDocs` project is in that folder, rather than living at the same root as `Batteries` (which implies that it is another Batteries lean_lib even though it isn't)

BatteriesDocs/lean-toolchain
```diff
@@ -0,0 +1 @@
+../lean-toolchain
```
digama0, line 1: I did not know you could do this. But I think it's not necessary, elan should pick the lean-toolchain from the root if you don't supply one here.

docs/lakefile.toml
digama0: > Is this setup better? (Do let me know since I will need to update CI and README accordingly. I can also revert if you don't think this is better.)

I still think the lean file should be in the folder. This is not standard lake layout, as indicated by the fact that you had to specify a `srcDir` for it. Normally the `srcDir` is `.` which is why the root file lives at the project root normally. In this package, the package root is `BatteriesDocs` so that's where the root file should be.

fgdorais: > > I don't understand how that's better?
> 
> It means everything related to the `BatteriesDocs` project is in that folder, rather than living at the same root as `Batteries` (which implies that it is another Batteries lean_lib even though it isn't)

I don't think about it that way. The `docs` folder is for integration with github pages. Users of this scheme may want to put an entire website in there in addition to the doc-gen4 documentation. The `BatteriesDocs.lean` file is just a placeholder. The directory where the lakefile.toml is for BatteriesDocs is irrelevant: trying to build BatteriesDocs from any other directory will just fail. 

I can't think of a scenario where one would want to add dependencies to `BatteriesDocs.lean` other than importing the project roots, but the modification I just did allows for that peculiar scenario without interfering with website files.

fgdorais: > > Is this setup better? (Do let me know since I will need to update CI and README accordingly. I can also revert if you don't think this is better.)
> 
> I still think the lean file should be in the folder. This is not standard lake layout, as indicated by the fact that you had to specify a `srcDir` for it. Normally the `srcDir` is `.` which is why the root file lives at the project root normally. In this package, the package root is `BatteriesDocs` so that's where the root file should be.

It's not a real package (hence `reservoir := false`). This will never be a "standard lake layout".

fgdorais: I think we might be homing in on an issue with lake. While this is not the case right now, I can imagine `BatteriesTest.lean` also needing its own lakefile to support currently impossible tests. I think lake does currently allow specifying the config directory but I'm not sure how that works. I'll experiment with that over the next few days...

BatteriesDocs/lean-toolchain
fgdorais: I don't know how portable github symliks are so that may well be a better idea!

docs/lakefile.toml
tydeu: > Maybe @tydeu can advise?

`lake -d <folder-with-lakefile> ...` will load the  workspace from the folder specified.

fgdorais: Can we put that as a requirement in the base lakefile? In other words, have `BatteriesDocs` require a different lakefile to build itself?

tydeu: Perhaps `require "batteries-docs" from "BatteriesDocs"` is what you are looking for?

fgdorais: No, that would reintroduce the circular dependencies.

tydeu: Oh, yeah. I am not sure what you want here, then. Either docs requires batteries or vice versa, and only one can be the root workspace  (at least at present).

fgdorais: PS: This is a scapegoat for a general solution to the problem of generating docs for a project without users having to depend on doc-gen4. This (partial) solution is also good in the sense that doc-gen4 could depend on this project without a circular dependency issue.

digama0: > I can't think of a scenario where one would want to add dependencies to `BatteriesDocs.lean` other than importing the project roots, but the modification I just did allows for that peculiar scenario without interfering with website files.

Let me put it this way: I want everything related to docs generation to be in one folder. I don't care if it's called `docs` or `BatteriesDocs` or something else, but I want it all neatly tucked away in one place where we can put a readme and so on. For reasons you have observed, `docs` is specifically problematic, so maybe not that. `BatteriesDocs` makes it look like a module directory, which it's not, so perhaps that's also a bad name and it should be `docgen` or `doc` or something else.

fgdorais: Done. There's a chance it will look a bit weird at https://leanprover-community.github.io/batteries/doc but I guess we can punt that to doc-gen4.

fgdorais: Henrik pointed out further simplifications that I think resolve all the issues discussed here.

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: lake build -q Batteries:docs
+
+      - name: Deploy Docs
+        run: |
+          git config user.name "leanprover-community-batteries-bot"
+          git config user.email "leanprover-community-batteries-bot@users.noreply.github.com"
+          git checkout -b docs
+          git add docs/doc docs/doc-data
+          git commit -m "chore: generate docs"
+          git push origin docs
```
fgdorais, line 38:
```suggestion
          git push origin docs --force
```

fgdorais:
The documentation setup is now fully in line with the new recommendations at [doc-gen4](https://github.com/leanprover/doc-gen4). The `docs` directory name and the `buildDir = "."` addition is to facilitate deployment to github pages.

PR #1029, opened 2024-11-08
feat: sugar for SatisfiesM

kim-em:
Enables:

```
example (xs : Array Expr) : MetaM { ts : Array Expr // ts.size = xs.size } := do
  let r ← satisfying (xs.size_mapM inferType)
  return r
```

by showing that for the basic monads and monad transformers, one can lift the `SatisfiesM` predicate to a monadic value in `Subtype`.

### DIFF
--- a/Batteries/Classes/SatisfiesM.lean
+++ b/Batteries/Classes/SatisfiesM.lean
@@ -5,6 +5,7 @@ Authors: Mario Carneiro, Kim Morrison
 -/
 import Batteries.Lean.EStateM
 import Batteries.Lean.Except
+import Batteries.Tactic.Lint
 
 /-!
 ## SatisfiesM
@@ -304,12 +305,13 @@ and an unsafe "proof" via `lcProof`.
 
 This is purely a performance optimization, and could be omitted.
 -/
+@[nolint unusedArguments]
 unsafe def unsafeSatisfying {m : Type u → Type v} [Functor m] [LawfulFunctor m] [MonadSatisfying m]
     {α} {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) : m {a // p a} :=
   (⟨·, lcProof⟩) <$> x
 
 /-- `unsafeSatisfying` is equivalent to `satisfying`. -/
-@[csimp] unsafe def satisfying_eq_unsafeSatisfying :
+@[csimp, nolint defLemma] unsafe def satisfying_eq_unsafeSatisfying :
     @satisfying = @unsafeSatisfying := by
   funext m _ _ _ _ p x h
   unfold unsafeSatisfying
--- a/Batteries/Classes/SatisfiesM.lean
+++ b/Batteries/Classes/SatisfiesM.lean
@@ -1,8 +1,10 @@
 /-
 Copyright (c) 2022 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Mario Carneiro
+Authors: Mario Carneiro, Kim Morrison
 -/
+import Batteries.Lean.EStateM
+import Batteries.Lean.Except
 
 /-!
 ## SatisfiesM
@@ -11,6 +13,13 @@ The `SatisfiesM` predicate works over an arbitrary (lawful) monad / applicative 
 and enables Hoare-like reasoning over monadic expressions. For example, given a monadic
 function `f : α → m β`, to say that the return value of `f` satisfies `Q` whenever
 the input satisfies `P`, we write `∀ a, P a → SatisfiesM Q (f a)`.
+
+For any monad equipped with `Satisfying m`
+one can lift `SatisfiesM` to a monadic value in `Subtype`,
+using `satisfying x h : m {a // p a}`, where `x : m α` and `h : SatisfiesM p x`.
+This includes `Option`, `ReaderT`, `StateT`, and `ExceptT`, and the Lean monad stack.
+(Although it is not entirely clear one should treat the Lean monad stack as lawful,
+even though Lean accepts this.)
 
 ## Notes
 
@@ -23,7 +32,7 @@ presumably requiring more syntactic support (and smarter `do` blocks) from Lean.
 Or it may be that such a solution will look different!
 This is an open research program, and for now one should not be overly ambitious using `SatisfiesM`.
 
-In particular lemmas about pure operations on data structures in `batteries` except for `HashMap`
+In particular lemmas about pure operations on data structures in `Batteries` except for `HashMap`
 should avoid `SatisfiesM` for now, so that it is easy to migrate to other approaches in future.
 -/
 
@@ -158,6 +167,23 @@ end SatisfiesM
   ⟨by revert x; intro | .ok _, ⟨.ok ⟨_, h⟩, rfl⟩, _, rfl => exact h,
    fun h => match x with | .ok a => ⟨.ok ⟨a, h _ rfl⟩, rfl⟩ | .error e => ⟨.error e, rfl⟩⟩
 
+@[simp] theorem SatisfiesM_EStateM_eq :
+    SatisfiesM (m := EStateM ε σ) p x ↔ ∀ s a s', x.run s = .ok a s' → p a := by
+  constructor
+  · rintro ⟨x, rfl⟩ s a s' h
+    match w : x.run s with
+    | .ok a s' => simp at h; exact h.1
+    | .error e s' => simp [w] at h
+  · intro w
+    refine ⟨?_, ?_⟩
+    · intro s
+      match q : x.run s with
+      | .ok a s' => exact .ok ⟨a, w s a s' q⟩ s'
+      | .error e s' => exact .error e s'
+    · ext s
+      rw [EStateM.run_map, EStateM.run]
+      split <;> simp_all
+
 @[simp] theorem SatisfiesM_ReaderT_eq [Monad m] :
     SatisfiesM (m := ReaderT ρ m) p x ↔ ∀ s, SatisfiesM p (x s) :=
   (exists_congr fun a => by exact ⟨fun eq _ => eq ▸ rfl, funext⟩).trans Classical.skolem.symm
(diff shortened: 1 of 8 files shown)


Batteries/Classes/SatisfiesM.lean
```diff
@@ -158,25 +167,128 @@ end SatisfiesM
...
   · exists ((fun | ⟨.ok a, h⟩ => .ok ⟨a, h _ rfl⟩ | ⟨.error e, _⟩ => .error e) <$> f : m _)
     show _ >>= _ = _; simp [← comp_map, ← bind_pure_comp]; congr; funext ⟨a, h⟩; cases a <;> rfl
+
+/--
+If a monad has `Satisfying m`, then we can lift a `h : SatisfiesM (m := m) p x` predicate
+to monadic value `satisfying x p : m { x // p x }`.
+
+Reader, state, and exception monads have `Satisfying` instances if the base monad does.
+-/
+class Satisfying (m : Type u → Type v) [Functor m] where
```
digama0, line 219:
```suggestion
class MonadSatisfying (m : Type u → Type v) [Functor m] where
```
WDYT?

Batteries/Classes/SatisfiesM.lean
```diff
@@ -158,25 +167,128 @@ end SatisfiesM
...
+  /-- Lift a `SatisfiesM` predicate to a monadic value. -/
+  satisfying {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) :
+    { y : m {a // p a} // Subtype.val <$> y = x }
```
digama0, line 222: Maybe this should be two fields?

digama0:
Another thing I was considering regarding this class is that if `Satisfying` took `LawfulMonad` as a parameter or extended it (such that you can't even say `satisfying` unless the monad is lawful), then it should be legal to implement `satisfying` using an `unsafeCast` since the constraints on it already tell you that it has to be the identity function. (You need lawfulness because otherwise the map in `satisfies.2` could be doing something nontrivial to the monadic value.)

Batteries/Classes/SatisfiesM.lean
kim-em: Done.

Batteries/Classes/SatisfiesM.lean
kim-em: Yes, done!

kim-em:
> Another thing I was considering regarding this class is that if `Satisfying` took `LawfulMonad` as a parameter or extended it (such that you can't even say `satisfying` unless the monad is lawful), then it should be legal to implement `satisfying` using an `unsafeCast` (i.e. using `implemented_by` on the typeclass projection function) since the constraints on it already tell you that it has to be the identity function. (You need lawfulness because otherwise the map in `satisfies.2` could be doing something nontrivial to the monadic value.)

@digama0, how do you feel about:

```
set_option linter.unusedVariables false in
/--
Since we assume `m` is lawful,
we can replace the implementation of `satisfying` at runtime using the identity function
and an unsafe "proof" via `lcProof`.
-/
unsafe def unsafeSatisfying {m : Type u → Type v} [Functor m] [LawfulFunctor m] [MonadSatisfying m]
    {α} {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) : m {a // p a} :=
  (⟨·, lcProof⟩) <$> x

/-- `unsafeSatisfying` is equivalent to `satisfying`. -/
@[csimp] unsafe def satisfying_eq_unsafeSatisfying :
    @satisfying = @unsafeSatisfying := by
  funext m _ _ _ _ p x h 
  unfold unsafeSatisfying
  conv => rhs; simp +singlePass [← MonadSatisfying.val_eq h]
  simp
```

instead of the `unsafeCast` / `implemented_by`? 

Pros: It "documents" the reason why it is safe to do the replacement. (Although of course given that Lean would accept `lcProof` inside the `@[csimp]` means that there is no real improvement in safety).
Cons: It's a bit strange! Maybe you have in mind some performance difference between `unsafeCast` and `lcProof` here?

Happy to switch it out for `@[implemented_by]`.

digama0:
I'm actually more concerned about the performance impact of the `<$>` rather than the unsafeCast. Also, proving theorems about unsafe definitions is a bit scuffed.

kim-em:
> I'm actually more concerned about the performance impact of the `<$>` rather than the unsafeCast. Also, proving theorems about unsafe definitions is a bit scuffed.

But no concrete change here, right? I don't see any way to avoid the `<$>`. Happy to remove the theorem, as I agree it is a bit misleading, but will leave as is unless you say otherwise!

digama0:
Well, with `unsafeCast` you can avoid the `<$>`

digama0:
It's actually somewhat safer, because the use of alpha in a nested context ensures that it will be boxed (so there will not be any difference in representation even if alpha is a primitive type).

Batteries/Classes/SatisfiesM.lean
```diff
@@ -158,25 +168,133 @@ end SatisfiesM
...
   · exists ((fun | ⟨.ok a, h⟩ => .ok ⟨a, h _ rfl⟩ | ⟨.error e, _⟩ => .error e) <$> f : m _)
     show _ >>= _ = _; simp [← comp_map, ← bind_pure_comp]; congr; funext ⟨a, h⟩; cases a <;> rfl
+
+/--
+If a monad has `MonadSatisfying m`, then we can lift a `h : SatisfiesM (m := m) p x` predicate
+to monadic value `satisfying x p : m { x // p x }`.
+
+Reader, state, and exception monads have `MonadSatisfying` instances if the base monad does.
+-/
+class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m]where
```
digama0, line 220:
```suggestion
class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m] where
```

BatteriesTest/satisfying.lean
```diff
@@ -0,0 +1,23 @@
+import Batteries.Lean.SatisfiesM
+import Batteries.Data.Array.Monadic
+
+open Lean Meta Array Elab Term Tactic Command
+
+-- For now these live in the test file, as it's not really clear we want people relying on them!
```
digama0, line 6: I think these instances should be next to the `MonadSatisfying` instances, since it makes no sense to have a `MonadSatisfying` instance without the `LawfulMonad` instance

Ericson2314:
Very happy this exists right around when I wrote https://github.com/leanprover-community/batteries/issues/1032 ! Crazy coincidence.

Thinking a bit further, I think the class can be simplified by inlining `SatisfiesM` and turning the existential into a universal (provided this doesn't fuck up the prop modality). I would write out what I mean exactly but I'm in my phone.

BatteriesTest/satisfying.lean
kim-em: I made a separate file, but agree they can be lifted out of the tests.

kim-em:
> Very happy this exists right around when I wrote #1032 ! Crazy coincidence.
> 
> Thinking a bit further, I think the class can be simplified by inlining `SatisfiesM` and currying the existential into a universal (provided this doesn't fuck up the prop modality). I would write out what I mean exactly but I'm in my phone.
> 
> (I am not sure what Lean/Batteries had to truncate a Type to a Prop. But that must be used to curry the existential.)

Sorry, not following you here.

Ericson2314:
```lean
class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m] where
  satisfying {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) : m {a // p a}
  val_eq {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) : Subtype.val <$> satisfying h = x
```

Now let's inline `SatisfiesM`:

```lean
class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m] where
  satisfying {p : α → Prop} {x : m α} (h : ∃ x' : m {a // p a}, Subtype.val <$> x' = x) : m {a // p a}
  val_eq {p : α → Prop} {x : m α} (h : ∃ x' : m {a // p a}, Subtype.val <$> x' = x) : Subtype.val <$> satisfying h = x
```

now let's curry:

```lean
class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m] where
  satisfying {p : α → Prop} {x : m α} [x'' : NonEmpty (m {a // p a})] (ext: Nonempty.elim x'' (fun x' => Subtype.val <$> x' = x) : m {a // p a}
  val_eq {p : α → Prop} {x : m α} [x'' : NonEmpty (m {a // p a})] (ext: Nonempty.elim x'' (fun x' => Subtype.val <$> x' = x) : Subtype.val <$> satisfying x'' ext = x
```

now let's simplify `val_eq`:

```lean
class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m] where
  satisfying {p : α → Prop} {x : m α} [x'' : NonEmpty (m {a // p a})] (ext: Nonempty.elim x'' (fun x' => Subtype.val <$> x' = x) : m {a // p a}
  val_eq {p : α → Prop} {x : m α} [x'' : NonEmpty (m {a // p a})] (ext: Nonempty.elim x'' (fun x' => Subtype.val <$> x' = x) : Nonempty.elim x'' (fun x' => satisfying x'' ext = x')
```

This avoids the second use of `<$>`, and shows that `satisfying` constructs the monadic value whose mere existence is provided as an argument.

A non-class version of  `NonEmpty` would be more appropriate for this.

digama0:
@Ericson2314  the "let's curry" step is not correct. The assumption `ext: Nonempty.elim x'' (fun x' => Subtype.val <$> x' = x)` is not type correct because `Nonempty.elim` can only produce elements of propositions and `Subtype.val <$> x' = x` is not an element of a proposition (it is a proposition itself). You would have to use choice to do this elimination, but that's also not correct because that's saying that an arbitrary other element of the type `m {a // p a}` is equivalent to the monadic value `x`, which generally will not be the case unless `m {a // p a}` is a subsingleton.

The usage of mere existence in `SatisfiesM` is quite deliberate. You could have a data carrying version of `SatisfiesM`, this would be equivalent in strength to the conjunction of `SatisfiesM` and `MonadSatisfying`, but I wanted to avoid that because that would require recursively constructing this value over the structure of the type, while `MonadSatisfying` allows you to skip the whole process and use an identity function.

Ericson2314:
> can only produce elements of propositions and Subtype.val <$> x' = x is not an element of a proposition (it is a proposition itself). 

True; my mistake.

I agree mere existence is well motivated. Can the function taking it be curried in some other way?

Alternatively, can we do this?

```lean
class MonadSatisfying (m : Type u → Type v) [Functor m] [LawfulFunctor m] where
  satisfying {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) : m {a // p a}
  val_eq {p : α → Prop} {x : m α} (h : SatisfiesM (m := m) p x) : let ⟨x', _⟩ := h ; satisfying h = x'
```

digama0:
> I agree mere existence is well motivated. Can the function taking it be curried in some other way?

No, currying the function would make it equivalent to the data carrying version and hence make this function useless (an identity function).

> Alternatively, can we do this?

No, the pattern match in the `let` is hiding a call to `Exists.recOn` and is also not type correct for the same reason as your previous version. Existential statements aren't actually storing any value `x'` which can meaningfully be used in such an equality. Once its in the existential package we completely forget which value `x'` we were talking about, and because this existential is not uniquely inhabited you can't recover the `x'` unless you have a function like `satisfying` to do it.

Ericson2314:
> No, currying the function would make it equivalent to the data carrying version and hence make this function useless (an identity function).

There's no way to get a proof-irrelevant universal quantification for this? That seems like an odd deficiency. I would think `NonEmpty` ought to work, since class vs inductive is more or less sugar, right? One should be able to dependently eliminate it into `_ : Prop`, no?

> Existential statements aren't actually storing any value `x'`

understood.

> which can meaningfully be used in such an equality. 

I don't get this because the equality is a prop, no? We shouldn't need proof relevance here any more than we need proof relevance on the RHS of `SatisfiesM`.

> Once its in the existential package we completely forget which value x' we were talking about

Yes we cannot case on it, but we do have the propositional equality from the RHS to reason about it nevertheless.

> because this existential is not uniquely inhabited you can't recover the x' unless you have a function like satisfying to do it.

But `val_eq` doesn't need to recover it! It is just prooving the proposition that `satisfying` recovered it.

Also, if we have

```lean
∃ x' : m {a // p a}, Subtype.val <$> x' = x
Subtype.val <$> satisfying h = x // via old `val_eq`
```
then it should be possible to prove
```
Subtype.val <$> satisfying h = Subtype.val <$> x' 
satisfying h = x'
```
(with a sufficiently lawful functor)?

So even if my new version doesn't make any proofs shorter (since the strategy should use the old version as a lemma anyways) it should still work?

digama0:
Let's take this to zulip

PR #1030, opened 2024-11-08
feat: improve induction code action

digama0:
* The code action now works even on `induction e using rec` custom eliminators
* It also respects custom eliminators via the `@[custom_eliminator]` attribute
* It also works even if the induction already has branches, it will only add branches that are uncovered to the end of the match
* If the induction has a `|_ => tac` arm, it will delete this arm and use `tac` instead of `sorry` to fill in all the missing branches

### DIFF
--- a/Batteries/CodeAction/Misc.lean
+++ b/Batteries/CodeAction/Misc.lean
@@ -3,11 +3,9 @@ Copyright (c) 2023 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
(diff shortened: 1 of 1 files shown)


fgdorais:
I'm thinking there is no meaningful way to write tests for this. Correct?

digama0:
At least currently, yes that's correct. Ideally we would get UI tests similar to those in lean core at some point, but we'll have to use `make` again or write a custom harness for that.

fgdorais (approved):
I tested locally and it does work as expected. The new features in the PR description should be in the documentation somewhere.

Since this `Misc` file is getting long, we may consider splitting this one out right before merging.

PR #1037, opened 2024-11-11
fix: add missing `initHeartbeats`

eric-wieser:
In only one case a `Core.Context` is created from within CoreM, and we inherit from the parent Context the `initHeartbeats`. Otherwise, we get the current heartbeats from IO.

It's not entirely clear to me what the intended semantics of `lintCore` are; should they inherit the outer heartbeat count (as with this patch), or get a new heartbeat budget for each declaration (prior to it)?

### DIFF
--- a/Batteries/Tactic/Lint/Frontend.lean
+++ b/Batteries/Tactic/Lint/Frontend.lean
@@ -100,6 +100,7 @@ def lintCore (decls : Array Name) (linters : Array NamedLinter) :
     CoreM (Array (NamedLinter × Std.HashMap Name MessageData)) := do
   let env ← getEnv
   let options ← getOptions -- TODO: sanitize options?
+  let initHeartbeats := (← read).initHeartbeats
 
   let tasks : Array (NamedLinter × Array (Name × Task (Option MessageData))) ←
     linters.mapM fun linter => do
@@ -108,7 +109,7 @@ def lintCore (decls : Array Name) (linters : Array NamedLinter) :
         BaseIO.asTask do
           match ← withCurrHeartbeats (linter.test decl)
               |>.run' mkMetaContext
-              |>.run' {options, fileName := "", fileMap := default} {env}
+              |>.run' {options, fileName := "", fileMap := default, initHeartbeats } {env}
               |>.toBaseIO with
           | Except.ok msg? => pure msg?
           | Except.error err => pure m!"LINTER FAILED:\n{err.toMessageData}"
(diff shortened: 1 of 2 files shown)


eric-wieser:
Sorry, I missed that message.

What are you expecting Lean core to do here? Because we are constructing a `Lean.Core.Context` manually from scratch, it's our responsibility to fill in all of its fields.

Perhaps @kmill should have a look at this and suggest if we can avoid this manual construction in the first place (and confirm whether the PR in its current state is still an improvement over the status quo).

Batteries/Tactic/Lint/Frontend.lean
```diff
@@ -108,7 +109,7 @@ def lintCore (decls : Array Name) (linters : Array NamedLinter) :
         BaseIO.asTask do
           match ← withCurrHeartbeats (linter.test decl)
               |>.run' mkMetaContext
-              |>.run' {options, fileName := "", fileMap := default} {env}
+              |>.run' {options, fileName := "", fileMap := default, initHeartbeats } {env}
```
kmill, line 112: It looks like the `withCurrHeartbeats` sets `initHeartbeats` from `IO.getNumHeartbeats`. What effect does initializing `initHeartbeats` here have? Can you come up with a test that validates this?

Batteries/Util/Cache.lean
```diff
@@ -64,8 +64,9 @@ def Cache.get [Monad m] [MonadEnv m] [MonadLog m] [MonadOptions m] [MonadLiftT B
       -- TODO: add customization option
       let options := maxHeartbeats.set options <|
         options.get? maxHeartbeats.name |>.getD 1000000
+      let initHeartbeats ← IO.getNumHeartbeats
       let res ← EIO.asTask <|
-        init {} |>.run' {} { options, fileName, fileMap } |>.run' { env }
+        init {} |>.run' {} { options, fileName, fileMap, initHeartbeats } |>.run' { env }
```
kmill, line 69: I see that `Lean.Core.wrapAsync` exists, and it does a bit more heartbeats arithmetic. `Lean.Core.CoreM.toIO` looks relevant.

Batteries/Tactic/Lint/Frontend.lean
eric-wieser: I think you're right, this has no effect.

Batteries/Util/Cache.lean
eric-wieser: `wrapAsync` indeed looks like the right tool, as otherwise the hearbeat counts on one thread are meaninglessly transferred to another thread.

fgdorais:
Have issues been fixed to everyone's satisfaction?

fgdorais:
I'm afraid I lost track of this one. What's the current status?

eric-wieser:
I need to update the description to reflect the new version, but I think the approach now aligns with @kmill's suggestion.

The mathlib failure suggests that either I have broken the heartbeat counting, or that we were previously erroneously resetting the heartbeat counter mid-linter.

PR #1050, opened 2024-11-14
fix: use two tokens for open private/export private

david-christiansen:
This allows users to include more than one space between the tokens, and will also allow building once #6012 is merged in Lean.


fgdorais:
Please ignore the auto-label workflow failure, we're still working on that script...

fgdorais:
Looks good but needs some tests.

david-christiansen:
Done!

fgdorais (approved):

PR #1054, opened 2024-11-19
chore: remove @[simp] attributes from monad-specific SatisfiesM lemmas

kim-em:
When experimenting with `SatisfiesM` recently, I've found these inconvenient, as they too eagerly invoke `.run`, when one might still want to be doing arbitrary-monad reasoning.


kim-em:
(I'm happy leaving these simp lemmas on for `Id`, `Option`, and even `Except`; I just want to remove @[simp] from the lemmas that introduce an extra `run` on the right hand side.)

PR #1055, opened 2024-11-19
refactor: add List.finRange and Array.finRange

fgdorais:
Harmonize with Mathlib, renaming Fin.list to List.finRange and Fin.enum to Array.finRange.

### DIFF
--- a/Batteries/Data/Array/Basic.lean
+++ b/Batteries/Data/Array/Basic.lean
@@ -175,6 +175,9 @@ def eraseIdx! (a : Array α) (i : Nat) : Array α :=
     have : Inhabited (Array α) := ⟨a⟩
     panic! s!"index {i} out of bounds"
 
+/-- `finRange n` is the array of all elements of `Fin n` in order. -/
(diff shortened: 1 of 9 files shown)


fgdorais:
This is a second attempt at #899. The Mathlib breakages are too much for now. On hiatus.

kim-em:
Perhaps it would be easier to change the definition in Mathlib first? I appreciate that this situation is super annoying.

We could also just rename Mathlib's definition to `finRange'` and then gradually switch it over.

(I am quite keen to have `finRange` fixed soon, so I am happy to help work on this.)

fgdorais:
Help is more than welcome! It looks like it needs a long work window to resolve the Mathlib issues, I don't see such a long window until mid December. I'm also happy with your proposed workaround, if Mathlib folks are on board.

kim-em:
@fgdorais, the Mathlib PR changing the definition of `finRange` has landed (note that it uses `ofFn fun x => x`, not `ofFn id`!).

Would you like to try again here? I'm very keen to have this in.

fgdorais:
In progress.

PR #1063, opened 2024-11-24
chore: replace Batteries.HashMap with a shim around Std.HashMap

kim-em:


Batteries/Data/HashMap/Basic.lean
```diff
@@ -345,7 +115,11 @@ hashMap.findEntry? "one" = some ("one", 1)
 hashMap.findEntry? "three" = none
 ```
 -/
-@[inline] def findEntry? (self : HashMap α β) (a : α) : Option (α × β) := self.1.findEntry? a
+@[inline,
+  deprecated "Use `m[a]?` instead. Previously this might have returned a key that was not \
+    identical to the input, if `==` ignores some part of the value. \
+    Now it just returns the input key."]
+def findEntry? (self : HashMap α β) (a : α) : Option (α × β) := (a, ·) <$> self[a]?
```
digama0, line 122: Recording that I don't think this is an acceptable regression, and contrariwise we will want to deprecate such a misleading function in the future if we have it. I think this should either be implemented directly here or implemented upstream and used here.

Batteries/Data/HashMap/Basic.lean
```diff
@@ -482,7 +252,7 @@ def toArray (self : HashMap α β) : Array (α × β) :=
   self.fold (init := #[]) fun r k v => r.push (k, v)
 
 /-- The number of buckets in the hash map. -/
-def numBuckets (self : HashMap α β) : Nat := self.1.buckets.1.size
+def numBuckets (self : HashMap α β) : Nat := self.1.1.buckets.size
```
TwoFX, line 255: It doesn't really make a difference, but rather than reaching into the guts of the hash map you could invoke `Std.HashMap.Internal.numBuckets` here.

Batteries/Data/HashMap/Basic.lean
```diff
@@ -297,7 +66,7 @@ hashMap.insert "three" 3 = {"one" => 1, "two" => 2, "three" => 3}
 hashMap.insert "two" 0 = {"one" => 1, "two" => 0}
 ```
 -/
-def insert (self : HashMap α β) (a : α) (b : β) : HashMap α β := ⟨self.1.insert a b, self.2.insert⟩
+def insert (self : HashMap α β) (a : α) (b : β) : HashMap α β := Std.HashMap.insert self a b
```
TwoFX, line 69: I'm surprised this wasn't `@[inline]` (considering that `erase` is).

Batteries/Data/HashMap/Basic.lean
```diff
@@ -333,19 +99,7 @@ The function `f` is passed the original key of the entry, along with the value i
 ```
 -/
 def modify (self : HashMap α β) (a : α) (f : α → β → β) : HashMap α β :=
```
TwoFX, line 101: Given that this is just a wrapper now, I think it should be `@[inline]`.

Batteries/Data/HashMap/Basic.lean
kim-em: It's now been implemented directly, thanks to @TwoFX actually knowing the full API.

kim-em:
We've decided to wait on this until some further verification lemmas have been written for `Std.HashMap`. These are now being actively worked on, so hopefully it won't be too long.

kim-em:
Those additional verification lemmas about HashMap.toList and HashMap.fold have now landed, so as far as I'm aware we have feature parity with downstream libraries based on the old HashMap, and I would like to proceed.

PR #1065, opened 2024-11-25
feat: add KMP frontend for lists

fgdorais:


kim-em:
I know this would be revising stuff that already exists, but could we have an API whereby one doesn't ever need to see `Matcher`, in the common case of just doing a single lookup?

(Perhaps there could be a function directly on List, that takes on optional Matcher argument, with default value that constructs the Matcher?)

fgdorais:
I'm not sure what these would return. Unlike `String` there is no `Sublist` type to work with.

fgdorais:
I added the same functions as for strings.

kim-em:
Great, thanks. I'm happy with those.

A naming question, however: `List.Sublist` already means *not-necessarily-contiguous* sublist, whereas we use the word "infix" for contiguous sublists. How would you feel about changing the List parts of the API to be consistent with this? (While leaving the String API in terms of Substring, and Array API in terms of Subarray of course.)

fgdorais:
Names changed.

PR #1071, opened 2024-12-02
feat: Define dependent version of `Fin.foldl`

quangvdao:
This PR aims to extend `Fin.foldl` and `Fin.foldr` to their dependent versions.

This is very much WIP; I only have time to define the basic definition `Fin.fold`. It is close enough to the definition of `Fin.foldl` that I think we could replace it upstream by the end of this PR.

Name suggestion: can we rename to `Fin.fold` and `Fin.foldRev` instead of `foldl` and `foldr`?

### DIFF
--- a/Batteries/Data/Fin/Fold.lean
+++ b/Batteries/Data/Fin/Fold.lean
@@ -45,3 +45,21 @@ theorem foldr_eq_foldr_finRange (f : Fin n → α → α) (x) :
 
 @[deprecated (since := "2024-11-19")]
 alias foldr_eq_foldr_list := foldr_eq_foldr_finRange
+
+variable (n : Nat) (α : Fin (n + 1) → Sort _)
+
+/-- Dependent version of `Fin.foldl`. -/
+def fold (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
+    loop 0 (Nat.zero_lt_succ n) init where
+  loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α (last n) :=
+    if h' : i < n then
+      loop (i + 1) (Nat.succ_lt_succ h') (f ⟨i, h'⟩ x)
+    else
+      haveI : ⟨i, h⟩ = last n := by
+        ext
+        simp only [val_last]
+        exact Nat.eq_of_lt_succ_of_not_lt h h'
+      _root_.cast (congrArg α this) x
+  termination_by n - i
+  decreasing_by
+    exact Nat.sub_add_lt_sub h' (Nat.lt_add_one 0)


Batteries/Data/Fin/Fold.lean
```diff
@@ -45,3 +45,21 @@ theorem foldr_eq_foldr_finRange (f : Fin n → α → α) (x) :
...
+      loop (i + 1) (Nat.succ_lt_succ h') (f ⟨i, h'⟩ x)
+    else
+      haveI : ⟨i, h⟩ = last n := by
+        ext
+        simp only [val_last]
+        exact Nat.eq_of_lt_succ_of_not_lt h h'
+      _root_.cast (congrArg α this) x
+  termination_by n - i
+  decreasing_by
+    exact Nat.sub_add_lt_sub h' (Nat.lt_add_one 0)
```
kim-em, line 65: You can just drop these lines.

Batteries/Data/Fin/Fold.lean
```diff
@@ -45,3 +45,21 @@ theorem foldr_eq_foldr_finRange (f : Fin n → α → α) (x) :
...
+def fold (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
+    loop 0 (Nat.zero_lt_succ n) init where
+  loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α (last n) :=
+    if h' : i < n then
+      loop (i + 1) (Nat.succ_lt_succ h') (f ⟨i, h'⟩ x)
+    else
+      haveI : ⟨i, h⟩ = last n := by
+        ext
+        simp only [val_last]
+        exact Nat.eq_of_lt_succ_of_not_lt h h'
```
kim-em, line 61:
```suggestion
      have : ⟨i, h⟩ = last n := by ext; simp; omega
```

kim-em:
Can you write the theorem relating it to the usual `Fin.fold`, with `\alpha := \Sigma i, \alpha i`?

kim-em:
Other directions that would be good:
* just as we have `Fin.fold` and `Nat.fold` and a theorem relating them, we should have the "unbundled" version of this for `Nat`, and the theorem
* we should have the monadic version as well.

quangvdao:
> Can you write the theorem relating it to the usual `Fin.fold`, with `\alpha := \Sigma i, \alpha i`?

Actually, I don't know how to state the theorem. A natural attempt is:
```
theorem fold_eq_foldl_sigma {α : Fin (n+1) → Sort _} (f : ∀ (i : Fin n), α i.castSucc → α i.succ)
    (x : α 0) :
    fold n α f x = foldl n (α := (i : Fin (n + 1)) × α i) (fun x i => f x.1 x.2) ⟨0, x⟩ := sorry
```
but we run into the problem of Lean not knowing that `x.1` is supposed to be the same as `i.castSucc` during execution.

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,69 @@ alias enum := Array.finRange
 
 @[deprecated (since := "2024-11-15")]
 alias list := List.finRange
+
+variable (n : Nat) (α : Fin (n + 1) → Sort _)
```
fgdorais, line 21: Avoid using `variable` in Batteries.

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,69 @@ alias enum := Array.finRange
...
+
+/-- Dependent version of `Fin.foldl`. -/
+@[inline] def fold (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
```
fgdorais, line 24: Define using the monadic version with `m := Id`.

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,69 @@ alias enum := Array.finRange
...
+  /-- Inner loop for `Fin.fold`. `Fin.fold.loop n α f i h x = f n (f (n-1) (... (f i x)))`  -/
+  loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α (last n) :=
+    if h' : i < n then
+      loop (i + 1) (Nat.succ_lt_succ h') (f ⟨i, h'⟩ x)
+    else
+      haveI : ⟨i, h⟩ = last n := by ext; simp; omega
+      _root_.cast (congrArg α this) x
+
+/-- Dependent version of `Fin.foldr`. -/
+@[inline] def foldRev (f : ∀ (i : Fin n), α i.succ → α i.castSucc) (init : α (last n)) : α 0 :=
```
fgdorais, line 35: Define using the monadic version with `m := Id`.

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,69 @@ alias enum := Array.finRange
...
+  loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α 0 :=
+    if h' : i > 0 then
+      loop (i - 1) (by omega) (f ⟨i - 1, by omega⟩
+        (by simpa [Nat.sub_one_add_one_eq_of_pos h'] using x))
+    else
+      haveI : ⟨i, h⟩ = 0 := by ext; simp; omega
+      _root_.cast (congrArg α this) x
+
+/-- Dependent version of `Fin.foldlM`. -/
+@[inline] def foldM [Monad m] (f : ∀ (i : Fin n), α i.castSucc → m (α i.succ)) (init : α 0)
```
fgdorais, line 48:
```suggestion
@[specialize] def foldM [Monad m] (f : ∀ (i : Fin n), α i.castSucc → m (α i.succ)) (init : α 0)
```

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,69 @@ alias enum := Array.finRange
...
+  -/
+  loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : m (α (last n)) :=
+    if h' : i < n then
+      (f ⟨i, h'⟩ x) >>= loop (i + 1) (Nat.succ_lt_succ h')
+    else
+      haveI : ⟨i, h⟩ = last n := by ext; simp; omega
+      _root_.cast (congrArg (fun i => m (α i)) this) (pure x)
+
+/-- Dependent version of `Fin.foldrM`. -/
+@[inline] def foldRevM [Monad m] (f : ∀ (i : Fin n), α i.succ → m (α i.castSucc)) (init : α (last n))
```
fgdorais, line 67:
```suggestion
@[specialize] def foldRevM [Monad m] (f : ∀ (i : Fin n), α i.succ → m (α i.castSucc)) (init : α (last n))
```

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -13,3 +13,96 @@ attribute [norm_cast] val_last
 /-! ### clamp -/
 
 @[simp] theorem coe_clamp (n m : Nat) : (clamp n m : Nat) = min n m := rfl
+
+/-! ### foldM -/
+
+variable {n : Nat} {α : Fin (n+1) → Type _} [Monad m]
```
fgdorais, line 19: Avoid using `variable` in Batteries.

fgdorais (changes requested):
`foldl` and `foldr` are standard names in functional programming, we should stick to that convention. Use `dfoldl`/`dfoldr` or `foldlDep`/`foldrDep`.

Batteries/Data/Fin/Basic.lean
quangvdao: What is the difference between `inline` and `specialize`?

Batteries/Data/Fin/Basic.lean
quangvdao: Hmm, I don't want to do this, because it would restrict me to `α : Type _` instead of `α : Sort _`. And I do need `α : Prop` for my project.

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,71 @@ alias enum := Array.finRange
...
+    else
+      haveI : ⟨i, h⟩ = 0 := by ext; simp; omega
+      _root_.cast (congrArg (fun i => m (α i)) this) (pure x)
+
+/-- Dependent version of `Fin.foldl`. -/
+@[specialize] def dfoldl (n : Nat) (α : Fin (n + 1) → Sort _)
+    (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
+  loop 0 (Nat.zero_lt_succ n) init where
+  /-- Inner loop for `Fin.dfoldl`. `Fin.dfoldl.loop n α f i h x = f n (f (n-1) (... (f i x)))` -/
+  loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α (last n) :=
```
fgdorais, line 68: Sorry, my previous suggestion had a mistake... `specialize` goes on the inner loop. Same for the others.
 
```suggestion
def dfoldl (n : Nat) (α : Fin (n + 1) → Sort _)
    (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
  loop 0 (Nat.zero_lt_succ n) init where
  /-- Inner loop for `Fin.dfoldl`. `Fin.dfoldl.loop n α f i h x = f n (f (n-1) (... (f i x)))` -/
  @[specialize] loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α (last n) :=
```

`specialize` makes lean generate new code for each `f`. It's similar to inlining `f` rather than the other way around.

Batteries/Data/Fin/Basic.lean
fgdorais: OK. It's a pity for all the code duplication...

Batteries/Data/Fin/Basic.lean
fgdorais: Answered in another comment.

Batteries/Data/Fin/Basic.lean
fgdorais: I think the inner loop should also be `semireducible`. There was a recent issue where it was a problem that it is `irreducible` because of the well-founded recursion.

quangvdao: That would have been an issue for the non-dependent versions as well, right? I don't see `semireducible` attribute for `Fin.foldl.loop`, say.

fgdorais: https://github.com/leanprover/lean4/blob/d9d54c1f99363f42d1485f6b67f281a5985f684e/src/Init/Data/Fin/Fold.lean#L16C1-L17C1

quangvdao: Thanks, I was looking at an older version. Will make changes

quangvdao: Actually, it's weird that only `Fin.foldl.loop` has `semireducible` attribute, but not other variants. I also added `semireducible` to only `Fin.dfoldl.loop`.

fgdorais: The better alternative would be to use `match` instead of `if ... then ... else ...` so that structural recursion can be used instead.

fgdorais:
@kim-em could you protect `Fin.cast` in Lean 4 when you have time? The few `_root_.cast` uses here are annoying.

fgdorais:
Looks good. Make sure to add `specialize` on the inner loops. If you have time, expand the docstrings using `Init.Data.Fin.Fold` as a model. I'll have another pass tomorrow after a good night of sleep!

fgdorais:
Avoid code duplication by defininig:
```lean4
@[inline] def dfoldrM [Monad m] (n : Nat) (α : Fin (n + 1) → Type _)
    (f : ∀ (i : Fin n), α i.succ → m (α i.castSucc)) (init : α (last n)) : m (α 0) := 
  dfoldr n (fun i => m (α i)) (fun i x => x >>= f i) (pure init)
```
Same for `dfoldlM`.

Batteries/Data/Fin/Basic.lean
```diff
@@ -17,3 +17,71 @@ alias enum := Array.finRange
...
+  loop n (Nat.lt_succ_self n) init where
+  /-- Inner loop for `Fin.dfoldr`.
+    `Fin.dfoldr.loop n α f i h x = f 0 (f 1 (... (f i x)))`  -/
+  @[semireducible] loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α 0 :=
+    if h' : i > 0 then
+      loop (i - 1) (by omega) (f ⟨i - 1, by omega⟩
+        (by simpa [Nat.sub_one_add_one_eq_of_pos h'] using x))
+    else
+      haveI : ⟨i, h⟩ = 0 := by ext; simp; omega
+      _root_.cast (congrArg α this) x
```
fgdorais, line 87: Use structural recursion:
```suggestion
  @[specialize] loop (i : Nat) (h : i < n + 1) (x : α ⟨i, h⟩) : α 0 :=
    match i with
    | i + 1 => loop i (Nat.lt_of_succ_lt h) (f ⟨i, Nat.lt_of_succ_lt_succ h⟩ x)
    | 0 => x
```

quangvdao:
> Avoid code duplication by defininig:
> 
> ```
> @[inline] def dfoldrM [Monad m] (n : Nat) (α : Fin (n + 1) → Type _)
>     (f : ∀ (i : Fin n), α i.succ → m (α i.castSucc)) (init : α (last n)) : m (α 0) := 
>   dfoldr n (fun i => m (α i)) (fun i x => x >>= f i) (pure init)
> ```
> 
> I'm not sure the same works for `dfoldlM`.

When I tried to do this, the theorems can't seem to infer the type of `α` and `x`.

fgdorais:
The `loop` lemmas are no longer relevant after this change. The other lemmas are immediate consequences of the `dfoldr` lemmas.

Batteries/Data/Fin/Fold.lean
```diff
@@ -1,12 +1,175 @@
...
+  | zero => simp only [dfoldl_zero, foldl_zero]
+  | succ n ih =>
+    simp only [dfoldl_succ, foldl_succ, Function.comp_apply, Function.comp_def]
+    congr; simp only [ih]
+
+/-! ### dfoldr -/
+
+theorem dfoldr_loop_zero (f : (i : Fin n) → α i.succ → α i.castSucc) (x) :
+    dfoldr.loop n α f 0 (Nat.zero_lt_succ n) x = x := by
+  rw [dfoldr.loop]
```
fgdorais, line 133: The simpler proof is now `rfl` but this lemma is no longer needed.

Batteries/Data/Fin/Fold.lean
```diff
@@ -1,12 +1,175 @@
...
+
+theorem dfoldr_loop_succ (f : (i : Fin n) → α i.succ → α i.castSucc) (h : i < n) (x) :
+      dfoldr.loop n α f (i+1) (Nat.add_lt_add_right h 1) x =
+        dfoldr.loop n α f i (Nat.lt_add_right 1 h) (f ⟨i, h⟩ x) := by
+  rw [dfoldr.loop]
```
fgdorais, line 138: The simpler proof is now `rfl` but this lemma is no longer needed.

Batteries/Data/Fin/Fold.lean
```diff
@@ -1,12 +1,175 @@
...
+
+theorem dfoldr_loop (f : (i : Fin (n+1)) → α i.succ → α i.castSucc) (h : i+1 ≤ n+1) (x) :
+      dfoldr.loop (n+1) α f (i+1) (Nat.add_lt_add_right h 1) x =
+        f 0 (dfoldr.loop n (α ∘ succ) (f ·.succ) i h x) := by
+  induction i with
+  | zero => simp [dfoldr_loop_succ, dfoldr_loop_zero]
+  | succ i ih => rw [dfoldr_loop_succ _ h, dfoldr_loop_succ _ (Nat.succ_lt_succ_iff.mp h),
+      ih (Nat.le_of_succ_le h)]; rfl
```
fgdorais, line 146:
```suggestion
  induction i with
  | zero => rfl
  | succ i ih => exact ih ..
```

Batteries/Data/Fin/Fold.lean
```diff
@@ -1,12 +1,175 @@
...
+
+@[simp] theorem dfoldr_zero (f : (i : Fin 0) → α i.succ → α i.castSucc) (x) :
+    dfoldr 0 α f x = x := dfoldr_loop_zero ..
```
fgdorais, line 149:
```suggestion
@[simp] theorem dfoldr_zero (f : (i : Fin 0) → α i.succ → α i.castSucc) (x) :
    dfoldr 0 α f x = x := rfl
```

fgdorais (changes requested):

fgdorais:
I completed the last few changes for you. Let me know if you would like to revert something.

fgdorais (approved):

fgdorais:
I just approved but I then noticed that the docstrings should not only refer to other functions. Please expand them and mention the relation with non-dependent functions as an additional note.

quangvdao:
I filled out the docstrings. Thanks for all the help!

kim-em:
> @kim-em could you protect `Fin.cast` in Lean 4 when you have time? The few `_root_.cast` uses here are annoying.

Done in https://github.com/leanprover/lean4/pull/6315

PR #1074, opened 2024-12-06
refactor: dependent fold for `Fin`

fgdorais:
Define `dfoldl` and `dfoldr` from their monadic versions to avoid code duplication. The cost is that they now require `Type _` instead of `Sort _`. However, induction principles are already available to deal with `Prop` so there is no overall loss.


quangvdao:
Sorry for the wait! Looks good to me.

PR #1078, opened 2024-12-23
feat: add BitVec.ofFn and lemmas

fgdorais:
- [x] Depends on #1089 
- [x] Depends on #1096


Batteries/Data/BitVec.lean
```diff
@@ -0,0 +1,42 @@
...
+    (v.shiftConcat b)[0] = b := by simp [getElem_shifConcat]
+
+@[simp] theorem getElem_shiftConcat_succ (v : BitVec n) (b : Bool) (h : i + 1 < n) :
+    (v.shiftConcat b)[i+1] = v[i] := by simp [getElem_shifConcat]
+
+/-- `ofFnAux f` returns the `BitVec m` whose `i`th bit is `f i` when `i < m` -/
+@[inline] def ofFnAux (m : Nat) (f : Fin n → Bool) : BitVec m :=
+  Fin.foldr n (fun i v => v.shiftConcat (f i)) 0
+
+/-- `ofFn f` returns the `BitVec n` whose `i`th bit is `f i` -/
```
kim-em, line 24: Doc-string should specifying little-endian. Probably should also be renamed to `ofFnLE`, too, and add `ofFnBE`. For now, at least, the plan is to treat the two endianness equally in the BitVec library.

Batteries/Data/BitVec.lean
```diff
@@ -0,0 +1,42 @@
...
+abbrev ofFn (f : Fin n → Bool) : BitVec n := ofFnAux n f
```
kim-em, line 25: New functions creating BitVecs should have simp lemmas for `toNat`, `toFin`, and ideally also `toInt`.

Batteries/Data/BitVec.lean
fgdorais: Did a quick fix. Implementation with `Fin.foldl` is probably a bit more efficient. I'll take a look soon to see if that's worthwhile.

Should I add lemmas for other BitVec getters?

Batteries/Data/BitVec.lean
fgdorais: Doesn't that need the `Nat`/`Fin`/`Int` functions to already exist in some way? I would probably just implement those using the BitVec version.

Batteries/Data/BitVec.lean
```diff
@@ -0,0 +1,59 @@
+/-
+Copyright (c) 2024 François G. Dorais. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
+
+namespace BitVec
+
+theorem getElem_shifConcat (v : BitVec n) (b : Bool) (i) (h : i < n) :
```
kim-em, line 9:
```suggestion
theorem getElem_shiftConcat (v : BitVec n) (b : Bool) (i) (h : i < n) :
```

Batteries/Data/BitVec.lean
fgdorais: I added `Nat.ofBits` and `Fin.ofBits`, little endian only since that's the way `Nat.testBit` works. I will try to think about what the right `Int` version should be.

fgdorais: Adding `Int.ofBits` would require upstreaming `Int.testBit` from Mathlib. Let me know if you think that's a good idea and I will get it done.

kim-em:
This seems to have grown into two separate PRs smushed together?

fgdorais:
Yes, I was working on the split earlier today. I will push the prequel PR soon. I'm also inclined to drop the `toInt` for now. There's a lot of technical issues that I keep bumping into and I can't seem to find reasonable pathways to link all the necessary endpoints.

Batteries/Data/BitVec.lean
fgdorais: I'm opting delaying the `Int` part from this PR. The `Nat` and `Fin` will shortly be moved to the prequel PR #1089

fgdorais:
Split completed, `Int` part postponed to another PR.

Batteries/Data/BitVec.lean
```diff
@@ -0,0 +1,88 @@
+/-
+Copyright (c) 2024 François G. Dorais. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
```
kim-em, line 5: Would you mind splitting this files as `Defs` and `Lemmas`?

Batteries/Data/BitVec/Basic.lean
```diff
@@ -0,0 +1,19 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
+
+import Batteries.Data.Fin.OfBits
+
+namespace BitVec
+
+/-- `ofFnLEAux f` returns the `BitVec m` whose `i`th bit is `f i` when `i < m`, little endian. -/
+@[inline] def ofFnLEAux (m : Nat) (f : Fin n → Bool) : BitVec m :=
```
hargoniX, line 12: I'm not quite sure whether we want to have these functions in general. There is no chance that `bv_decide` can ever support this kind of `BitVec` as it would require reasoning about a function `Fin n -> Bool` which it cannot do. Whether this is an inclusion/exclusion criterium for `BitVec` stuff I don't know.

Batteries/Data/BitVec/Lemmas.lean
```diff
@@ -0,0 +1,78 @@
...
+Copyright (c) 2024 François G. Dorais. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
+
+import Batteries.Data.BitVec.Basic
+
+namespace BitVec
+
+theorem getElem_shiftConcat (v : BitVec n) (b : Bool) (i) (h : i < n) :
```
hargoniX, line 11: Such basic lemmas about `shiftConcat` should live in core if possible. I would suggest to coordinate with @bollu or @tobiasgrosser on this matter.

Batteries/Data/BitVec/Lemmas.lean
```diff
@@ -0,0 +1,78 @@
...
+    rw [Fin.foldr_succ, toNat_shiftConcat, Nat.shiftLeft_eq, Nat.pow_one, Nat.ofBits_succ, ih,
+      ← Nat.mod_add_div (Nat.ofBits (f ∘ Fin.succ)) (2 ^ m), Nat.mul_add 2, Nat.add_right_comm,
+      Nat.mul_left_comm, Nat.add_mul_mod_self_left, Nat.mul_comm 2]
+    rfl
+
+@[simp] theorem toFin_ofFnLEAux (m : Nat) (f : Fin n → Bool) :
+    (ofFnLEAux m f).toFin = Fin.ofNat' (2 ^ m) (Nat.ofBits f) := by
+  ext; simp
+
+@[simp] theorem toNat_ofFnLE (f : Fin n → Bool) : (ofFnLE f).toNat = Nat.ofBits f := by
```
hargoniX, line 37: Generally speaking we want a:
- `toNat`
- `toFin`
- `toInt`
- `getElem`
- `getLsbD`
- `getMsbD`
- `msb`

lemma for all `BitVec` operations to have at least a basic API that's available for all of them, not all of these seem to be given for either of these functions so far, could you add them?

fgdorais: Done.

Batteries/Data/BitVec/Basic.lean
fgdorais: I'm using these for linear algebra over the 2-element field, so this is not completely related to `bv_decide`.

Batteries/Data/BitVec/Lemmas.lean
kim-em: Yes please.

fgdorais: Rather than guessing, any specific place these should be added?

fgdorais: Done! [lean4#6875](https://github.com/leanprover/lean4/pull/6875)

PR #1084, opened 2025-01-04
chore: post reminder to create adaptation PR

kim-em:
This PR copies across the bot from Mathlib that posts a reminder to create a PR to the `bump/v4.X.0`, and provides the script to automate that process.


kim-em:
I *think* that the required secret is already installed here, but it dates from 2 years ago so may be out of date. As usual, we will test bots in production.

PR #1111, opened 2025-02-01
feat: add `List.findMap?` and lemmas

fgdorais:


fgdorais:
The `proof_wanted` are to outline a basic API. I'll probably fill-in or add more by the time this is out of WIP.

PR #1129, opened 2025-02-10
chore: fix simp direction for `monadLift_bind`

quangvdao:
This PR reverses the simp direction of lawful `monadLift` operations, so that they got pulled inward rather than outward. This is similar to the direction in classes like `MulHomClass` and also consistent with the [Haskell documentation](https://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-Class.html#t:MonadTrans).

We also duplicate the lemmas for `monadLift` and `liftM` to help in rewriting, which matches on syntax only. We set the simp attribute for lemmas about `liftM` only, since simp can look into definitions.

Finally, we add the canonical lifting operation from `Id` to any monad `m` via its `pure` function.


quangvdao:
@kim-em sorry to bother you again, but this PR should be quick to review

PR #1130, opened 2025-02-10
chore: remove explicit universes from `EStateM` results

eric-wieser:
This makes the code work both with and without leanprover/lean4#3010, making it easier to test/merge/revert that PR.

Right now the files are inconsistent regarding the use of implicit universe arguments, and so with the Lean change above, the downstream file ends up more general than the upstream one.

### DIFF
--- a/Batteries/Lean/EStateM.lean
+++ b/Batteries/Lean/EStateM.lean
@@ -9,22 +9,22 @@ namespace EStateM
 namespace Result
 
 /-- Map a function over an `EStateM.Result`, preserving states and errors. -/
-def map {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) : Result ε σ β :=
+def map {ε σ α β} (f : α → β) (x : Result ε σ α) : Result ε σ β :=
   match x with
   | .ok a s' => .ok (f a) s'
   | .error e s' => .error e s'
 
-@[simp] theorem map_ok {ε σ α β : Type u} (f : α → β) (a : α) (s : σ) :
+@[simp] theorem map_ok {ε σ α β} (f : α → β) (a : α) (s : σ) :
     (Result.ok a s : Result ε σ α).map f = .ok (f a) s := rfl
 
-@[simp] theorem map_error {ε σ α β : Type u} (f : α → β) (e : ε) (s : σ) :
+@[simp] theorem map_error {ε σ α β} (f : α → β) (e : ε) (s : σ) :
     (Result.error e s : Result ε σ α).map f = .error e s := rfl
 
-@[simp] theorem map_eq_ok {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) (b : β) (s : σ) :
+@[simp] theorem map_eq_ok {ε σ α β} (f : α → β) (x : Result ε σ α) (b : β) (s : σ) :
     x.map f = .ok b s ↔ ∃ a, x = .ok a s ∧ b = f a := by
   cases x <;> simp [and_assoc, and_comm, eq_comm]
 
-@[simp] theorem map_eq_error {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) (e : ε) (s : σ) :
+@[simp] theorem map_eq_error {ε σ α β} (f : α → β) (x : Result ε σ α) (e : ε) (s : σ) :
     x.map f = .error e s ↔ x = .error e s := by
   cases x <;> simp [eq_comm]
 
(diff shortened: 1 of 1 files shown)


Batteries/Lean/EStateM.lean
```diff
@@ -9,22 +9,22 @@ namespace EStateM
 namespace Result
 
 /-- Map a function over an `EStateM.Result`, preserving states and errors. -/
-def map {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) : Result ε σ β :=
+def map {ε σ α β} (f : α → β) (x : Result ε σ α) : Result ε σ β :=
```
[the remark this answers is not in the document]
eric-wieser: The proofs no longer compile if I make this change globally. I think the idea here is to put the arguments in a convenient order, which autoImplicits do not.

Batteries/Lean/EStateM.lean
```diff
@@ -33,7 +33,7 @@ end Result
 @[simp] theorem run_map (f : α → β) (x : EStateM ε σ α) :
     (f <$> x).run s = (x.run s).map f := rfl
 
-@[ext] theorem ext {ε σ α : Type u} (x y : EStateM ε σ α) (h : ∀ s, x.run s = y.run s) : x = y := by
+@[ext] theorem ext {ε σ α} (x y : EStateM ε σ α) (h : ∀ s, x.run s = y.run s) : x = y := by
```
fgdorais, line 36:
```suggestion
@[ext] theorem ext {ε σ α} {x y : EStateM ε σ α} (h : ∀ s, x.run s = y.run s) : x = y := by
```

Batteries/Lean/EStateM.lean
```diff
@@ -9,22 +9,22 @@ namespace EStateM
...
+@[simp] theorem map_error {ε σ α β} (f : α → β) (e : ε) (s : σ) :
     (Result.error e s : Result ε σ α).map f = .error e s := rfl
 
-@[simp] theorem map_eq_ok {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) (b : β) (s : σ) :
+@[simp] theorem map_eq_ok {ε σ α β} (f : α → β) (x : Result ε σ α) (b : β) (s : σ) :
     x.map f = .ok b s ↔ ∃ a, x = .ok a s ∧ b = f a := by
   cases x <;> simp [and_assoc, and_comm, eq_comm]
 
-@[simp] theorem map_eq_error {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) (e : ε) (s : σ) :
+@[simp] theorem map_eq_error {ε σ α β} (f : α → β) (x : Result ε σ α) (e : ε) (s : σ) :
```
fgdorais, line 27:
```suggestion
@[simp] theorem map_eq_error {ε σ α β} (f : α → β) {x : Result ε σ α} {e : ε} {s : σ} :
```

Batteries/Lean/EStateM.lean
```diff
@@ -9,22 +9,22 @@ namespace EStateM
...
-@[simp] theorem map_ok {ε σ α β : Type u} (f : α → β) (a : α) (s : σ) :
+@[simp] theorem map_ok {ε σ α β} (f : α → β) (a : α) (s : σ) :
     (Result.ok a s : Result ε σ α).map f = .ok (f a) s := rfl
 
-@[simp] theorem map_error {ε σ α β : Type u} (f : α → β) (e : ε) (s : σ) :
+@[simp] theorem map_error {ε σ α β} (f : α → β) (e : ε) (s : σ) :
     (Result.error e s : Result ε σ α).map f = .error e s := rfl
 
-@[simp] theorem map_eq_ok {ε σ α β : Type u} (f : α → β) (x : Result ε σ α) (b : β) (s : σ) :
+@[simp] theorem map_eq_ok {ε σ α β} (f : α → β) (x : Result ε σ α) (b : β) (s : σ) :
```
fgdorais, line 23:
```suggestion
@[simp] theorem map_eq_ok {ε σ α β} {f : α → β} {x : Result ε σ α} {b : β} {s : σ} :
```

fgdorais (changes requested):

Batteries/Lean/EStateM.lean
eric-wieser: What do you think about using `⦃⦄` binders for the `ext` lemmas? There's some precedent for that, as it means that these can be type-cast as `Function.Injective`?

fgdorais: In what sense does this math `Function.Injective`? I see `(fun s => x.run s) = (fun s => y.run s) -> x = y` but that doesn't seem very useful. I'm probably missing something.

eric-wieser: In precisely that sense. Agreed that it isn't very useful here (which is why I took your suggestion), but it's useful in some cases and mostly harmless in the rest, so perhaps sensible as a general pattern to consider in the future.

fgdorais (approved):

PR #1152, opened 2025-03-03
feat: Laws for monads with `Alternative` instances.

dtumad:
This PR adds type-classes `AlternativeMonad`, `LawfulAlternative`, and `LawfulAlternativeLift` that enforce laws regarding `failure` and `orElse`.

The file OptionT.lean contains lemmas that currently are part of mathlib, and would need to be removed there when this is merged, but they are needed to prove the instances for `OptionT`. Similarly for the functions `succeeds`, 

Additional instances for `Set` and `List` need to be written to mathlib as that is where the main monad instances for those types are.


Batteries/Control/AlternativeMonad.lean
```diff
@@ -0,0 +1,204 @@
...
+
+universe u v w
+
+section Alternative
+
+variable {F : Type → Type v} [Alternative F]
+
+-- [todo] add notation for `Functor.mapConst` and port `Functor.mapConstRev`
+/-- Returns `pure true` if the computation succeeds and `pure false` otherwise. -/
+def succeeds {α} (x : F α) : F Bool :=
```
eric-wieser, line 45: Upstreaming these seems orthogonal to the rest of the PR; I'd suggest splitting them into a separate PR, or leaving them in mathlib

Batteries/Control/AlternativeMonad.lean
```diff
@@ -0,0 +1,167 @@
...
+* `failure >>= g = failure`
+* `x <|> failure = x`
+* `failure <|> y = y`
+* `x <|> y <|> z = (x <|> y) <|> z`
+* `f <$> (x <|> y) = (f <$> x <|> f <$> y)`
+
+`Option`/`OptionT` are the most basic examples, but transformers like `StateT` also preserve
+the lawfulness of this on the underlying monad.
+
+The law `do _ ← x; failure = failure` is true for monads like `Option` and `List` that don't
```
eric-wieser, line 25: In the symbolic language above, this is `seqRight_failure : x *> failure = failure`

Batteries/Control/AlternativeMonad.lean
```diff
@@ -0,0 +1,167 @@
...
+  rw [map_eq_pure_bind, failure_bind]
+
+@[simp] theorem mapConst_failure {α β} (y : β) : Functor.mapConst y (failure : m α) = failure := by
+  rw [LawfulFunctor.map_const, Function.comp_apply, map_failure]
+
+@[simp] theorem mapConst_orElse {α β} (x x' : m α) (y : β) :
+    Functor.mapConst y (x <|> x') = (Functor.mapConst y x <|> Functor.mapConst y x') := by
+  simp only [map_const, Function.comp_apply, map_orElse]
+
+@[simp] theorem failure_seq {α β} (x : m α) : (failure : m (α → β)) <*> x = failure := by
```
eric-wieser, line 76: Could you add the versions for `*>` and `<*` as well?

Batteries/Control/AlternativeMonad.lean
```diff
@@ -0,0 +1,175 @@
...
+universe u v w
+
+/-- `AlternativeMonad m` means that `m` has both a `Monad` and `Alternative` instance,
+which both share the same underlying `Applicative` instance.
+The main example is `Option`, but many monad transformers also preserve or add this structure. -/
+class AlternativeMonad (m : Type u → Type v) extends Alternative m, Monad m
+
+/-- `LawfulAlternative m` means that the `failure` operation on `m` behaves naturally
+with respect to the `bind` and `orElse` operators. -/
+class LawfulAlternative (m : Type u → Type v) [AlternativeMonad m] extends LawfulMonad m where
```
eric-wieser, line 46:
```suggestion
class LawfulAlternative (m : Type u → Type v) [AlternativeMonad m] : Prop extends LawfulMonad m where
```
and similarly below for the `Lift` version.

Batteries/Control/OptionT.lean
```diff
@@ -0,0 +1,95 @@
...
+theorem run_map (f : α → β) [LawfulMonad m] : (f <$> x).run = Option.map f <$> x.run := by
+  rw [← bind_pure_comp _ x.run]
+  change x.run >>= (fun
+                     | some a => OptionT.run (pure (f a))
+                     | none   => pure none) = _
+  apply bind_congr
+  intro a; cases a <;> simp [Option.map, Option.bind]
+
+@[simp] theorem run_monadLift {n} [MonadLiftT n m] (x : n α) :
+    (monadLift x : OptionT m α).run = (monadLift x : m α) >>= fun a => pure (some a) := rfl
```
eric-wieser, line 47: Or maybe
```suggestion
    (monadLift x : OptionT m α).run = some <$> (monadLift x : m α) := rfl
```
(untested)

dtumad: This does require the base monad `m` be lawful but I don't think that's a very important generalization, and the map seems cleaner so I've changed it to that

Rob23oba:
I wouldn't immediately go for defining `LawfulAlternative` on `AlternativeMonad`s, the `failure_bind` rule can be recovered from `map_failure` and `failure_seq`:
```lean
class LawfulAlternative (m : Type u → Type v) [Alternative m] extends LawfulApplicative m where
  map_failure {α β : Type u} (f : α → β) : f <$> (failure : m α) = failure
  failure_seq {α β : Type u} (x : m α) : (failure : m (α → β)) <*> x = failure
  -- ...

export LawfulAlternative (map_failure failure_seq)

class AlternativeMonad (m : Type u → Type v) extends Alternative m, Monad m

theorem failure_bind [AlternativeMonad m] [LawfulMonad m] [LawfulAlternative m] (x : α → m β) : failure >>= x = failure := by
  conv => lhs; rw [← map_failure PEmpty.elim]
  conv => rhs; rw [← bind_pure failure]
  conv => rhs; rw [← map_failure PEmpty.elim]
  simp only [bind_map_left]
  congr 1; funext a
  exact a.elim
```

dtumad:
@Rob23oba updated to that definition for the class, using `PEmpty.elim` in the proof like that is a very neat trick.

dtumad:
What's the procedure for updating mathlib after upstreaming these `OptionT` lemmas? I'm not sure how to handle coordinating the two changes, should I just open a mathlib PR with the relevant lemmas removed? Or remove the imports from `Batteries.lean` so that mathlib doesn't see the duplicate lemmas when building mathlib?

Batteries/Control/AlternativeMonad.lean
```diff
@@ -73,21 +76,33 @@ variable {m : Type u → Type v} [AlternativeMonad m] [LawfulAlternative m] {α
...
 @[simp] theorem failure_seqRight (x : m α) : (failure : m β) *> x = failure := by
   simp only [seqRight_eq, map_failure, failure_seq]
 
-end LawfulAlternative
+end Alternative
+
+section AlternativeMonad
 
-section LawfulAlternativeLift
+variable{m : Type u → Type v} [AlternativeMonad m]
```
eric-wieser, line 89:
```suggestion
variable {m : Type u → Type v} [AlternativeMonad m]
```

Batteries/Control/AlternativeMonad.lean
```diff
@@ -0,0 +1,190 @@
...
+  orElse_assoc _ _ _ := by
+    simp only [OptionT.ext_iff, run_orElse, Option.elimM, bind_assoc]
+    refine bind_congr fun | some _ => by simp | none => rfl
+  map_orElse x y f := by
+    simp only [OptionT.ext_iff, run_map, run_orElse, map_bind, bind_map_left, Option.elimM]
+    refine bind_congr fun | some _ => by simp | none => rfl
+
+end OptionT
+
+namespace StateT
```
eric-wieser, line 152: Is this true for [`StateRefT'`](https://leanprover-community.github.io/mathlib4_docs/Init/Control/StateRef.html#StateRefT') too?

Rob23oba: I'd think this is also true for `StateRefT'` / `ReaderT`, those are usually the simplest instances.

eric-wieser: ReaderT is already here.

Rob23oba: Yeah right then this just a matter of "unfold ReaderT; infer_instance"

eric-wieser (approved):
LGTM, thanks!

You should get a mathlib adaption PR ready to go before this gets merged, so as not to hold up merging other PRs.

fgdorais:
This needs a Mathlib adaptation before merging. Ideally, this just involves pushing changes to the Mathlib branch [batteries-pr-testing-1152](https://github.com/leanprover-community/mathlib4/compare/batteries-pr-testing-1152) until it works. Let us know if you have questions of if you run into complications.

fgdorais:
Thanks! Could you create an actual Mathlib PR out of [batteries-pr-testing-1152](https://github.com/leanprover-community/mathlib4/tree/batteries-pr-testing-1152). I think this needs a review on the Mathlib end. You may have to ping someone to review it. Ask @eric-wieser if they are interested or if they have alternate recommendations for that review. Once that's done, ping back here since Batteries maintainers won't be notified automatically.

eric-wieser:
I'd be happy to review it

dtumad:
Sorry about the slow response. [This PR has the needed changes](https://github.com/leanprover-community/mathlib4/pull/22859/files). Should I change the mathlib lakefile back to main batteries repo before this gets merged here?

Batteries/Control/OptionT.lean
```diff
@@ -0,0 +1,110 @@
...
+
+section Monad
+
+variable [Monad m]
+
+@[simp] theorem run_pure (a) : (pure a : OptionT m α).run = pure (some a) := rfl
+
+@[simp]
+theorem run_bind (f : α → OptionT m β) :
+    (x >>= f).run = Option.elimM x.run (pure none) (run ∘ f) := by
```
eric-wieser, line 52: Can you mention in the description that these were changed to use `elimM` as part of upstreaming from mathlib?

Batteries/Control/OptionT.lean
```diff
@@ -0,0 +1,110 @@
...
+theorem run_map (f : α → β) [LawfulMonad m] : (f <$> x).run = Option.map f <$> x.run := by
+  rw [← bind_pure_comp _ x.run]
+  change x.run >>= _ = _
+  exact bind_congr fun |some _ => rfl | none => rfl
+
+@[simp] theorem run_monadLift {n} [LawfulMonad m] [MonadLiftT n m] (x : n α) :
+    (monadLift x : OptionT m α).run = some <$> (monadLift x : m α) := (map_eq_pure_bind _ _).symm
+
+protected theorem mapConst_eq_map_const (y : β) (x : OptionT m α) :
+    Functor.mapConst y x = Function.const α y <$> x := rfl
```
eric-wieser, line 67: Is this necessary? There is a global `mapConst_eq_map_const` lemma that is provided by the `LawfulMonad` instead immediately below.

dtumad: No, and the proof below is just `run_map _ _` without using `rw`

Batteries/Control/OptionT.lean
```diff
@@ -47,6 +47,8 @@ variable [Monad m]
 
 @[simp] theorem run_pure (a) : (pure a : OptionT m α).run = pure (some a) := rfl
 
+/-- This was changed during porting from mathlib to use Option.elimM rather than explicit
+pattern matching, aligning with `run_seq` and `run_orElse`. -/
```
eric-wieser, line 51: This belongs in the PR description, not the file.

eric-wieser:
@fgdorais, I'm happy with the adaptation PR, just the stray comment to delete above.

Batteries/Control/OptionT.lean
dtumad: Ah sorry, that makes sense

fgdorais:
I made a few style changes because Batteries prefers to avoid section variables and also makes systematic use of autoimplicits. Please have a quick look at my changes in case there are instances where Lean doesn't automatically make the right choice.

fgdorais (approved):

PR #1162, opened 2025-03-15
Don't use implicit lambdas in by_contra

alexjbest:


alexjbest:
@urkud is this what you wanted?

JovanGerb:
Can you apply the same fix to `absurd`?

Batteries/Tactic/Init.lean
```diff
@@ -42,8 +42,8 @@ macro (name := byContra) tk:"by_contra" e?:(ppSpace colGt binderIdent)? : tactic
     | none => Unhygienic.run `(_%$tk)
   `(tactic| first
     | guard_target = Not _; intro $e:term
-    | refine Decidable.byContradiction fun $e => ?_
-    | refine Classical.byContradiction fun $e => ?_)
+    | refine @Decidable.byContradiction _ fun $e => ?_
```
eric-wieser, line 45: This is missing an argument I think

alexjbest: Ah yes I suppose it is!

eric-wieser: On zulip there are now some other fixes which don't rely 
```suggestion
    | refine (Decidable.byContradiction fun $e => ?_ :)
```
I think works, while avoiding having to count arguments.

fgdorais:
These are very nice changes but, as @JovanGerb pointed out, `absurd` could be adapted in a similar manner, and there needs to be a [Mathlib adaptation](https://github.com/leanprover-community/batteries?tab=readme-ov-file#mathlib-adaptations).

eric-wieser:
https://github.com/leanprover-community/mathlib4/pull/23221 should fix the mathlib build, and doesn't require this to be merged first.

Batteries/Tactic/Init.lean
eric-wieser: I pushed this fix.

eric-wieser:
> `absurd` could be adapted in a similar manner

I've applied this change.

PR #1171, opened 2025-03-21
chore: generalize `TransCmp` -> `OrientedCmp`

JovanGerb:
This PR generalizes some typeclasses. They were found using a linter.

[#mathlib4 > Linter for generalizing type class hypotheses](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Linter.20for.20generalizing.20type.20class.20hypotheses)

### DIFF
--- a/Batteries/Classes/Order.lean
+++ b/Batteries/Classes/Order.lean
@@ -20,26 +20,24 @@ class OrientedCmp (cmp : α → α → Ordering) : Prop where
 
 namespace OrientedCmp
 
-variable {α} {cmp : α → α → Ordering} [OrientedCmp cmp] {x y : α}
-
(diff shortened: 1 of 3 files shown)


fgdorais:
Could you split the `SatisfiesM` part into a separate PR? That part can be merged immediately.

The `cmp` stuff is good but that old API could use a more substantial review, so it might take more time to merge. Also, keep in mind that Batteries makes systematic use of autoimplicits and tends to avoid section variables (with a few exceptions in both cases).

JovanGerb:
All right, I split off the satisfiesM change in #1172.

fgdorais:
The `cmp` API does still need a serious review but the revised changes are minimal enough to merge regardless. Thank you for the edits!

PR #1178, opened 2025-03-27
refactor: runLinter Lake adaptions

tydeu:
Forward-compatible adaption of the `runLinter.lean` default module detection that avoids use of Lake APIs that are undergoing breaking refactors.


tydeu:
Closing as an equivalent adaption has already been merged as part of a Lean bump.

PR #1181, opened 2025-03-30
chore: adjust cron jobs for nightly-testing

kim-em:


kim-em:
Ideally this would be on a push basis, but CI is complicated enough as is.

PR #1194, opened 2025-04-07
chore: mark `Nat.sqrt.iter` as semireducible

b-mehta:
This definition has become irreducible in v4.19.0-rc2, which blocks reduction of `Nat.sqrt`, as well as `IsSquare` and `Nat.minFac` in mathlib, both functions which are used for decidability there. Furthermore, marking `Nat.sqrt.iter` as `semireducible` locally does not make those reduce appropriately. Further still, `decide +kernel` - which is meant to ignore irreducible tags - fails to see through `Nat.sqrt.iter`, so the only solution is this one.

### DIFF
--- a/Batteries/Data/Nat/Basic.lean
+++ b/Batteries/Data/Nat/Basic.lean
@@ -94,7 +94,7 @@ def sqrt (n : Nat) : Nat :=
   iter n (n / 2)
 where
   /-- Auxiliary for `sqrt`. If `guess` is greater than the integer square root of `n`,
-  returns the integer square root of `n`. 
+  returns the integer square root of `n`.
 
   By default this well-founded recursion would be irreducible.
   This prevents use `decide` to resolve `Nat.sqrt n` for small values of `n`,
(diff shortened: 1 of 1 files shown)


Batteries/Data/Nat/Basic.lean
```diff
@@ -94,8 +94,12 @@ def sqrt (n : Nat) : Nat :=
   iter n (n / 2)
 where
   /-- Auxiliary for `sqrt`. If `guess` is greater than the integer square root of `n`,
-  returns the integer square root of `n`. -/
-  iter (n guess : Nat) : Nat :=
+  returns the integer square root of `n`.
```
b-mehta, line 97:
```suggestion
  returns the integer square root of `n`.
```

fgdorais:
LGTM! Is there a chance `Nat.sqrt` could be made structural using fuel?

b-mehta:
Yes, I expect so, but that would be a more difficult change! Something like
```
def sqrtAux (n guess : Nat) (fuel : Nat) : Nat :=
  match fuel with
  | 0 => guess
  | k + 1 =>
    let next := (guess + n / guess) / 2
    (guess.ble next).casesOn (sqrtAux n next k) guess

def sqrt (n : Nat) : Nat :=
  (n.ble 1).casesOn (sqrtAux n (n / 2) n) n
 ```
(by @Rob23oba) should work. 
It would, however, require nontrivial changes in mathlib since the correctness proofs of Nat.sqrt would change, and my aim with this PR was to be as unobtrusive as possible.

PR #1204, opened 2025-04-10
chore: upstream shake from Mathlib to Batteries

Vierkantor:
The `shake` command doesn't depend on Mathlib so let's upstream it and make it available for other projects.

Zulip discussion: https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Spinning.20off.20.60mathlib-linters.60.20and.20.60shake.60.20repo

### DIFF
--- a/lakefile.toml
+++ b/lakefile.toml
@@ -22,3 +22,8 @@ supportInterpreter = true
 [[lean_exe]]
 name = "test"
 srcDir = "scripts"
+
+[[lean_exe]]
+name = "shake"
+root = "Shake.Main"
+supportInterpreter = true
--- a/Shake/Main.lean
+++ b/Shake/Main.lean
@@ -3,8 +3,10 @@ Copyright (c) 2023 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
+import Lean.Environment
+import Lean.Parser.Module
 import Lean.Util.FoldConsts
-import Lean
+import Lean.Util.Paths
 
 /-! # `lake exe shake` command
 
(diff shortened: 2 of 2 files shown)


lakefile.toml
```diff
@@ -22,3 +22,8 @@ supportInterpreter = true
 [[lean_exe]]
 name = "test"
 srcDir = "scripts"
+
```
grunweg, line 25: mathlib's lakefile.lean had a comment explaining what shake does. Can you preserve it, and be it as a normal comment, please?

The old comment was:
./-- `lake exe shake` checks files for unnecessary imports. -/

grunweg (commented):
LGTM, with one small comment. (Does this work; have to tested it locally? Does this require a noshake.json file, for instance?)

Vierkantor:
I have tested that shake does sensible things, and indeed it reports quite a few files in Batteries can be shaken! The default target to run `shake` on is still Mathlib, currently working on making Lake tell me the default target instead. But I think we can leave that to a follow-up PR.

lakefile.toml
Vierkantor: Not entirely sure what you mean by "normal comment", since I know only one type of comment in TOML.
```suggestion

# `lake exe shake` checks files for unnecessary imports.
```

grunweg:
Sure, I agree. I'll add that to the PR description.

grunweg (approved):
I prefer for the review suggestion to be applied - otherwise, this LGTM.

fgdorais:
What is the status of this PR?

bryangingechen:
It looks to me like the discussion on Zulip about whether `shake` should be in its own repository is still unresolved.

PR #1205, opened 2025-04-10
feat: Shake autodetects which module(s) to shake

Vierkantor:
There were still a few references to Mathlib in Shake, in particular it had Mathlib hardcoded as the default module to shake. Instead ask Lake what the default targets are.

(Also remove a reference to Mathlib from the help text. We still have Mathlib as an example usage, which I think is innocent enough and can stay.)

Depends on: #1204

### DIFF
--- a/Shake/Main.lean
+++ b/Shake/Main.lean
@@ -3,8 +3,10 @@ Copyright (c) 2023 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
+import Lean.Environment
+import Lean.Parser.Module
 import Lean.Util.FoldConsts
-import Lean
+import Lean.Util.Paths
 
 /-! # `lake exe shake` command
 
--- a/Shake/Main.lean
+++ b/Shake/Main.lean
@@ -3,10 +3,12 @@ Copyright (c) 2023 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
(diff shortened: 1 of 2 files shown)


Shake/Main.lean
```diff
@@ -3,10 +3,12 @@ Copyright (c) 2023 Mario Carneiro. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
+import Lake.Util.Error
 import Lean.Environment
 import Lean.Parser.Module
 import Lean.Util.FoldConsts
 import Lean.Util.Paths
+import Lake.CLI.Main
```
grunweg, line 11: Would you like to alphabetise the new imports? (optional)

Shake/Main.lean
```diff
@@ -451,12 +453,32 @@ def main (args : List String) : IO UInt32 := do
       IO.println "There are out of date oleans. Run `lake build` or `lake exe cache get` first"
       IO.Process.exit 1
 
+  -- Determine default module(s) to run Shake on
```
grunweg, line 456: Minor: do we capitalise shake or not? (I would instinctively lower-case it.) Also optional.

grunweg (commented):
The third commit looks good to me, but I'd like another pair of eyes on the `defaultTargetModules` computation.

grunweg:
My understanding is that the equivalent PR has landed in mathlib, so this PR can be closed. (If shake were to be upstreamed to batteries, #1204 should be updated.) That said, my reading of the zulip discussion was also that moving this to batteries had no consensus, and a separate repository might be better.

PR #1214, opened 2025-04-28
chore(Tactic/Lint/Simp): add local hypotheses to simp context

jcommelin:
This PR adds the local hypotheses to the simp context when checking simp-normal form.
That means we effectively test if `simp [*]` will rewrite the left-hand side to the right-hand side,
instead of just `simp` (or `dsimp` in the case of rfl theorems).


JovanGerb:
This PR was superseded by #1220, so it can now be closed.

PR #1220, opened 2025-05-01
feat: simpNF to check lemmas with conditions

### LINKED ISSUE #365: simpNF linter does not handle conditional simp lemmas correctly
The linter fails here, but I claim it should succeed:
```lean
import Std

@[simp] theorem cast_fin_h {n m : Nat} (h : m = n) (a : Fin (m + 2)) : (cast (by rw [h]) a : Fin (n + 2)) = (a : Nat) := by cases h; rfl

#lint
#check @cast_fin_h /- Left-hand side does not simplify, when using the simp lemma on itself.
This usually means that it will never apply.
 -/

-- but `simp` works just fine
example {n m : Nat} (h : m = n) (a : Fin (m + 2)) : (cast (by rw [h]) a : Fin (n + 2)) = (a : Nat) := by
  simp [h]
```

I think this code is falsely concluding that this is not a conditional simp lemma:
https://github.com/leanprover/std4/blob/c14f6a65b2c08caa38e1ab5db83451460d6cde3e/Std/Tactic/Lint/Simp.lean#L35-L45

The check for "appears on the LHS" should not visit `Prop` subexpressions, as those aren't found by unification.

JovanGerb:
This PR expands the `simpNF` linter to verify that lemmas with conditions are in simp normal form and apply to themselves.

This was first tried at #1214

[#mathlib4 > simpNF lint with hyps](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/simpNF.20lint.20with.20hyps)

closes #365


Batteries/Tactic/Lint/Simp.lean
```diff
@@ -142,12 +143,41 @@ If that's not the case try reordering lemmas or adding @[priority].
...
+                simplify ldecl.type (← Simp.Context.mkDefault)
+            unless ← isSimpEq hType' ldecl.type do
+              hints := hints ++ m!"
+The simp lemma may be invalid because hypothesis {name} simplifies from
+  {ldecl.type}
+to
+  {hType'}
+using
+  {← formatLemmas stats.usedTheorems simpName none}
+Try to change the hypothesis to the simplified term!"
```
digama0:
```suggestion
              hints := hints ++ m!"\
                \nThe simp lemma may be invalid because hypothesis {name} simplifies from\
                \n  {ldecl.type}\
                \nto\
                \n  {hType'}\
                \nusing\
                \n  {← formatLemmas stats.usedTheorems simpName none}\
                \nTry to change the hypothesis to the simplified term!"
```

PR #1232, opened 2025-05-17
feat: some lemmas about `Vector.get`

eric-wieser:


fgdorais:
Looks good but where do you need `get` specifically instead of `getElem`?

fgdorais (approved):

eric-wieser:
It's handy when converting back and forth to Fin vectors.

PR #1234, opened 2025-05-21
feat: add `Vector.tail` API

linesthatinterlace:
There are no lemmas for interacting with `Vector.tail`. This PR adds some basic API for it. (I also think we should define it using a match rather than an if-then-else but that's a core problem).

PR #1236, opened 2025-05-23
fix: Rename `tail` lemmas

linesthatinterlace:
Renames two lemmas added in  #1234 that were misnamed.

### DIFF
--- a/Batteries/Data/Vector/Lemmas.lean
+++ b/Batteries/Data/Vector/Lemmas.lean
@@ -51,8 +51,8 @@ theorem tail_eq_of_ne_zero [NeZero n] {v : Vector α n} :
 @[simp] theorem toList_tail {v : Vector α n} :
     v.tail.toList = v.toList.tail :=
   match n with
-  | 0 => by simp only [tail_eq_of_zero, Vector.eq_empty, List.tail_nil]
-  | _ + 1 => by simp only [tail_eq_of_ne_zero, Vector.toList_eraseIdx, List.eraseIdx_zero]
+  | 0 => by simp [Vector.eq_empty]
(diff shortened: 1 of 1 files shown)


Batteries/Data/Vector/Lemmas.lean
```diff
@@ -43,23 +43,25 @@ theorem toArray_injective : ∀ {v w : Vector α n}, v.toArray = w.toArray → v
...
+theorem tail_eq_of_ne_zero [NeZero n] {v : Vector α n} :
     v.tail = v.eraseIdx 0 n.pos_of_neZero := dif_pos _
 
 @[simp] theorem toList_tail {v : Vector α n} :
     v.tail.toList = v.toList.tail :=
   match n with
-  | 0 => by simp only [tail_zero, Vector.eq_empty, List.tail_nil]
-  | _ + 1 => by simp only [tail_ne_zero, Vector.toList_eraseIdx, List.eraseIdx_zero]
+  | 0 => by simp only [tail_eq_of_zero, Vector.eq_empty, List.tail_nil]
+  | _ + 1 => by simp only [tail_eq_of_ne_zero, Vector.toList_eraseIdx, List.eraseIdx_zero]
```
fgdorais, line 55: Please change these to ordinary simps since they are terminal.

linesthatinterlace: Pushed a change - is that what you mean?

fgdorais: I mean they should be `simp` and not `simp only`.

linesthatinterlace: Right - I did that and then omitted the lemmas which are already in the simp set.

PR #1237, opened 2025-05-24
feat: add wellfounded part of a relation

fgdorais:
I was pretty sure this already existed but I couldn't find it anywhere.


fgdorais:
Not quite, `wrap` assumes that the relation is wellfounded. This instance applies to any relation.

mattrobball:
How is this different from the first declaration in the file?

fgdorais:
Well, I guess I searched every *other* file. lol

PR #1262, opened 2025-06-05
chore(Data/Vector): lemmas about finIdxOf?

eric-wieser:
These match the API for `Array` in https://github.com/leanprover/lean4/pull/7913.

Motivated by [#Is there code for X? > Concrete permutation @ 💬](https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Concrete.20permutation/near/522461859)

### DIFF
--- a/Batteries/Data/Vector/Lemmas.lean
+++ b/Batteries/Data/Vector/Lemmas.lean
@@ -117,12 +117,45 @@ theorem finIdxOf?_eq_some_iff [BEq α] [LawfulBEq α] {v : Vector α n} {a : α}
   obtain ⟨xs, rfl⟩ := v
   simp
 
+-- forward-port of https://github.com/leanprover/lean4/pull/8678
 @[simp]
-theorem isSome_finIdxOf? [BEq α] [LawfulBEq α] {v : Vector α n} {a : α} :
-    (v.finIdxOf? a).isSome ↔ a ∈ v := by
+theorem _root_.List.isSome_finIdxOf?_eq [BEq α] [PartialEquivBEq α] {l : List α} {a : α} :
+    (l.finIdxOf? a).isSome = l.contains a := by
+  induction l with
+  | nil => simp
+  | cons x xs ih =>
+    simp only [List.finIdxOf?_cons]
+    split <;> simp_all [BEq.comm]
+
+-- forward-port of https://github.com/leanprover/lean4/pull/8678
+@[simp]
+theorem _root_.List.isNone_finIdxOf?_eq [BEq α] [PartialEquivBEq α] {l : List α} {a : α} :
+    (l.finIdxOf? a).isNone = !l.contains a := by
+  rw [← List.isSome_finIdxOf?_eq, Option.not_isSome]
+
+
+-- forward-port of https://github.com/leanprover/lean4/pull/8678
+@[simp]
+theorem _root_.Array.isSome_finIdxOf?_eq [BEq α] [PartialEquivBEq α] {xs : Array α} {a : α} :
+    (xs.finIdxOf? a).isSome = xs.contains a := by
+  rcases xs with ⟨xs⟩
+  simp [Array.size]
+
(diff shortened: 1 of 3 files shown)


Batteries/Data/Vector/Lemmas.lean
```diff
@@ -96,3 +98,31 @@ theorem getElem_tail {v : Vector α n} {i : Nat} (hi : i < n - 1) :
...
+
+@[simp]
+theorem finIdxOf?_eq_none_iff [BEq α] [LawfulBEq α] {v : Vector α n} {a : α} :
+    v.finIdxOf? a = none ↔ a ∉ v := by
+  obtain ⟨xs, rfl⟩ := v
+  simp
+
+@[simp]
+theorem finIdxOf?_eq_some_iff [BEq α] [LawfulBEq α] {v : Vector α n} {a : α} {i : Fin n} :
+    v.finIdxOf? a = some i ↔ v.get i = a ∧ ∀ j < i, ¬v.get j = a := by
```
fgdorais, line 116: Maybe add a shortcut lemma to get just `v.get i = a`?

Batteries/Data/Vector/Lemmas.lean
```diff
@@ -96,3 +98,31 @@ theorem getElem_tail {v : Vector α n} {i : Nat} (hi : i < n - 1) :
...
+
+@[simp]
+theorem isSome_finIdxOf? [BEq α] [LawfulBEq α] {v : Vector α n} {a : α} :
+    (v.finIdxOf? a).isSome ↔ a ∈ v := by
+  obtain ⟨v, rfl⟩ := v
+  simp
+
+theorem isNone_finIdxOf? [BEq α] [LawfulBEq α] {v : Vector α n} {a : α} :
+    (v.finIdxOf? a).isNone = ¬ a ∈ v := by
+  simp
```
fgdorais: Should this be `↔` instead of `=`? Doesn't make a real difference but with `=` I would expect `v` to be explicit in order to allow for dot-notation to work. Perhaps we should have both?

fgdorais (commented):
Looks good to me! Just a few thoughts and questions. Once answered, I will move this to `will-merge-soon` for a day or two to allow others to chime in before merging.

Batteries/Data/Vector/Lemmas.lean
eric-wieser: `isNone` is a boolean, so `=` is the stronger lemma.

eric-wieser: The argument implicitness is copied from the `Array` API.

Batteries/Data/Vector/Lemmas.lean
eric-wieser: We don't have this for `Array`, so I figured it wasn't worth adding.

Batteries/Data/Vector/Lemmas.lean
fgdorais: I think `isSome_finIdxOf?` and `isSome_finIdxOf?` should match. All things being equal, I prefer `Iff` since I don't understand why `((v.finIdxOf? a).isNone = true) = ¬a ∈ v` is any stronger than the `Iff` version.

Batteries/Data/Vector/Lemmas.lean
fgdorais: Let's add it. I think it's missing for `Array` but I'll take care of that.

Batteries/Data/Vector/Lemmas.lean
eric-wieser: I've made https://github.com/leanprover/lean4/pull/8678 to fix this inconsistency upstream.

eric-wieser: I've forward-ported this here, which now means the `=` is in `Bool` not `Prop`.

PR #1263, opened 2025-06-05
fix: do not create Decidable subgoals from by_contra

### LINKED ISSUE #1196: by_contra returns decidability goal
In the second example below, `by_contra` generates an unexpected and undesired second goal.
```lean
import Batteries

example : True := by
  by_contra h
  -- 1 goal, as expected

example : True ∧ True := by
  constructor
  · skip
  · by_contra h
    -- first goal is `Decidable True` before the expected goal.
```

eric-wieser:
This works around https://github.com/leanprover/lean4/issues/8643.

Also adjusts a test to actually verify that the proof uses decidability, rather than using proof irrelevence.

Fixes #1196

### DIFF
--- a/Batteries/Tactic/Init.lean
+++ b/Batteries/Tactic/Init.lean
@@ -42,8 +42,8 @@ macro (name := byContra) tk:"by_contra" e?:(ppSpace colGt binderIdent)? : tactic
     | none => Unhygienic.run `(_%$tk)
   `(tactic| first
     | guard_target = Not _; intro $e:term
-    | refine (Decidable.byContradiction fun $e => ?_ :)
-    | refine (Classical.byContradiction fun $e => ?_ :))
+    | refine @Decidable.byContradiction _ _ fun $e => ?_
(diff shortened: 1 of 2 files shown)


fpvandoorn (approved):
Thanks, LGTM!

fgdorais:
[Temporarily removing my maintainer hat and putting on my logician hat]

The class of stable propositions, i.e. propositions `p` that satisfy `¬¬p →p`, is strictly larger than the class of decidable propositions in plain intuitionistic logic. For example, all negated propositions are stable. This is exactly the class of propositions for which `by_contra` makes sense without any assumptions. Why don't we revise `by_contra` to use this class by design, that would avoid special cases for negated formulas and such. 

I've implemented this and other useful logic classes at https://github.com/fgdorais/logic4 (Apologies for the lack of documentation and comments, this repo was always intended as a personal experiment.)

[Restoring my maintainer hat]

This PR looks good to me but I hope that what my logician alter ego said will not go unheard.

fgdorais (approved):

PR #1267, opened 2025-06-07
feat: lemmas about `Nat.toDigits`

marcusrossel:
As discussed in https://github.com/leanprover/lean4/pull/8673.

### DIFF
--- a/Batteries/Data/Nat/Digits.lean
+++ b/Batteries/Data/Nat/Digits.lean
@@ -6,13 +6,13 @@ Authors: Marcus Rossel
 
 namespace Nat
 
-theorem isDigit_digitChar_iff_lt : n.digitChar.isDigit ↔ (n < 10) :=
+@[simp]
+theorem isDigit_digitChar : n.digitChar.isDigit = decide (n < 10) :=
   match n with
-  | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 => ⟨by decide, by simp [*, digitChar]⟩
+  | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 => by simp [digitChar]
   | _ + 10 => by
-    refine ⟨fun h => ?_, fun _ => by contradiction⟩
-    simp only [digitChar, Nat.reduceEqDiff] at h
-    (repeat' split at h) <;> contradiction
+    simp only [digitChar, ↓reduceIte, Nat.reduceEqDiff]
+    (repeat' split) <;> simp
 
 private theorem isDigit_of_mem_toDigitsCore_10
     (hf : n < fuel) (hc : c ∈ cs → c.isDigit) (h : c ∈ toDigitsCore 10 fuel n cs) :
@@ -24,7 +24,7 @@ private theorem isDigit_of_mem_toDigitsCore_10
     case' isFalse => apply ih (by omega) (fun h => ?_) h
     all_goals
       cases h
-      next => have := isDigit_digitChar_iff_lt.mpr (mod_lt n <| by decide); simp_all
+      next => simp [mod_lt]
       next hm => exact hc hm
 
 theorem isDigit_of_mem_toDigits_10 (h : c ∈ toDigits 10 n) : c.isDigit :=
--- a/Batteries/Data/Nat/Digits.lean
+++ b/Batteries/Data/Nat/Digits.lean
@@ -38,6 +38,10 @@ private theorem toDigitsCore_of_lt_base (hb : n < b) (hf : n < fuel) :
 theorem toDigits_of_lt_base (h : n < b) : toDigits b n = [n.digitChar] :=
   toDigitsCore_of_lt_base h (lt_succ_self _)
 
+theorem toDigits_zero : (b : Nat) → toDigits b 0 = ['0']
+  | 0     => rfl
+  | _ + 1 => toDigits_of_lt_base (zero_lt_succ _)
+
 private theorem toDigitsCore_append :
     toDigitsCore b fuel n cs₁ ++ cs₂ = toDigitsCore b fuel n (cs₁ ++ cs₂) := by
   induction fuel generalizing n cs₁ with simp only [toDigitsCore]
--- a/Batteries/Data/Nat/Digits.lean
+++ b/Batteries/Data/Nat/Digits.lean
@@ -6,10 +6,13 @@ Authors: Marcus Rossel
 
 namespace Nat
 
-theorem isDigit_digitChar_of_lt (h : n < 10) : n.digitChar.isDigit := by
+theorem isDigit_digitChar_iff_lt : n.digitChar.isDigit ↔ (n < 10) :=
   match n with
-  | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 => simp [*, digitChar]
-  | _ + 10                                => contradiction
+  | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 => ⟨by decide, by simp [*, digitChar]⟩
+  | _ + 10 => by
+    refine ⟨fun h => ?_, fun _ => by contradiction⟩
+    simp only [digitChar, Nat.reduceEqDiff] at h
+    (repeat' split at h) <;> contradiction
 
 private theorem isDigit_of_mem_toDigitsCore_10
     (hf : n < fuel) (hc : c ∈ cs → c.isDigit) (h : c ∈ toDigitsCore 10 fuel n cs) :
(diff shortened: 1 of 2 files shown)


fgdorais:
Please summarize the discussion rather than just linking to it. That makes it easier for people here to catch up.

Batteries/Data/Nat/Digits.lean
```diff
@@ -0,0 +1,31 @@
+/-
+Copyright (c) 2025 Marcus Rossel. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Marcus Rossel
+-/
+
+namespace Nat
+
+theorem isDigit_digitChar_iff_lt : n.digitChar.isDigit ↔ (n < 10) :=
```
eric-wieser: Arguably this should be 
```suggestion
@[simp]
theorem isDigit_digitChar : n.digitChar.isDigit = n.blt 10 :=
```

Rob23oba: `Nat.blt` though..? I'd rather do `decide (n < 10)` in that case

eric-wieser: I don't mind much about the RHS, I just assumed that blt was simp-normal

Rob23oba: `Nat.blt` is only used in the construction of `DecidableLT Nat` (and in some other places where fast kernel reduction matters)

eric-wieser: I'm surprised that there's not a simp lemma in either direction

marcusrossel:
I've added the lemmas `toDigits_10_of_lt_10` and `toDigits_10_natAbs_ofNat`, and generalized the title of this PR accordingly.

Batteries/Data/Nat/Digits.lean
```diff
@@ -0,0 +1,36 @@
...
+  next => contradiction
+  next fuel ih =>
+    split at h
+    case' isFalse => apply ih (by omega) (fun h => ?_) h
+    all_goals
+      cases h
+      next => simp [mod_lt]
+      next hm => exact hc hm
+
+theorem isDigit_of_mem_toDigits_10 (h : c ∈ toDigits 10 n) : c.isDigit :=
```
eric-wieser: This holds more generally as
```suggestion
theorem isDigit_of_mem_toDigits (k) (hk : k ≤ 10) (h : c ∈ toDigits k n) : c.isDigit :=
```
right?

marcusrossel: This also requires `0 < k`. 

Side note: Base `1` currently returns garbage values which happen to be all `0`s, so it's covered by the proposed theorem. Should we expect base `1` to actually return the correct number of `0`s for a unary representation of the input?

marcusrossel: I just noticed, the implementation for base `1` is in fact incorrect. E.g. `toDigits 1 3` returns `['0', '0', '0', '0']`. It only terminates in the first place, because the fuel counts down.

eric-wieser: If you want that unary representation, you should use `Nat.digits` from Mathlib instead of `Nat.toDigits`.

Batteries/Data/Nat/Digits.lean
```diff
@@ -0,0 +1,34 @@
...
+    c.isDigit := by
+  induction fuel generalizing n cs <;> rw [toDigitsCore] at h
+  next => exact hc h
+  next fuel ih =>
+    split at h
+    case' isFalse => apply ih (fun h => ?_) h
+    all_goals
+      cases h
+      next => simp [Nat.lt_of_lt_of_le (mod_lt _ hb₁) hb₂]
+      next hm => exact hc hm
```
fgdorais: This style is easier to read than using `next =>`:
```suggestion
  induction fuel generalizing n cs with rw [toDigitsCore] at h
  | zero => exact hc h
  | succ fuel ih => 
    split at h
    case' isFalse => apply ih (fun h => ?_) h
    all_goals
      cases h with
      | head => simp [Nat.lt_of_lt_of_le (mod_lt _ hb₁) hb₂]
      | tail _ hm => exact hc hm
```

Batteries/Data/Nat/Digits.lean
marcusrossel: This is being discussed at https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/.60Nat.2EtoDigits.60.20base.201/with/523028786 now.

marcusrossel:
I've added the lemma `toDigits_append_toDigits`.

Rob23oba:
I think there should be some general characterisation lemmas to avoid using `Nat.toDigitsCore`, e.g.:
```lean
theorem toDigits_base_zero (n : Nat) : toDigits 0 n = [n.digitChar]
theorem toDigits_base_one (n : Nat) : toDigits 1 n = List.replicate (n + 1) '0'
theorem toDigits_of_lt_base (h : n < b) : toDigits b n = [n.digitChar]
theorem toDigits_of_base_le (h : 1 < b) (h' : b ≤ n) : toDigits b n = toDigits b (n / b) ++ [(n % b).digitChar]
```
Then some of the others might be able to be proven with these instead of `toDigitCore`.

marcusrossel:
@Rob23oba Is it desirable to have characterisation lemmas for bases `0` and `1`? I feel like this almost encourages relying on these "garbage values".

Batteries/Data/Nat/Digits.lean
```diff
@@ -17,18 +17,61 @@ theorem isDigit_digitChar : n.digitChar.isDigit = decide (n < 10) :=
...
+      have : b ≤ b * n + d := Nat.le_trans (Nat.le_mul_of_pos_right _ hn) (le_add_right _ _)
+      cases Nat.div_eq_zero_iff.mp h <;> omega
+    case isFalse =>
+      have h : (b * n + d) / b = n := by
+        rw [mul_add_div (by omega), Nat.div_eq_zero_iff.mpr (.inr hd), Nat.add_zero]
+      have := (Nat.lt_mul_iff_one_lt_left hn).mpr hb₁
+      simp only [toDigitsCore_of_lt_base hd hdf, mul_add_mod_self_left, mod_eq_of_lt hd, h]
+      apply toDigitsCore_eq_of_lt_fuel hb₁ hnf (by omega)
+
+theorem toDigits_append_toDigits (hb₁ : 1 < b) (hb₂ : b ≤ 10) (hn : 0 < n) (hd : d < b) :
```
eric-wieser: I think this is true with a bound of 16?

Rob23oba: It's true for all bases starting at 2 and that has been fixed already

fgdorais:
I'm on the fence about this PR. Several lemmas belong in `Char` and not `Nat`, but the main issue is that the functions involved are right on the edge with "implementation details". 

I do think it's important to have theorems to prove facts about these. I would recommend a different approach:

1. Add a function that calculates the representation in an arbitrary base `b ≥ 2` of a `Nat` (as a `List (Fin b)` or an `Array (Fin b)` or both, whichever is most convenient).
2. Add theorems that prove the basic arithmetic properties of the base `b` representation for the function from item 1.
3. Prove the correctness of core's implementation by proving that it is a `map` of the function from item 1 for relevant bases.

I don't think string representation is too relevant. For example, Batteries will eventually need to support Base-64 encoding, which is incompatible with traditional representation in bases 2, 8, 10, 16.

eric-wieser:
> 1. Add a function that calculates the representation in an arbitrary base `b ≥ 2` of a `Nat` (as a `List (Fin b)` or an `Array (Fin b)` or both, whichever is most convenient).

Isn't this basically the `Nat.digits` function that's already in mathlib?

fgdorais:
Perhaps. If so, that should make it easier to adapt this PR!

eric-wieser:
I'm not sure I understand the proposal; surely we don't want a third version of this function?

Is the suggestion to upstream `Nat.digits`?  If so, it probably would save a lot of review time to just copy the file exactly from mathlib, and then make a subsequence cleanup if necessary.

fgdorais:
I checked out `Nat.digits` in Mathlib. It is not suitable for Batteries but what is needed would be a more efficient drop-in replacement for `Nat.digitsAux`. I'm happy to do that in the next couple of weeks.

@marcusrossel How urgent is your need for this PR?

marcusrossel:
@fgdorais not urgent at all :)
I'd also be happy to help with the version you have in mind.

fgdorais:
WIP PR #1293 adds the functions I'm thinking about. The `toDigitsBE` function is the main one needed here. However, I haven't finished proving its correctness. `toDigitsLE .. |>.reverse` can be used in the mean time but that function is not as efficient as `toDigitsBE`.

@marcusrossel If you are still inclined to help, you can start reviewing this PR when you have time. The parts leading up to `toDigitsBE` are unlikely to change in non-trivial ways before final version.

marcusrossel:
I'll be rather busy the next 2 weeks, but if it's not too late by then, I'll take a look :)

marcusrossel:
Uh, @fgdorais was this merge intentional?

PR #1268, opened 2025-06-09
fix: use lakefile.toml instead of lakefile.lean in discover workflow

kim-em:
This fixes the discover-lean-pr-testing workflow which was incorrectly filtering branches by looking for changes to `lakefile.lean`, but Batteries uses `lakefile.toml`.

## Problem
The workflow was copied from Mathlib's equivalent workflow, but this key difference wasn't properly adapted. This caused the branch relevance detection to work incorrectly - branches that should have been automatically merged were either being skipped or incorrectly flagged as needing manual review.

## Solution
Changed the file filtering logic to use `lakefile.toml` instead of `lakefile.lean` in the branch diff check.


kim-em:
Full disclosure: this is an AI generated PR (note commit author).

PR #1291, opened 2025-06-26
feat: fix some issues with `equals t => tac`

plp127:
This PR fixed some issues with the `conv` tactic `equals t => tac`:
- It gives bad error messages when `t` has the wrong type. For example,
  ```
  type mismatch
    this
  has type
    ?m.6391 = 1 : Prop
  but is expected to have type
    3 = ?m.6381 : Prop
  ```


fgdorais:
This needs tests.

plp127:
awaiting-review

PR #1293, opened 2025-06-26
feat: add `Nat.Digits`

fgdorais:
This PR adds functions for calculating digit representation in any base. Both little-endian and big-endian versions are provided.

### DIFF
--- a/Batteries/Data/Nat/Digits.lean
+++ b/Batteries/Data/Nat/Digits.lean
@@ -1,0 +1,307 @@
+/-
+Copyright (c) 2025 François G. Dorais. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
+module
+
+public import Batteries.Tactic.Basic
+public import Batteries.Data.Nat.Basic
+
+@[expose] public section
+
+namespace Nat
+
+/-! # Little-Endian (`LE`) Functions -/
+
+/--
+Construct a natural number from a little-endian list of digits in the given base, represented
+as elements of `Fin base`:
+```
+ofDigitsLE (base := 10) [1,2,3] = 321 -- 1*10^0 + 2*10^1 + 3*10^2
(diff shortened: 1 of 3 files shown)


Batteries/Data/Nat/Digits.lean
```diff
@@ -0,0 +1,303 @@
...
+theorem ofDigitsBE_append {l₁ l₂ : List (Fin base)} :
+    ofDigitsBE (l₁ ++ l₂) = ofDigitsBE l₁ * base ^ l₂.length + ofDigitsBE l₂ := by
+  induction l₁ with
+  | nil => simp
+  | cons d₁ l₁ ih =>
+    simp +arith [ih, Nat.add_mul, Nat.pow_add', Nat.mul_assoc]
+
+@[simp]
+theorem ofDigitsBE_reverse_eq_ofDigitsLE {l : List (Fin base)} :
+    ofDigitsBE l.reverse = ofDigitsLE l := by
```
eric-wieser, line 187: I claim that because of this there really is no point having both versions of the function.

fgdorais: There is:

* The LE version has an obvious user-friendly recursion pattern but is highly inefficient. (Stack overflow happens before 7000 digits.)
* The BE version is not user-friendly but it is naturally tail-recursive.
* The BE version is used via `csimp` to make an efficient implementation of the LE version.
* The BE version has an obvious use case that decimal numbers are in BE format.

fgdorais: The names used here don't conflict with Mathlib at all. The expected blend-in is to use `toDigitsLE` (once `csimp` implemented as the reverse of `toDigitsBE`) as a more efficient replacement for Mathlib's `digitsAux`, the whole thing should be invisible to Mathlib users.

linesthatinterlace:
You might want to consider that batteries#799 ought to be compatible with this. I am sure there's a lemma like this in Mathlib somewhere but I can't find it.

PR #1294, opened 2025-06-28
feat: make `unusedHavesSuffices` linter only apply to propositions

kmill:
The `unusedHavesSuffices` linter error talks about proofs, so this PR adds in a check that the `have` is indeed introducing a proposition. This will be necessary in the next release because Lean will convert `let`s to `have`s automatically to cache which ones are nondependent, and, for example, `do` notation can elaborate into expressions that have unused `let`s, which we don't want to be flagged by this linter.

After the next Lean release, this linter should be updated look for nondependent `Expr.letE` expressions rather than `letFun`s.

### DIFF
--- a/Batteries/Tactic/Lint/Misc.lean
+++ b/Batteries/Tactic/Lint/Misc.lean
@@ -225,7 +225,7 @@ with rfl when elaboration results in a different term than the user intended. -/
       return none
 
 /--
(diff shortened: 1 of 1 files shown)


fgdorais:
Aside: This linter already had a lot of false positives with known workarounds. I think it needs a full review.

PR #1307, opened 2025-07-01
feat: add code action to generate `match` cases

MoritzBeroRoos:
A small adaption of the existing code action for holes in this Situation: 
```lean
def foo : Expr → Unit := _
```

gives us a code action that works for 
```lean
def myfun2 (n:Nat) : Nat :=
  match n
```
and produces:
```lean
def myfun2 (n:Nat) : Nat :=
  match n with
  | 0 => _
  | n + 1 => _
```


MoritzBeroRoos:
are the underscores in CodeAction.lean a problem? I used this file to put some examples there

fgdorais:
I'm not sure it's possible but is there a way to add test cases for this?

MoritzBeroRoos:
I don't know, but i believe not, since i don't see any tests for other code actions. Also since code actions are a vscode feature, existence of any testing automation would be unlikely.

MoritzBeroRoos:
i have removed the issues with the long lines. It builds locally for me now.
If the build still fails now, i have no idea what it ism the error message from the github log didn't tell me why.

fgdorais:
You can run `lake test` and `lake lint` locally. Looks like some defs are missing docstrings.

Batteries/CodeAction/Misc.lean
```diff
@@ -267,6 +267,125 @@ def findTermInfo? (node : InfoTree) (stx : Term) : Option TermInfo :=
...
+inspired by InfoTree.findInfo? -/
+partial def findAllInfos (p : Info → Bool) (t : InfoTree) : Array Info :=
+  let rec loop (t : InfoTree) (acc : Array Info) : Array Info :=
+    match t with
+    | .context _ childTree => loop childTree acc
+    | .node info children  =>
+      let acc' := if p info then acc.push info else acc
+      children.foldl (fun currentAcc child => loop child currentAcc) acc'
+    | .hole _              => acc
+  loop t #[]
```
fgdorais: That should fix it:
```suggestion
partial def findAllInfos (p : Info → Bool) (t : InfoTree) : Array Info :=
  loop t #[]
where
  /-- Inner loop for `findAllInfos` -/
  loop (t : InfoTree) (acc : Array Info) : Array Info :=
    match t with
    | .context _ childTree => loop childTree acc
    | .node info children  =>
      let acc' := if p info then acc.push info else acc
      children.foldl (fun currentAcc child => loop child currentAcc) acc'
    | .hole _              => acc
```

MoritzBeroRoos:
Thank you, very good fix!
It now builds, tests, and lints locally for me.

MoritzBeroRoos:
`Is there a way to support multiple discriminants? That would be helpful since it is time consuming to get all the cases for these.`
Would this be the expected ordering for e.g. 2 discriminants?

```lean4
def myfun2 (n:Nat) (m:Nat): Nat :=
  match n,m with
  | .zero, .zero => 0
  | .zero, .succ m' => 1
  | .succ n', .zero => 2
  | .succ n', .succ m' => 3
  ```

**Comment by fgdorais** [comment]
This is excellent!

There's an issue with placeholder names:
```lean
example (l r : List Nat) : Empty :=
  match l, r with
```
gives this:
```lean
example (l r : List Nat) : Empty :=
  match l, r with
  | [], [] => _
  | [], head :: tail => _
  | head :: tail, [] => _
  | head :: tail, head :: tail => _
```
This is not valid because `head` and `tail` are reused for different matches.

I don't think it's a burden on users to edit the pattern names (they probably will want to do that anyway). But there probably is a better automatic disambiguation solution, especially one that would make it easier to use find/replace instead of manual editing.

MoritzBeroRoos:
Should we just add primes `tail'''` until we get a not-already-used name in each line?
So generate this instead:
```
example (l r : List Nat) : Empty :=
  match l, r with
  | [], [] => _
  | [], head :: tail => _
  | head :: tail, [] => _
  | head :: tail, head' :: tail' => _
  ```

**Comment by fgdorais** [comment]
The second version is better for renaming. I think I'd prefer `_1`, `_2`, `_3` over primes.

**Comment by fgdorais** [comment]
The most important thing is that the generated code is valid. This should be the case with the following scheme:

1. If there is only one discriminant then the current scheme is fine.
2. If there are two or more discriminants, add suffix `_i` where `i` is the index of the discriminant (including the first discriminant).

**Comment by fgdorais** [comment]
I'm afraid I must veto primes. 

Esthetics is not a concern here since users are encouraged to edit the names, so an easily predictable and consistent scheme is better.

Primes are unfriendly to some users. Imagine what `n'''''''` sounds like to a programmer who needs to use a screen reader.

**Comment by MoritzBeroRoos** [comment]
Can we at least do some scheme, that doesn't need `_n`s for 2 or more discriminants if there is no actual collision? Because that is probably the most common case.
Maybe instead of n primes `''...''` we put `_n` in the scheme described above?

**Comment by fgdorais** [comment]
Let's keep things simple, consistent and easily predictable.

File: Batteries/CodeAction/Match.lean [hunk 2]
```diff
@@ -0,0 +1,205 @@
...
+    | .context _ childTree => loop childTree acc
+    | .node info children  =>
+      let acc' := if p info then acc.push info else acc
+      children.foldl (fun currentAcc child => loop child currentAcc) acc'
+    | .hole _              => acc
+
+/-- Computes the cartesian product over a list of lists.
+i.e. cartesian_product [[1, 2], [3, 4]] =  [[1, 3], [1, 4], [2, 3], [2, 4]].
+-/
+def cartesian_product {α : Type} : List (List α) → List (List α)
```
fgdorais: This is called `List.sections`.

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
...
+  | .none, n_2 + 1 => _
+  | .some val_1, 0 => _
+  | .some val_1, n_2 + 1 => _
+```
+-/
+@[command_code_action] --I couldn't make this work with '@[command_code_action Parser.Term.match]':
+                       --It never fires. So i filter it myself in Step 1.
+def matchExpand : CommandCodeAction := fun CodeActionParams snap ctx node => do
+  --dbg_trace "--------------------------------------------------------------"
+  --dbg_trace (←node.format ctx)
```
fgdorais: Delete this and similar debugging code later.

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
...
+  | n + 1 => _
+```
+Also has support for multiple discriminants, e.g.
+```
+def myfun3 (o: Option Bool) (m:Nat): Nat :=
+  match o, m with
+```
+can be expanded into
+```
+def myfun3 (o: Option Bool) (m:Nat): Nat :=
```
fgdorais:
```suggestion
def myfun3 (o : Option Bool) (m : Nat) : Nat :=
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
...
+/--
+Invoking tactic code action "Generate a list of equations for this match." in the
+following:
+```lean
+def myfun2 (n:Nat) : Nat :=
+  match n
+```
+produces:
+```lean
+def myfun2 (n:Nat) : Nat :=
```
fgdorais:
```suggestion
def myfun2 (n : Nat) : Nat :=
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
...
+      arguments these might clash, so we fix it by appending a suffix like `_2` -
+      you will probably want to rename these suffixed names yourself. -/
+      str := str ++ if arg.hasNum || arg.isInternal then " _" else s!" {arg}{suffix}"
+    return str
+
+/--
+Invoking tactic code action "Generate a list of equations for this match." in the
+following:
+```lean
+def myfun2 (n:Nat) : Nat :=
```
fgdorais:
```suggestion
def myfun2 (n : Nat) : Nat :=
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
...
+produces:
+```lean
+def myfun2 (n:Nat) : Nat :=
+  match n with
+  | 0 => _
+  | n + 1 => _
+```
+Also has support for multiple discriminants, e.g.
+```
+def myfun3 (o: Option Bool) (m:Nat): Nat :=
```
fgdorais:
```suggestion
def myfun3 (o : Option Bool) (m : Nat) : Nat :=
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
+import Batteries.CodeAction.Misc
+
+namespace Batteries.CodeAction
+
+open Lean Meta Elab Server RequestM CodeAction
+
+
+/-- Filter for the info-nodes to find the match-nodes
+-/
+def isMatchTerm : (info: Info) → Bool
```
fgdorais:
```suggestion
def isMatchTerm : Info → Bool
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,205 @@
...
+def cartesian_product {α : Type} : List (List α) → List (List α)
+| [] => [[]]
+| xs :: xss =>
+  let sub_products := cartesian_product xss
+  (xs.map (fun val => sub_products.map (fun sub => val :: sub))).flatten
+
+/-- From a constructor-name e.g. 'Option.some' construct the corresponding match pattern, e.g.
+'.some val'. We implement special cases for Nat and List to produce 'n + 1' instead of 'Nat.succ n'.
+-/
+def pattern_from_constructor (ctor : Name) (env: Environment) (suffix : String): Option String := do
```
fgdorais:
```suggestion
def pattern_from_constructor (ctor : Name) (env : Environment) (suffix : String) : Option String := do
```

Batteries/CodeAction/Match.lean
MoritzBeroRoos: done

Batteries/CodeAction/Match.lean
MoritzBeroRoos: i didn't know about `List.sections`, but after having taken a look, it generates the wrong order for the resulting List.
In `List.sections`, the first components vary the fastest, but for the constructor patterns we want the last components to vary the fastest.

MoritzBeroRoos: Maybe we should rename `cartesian_product` and put it where List.sections is?

fgdorais: That would be a separate PR. You could also just use `xss.reverse.sections.map reverse`.

MoritzBeroRoos: `You could also just use xss.reverse.sections.map reverse`
Right, done.

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
+import Batteries.CodeAction.Misc
+import Batteries.Data.List
+
+namespace Batteries.CodeAction
+
+open Lean Meta Elab Server RequestM CodeAction
+
+
+/-- Filter for the info-nodes to find the match-nodes
+-/
```
fgdorais:
```suggestion
/-- Filter for the info-nodes to find the match-nodes. -/
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+def isMatchTerm : Info → Bool
+  | .ofTermInfo i => i.stx.isOfKind ``Lean.Parser.Term.match
+  | _ => false
+
+/-- Returns the String.range that encompasses 'match e (with)'.
+-/
```
fgdorais:
```suggestion
/-- Returns the String.range that encompasses `match e (with)`. -/
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+    if let some withStx := (matchStx.getArgs.find? (fun s => s.isAtom && s.getAtomVal == "with"))
+      then return ⟨startPos, ←withStx.getTailPos?⟩
+    else
+      let lastMatchDiscr ← discrs.back?
+      return ⟨startPos, ←lastMatchDiscr.raw.getTailPos?⟩
+  | _ => none
+
+/-- Flattens an Infotree into an array of Info-nodes that fulfill p,
+inspired by InfoTree.findInfo?
+-/
```
fgdorais:
```suggestion
/-- Flattens an Infotree into an array of Info-nodes that fulfill p. -/
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+    match t with
+    | .context _ childTree => loop childTree acc
+    | .node info children  =>
+      let acc' := if p info then acc.push info else acc
+      children.foldl (fun currentAcc child => loop child currentAcc) acc'
+    | .hole _              => acc
+
+/-- From a constructor-name e.g. 'Option.some' construct the corresponding match pattern, e.g.
+'.some val'. We implement special cases for Nat and List to produce 'n + 1' instead of 'Nat.succ n'.
+-/
```
fgdorais: How about two other special cases for `Bool` and `Option` that omits the dot?

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+    for arg in explicit_args do
+      /- This takes the variable names `arg` which were used in the
+      inductive type specification. When using this action with multiple (same-type)
+      arguments these might clash, so we fix it by appending a suffix like `_2` -
+      you will probably want to rename these suffixed names yourself. -/
+      str := str ++ if arg.hasNum || arg.isInternal then " _" else s!" {arg}{suffix}"
+    return str
+
+/--
+Invoking tactic code action "Generate a list of equations for this match." in the
```
fgdorais:
```suggestion
Invoking tactic code action "Generate a list of alternatives for this match." in the
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+def myfun3 (o : Option Bool) (m : Nat) : Nat :=
+  match o, m with
+  | .none, 0 => _
+  | .none, n_2 + 1 => _
+  | .some val_1, 0 => _
+  | .some val_1, n_2 + 1 => _
+```
+-/
+@[command_code_action] --I couldn't make this work with '@[command_code_action Parser.Term.match]':
+                       --It never fires. So i filter it myself in Step 1.
```
fgdorais: Adjust for length:
```suggestion
@[command_code_action] 
/- I couldn't make this work with `@[command_code_action Parser.Term.match]`: it never fires. So I filter it myself in Step 1. -/
```
IIRC that parser has special handling. Ask on Zulip, the answers may be usefully informative.

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+def matchExpand : CommandCodeAction := fun CodeActionParams snap ctx node => do
+  -- 1. Find ALL ofTermInfo Info nodes that are of kind `Term.match`
```
fgdorais:
```suggestion
  /- 1. Find ALL ofTermInfo Info nodes that are of kind `Term.match` -/
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+    | `(matchDiscr| $t: term) => some t
+    | `(matchDiscr| $_:ident : $t: term) => some t
+    | _ => none
+    ) | return #[]
+
+  -- Get a Bool, that tells us if "with" is already typed in:
+  let withPresent :=
+    (matchInfo.stx.getArgs.find? (fun s => s.isAtom && s.getAtomVal == "with")).isSome
+
+  /- Construct a list containing for each discriminant its list of constructornames: -/
```
fgdorais:
```suggestion
  /- Construct a list containing for each discriminant its list of constructor names: -/
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+  let mut constructors : List (List Name) := []
+  for discrTerm in discrTerms do
+    let some (info : TermInfo) := findTermInfo? node discrTerm | return #[]
+    let ty ← info.runMetaM ctx (Lean.Meta.inferType info.expr)
+    let .const name _ := (← info.runMetaM ctx (whnf ty)).getAppFn | return #[]
+    -- Find the inductive constructors of e:
+    let some (.inductInfo val) := snap.env.find? name | return #[]
+    constructors := constructors.concat (val.ctors)
```
fgdorais: `val.ctors :: constructors` would be more efficient and the constructors list would already be reversed for the usage below.

Batteries/CodeAction/Misc.lean
```diff
@@ -258,7 +258,8 @@ def getElimExprNames (elimType : Expr) : MetaM (Array (Name × Array Name)) := d
           altsInfo := altsInfo.push (xDecl.userName, args)
     pure altsInfo
 
-/-- Finds the `TermInfo` for an elaborated term `stx`. -/
+/-- Finds the `TermInfo` for an elaborated term `stx`.
+-/
```
fgdorais:
```suggestion
/-- Finds the `TermInfo` for an elaborated term `stx`. -/
```

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,192 @@
...
+      return ⟨startPos, ←lastMatchDiscr.raw.getTailPos?⟩
+  | _ => none
+
+/-- Flattens an Infotree into an array of Info-nodes that fulfill p,
+inspired by InfoTree.findInfo?
+-/
+partial def findAllInfos (p : Info → Bool) (t : InfoTree) : Array Info :=
+  loop t #[]
+where
+  /-- Inner loop for `findAllInfos` -/
```
fgdorais:
```suggestion
  /-- Inner loop for `findAllInfos`. -/
```

fgdorais (changes requested):
A few style edits and two plausible improvements.

Batteries/CodeAction/Match.lean
MoritzBeroRoos: done

Batteries/CodeAction/Match.lean
MoritzBeroRoos: i asked and Mario answered, that it doesn't fire since `match` is a term and not a command.
We might get a corresponding `term_code_action` from him at some point in the future, but for know we should keep this.

fgdorais: Can you add a link to Mario's answer for future reference.

MoritzBeroRoos: [#mathlib4 > Tactic code actions @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Tactic.20code.20actions/near/529916024)

Batteries/CodeAction/Match.lean
```diff
@@ -0,0 +1,196 @@
...
+  match o, m with
+```
+can be expanded into
+```
+def myfun3 (o : Option Bool) (m : Nat) : Nat :=
+  match o, m with
+  | .none, 0 => _
+  | .none, n_2 + 1 => _
+  | .some val_1, 0 => _
+  | .some val_1, n_2 + 1 => _
```
fgdorais: This example is no longer correct?

fgdorais (changes requested):
Just one more!

Batteries/CodeAction/Match.lean
MoritzBeroRoos: right ofc, since i implemented the special syntax for `Option`.
It's fixed now.

fgdorais (approved):
Thank you very much for your patience and responsiveness through the review process.

Since `match` is such a versatile tool, it's impossible for a code action to handle all possible use cases perfectly. There is definitely room for improvements but this is an excellent starting point. Thank you for this excellent contribution!

Since code actions are difficult to test, I will ping an additional review before merging. (Lots of people are on vacation right now so if that doesn't happen in the next couple of weeks I will probably merge anyway.)

PR #1308, opened 2025-07-01
feat: add `@[simp]` lemma `WellFounded.val_wrap`

plp127:
Adds `WellFounded.val_wrap` so `simp` can unfold it.

The reason for this is so that when you use `termination_by otherWF.wrap n`, when simplifying the decreasing goal you don't get stuck reducing `(wrap _ _).val`.

Here's an example:
```
import Batteries

def rel : Nat → Nat → Prop := sorry

theorem wf : WellFounded rel := sorry

@[simp]
theorem rel_add_one : ∀ n, rel (n + 1) n := sorry

def k : Nat → Nat := fun i => k (i + 1)
termination_by i => wf.wrap i
```
Without `WellFounded.val_wrap`, `decreasing_tactic` will fail to prove termination.


plp127:
The build errors don't seem to be related to the changes?

fgdorais:
What is the use case for this?

fgdorais:
I would say no since it breaks mathlib as a side-effect. Did you investigate that problem?

plp127:
Well I found out why CI is failing

plp127:
awaiting-review

kim-em:
I reset `batteries-pr-testing-1308` to match Mathlib's master, and restores the batteries dependency to point at this PR.

plp127:
I still have no idea why mathlib build is failing

fgdorais:
I finally got around to running the essential checks manually. All checks passed. These tests were not as thorough as our CI tests (when they work!) but I am confident this PR is safe to merge.

PR #1283, opened 2025-06-18
feat: add case insensitive comparison for strings (ASCII only)

ammkrn:
Adds ASCII-only case insensitive comparison to `Char` and `String`.

PR #1312, opened 2025-07-03
feat: add case insensitive comparison for ASCII characters

ammkrn:
This is a separate PR split off from #1283 as requested.


Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
 
 instance : Batteries.LawfulOrd Char := .compareOfLessAndEq
   (fun _ => Nat.lt_irrefl _) Nat.lt_trans Nat.not_lt Char.le_antisymm
+
+namespace Char
+
+theorem isUpper_or_isLower_iff_isAlpha {c : Char} : c.isUpper ∨ c.isLower ↔ c.isAlpha := by
```
fgdorais: The opposite iff direction seems more useful.

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+  simp [Char.isAlpha]
+
+theorem toLower_eq_of_isLower (c : Char) : c.isLower → c.toLower = c := by
```
fgdorais:
```suggestion
theorem toLower_eq_of_isLower {c : Char} : c.isLower → c.toLower = c := by
```

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+  omega
+
+theorem toLower_eq_of_not_isAlpha {c : Char} : ¬c.isAlpha → c.toLower = c := by
+  simp only [isAlpha, isUpper, ge_iff_le, isLower, Bool.or_eq_true, Bool.and_eq_true,
+    decide_eq_true_eq, not_or, not_and, UInt32.not_le, toLower, ite_eq_right_iff, and_imp]
+  intro h0 _ h2 h3
+  have _ : c.val.toNat ≤ 90 ∧ 90 < c.val.toNat:= ⟨h3, h0 h2⟩
+  omega
+
+theorem toLowerIsAlpha_iff_isAlpha {c : Char} : c.isAlpha ↔ c.toLower.isAlpha := by
```
fgdorais:
```suggestion
theorem isAlpha_toLower_iff_isAlpha {c : Char} : c.toLower.isAlpha ↔ c.isAlpha := by
```

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+      exact isUpper_or_isLower_iff_isAlpha.mp (.inr ‹_›)
+  case _ =>
+    by_cases hIsAlpha : c.isAlpha
+    . intro _; assumption
+    . rw [toLower_eq_of_not_isAlpha hIsAlpha]; exact id
+
+/--
+Prop-valued comparison of two `Char`s for *ascii*-case insensitive equality.
+-/
+def eqIgnoreAsciiCase (c₁ c₂ : Char) : Prop := c₁.toLower = c₂.toLower
```
fgdorais: We don't need both the Prop and Bool variants, Bool is better.

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+
+/--
+Bool-valued comparison of two `Char`s for *ascii*-case insensitive equality.
+-/
+def beqIgnoreAsciiCase (c₁ c₂ : Char) : Bool := c₁.toLower == c₂.toLower
+
+theorem eqIgnoreAsciiCase_iff_beqIgnoreAsciiCase (c₁ c₂ : Char) :
+  c₁.eqIgnoreAsciiCase c₂ ↔ (c₁.beqIgnoreAsciiCase c₂ = true) := by
+  simp [eqIgnoreAsciiCase, beqIgnoreAsciiCase]
```
fgdorais: Delete. See comment at `eqIngoreAsciiCase`.

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+
+theorem eqIgnoreAsciiCase.eqv : Equivalence eqIgnoreAsciiCase := {
+  refl _ := rfl
+  trans := fun h1 h2 => by simp only [eqIgnoreAsciiCase]; exact h1 ▸ h2
+  symm := by simp [eqIgnoreAsciiCase]; exact Eq.symm
+}
+
+instance eqIgnoreAsciiCase.isSetoid : Setoid Char:= .mk eqIgnoreAsciiCase eqIgnoreAsciiCase.eqv
```
fgdorais:
```suggestion
def beqIgnoreAsciiCase.setoid : Setoid Char := .mk _ beqIgnoreAsciiCase.eqv
```
Making this an instance basically says that this is the only relevant case for `Setoid Char`. This is a strong claim. It's better to make it a def and let users declare local instances when needed.

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+
+theorem eqIgnoreAsciiCase_iff_beqIgnoreAsciiCase (c₁ c₂ : Char) :
+  c₁.eqIgnoreAsciiCase c₂ ↔ (c₁.beqIgnoreAsciiCase c₂ = true) := by
+  simp [eqIgnoreAsciiCase, beqIgnoreAsciiCase]
+
+theorem eqIgnoreAsciiCase.eqv : Equivalence eqIgnoreAsciiCase := {
+  refl _ := rfl
+  trans := fun h1 h2 => by simp only [eqIgnoreAsciiCase]; exact h1 ▸ h2
+  symm := by simp [eqIgnoreAsciiCase]; exact Eq.symm
+}
```
fgdorais:
```suggestion
theorem beqIgnoreAsciiCase.eqv : Equivalence (beqIgnoreAsciiCase · ·) := {
  refl _ := BEq.rfl
  trans _ _ := by simp_all [beqIgnoreAsciiCase]
  symm := by simp_all [beqIgnoreAsciiCase]}
```
See comment at `eqIgnoreAsciiCase`.

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,67 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+
+instance eqIgnoreAsciiCase.isSetoid : Setoid Char:= .mk eqIgnoreAsciiCase eqIgnoreAsciiCase.eqv
+
+end Char
```
fgdorais: Delete. There is no need to close namespaces at the end of a file.

fgdorais (changes requested):
First pass.

Regarding `ASCII`/`Ascii`: `ASCII` is systematically used in docstrings and error messages, but `Ascii` (or `ascii` at the start) seems to be preferred in identifiers.

fgdorais:
I've had a bit of time to think about names for things.

I think adding `asciiCaseFold` as an abbrev for `toLower`. This is what the [Unicode case folding algorithm](https://www.unicode.org/Public/16.0.0/ucd/CaseFolding.txt) does for ASCII characters. This further clarifies the ASCII limitation and it settles the otherwise arbitrary choice of `toLower` vs `toUpper`.

I think `AsciiCaseInsensitive` is better than `IgnoreAsciiCase` for linguistic reasons. Also, if you have time, it would be nice to have a `cmpAsciiCaseInsensitive` for sorting purposes.

What do you think?

ammkrn:
This is very helpful, thank you. 

I like the idea of using `asciiCaseFold`. My only concern would the visibility/searchability of abbrevs, but I'm not very up to date on the details of doc generation and the more modern search tools, so this may be a non-issue in practice.

`beqAsciiCaseInsensitive` is fine. One other alternative, let me know what you think of `beqModAsciiCase`. This would get rid of the (probably personal) hang-up about potentially misreading "AsciiCaseInsensitive" as "ascii && caseInsensitive".

fgdorais:
Note that the PR author expected to implement the requested changes or explain why not.

fgdorais:
I still haven't figured out what's happening with the Mathlib CI. We made a lot of CI changes recently so I don't think it's a problem on the PR end. Let's set that aside for a bit.

Batteries/Data/Char.lean
```diff
@@ -11,3 +12,73 @@ theorem Char.le_antisymm_iff {x y : Char} : x = y ↔ x ≤ y ∧ y ≤ x :=
...
+namespace Char
+
+/--
+Unicode defines a "case folding" algorithm in which upper-case ASCII characters are
+mapped onto their lower-case counterparts.
+
+This abbrev standardizes on one direction (it answers the question of whether
+one should choose e.g. `toUpper` or `toLower` with respect to case insensitive logic).
+-/
+abbrev asciiCaseFold := Char.toLower
```
fgdorais:
````suggestion
/-- Case folding for ASCII characters only.

Alphabetic ASCII characters are mapped to their lowercase form, all other characters are left unchanged. This agrees with the Unicode case folding algorithm for ASCII characters.

```
#eval caseFoldAsciiOnly 'A' == 'a'
#eval caseFoldAsciiOnly 'a' == 'a'
#eval caseFoldAsciiOnly 'À` == 'À`
#eval caseFoldAsciiOnly `à` == `à`
#eval caseFoldAsciiOnly `$` == `$`
```
-/
abbrev caseFoldAsciiOnly := Char.toLower
````

Comment (fgdorais) [reply Batteries/Data/Char.lean]: I think the easiest naming scheme for this PR is "basic name" + `AsciiOnly`, where "basic name" is the name that would make sense if it was a full Unicode implementation (or if there were only ASCII characters). The docstring should explain what `AsciiOnly` precisely means, with examples, as illustrated above.

Comment (ammkrn) [reply Batteries/Data/Char.lean]: That works for me, thanks. I've made the appropriate changes.

**Comment by fgdorais** [comment]
Excellent! Could you add a BatteriesTest file that contains tests for all the docstring examples like:
```lean
#guard cmpCaseInsensitiveAsciiOnly 'a' 'A' == .eq
```
This will ensure that the documentation is and remains accurate.

Also note that I added #1333 that contains a comprehensive suite of lemmas to reason about ASCII casing functions. I would appreciate if you have time to review that PR.

fgdorais:
Great! Let's merge this into #1283 to see if anything breaks.

Batteries/Data/Char.lean
```diff
@@ -156,3 +157,74 @@ theorem toUpper_eq_of_isLower {c : Char} (h : c.isLower) : c.toUpper = ofNat (c.
...
+```
+#eval beqCaseInsensitiveAsciiOnly 'a' 'A' -- true
+#eval beqCaseInsensitiveAsciiOnly 'a' 'a' -- true
+#eval beqCaseInsensitiveAsciiOnly '$' '$' -- true
+#eval beqCaseInsensitiveAsciiOnly 'a' 'b' -- false
+#eval beqCaseInsensitiveAsciiOnly 'γ' 'Γ' -- false
+#eval beqCaseInsensitiveAsciiOnly 'ä' 'Ä' -- false
+```
+-/
+def beqCaseInsensitiveAsciiOnly (c₁ c₂ : Char) : Bool := c₁.toLower == c₂.toLower
```
fgdorais:
```suggestion
def beqCaseInsensitiveAsciiOnly (c₁ c₂ : Char) : Bool :=
  c₁.caseFoldAsciiOnly == c₂.caseFoldAsciiOnly
```

Batteries/Data/Char.lean
```diff
@@ -156,3 +157,74 @@ theorem toUpper_eq_of_isLower {c : Char} (h : c.isLower) : c.toUpper = ofNat (c.
...
+```
+#eval cmpCaseInsensitiveAsciiOnly 'a' 'A' -- eq
+#eval cmpCaseInsensitiveAsciiOnly 'a' 'a' -- eq
+#eval cmpCaseInsensitiveAsciiOnly '$' '$' -- eq
+#eval cmpCaseInsensitiveAsciiOnly 'a' 'b' -- lt
+#eval cmpCaseInsensitiveAsciiOnly 'γ' 'Γ' -- gt
+#eval cmpCaseInsensitiveAsciiOnly 'ä' 'Ä' -- gt
+```
+-/
+def cmpCaseInsensitiveAsciiOnly (c₁ c₂ : Char) : Ordering := Ord.compare c₁.toLower c₂.toLower
```
fgdorais:
```suggestion
def cmpCaseInsensitiveAsciiOnly (c₁ c₂ : Char) : Ordering := 
  compare c₁.caseFoldAsciiOnly c₂.caseFoldAsciiOnly
```

Batteries/Data/Char.lean
```diff
@@ -156,3 +157,74 @@ theorem toUpper_eq_of_isLower {c : Char} (h : c.isLower) : c.toUpper = ofNat (c.
...
+  intro h0 _ _ h3
+  have _ : 97 ≤ c.val.toNat ∧ c.val.toNat ≤ 90 := ⟨h0, h3⟩
+  omega
+
+theorem toLower_eq_of_not_isAlpha {c : Char} : ¬c.isAlpha → c.toLower = c := by
+  simp only [isAlpha, isUpper, ge_iff_le, isLower, Bool.or_eq_true, Bool.and_eq_true,
+    decide_eq_true_eq, not_or, not_and, UInt32.not_le, toLower, ite_eq_right_iff, and_imp]
+  intro h0 _ h2 h3
+  have _ : c.val.toNat ≤ 90 ∧ 90 < c.val.toNat:= ⟨h3, h0 h2⟩
+  omega
```
fgdorais: Let's temporarily remove these lemmas since #1333 should cover these.

fgdorais (changes requested):

fgdorais:
After the last few edits above, you will need to merge main again since I reorganized `Data.Char` a bit.

Once that's done, merge this branch into the #1283 branch to see if anything breaks. I don't expect major issues so we will be ready to merge after that!

Batteries/Data/Char.lean
ammkrn: I've removed them in order to keep things moving, but I was not able to close those goals with just `simp` or `apply?`, so at least by this (admittedly very crude) metric they may not necessarily be duplicative.

fgdorais: Yes, I agree, we may need to add some back.

fgdorais (approved):

PR #1313, opened 2025-07-03
feat: Laws for `MonadStateOf` instances

dtumad:
This PR defines a `LawfulMonadStateOf` type-class for instances of `MonadStateOf` where the `get` and `set` operations are compatible in the expected ways, and adds instances for basic examples like `StateT` and `EStateM`.

### DIFF
--- a/Batteries/Control/LawfulMonadState.lean
+++ b/Batteries/Control/LawfulMonadState.lean
@@ -56,6 +56,8 @@ attribute [simp] get_bind_const get_bind_get_bind get_bind_set_bind set_bind_get
     (fun _ => x) <$> get (m := m) = pure x := by
   rw [map_eq_pure_bind, get_bind_const]
 
+theorem get_bind_get : (do let _ ← get (m := m); get) = get := get_bind_const get
+
 @[simp] theorem get_bind_set [LawfulMonad m] :
     (do let s ← get (m := m); set s) = return PUnit.unit := by
   simpa only [bind_pure_comp, id_map', get_map_const] using
@@ -96,6 +98,7 @@ theorem getModify_eq [LawfulMonad m] (f : σ → σ) :
     getModify (m := m) f = do let s ← get; set (f s); return s := by
   rw [getModify, modifyGet_eq, bind_map_left]
 
+/-- Version of `modifyGet_eq` that preserves an call to `modify`. -/
 theorem modifyGet_eq' [LawfulMonad m] (f : σ → α × σ) :
     modifyGet (m := m) f = do let s ← get; modify (Prod.snd ∘ f); return (f s).fst := by
   simp [modify_eq, modifyGet_eq]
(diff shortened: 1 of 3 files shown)


Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,225 @@
...
+    (do let s ← get; let s' ← get; mx s s') = (do let s ← get; mx s s)
+  /-- Setting the monad state to its current value has no effect. -/
+  get_bind_set_bind {α} (mx : σ → PUnit → m α) :
+    (do let s ← get; let u ← set s; mx s u) = (do let s ← get; mx s PUnit.unit)
+  /-- Setting and then returning the monad state is the same as returning the set value. -/
+  set_bind_get (s : σ) : (do set (m := m) s; get) = (do set s; return s)
+  /-- Setting the monad twice is the same as just setting to the final state. -/
+  set_bind_set (s s' : σ) : (do set (m := m) s; set s') = set s'
+
+variable {σ : Type _} {m : Type _ → Type _} [Monad m] [MonadStateOf σ m] [LawfulMonadStateOf σ m]
```
fgdorais: `variable` declarations outside a section or namespace are best avoided in Batteries. Can you localize these to one of the namespaces below? (If there are universe issues, they should be explained and documented.)

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,225 @@
...
+
+/-- The namespaced `MonadStateOf.modifyGet` is equal to the `MonadState` provided `modifyGet`. -/
+@[simp] theorem monadStateOf_modifyGet_eq_modifyGet [MonadStateOf σ m]
+    (f : σ → α × σ) : (MonadStateOf.modifyGet f : m α) = modifyGet f := rfl
+
+/-- Class for well behaved state monads, extending the base `MonadState` type.
+Requires that `modifyGet` is equal to the same definition with only `get` and `set`,
+that `get` is idempotent if the result isn't used, and that `get` after `set` returns
+exactly the value that was previously `set`. -/
+class LawfulMonadStateOf (σ : Type _) (m : Type _ → Type _) [Monad m] [MonadStateOf σ m] where
```
fgdorais: Is there a utility for a `LawfulMonadStateOf` that is not a `LawfulMonad` too?

dtumad: Not that I'm aware of, just instinct to minimize assumptions in things. I can add that in, I'm not sure if `extends` might lead to diamond issues with `EStateM` getting a similar instance for lawful error behavior.

fgdorais: These are in `Prop` so there are no diamond issues. There is no serious issue either way, it's just unusual that `LawfulMonad*` doesn't imply `LawfulMonad`. Maybe rename to `LawfulStateOf` to avoid the potential confusion?

fgdorais: Actually, almost all the lemmas need to add `LawfulMonad`, so it makes more sense to have `LawfulMonadStateOf` extend `LawfulMonad`.

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,235 @@
...
+    (do modify (m := m) f; modify g) = (do modify (m := m) g; modify f) := by
+  simp [modify_bind_modify, h]
+
+theorem modify_bind_get (f : σ → σ) :
+    (do modify (m := m) f; get) = (do let s ← get; modify f; return (f s)) := by
+  simp [modify_eq]
+
+end modify
+
+end LawfulMonadStateOf
```
fgdorais: There is no need to end the namespace. Just add a header indicating that the following are basic instances.

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,235 @@
...
+
+namespace StateT
```
fgdorais: Namespaces have semantic meaning that is not used here. A comment would have the same effect,

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,235 @@
...
+  modifyGet_eq f := StateT.ext fun s => by simp
+  get_bind_const mx := StateT.ext fun s => by simp
+  get_bind_get_bind mx := StateT.ext fun s => by simp
+  get_bind_set_bind mx := StateT.ext fun s => by simp
+  set_bind_get s := StateT.ext fun s => by simp
+  set_bind_set s s' := StateT.ext fun s => by simp
+
+end StateT
+
+namespace StateCpsT
```
fgdorais: Same as above.

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,235 @@
...
+  modifyGet_eq _ := rfl
+  get_bind_const _ := rfl
+  get_bind_get_bind _ := rfl
+  get_bind_set_bind _ := rfl
+  set_bind_get _ := rfl
+  set_bind_set _ _ := rfl
+
+end StateCpsT
+
+namespace EStateM
```
fgdorais: Same as above.

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,235 @@
...
+  modifyGet_eq _ := rfl
+  get_bind_const _ := rfl
+  get_bind_get_bind _ := rfl
+  get_bind_set_bind _ := rfl
+  set_bind_get _ := rfl
+  set_bind_set _ _ := rfl
+
+end EStateM
+
+section MonadLift
```
fgdorais: A comment would work just as well as a section here.

Batteries/Control/LawfulMonadState.lean
```diff
@@ -0,0 +1,235 @@
...
+
+@[simp] theorem liftM_modifyGet {m n} [MonadStateOf σ m] [MonadLift m n]
+    (f : σ → α × σ) : (liftM (modifyGet (m := m) f) : n _) = modifyGet f := rfl
+
+@[simp] theorem liftM_getModify {m n} [MonadStateOf σ m] [MonadLift m n]
+    (f : σ → σ) : (liftM (getModify (m := m) f) : n _) = getModify f := rfl
+
+end MonadLift
+
+namespace ReaderT
```
fgdorais: Same as above.

fgdorais (changes requested):
Minor changes.

fgdorais (approved):
Looks great! Thank you for this excellent contribution!

Some of our CI are finicky, I will run them manually tomorrow and then merge.

PR #1331, opened 2025-07-18
feat: add finite and well-founded streams

fgdorais:
Add classes for provably finite streams and connect streams with standard library iterators.


kim-em:
Could you explain how this relates to the core iterators? What is the use case that is not being filled by those?

PR #1332, opened 2025-07-19
feat: coding functions for `Fin`

fgdorais:
This PR adds functions for coding common finite types into `Fin`.

Depends on:
- [x] #1460


fgdorais:
This PR is split from #1007 to avoid Mathlib issues discussed there.

Batteries/Data/Fin/Basic.lean
```diff
@@ -86,6 +86,18 @@ This is the dependent version of `Fin.foldl`. -/
     (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
   dfoldlM (m := Id) n α f init
 
+/-- Sum of a list indexed by `Fin n`. -/
+protected def sum [OfNat α (nat_lit 0)] [Add α] (x : Fin n → α) : α :=
+  foldr n (x · + ·) 0
+
+/-- Product of a list indexed by `Fin n`. -/
+protected def prod [OfNat α (nat_lit 1)] [Mul α] (x : Fin n → α) : α :=
+  foldr n (x · * ·) 1
```
kim-em: `Zero` and `One` are now available in core Lean.

Batteries/Data/Fin/Basic.lean
```diff
@@ -86,6 +86,18 @@ This is the dependent version of `Fin.foldl`. -/
...
+
+/-- Count the number of true values of a decidable predicate on `Fin n`. -/
+protected def count (P : Fin n → Prop) [DecidablePred P] : Nat :=
+  Fin.sum (if P · then 1 else 0)
```
kim-em: Bool rather than Prop?

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,538 @@
...
+  | 0 => rfl
+  | 1 => rfl
+
+@[simp] theorem decodeBool_encodeBool : (x : Bool) → decodeBool (encodeBool x) = x
+  | false => rfl
+  | true => rfl
+
+/-- Encode a character value as a `Fin` type. -/
+def encodeChar (c : Char) : Fin 1112064 :=
+  have : c.toNat < 1114112 :=
```
kim-em: Magic constants again.

Batteries/Data/Fin/Basic.lean
fgdorais: Makes sense.

Batteries/Data/Fin/Basic.lean
fgdorais: Thanks! I hadn't noticed.

Batteries/Data/Fin/Coding.lean
fgdorais: Documented.

fgdorais:
@kim-em Is it fine to merge?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -14,6 +14,44 @@ attribute [norm_cast] val_last
...
+/-! ### sum -/
+
+@[simp] theorem sum_zero [Zero α] [Add α] (x : Fin 0 → α) :
+    Fin.sum x = 0 := by
+  simp [Fin.sum]
+
+theorem sum_succ [Zero α] [Add α] (x : Fin (n + 1) → α) :
+    Fin.sum x = x 0 + Fin.sum (x ∘ Fin.succ) := by
+  simp [Fin.sum, foldr_succ]
+
```
kim-em, line 53: Not needed here, but it would be good to have the lemmas relating this to `List.sum`.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -14,6 +14,44 @@ attribute [norm_cast] val_last
...
+  split <;> simp [Fin.count, Fin.sum_succ, Nat.one_add, Function.comp_def, *]
+
+theorem count_le (p : Fin n → Bool) : Fin.count p ≤ n := by
+  induction n with
+  | zero => simp
+  | succ n ih =>
+    rw [count_succ]
+    split
+    · simp [ih]
+    · apply Nat.le_trans _ (Nat.le_succ n); simp [ih]
```
kim-em:
```suggestion
@[simp, grind =] theorem count_zero (p : Fin 0 → Bool) : Fin.count p = 0 := by
  simp [Fin.count]

theorem count_succ (p : Fin (n + 1) → Bool) : Fin.count p =
    if p 0 then Fin.count (fun i => p i.succ) + 1 else Fin.count (fun i => p i.succ) := by
  split <;> simp [Fin.count, Fin.sum_succ, Nat.one_add, Function.comp_def, *]

@[grind]
theorem count_le (p : Fin n → Bool) : Fin.count p ≤ n := by
  induction n with grind [count_succ]
```

kim-em: Presumably `count_succ` can be proved nicely using `grind` as well, but it revealed a bug: leanprover/lean4#10292.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+Copyright (c) 2024 François G. Dorais. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: François G. Dorais
+-/
+import Batteries.Data.Char
+import Batteries.Data.Fin.Lemmas
+import Batteries.Tactic.Basic
+import Batteries.Tactic.Trans
+import Batteries.Tactic.Lint
+
```
kim-em, line 15: Module doc, please, explaining what this is for.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+  | 1 => rfl
+
+@[simp] theorem decodeBool_encodeBool : (x : Bool) → decodeBool (encodeBool x) = x
+  | false => rfl
+  | true => rfl
+
+/-- Encode a character value as a `Fin` type.
+
+Valid character code points range from 0 to 1114112 (U+10FFFF) excluding the surrogate range from
+55296 (U+D800) to 57343 (U+DFFF). This results in 1112064 valid code points.
```
kim-em: Please give some reference for these magic constants.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+-/
+def encodeChar (c : Char) : Fin 1112064 :=
+  have : c.toNat < 1114112 :=
+    match c.valid with
+    | .inl h => Nat.lt_trans h (by decide)
+    | .inr h => h.right
+  if _ : c.toNat < 55296 then
+    ⟨c.toNat, by omega⟩
```
kim-em: Eventually we will want to deprecate `omega` except for internal use. All but one of the `omega`s here are currently replaceable by `grind`. No need to do this in this PR, however.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+      | .inl h => contradiction
+      | .inr h =>
+        constructor
+        · exact Nat.add_one_le_of_lt h.left
+        · exact h.right
+    have h1 : ¬ x.toNat - 2048 < 55296 := by omega
+    have h2 : 2048 ≤ x.toNat := by omega
+    simp only [dif_neg h1, Char.ofNatAux, Nat.sub_add_cancel h2]; rfl
+
+/-- Decode an optional `Fin` types. -/
```
kim-em:
```suggestion
/-- Decode an optional `Fin` type. -/
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+def encodeOption : Option (Fin n) → Fin (n+1)
+  | none => 0
+  | some ⟨i, h⟩ => ⟨i+1, Nat.succ_lt_succ h⟩
+
+/-- Decode an optional `Fin` types. -/
```
kim-em:
```suggestion
/-- Decode an optional `Fin` type. -/
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+  match n, f, x with
+  | 0, _, ⟨_, h⟩ => False.elim (by simp at h)
+  | n+1, f, ⟨x, hx⟩ =>
+    if hx0 : x < f 0 then
+      ⟨0, ⟨x, hx0⟩⟩
+    else
+      have hxf : x - f 0 < Fin.sum (f ∘ succ) := by
+        apply Nat.sub_lt_left_of_lt_add
+        · exact Nat.le_of_not_gt hx0
+        · rw [← sum_succ]; exact hx
```
kim-em:
```suggestion
      have hxf : x - f 0 < Fin.sum (f ∘ succ) := by grind [sum_succ]   
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+  | zero => absurd x.is_lt; simp
+  | succ n ih =>
+    simp only [decodeSigma]
+    split
+    · rfl
+    · next h1 =>
+      have h2 : x - f 0 < Fin.sum (f ∘ succ) := by
+        apply Nat.sub_lt_left_of_lt_add
+        · exact Nat.le_of_not_gt h1
+        · rw [← sum_succ]; exact x.is_lt
```
kim-em:
```suggestion
      have h2 : x - f 0 < Fin.sum (f ∘ succ) := by grind [sum_succ]
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+  | succ n ih =>
+    simp only [decodeSigma]
+    match x with
+    | ⟨⟨0, _⟩, x, hx⟩ =>
+      split
+      · rfl
+      · next h =>
+        absurd h
+        simp only [encodeSigma]
+        exact hx
```
kim-em:
```suggestion
      · grind [encodeSigma]
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+
+/-- Decode a function between `Fin` types. -/
+@[pp_nodot] def decodeFun : {m : Nat} → Fin (n ^ m) → Fin m → Fin n
+  | 0, _ => (nomatch .)
+  | m+1, ⟨k, hk⟩ =>
+    have hn : n > 0 := by
+      apply Nat.zero_lt_of_ne_zero
+      intro h
+      rw [h, Nat.pow_succ, Nat.mul_zero] at hk
+      contradiction
```
kim-em:
```suggestion
    have hn : n > 0 := by grind
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+
+/-- Encode a dependent product of `Fin` types. -/
+def encodePi (f : Fin n → Nat) (x : (i : Fin n) → Fin (f i)) : Fin (Fin.prod f) :=
+  match n, f, x with
+  | 0, _, _ => ⟨0, by simp [Fin.prod]⟩
+  | _+1, f, x =>
+    match encodePi ((f ∘ succ)) (fun ⟨i, hi⟩ => x ⟨i+1, Nat.succ_lt_succ hi⟩) with
+    | ⟨k, hk⟩ => Fin.mk (f 0 * k + (x 0).val) $ calc
+      _ < f 0 * k + f 0 := Nat.add_lt_add_left (x 0).isLt ..
+      _ = f 0 * (k + 1) := Nat.mul_succ ..
```
kim-em:
```suggestion
      _ < f 0 * k + f 0 := by grind
      _ = f 0 * (k + 1) := by grind
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+/-- Decode a dependent product of `Fin` types. -/
+def decodePi (f : Fin n → Nat) (x : Fin (Fin.prod f)) : (i : Fin n) → Fin (f i) :=
+  match n, f, x with
+  | 0, _, _ => (nomatch ·)
+  | n+1, f, ⟨x, hx⟩ =>
+    have h : f 0 > 0 := by
+      apply Nat.zero_lt_of_ne_zero
+      intro h
+      rw [prod_succ, h, Nat.zero_mul] at hx
+      contradiction
```
kim-em:
```suggestion
    have h : f 0 > 0 := by grind [prod_succ]
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,546 @@
...
+@[nolint unusedArguments]
+def encodePUnit : PUnit → Fin 1
+  | .unit => 0
+
+/-- Decode a unit value as a `Fin` type. -/
+@[pp_nodot] def decodePUnit : Fin 1 → PUnit
+  | 0 => .unit
+
+/-- Encode a unit value as a `Fin` type. -/
+abbrev encodeUnit : Unit → Fin 1 := encodePUnit
```
eric-wieser: Given the pattern in this file, should there just be a `Fin.Encodable` typeclass which groups together `encode`, `decide`, and the inverse laws for each type? Ultimately this ends up reimplementing `FinEnum` from Mathlib, but pretty much every instance there is computationally lousy, so a replacement probably is in order.

fgdorais: This is #1007

eric-wieser: I guess my question is whether we really want to expose this API twice, once through the typeclass API and once through a different function for each type. Which spelling is simp-normal?

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,539 @@
...
+namespace Fin
+
+/-- Encode a unit value as a `Fin` type. -/
+@[nolint unusedArguments]
+def encodePUnit : PUnit → Fin 1
+  | .unit => 0
+
+/-- Decode a unit value as a `Fin` type. -/
+@[pp_nodot] def decodePUnit : Fin 1 → PUnit
+  | 0 => .unit
```
eric-wieser: These should be `simp` I think?

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,539 @@
...
+
+/-- Encode a boolean value as a `Fin` type. -/
+def encodeBool : Bool → Fin 2
+  | false => 0
+  | true => 1
+
+/-- Decode a boolean value as a `Fin` type. -/
+@[pp_nodot] def decodeBool : Fin 2 → Bool
+  | 0 => false
+  | 1 => true
```
eric-wieser, line 68: `@[simp]` here too?

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,539 @@
...
+  | true => rfl
+
+/-- Encode a character value as a `Fin` type.
+
+Valid character code points range from 0 to 1114112 (U+10FFFF) excluding the surrogate range from
+55296 (U+D800) to 57343 (U+DFFF). This results in 1112064 valid code points.
+(See [unicode scalar value](https://www.unicode.org/glossary/#unicode_scalar_value).)
+-/
+def encodeChar (c : Char) : Fin 1112064 :=
+  have : c.toNat < 1114112 :=
```
eric-wieser: This looks like it would be a good lemma in the `Char` namespace.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,539 @@
...
+    match c.valid with
+    | .inl h => Nat.lt_trans h (by decide)
+    | .inr h => h.right
+  if _ : c.toNat < 55296 then
+    ⟨c.toNat, by grind⟩
+  else
+    ⟨c.toNat - 2048, by grind⟩
```
eric-wieser: Perhaps for clarity
```suggestion
    ⟨c.toNat - (0xE000 - 0xD800), by grind⟩
```

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,539 @@
...
+Valid character code points range from 0 to 1114112 (U+10FFFF) excluding the surrogate range from
+55296 (U+D800) to 57343 (U+DFFF). This results in 1112064 valid code points.
+(See [unicode scalar value](https://www.unicode.org/glossary/#unicode_scalar_value).)
+-/
+def encodeChar (c : Char) : Fin 1112064 :=
+  have : c.toNat < 1114112 :=
+    match c.valid with
+    | .inl h => Nat.lt_trans h (by decide)
+    | .inr h => h.right
+  if _ : c.toNat < 55296 then
```
eric-wieser:
```suggestion
  if _ : c.toNat < 0xD800 then
```
since this range is almost always represented in hex.

Batteries/Data/Fin/Coding.lean
fgdorais: I see. This kind of concerns is ultimately what led to splitting this from #1007. This PR is intended as a low-level API, to save people from reinventing the wheel. The current situation is that #1007 is shelved for the forseeable future and that the functions here will be used to reimplement the internal part of `FinEnum` in Mathlib. I would hope that this implementation does not expose these internals and that Mathlib users would be expected to avoid the functions here and use `FinEnum` instead.

Batteries/Data/Fin/Coding.lean
kim-em: Maybe even with a `def` for the magic constant.

Batteries/Data/Fin/Coding.lean
kim-em: This sounds reasonable. Can it go in the module doc?

fgdorais: Done.

Batteries/Data/Fin/Coding.lean
fgdorais: Done. I also fixed the constants for `Char.all`/`Char.any` at the same time.

fgdorais:
Are you saying that this is enough to make `(en|de)codePUnit`, `(en|de)codeBool`, `(en|de)codeOrdering` simp?

Would you like me to add monotonicity as a feature? It didn't occur to me that would be desirable but I'm happy to add it.

Batteries/Data/Fin/Basic.lean
```diff
@@ -83,6 +83,18 @@ This is the dependent version of `Fin.foldl`. -/
     (f : ∀ (i : Fin n), α i.castSucc → α i.succ) (init : α 0) : α (last n) :=
   dfoldlM (m := Id) n α f init
 
+/-- Sum of a list indexed by `Fin n`. -/
```
wrenna-robson: I would not describe this as a list indexed by Fin n.

Batteries/Data/Fin/Basic.lean
```diff
@@ -83,6 +83,18 @@ This is the dependent version of `Fin.foldl`. -/
...
+protected def sum [Zero α] [Add α] (x : Fin n → α) : α :=
+  foldr n (x · + ·) 0
+
+/-- Product of a list indexed by `Fin n`. -/
+protected def prod [One α] [Mul α] (x : Fin n → α) : α :=
+  foldr n (x · * ·) 1
+
+/-- Count the number of true values of a decidable predicate on `Fin n`. -/
```
wrenna-robson, line 99: I note that actually this has to be < n (obviously) and I wonder if that is better captured by making this a Fin n. You will never end up wrapping!

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,548 @@
...
+/-- Encode `Fin m × Fin n` into `Fin (m * n)`. -/
+def encodeProd : Fin m × Fin n → Fin (m * n)
+  | (⟨i, hi⟩, ⟨j, hj⟩) => Fin.mk (n * i + j) <| calc
+    _ < n * i + n := Nat.add_lt_add_left hj ..
+    _ = n * (i + 1) := Nat.mul_succ ..
+    _ ≤ n * m := Nat.mul_le_mul_left n (Nat.succ_le_of_lt hi)
+    _ = m * n := Nat.mul_comm ..
+
+/-- Decode `Fin m × Fin n` from `Fin (m * n)`. -/
+@[pp_nodot] def decodeProd (z : Fin (m * n)) : Fin m × Fin n :=
```
wrenna-robson, line 184: This duplicates `finProdFinEquiv` in Mathlib. That is fine, but I think at the very least one should consider porting over `divNat` and `modNat` from Mathlib for this purpose. I worry this is too "raw" a definition.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,548 @@
...
+    have hn : 0 < n := Nat.zero_lt_of_lt (x 0).is_lt
+    split
+    · ext; simp [Nat.mod_eq_of_lt]
+    · next i hi =>
+      have : decodeFun (encodeFun fun k => x k.succ) ⟨i, Nat.lt_of_succ_lt_succ hi⟩
+          = x ⟨i+1, hi⟩ := by rw [ih]; rfl
+      simp [← this, Nat.mul_add_div hn, Nat.div_eq_of_lt]
+
+/-- Encode `(i : Fin n) → Fin (f i)` into `Fin (Fin.prod f)`. -/
+def encodePi (f : Fin n → Nat) (x : (i : Fin n) → Fin (f i)) : Fin (Fin.prod f) :=
```
wrenna-robson, line 295: See [finPiFinEquiv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/BigOperators/Fin.html#finPiFinEquiv).

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,548 @@
...
+      simp [Sigma.ext_iff, Fin.ext_iff] at h ⊢
+      constructor
+      · conv => rhs; rw [← h.1]
+        apply congrArg
+        apply congrArg Sigma.fst
+        congr
+      · apply HEq.trans _ h.2; congr
+
+/-- Encode `Fin m → Fin n` into `Fin (n ^ m)`. -/
+def encodeFun : {m : Nat} → (Fin m → Fin n) → Fin (n ^ m)
```
wrenna-robson, line 257: See [finFunctionFinEquiv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/BigOperators/Fin.html#finFunctionFinEquiv).

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,548 @@
...
+@[simp] theorem encodeProd_decodeProd (x : Fin (m * n)) : encodeProd (decodeProd x) = x := by
+  simp [encodeProd, decodeProd, decodeProd.left, decodeProd.right, Nat.div_add_mod]
+
+@[simp] theorem decodeProd_encodeProd (x : Fin m × Fin n) : decodeProd (encodeProd x) = x := by
+  match x with
+  | ⟨⟨_, _⟩, ⟨_, h⟩⟩ => simp [encodeProd, decodeProd, decodeProd.left, decodeProd.right,
+    Nat.mul_add_div (Nat.zero_lt_of_lt h), Nat.div_eq_of_lt h, Nat.mod_eq_of_lt h]
+
+/-- Encode `(i : Fin n) × Fin (f i)` into `Fin (Fin.sum f)`. -/
+def encodeSigma (f : Fin n → Nat) (x : (i : Fin n) × Fin (f i)) : Fin (Fin.sum f) :=
```
wrenna-robson, line 194: See [finSigmaFinEquiv](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/BigOperators/Fin.html#finSigmaFinEquiv).

Batteries/Data/Fin/Basic.lean
fgdorais: You mean ≤ n and Fin (n+1)?

Batteries/Data/Fin/Coding.lean
fgdorais: This is the idea. These are low-level functions intended to replace the internal parts of `FinEquiv` in Mathlib. It's okay for these to be raw since they are not meant to be exposed to the general public. I do like `divNat` and `modNat` though.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,532 @@
...
+    · rfl
+    · cases he
+  · next he =>
+    split at he
+    · cases he
+    · cases he; rfl
+
+/-- Encode `Fin n ⊕ Fin m` into `Fin (n + m)`. -/
+def encodeSum : Sum (Fin n) (Fin m) → Fin (n + m)
+  | .inl x => x.castLE (Nat.le_add_right _ _)
```
wrenna-robson: Use [Fin.castAdd](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Fin/Basic.html#Fin.castAdd) instead.

Batteries/Data/Fin/Coding.lean
```diff
@@ -0,0 +1,532 @@
...
+  | .inr x => x.natAdd n
+
+/-- Decode `Fin n ⊕ Fin m` from `Fin (n + m)`. -/
+@[pp_nodot] def decodeSum (x : Fin (n + m)) : Sum (Fin n) (Fin m) :=
+  if h : x < n then
```
wrenna-robson: Use [Fin.addCases](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Fin/Lemmas.html#Fin.addCases) to define this.

PR #1333, opened 2025-07-19
feat: add ASCII character casing lemmas

fgdorais:
This PR adds lemmas relating `Char.isLower`, `Char.isUpper`, `Char.toLower`, `Char.toUpper`.

### DIFF
--- a/Batteries/Data/Char.lean
+++ b/Batteries/Data/Char.lean
@@ -1,13 +1,152 @@
 /-
 Copyright (c) 2022 Jannis Limperg. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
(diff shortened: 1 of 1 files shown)


Rob23oba:
Hmm I'm not sure how easy it would be to prove these if we actually implemented `Char.toLower` and `Char.toUpper` correctly (as in converting unicode letters).

fgdorais:
The only way to prove such properties for Unicode simple case mapping is by exhaustion. Many of these are not true for Unicode. There are also locale complications and that some code points case map to multiple characters: https://www.unicode.org/Public/16.0.0/ucd/SpecialCasing.txt

I edited the title to mention ASCII only.

PR #1337, opened 2025-07-22
feat: lemmas about running `EStateM` computations

dtumad:
This PR fills in lemmas about `EStateM.run` and `EStateM.run'` applied to various standard monad operations, as well as those for error and state monads.

### DIFF
--- a/Batteries/Lean/EStateM.lean
+++ b/Batteries/Lean/EStateM.lean
@@ -6,6 +6,10 @@ Authors: Kim Morrison
 
 namespace EStateM
 
+variable {ε σ α β : Type _} (s : σ)
+
+open Backtrackable
+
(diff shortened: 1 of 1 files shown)


dtumad:
I'm not sure if using match statements in the lemma statements like this is the right thing to be doing, versus using things like `EStateM.Result.rec` directly.

Batteries/Lean/EStateM.lean
```diff
@@ -6,6 +6,10 @@ Authors: Kim Morrison
 
 namespace EStateM
 
+variable {ε σ α β : Type _} (s : σ)
```
fgdorais: This is not necessary since auto-implicits are encouraged in Batteries.

PR #1354, opened 2025-07-29
feat: add `open_problem` syntax variant of `proof_wanted`

edegeltje:
As discussed [here](https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Documenting.20open.20problems) on zulip, it can be useful to be able to distinguish syntactically between problems to which no solution is known vs to which a solution is known, but hasn't been formalized yet. This new syntax (which is semantically the same as the `proof_wanted` syntax) allows to make this distinction.

### DIFF
--- a/Batteries/Util/ProofWanted.lean
+++ b/Batteries/Util/ProofWanted.lean
@@ -11,10 +11,15 @@ open Lean Parser Elab Command
 
 /-- This proof would be a welcome contribution to the library!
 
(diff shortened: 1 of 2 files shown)


edegeltje:
awaiting-review

fgdorais:
I'm sorry, this is not a very compelling argument. A more elaborate PR that allows users to easily make their own `proof_wanted` customizations would be welcome.

PR #1369, opened 2025-08-07
chore: propose deprecation of List.fillNones

kim-em:
Has this ever been used? Is it common in other languages' List APIs? It's just cost me half an hour and I'd love to avoid having it cost me more!

Anyone wanting to keep this function could please assist by reverting 0de31883 on the nightly-testing branch, and then removing the reliance on the now private `filterMapTR.go`. (I'm alternatively happy to review a PR refactoring the relevant code in Lean to make that non-private, by making it a top-level declaration. But either this PR or that needs to happen before the next release.)

### DIFF
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -829,7 +829,8 @@ dropped from `xs`.
 fillNones [none, some 1, none, none] [2, 3] = [2, 1, 3]
 ```
 -/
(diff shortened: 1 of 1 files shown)


kim-em:
> I don't know of any use and I don't mind deprecating, but replacing `filterMapTR.go id as acc` with `acc ++ filterMap id as` should work fine.

This mixes up List and Array.

PR #1387, opened 2025-08-24
chore: fixed name `is_lt_iff_lt` to `isLT_iff_lt`

qawbecrdtey:
This is a single fix in naming `LawfulLTCmp.eq_lt_iff_lt`.


fgdorais:
Thanks for the PR but `eq_lt` is correct according to naming conventions.

PR #1395, opened 2025-08-30
feat: better Newton starting point in `Nat.sqrt`

kim-em:
This enables us to run `Nat.sqrt` on large (e.g. million digit) numbers.


kim-em:
> Shouldn't we be using GMP for this?

The GMP story with Lean is ... complicated. We shouldn't assume that it is always available, unfortunately.

PR #1052, opened 2024-11-17
refactor: define `List.IsChain`, deprecate `Chain` and `Chain'`

urkud:

PR #1421, opened 2025-09-19
refactor: define List.IsChain, deprecate Chain and Chain'

linesthatinterlace:
Will require mathlib update. Deprecates #1052 which now appears to be inactive.

### DIFF
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -555,29 +555,54 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
 def pwFilter (R : α → α → Prop) [DecidableRel R] (l : List α) : List α :=
   l.foldr (fun x IH => if ∀ y ∈ IH, R x y then x :: IH else IH) []
 
-section Chain
+section IsChain
 
-variable (R : α → α → Prop)
+/-- `IsChain R l` means that `R` holds between adjacent elements of `l`.
+```
+IsChain R [a, b, c, d] ↔ R a b ∧ R b c ∧ R c d
+``` -/
+inductive IsChain (R : α → α → Prop) : List α → Prop where
+  /-- A list of length 0 is a chain. -/
+  | nil : IsChain R []
+  /-- A list of length 1 is a chain. -/
+  | singleton (a : α) : IsChain R [a]
+  /-- If `a` relates to `b` and `b::l` is a chain, then `a :: b :: l` is also a chain. -/
+  | cons₂ (hr : R a b) (h : IsChain R (b :: l)) : IsChain R (a :: b :: l)
+
+
+variable {R : α → α → Prop}
+
+attribute [simp, grind ←] IsChain.nil
+attribute [simp, grind ←] IsChain.singleton
+
+@[simp, grind =] theorem isChain_cons₂ : IsChain R (a :: b :: l) ↔ R a b ∧ IsChain R (b :: l) :=
+  ⟨fun | .cons₂ hr h => ⟨hr, h⟩, fun ⟨hr, h⟩ => .cons₂ hr h⟩
+
+instance instDecidableIsChain [h : DecidableRel R] (l : List α) : Decidable (l.IsChain R) :=
+  match l with
+  | [] => isTrue .nil
+  | a :: l => go a l
+  where
+    go (a : α) (l : List α) : Decidable ((a :: l).IsChain R) :=
+      match l with
+      | [] => isTrue <| .singleton a
+      | b :: l => haveI := (go b l); decidable_of_iff'  _ isChain_cons₂
 
 /-- `Chain R a l` means that `R` holds between adjacent elements of `a::l`.
 ```
 Chain R a [b, c, d] ↔ R a b ∧ R b c ∧ R c d
 ``` -/
-inductive Chain : α → List α → Prop
-  /-- A chain of length 1 is trivially a chain. -/
-  | nil {a : α} : Chain a []
-  /-- If `a` relates to `b` and `b::l` is a chain, then `a :: b :: l` is also a chain. -/
-  | cons : ∀ {a b : α} {l : List α}, R a b → Chain b l → Chain a (b :: l)
+@[deprecated IsChain (since := "2025-09-19")]
+def Chain : (α → α → Prop) → α → List α → Prop := (IsChain · <| · :: ·)
 
 /-- `Chain' R l` means that `R` holds between adjacent elements of `l`.
 ```
 Chain' R [a, b, c, d] ↔ R a b ∧ R b c ∧ R c d
 ``` -/
-def Chain' : List α → Prop
-  | [] => True
-  | a :: l => Chain R a l
+@[deprecated IsChain (since := "2025-09-19")]
+def Chain' : (α → α → Prop) → List α → Prop := (IsChain · ·)
 
-end Chain
+end IsChain
 
 /-- `eraseDup l` removes duplicates from `l` (taking only the first occurrence).
 Defined as `pwFilter (≠)`.
--- a/Batteries/Data/List/Lemmas.lean
+++ b/Batteries/Data/List/Lemmas.lean
@@ -330,43 +330,91 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
       refine ⟨fun h => h₀ _ (mem_of_mem_drop h) rfl, ?_⟩
       exact disjoint_take_drop h₁ (Nat.le_of_succ_le_succ h)
 
-/-! ### Chain -/
+/-! ### Pairwise -/
 
-attribute [simp, grind ←] Chain.nil
+attribute [simp, grind ←] Pairwise.nil
 
-@[simp, grind =]
-theorem chain_cons {a b : α} {l : List α} : Chain R a (b :: l) ↔ R a b ∧ Chain R b l :=
-  ⟨fun p => by cases p with | cons n p => exact ⟨n, p⟩,
-   fun ⟨n, p⟩ => p.cons n⟩
+protected theorem Pairwise.isChain (p : Pairwise R l) : IsChain R l := by
+  induction p with
+  | nil => grind
+  | cons _ l => cases l with grind
 
-theorem rel_of_chain_cons {a b : α} {l : List α} (p : Chain R a (b :: l)) : R a b :=
-  (chain_cons.1 p).1
+/-! ### IsChain -/
 
-theorem chain_of_chain_cons {a b : α} {l : List α} (p : Chain R a (b :: l)) : Chain R b l :=
-  (chain_cons.1 p).2
+attribute [simp, grind ←] IsChain.nil
+attribute [simp, grind ←] IsChain.singleton
 
-theorem Chain.imp' {R S : α → α → Prop} (HRS : ∀ ⦃a b⦄, R a b → S a b) {a b : α}
-    (Hab : ∀ ⦃c⦄, R a c → S b c) {l : List α} (p : Chain R a l) : Chain S b l := by
-  induction p generalizing b with grind
+@[deprecated (since := "2025-09-19")]
+alias chain_cons := isChain_cons₂
 
-theorem Chain.imp {R S : α → α → Prop} (H : ∀ a b, R a b → S a b) {a : α} {l : List α}
-    (p : Chain R a l) : Chain S a l :=
-  p.imp' H (H a)
+theorem rel_of_isChain_cons₂ {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) : R a b :=
+  (isChain_cons₂.1 p).1
 
-protected theorem Pairwise.chain (p : Pairwise R (a :: l)) : Chain R a l := by
-  let ⟨r, p'⟩ := pairwise_cons.1 p; clear p
-  induction p' generalizing a with grind
+@[deprecated (since := "2025-09-19")]
+alias rel_of_chain_cons := rel_of_isChain_cons₂
+
+theorem isChain_of_isChain_cons₂ {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) :
+    IsChain R (b :: l) := (isChain_cons₂.1 p).2
+
+@[deprecated (since := "2025-09-19")]
+alias chain_of_chain_cons := isChain_of_isChain_cons₂
+
+@[grind]
+theorem IsChain.imp {S : α → α → Prop} (H : ∀ ⦃a b⦄, R a b → S a b) {l : List α} (p : IsChain R l) :
+    IsChain S l := by induction p with grind
+
+theorem IsChain.imp' {S : α → α → Prop} (HRS : ∀ ⦃a b⦄, R a b → S a b)
+    (Hab : ∀ ⦃c⦄, R a c → S b c) {l : List α} : IsChain R (a :: l) → IsChain S (b :: l) := by
+  cases l with grind
+
+@[deprecated (since := "2025-09-19")]
+alias Chain.imp := IsChain.imp
+
+@[deprecated (since := "2025-09-19")]
+alias Chain.imp' := IsChain.imp'
+
+@[deprecated (since := "2025-09-19")]
+protected alias Pairwise.chain := Pairwise.isChain
+
+protected theorem IsChain.pairwise (ht : ∀ ⦃a b c⦄, R a b → R b c → R a c) (c : IsChain R l) :
+    Pairwise R l := by
+  induction c with
+  | nil | singleton => grind
+  | cons₂ hr h p =>
+    simp at p ⊢
+    exact ⟨⟨hr, fun _ ha => ht hr <| p.1 _ ha⟩, p⟩
+
+theorem isChain_iff_pairwise (ht : ∀ ⦃a b c⦄, R a b → R b c → R a c) :
+    IsChain R l ↔ Pairwise R l := ⟨IsChain.pairwise ht, Pairwise.isChain⟩
+
+theorem isChain_iff_getElem {l : List α} :
+    IsChain R l ↔ ∀ (i : Nat) (_hi : i + 1 < l.length), R l[i] l[i + 1] := by
+  induction l with
+  | nil => grind
+  | cons a l IH => cases l with | nil => grind | cons b l => simp [IH, Nat.forall_lt_succ_left']
+
+theorem IsChain.getElem {l : List α} (c : IsChain R l) (i : Nat) (hi : i + 1 < l.length) :
+    R l[i] l[i + 1] := isChain_iff_getElem.mp c _ _
 
 /-! ### range', range -/
 
-theorem chain_succ_range' : ∀ s n step : Nat,
-    Chain (fun a b => b = a + step) s (range' (s + step) n step)
-  | _, 0, _ => Chain.nil
-  | s, n + 1, step => (chain_succ_range' (s + step) n step).cons rfl
+theorem isChain_range' (s : Nat) : ∀ n step : Nat,
+    IsChain (fun a b => b = a + step) (range' s n step)
+  | 0, _ => .nil
+  | 1, _ => .singleton _
+  | n + 2, step => (isChain_range' (s + step) (n + 1) step).cons₂ rfl
 
+@[deprecated isChain_range' (since := "2025-09-19")]
+theorem chain_succ_range' (s : Nat) (n : Nat) (step: Nat) :
+    IsChain (fun a b => b = a + step) (s :: range' (s + step) n step) := isChain_range'  _ (n + 1) _
+
+theorem isChain_lt_range' (s n : Nat) {step} (h : 0 < step) :
+    IsChain (· < ·) (range' s n step) :=
+  (isChain_range' s n step).imp fun _ _ e => e.symm ▸ Nat.lt_add_of_pos_right h
+
+@[deprecated isChain_lt_range' (since := "2025-09-19")]
 theorem chain_lt_range' (s n : Nat) {step} (h : 0 < step) :
-    Chain (· < ·) s (range' (s + step) n step) :=
-  (chain_succ_range' s n step).imp fun _ _ e => e.symm ▸ Nat.lt_add_of_pos_right h
+    IsChain (· < ·) (s :: range' (s + step) n step) := isChain_lt_range' _ (n + 1) h
 
 /-! ### indexOf and indexesOf -/
 


Batteries/Data/List/Basic.lean
```diff
@@ -555,29 +555,54 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
...
 ```
 Chain R a [b, c, d] ↔ R a b ∧ R b c ∧ R c d
 ``` -/
-inductive Chain : α → List α → Prop
-  /-- A chain of length 1 is trivially a chain. -/
-  | nil {a : α} : Chain a []
-  /-- If `a` relates to `b` and `b::l` is a chain, then `a :: b :: l` is also a chain. -/
-  | cons : ∀ {a b : α} {l : List α}, R a b → Chain b l → Chain a (b :: l)
+@[deprecated IsChain (since := "2025-09-19")]
+def Chain : (α → α → Prop) → α → List α → Prop := (IsChain · <| · :: ·)
```
Rob23oba, line 590: Let's maybe add some deprecated versions of `Chain.nil` and `Chain.cons` as in
```suggestion

set_option linter.deprecated false in
@[deprecated IsChain.singleton (since := "2025-09-19"), match_pattern]
def Chain.nil {a : α} : Chain a [] := Chain.singleton a

@[deprecated (since := "2025-09-19"), match_pattern]
alias Chain.cons := IsChain.cons₂

```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,91 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+@[deprecated (since := "2025-09-19")]
+alias rel_of_chain_cons := rel_of_isChain_cons₂
+
+theorem isChain_of_isChain_cons₂ {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) :
+    IsChain R (b :: l) := (isChain_cons₂.1 p).2
+
+@[deprecated (since := "2025-09-19")]
+alias chain_of_chain_cons := isChain_of_isChain_cons₂
+
+@[grind]
```
Rob23oba: I'm not convinced this is a good `grind` lemma, this simply matches on all pairs of `IsChain R l` in the context (growing quadratically)

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,91 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+@[deprecated (since := "2025-09-19")]
+alias Chain.imp := IsChain.imp
+
+@[deprecated (since := "2025-09-19")]
+alias Chain.imp' := IsChain.imp'
+
+@[deprecated (since := "2025-09-19")]
+protected alias Pairwise.chain := Pairwise.isChain
+
+protected theorem IsChain.pairwise (ht : ∀ ⦃a b c⦄, R a b → R b c → R a c) (c : IsChain R l) :
```
Rob23oba: It would probably be nice to have a variant where the transitivity proof is automatic, as in `[Trans R R R]`.

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,91 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    Pairwise R l := by
+  induction c with
+  | nil | singleton => grind
+  | cons₂ hr h p =>
+    simp at p ⊢
```
Rob23oba: Non-terminal simp

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,91 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+alias rel_of_chain_cons := rel_of_isChain_cons₂
+
+theorem isChain_of_isChain_cons₂ {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) :
+    IsChain R (b :: l) := (isChain_cons₂.1 p).2
+
+@[deprecated (since := "2025-09-19")]
+alias chain_of_chain_cons := isChain_of_isChain_cons₂
+
+@[grind]
+theorem IsChain.imp {S : α → α → Prop} (H : ∀ ⦃a b⦄, R a b → S a b) {l : List α} (p : IsChain R l) :
```
fgdorais: Use `Subrelation`.

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,91 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain S l := by induction p with grind
+
+theorem IsChain.imp' {S : α → α → Prop} (HRS : ∀ ⦃a b⦄, R a b → S a b)
```
fgdorais: Same.

Batteries/Data/List/Basic.lean
```diff
@@ -555,29 +555,54 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
...
+inductive IsChain (R : α → α → Prop) : List α → Prop where
+  /-- A list of length 0 is a chain. -/
+  | nil : IsChain R []
+  /-- A list of length 1 is a chain. -/
+  | singleton (a : α) : IsChain R [a]
+  /-- If `a` relates to `b` and `b::l` is a chain, then `a :: b :: l` is also a chain. -/
+  | cons₂ (hr : R a b) (h : IsChain R (b :: l)) : IsChain R (a :: b :: l)
+
+
+variable {R : α → α → Prop}
```
fgdorais: Batteries tends to avoid `variable`. I don't think this one is necessary.

Batteries/Data/List/Basic.lean
```diff
@@ -555,29 +555,54 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
...
+```
+IsChain R [a, b, c, d] ↔ R a b ∧ R b c ∧ R c d
+``` -/
+inductive IsChain (R : α → α → Prop) : List α → Prop where
+  /-- A list of length 0 is a chain. -/
+  | nil : IsChain R []
+  /-- A list of length 1 is a chain. -/
+  | singleton (a : α) : IsChain R [a]
+  /-- If `a` relates to `b` and `b::l` is a chain, then `a :: b :: l` is also a chain. -/
+  | cons₂ (hr : R a b) (h : IsChain R (b :: l)) : IsChain R (a :: b :: l)
```
fgdorais: I like saving keystrokes too but this should be `cons_cons` and also below.

Batteries/Data/List/Basic.lean
```diff
@@ -555,29 +555,54 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
 def pwFilter (R : α → α → Prop) [DecidableRel R] (l : List α) : List α :=
   l.foldr (fun x IH => if ∀ y ∈ IH, R x y then x :: IH else IH) []
 
-section Chain
+section IsChain
```
fgdorais: This `section` is not necessary. A header comment is fine here.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Note that this does mean `grind` can no longer do the proof (I assume because it doesn't know what to do with Subrelation). Is this OK?

Batteries/Data/List/Basic.lean
wrenna-robson: Actually even heads don't appear to be the norm in this file. I'll remove it.

Batteries/Data/List/Basic.lean
wrenna-robson: Alright.

Batteries/Data/List/Lemmas.lean
Rob23oba: I'm not convinced, there is almost no API for `Subrelation` and e.g. `Pairwise.imp` also uses regular forall. Regardless, this shouldn't be tagged `grind`.

Batteries/Data/List/Lemmas.lean
wrenna-robson: `simp only` is OK, right?

Batteries/Data/List/Lemmas.lean
wrenna-robson: Do we need both versions? The `Trans` version would really seem to suffice. If we do have both, what is the naming convention?

Batteries/Data/List/Lemmas.lean
wrenna-robson: Fair enough.

linesthatinterlace:
Add some comments.

Batteries/Data/List/Lemmas.lean
wrenna-robson: I have removed the `grind`. I think I agree that using `Subrelation` here is not necessary.

Batteries/Data/List/Basic.lean
wrenna-robson: [dropLast_cons₂](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Basic.html#List.dropLast_cons%E2%82%82)

Note that we do have things like this in Mathlib (not sure about Batteries).

Batteries/Data/List/Lemmas.lean
fgdorais: Hm. I didn't realize `Subrelation` wasn't reducible. Seems that should be fixed.

fgdorais: I'm currently checking whether it makes sense to make `Subrelation` reducible [lean4#10464](https://github.com/leanprover/lean4/pull/10464). If so, there will be no difference between using `Subrelation` and spelling it out. However, this change cannot be forward ported here, so this PR will stick with the spelled-out version.

In any case, it would also be nice to have a theorem for `InvImage`.

Batteries/Data/List/Basic.lean
fgdorais: I'd rather stick to standard naming conventions than to promote non-standard ones; `cons²` would make more sense to me but I'd rather not invent new standards either.

Batteries/Data/List/Lemmas.lean
fgdorais: Looks like it doesn't break anything in Lean, Batteries and Mathlib. Let's see if core accepts the PR, but there is no need to worry about that here.

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,95 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+
+@[deprecated (since := "2025-09-19")]
+alias chain_of_chain_cons := isChain_of_isChain_cons_cons
+
+theorem IsChain.imp {S : α → α → Prop} (H : ∀ ⦃a b : α⦄, R a b → S a b) {l : List α}
+    (p : IsChain R l) : IsChain S l := by induction p with grind
+
+theorem IsChain.imp' {S : α → α → Prop} (HRS : ∀ ⦃a b : α⦄, R a b → S a b)
+    (Hab : ∀ ⦃c⦄, R a c → S b c) {l : List α} : IsChain R (a :: l) → IsChain S (b :: l) := by
+  cases l with grind [IsChain.imp]
```
fgdorais: This theorem is a bit too niche for my taste. I don't mind keeping it and I'm sure it was useful somewhere at some time, but I couldn't find any current uses in Batteries or Mathlib. I propose we deprecate `Chain.imp'` without a drop-in replacement. If CI finds a subtle use case then we can reevaluate before merging.

fgdorais: Here is a more useful lemma that fills in the gap:
```lean
theorem IsChain.change_head (h : ∀ c, R a c → R b c) : 
    IsChain R (a :: l) → IsChain R (b :: l) := by
  cases l <;> grind
```

Batteries/Data/List/Lemmas.lean
fgdorais: I agree that we don't need both versions: it's difficult to imagine a real situation where one would apply this theorem, locally prove that the relation is transitive, and then decide that the fact that this relation is transitive is not worth noting as a separate lemma.

In any case, if we eventually decide on having both versions, the spelled-out version should be `pairwise_of_trans` (or similar) and the `Trans` version should be just `pairwise`.

Batteries/Data/List/Lemmas.lean
fgdorais: Since `imp` is (short but) definitively non-descriptive. This should eventually be renamed `IsChain.subrelation` (or similar) but there is no need to do this here since it's easier to handle all similar uses of `imp` at the same time.

Batteries/Data/List/Lemmas.lean
Rob23oba: If we want to stay more true to the naming convention, something like `IsChain.cons_imp`. Then we can deprecate `IsChain.imp'` for `IsChain.imp` and `IsChain.cons_imp`.

Batteries/Data/List/Lemmas.lean
wrenna-robson: I will leave this here then and we can revisit when that change to core goes through.

Batteries/Data/List/Lemmas.lean
wrenna-robson: I have found a use for this in one place, but I agree it isn't an ideal name.

fgdorais: Where?

wrenna-robson: It was useful when adapting the proof of `Encodable.raise'_chain` in mathlib4-nightly-testing#70. But that literally is the only place I found to use it.

wrenna-robson: However I just managed to factor it out, on second thoughts. I've called your `change_head` `cons_of_imp_of_cons`.

fgdorais: I was hoping to get rid of `imp` but I think we're stuck with it for the moment.

Batteries/Data/List/Basic.lean
```diff
@@ -555,29 +555,55 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
...
+  ⟨fun | .cons_cons hr h => ⟨hr, h⟩, fun ⟨hr, h⟩ => .cons_cons hr h⟩
 
-variable (R : α → α → Prop)
+instance instDecidableIsChain {R : α → α → Prop} [h : DecidableRel R] (l : List α) :
+    Decidable (l.IsChain R) := match l with | [] => isTrue .nil | a :: l => go a l
+  where
+    go (a : α) (l : List α) : Decidable ((a :: l).IsChain R) :=
+      match l with
+      | [] => isTrue <| .singleton a
+      | b :: l => haveI := (go b l); decidable_of_iff'  _ isChain_cons_cons
```
fgdorais:
```suggestion
      | b :: l => haveI := (go b l); decidable_of_iff' _ isChain_cons_cons
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
-theorem Chain.imp {R S : α → α → Prop} (H : ∀ a b, R a b → S a b) {a : α} {l : List α}
-    (p : Chain R a l) : Chain S a l :=
-  p.imp' H (H a)
+@[deprecated (since := "2025-09-19")]
+alias chain_cons := isChain_cons_cons
 
-protected theorem Pairwise.chain (p : Pairwise R (a :: l)) : Chain R a l := by
-  let ⟨r, p'⟩ := pairwise_cons.1 p; clear p
-  induction p' generalizing a with grind
+theorem rel_of_isChain_cons_cons {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) : R a b :=
```
fgdorais:
```suggestion
theorem rel_of_isChain_cons_cons (p : IsChain R (a :: b :: l)) : R a b :=
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+  (isChain_cons_cons.1 p).1
+
+@[deprecated (since := "2025-09-19")]
+alias rel_of_chain_cons := rel_of_isChain_cons_cons
+
+theorem isChain_cons_of_isChain_cons_cons {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) :
```
fgdorais:
```suggestion
theorem isChain_cons_of_isChain_cons_cons (p : IsChain R (a :: b :: l)) :
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain R (b :: l) := (isChain_cons_cons.1 p).2
+
+@[deprecated (since := "2025-09-19")]
+alias chain_of_chain_cons := isChain_cons_of_isChain_cons_cons
+
+theorem isChain_of_isChain_cons {b : α} {l : List α} (p : IsChain R (b :: l)) :
```
fgdorais:
```suggestion
theorem isChain_of_isChain_cons (p : IsChain R (b :: l)) :
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain R l := by
+  cases l
+  · exact .nil
+  · exact isChain_cons_of_isChain_cons_cons p
+
+theorem isChain_of_isChain_cons_cons {a b : α} {l : List α} (p : IsChain R (a :: b :: l)) :
```
fgdorais:
```suggestion
theorem isChain_of_isChain_cons_cons (p : IsChain R (a :: b :: l)) :
```

Batteries/Data/List/Lemmas.lean
fgdorais: This is ultimately redundant but it will be a slow deprecation process...

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain R l := isChain_of_isChain_cons (isChain_of_isChain_cons p)
+
+theorem IsChain.imp {S : α → α → Prop} (H : ∀ ⦃a b : α⦄, R a b → S a b) {l : List α}
```
fgdorais:
```suggestion
theorem IsChain.imp (H : ∀ ⦃a b : α⦄, R a b → S a b)
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+
+@[deprecated (since := "2025-09-19")]
+alias Chain.imp := IsChain.imp
+
+theorem IsChain.cons_of_imp_of_cons (h : ∀ c, R a c → R b c) :
+    IsChain R (a :: l) → IsChain R (b :: l) := by cases l <;> grind
+
+@[deprecated "Use IsChain.imp and IsChain.change_head" (since := "2025-09-19")]
+theorem Chain.imp' {S : α → α → Prop} (HRS : ∀ ⦃a b : α⦄, R a b → S a b)
+    (Hab : ∀ ⦃c⦄, R a c → S b c) {l : List α} : IsChain R (a :: l) → IsChain S (b :: l) := by
```
fgdorais:
```suggestion
theorem Chain.imp' (HRS : ∀ ⦃a b : α⦄, R a b → S a b)
    (Hab : ∀ ⦃c⦄, R a c → S b c) : IsChain R (a :: l) → IsChain S (b :: l) := by
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+  | 1, _ => .singleton _
+  | n + 2, step => (isChain_range' (s + step) (n + 1) step).cons_cons rfl
+
+@[deprecated isChain_range' (since := "2025-09-19")]
+theorem chain_succ_range' (s : Nat) (n : Nat) (step: Nat) :
+    IsChain (fun a b => b = a + step) (s :: range' (s + step) n step) := isChain_range'  _ (n + 1) _
+
+theorem isChain_lt_range' (s n : Nat) {step} (h : 0 < step) :
+    IsChain (· < ·) (range' s n step) :=
+  (isChain_range' s n step).imp fun _ _ e => e.symm ▸ Nat.lt_add_of_pos_right h
```
fgdorais:
```suggestion
  (isChain_range' s n step).imp fun | _, _, rfl => Nat.lt_add_of_pos_right h
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
-  | _, 0, _ => Chain.nil
-  | s, n + 1, step => (chain_succ_range' (s + step) n step).cons rfl
+theorem isChain_range' (s : Nat) : ∀ n step : Nat,
+    IsChain (fun a b => b = a + step) (range' s n step)
+  | 0, _ => .nil
+  | 1, _ => .singleton _
+  | n + 2, step => (isChain_range' (s + step) (n + 1) step).cons_cons rfl
+
+@[deprecated isChain_range' (since := "2025-09-19")]
+theorem chain_succ_range' (s : Nat) (n : Nat) (step: Nat) :
```
fgdorais:
```suggestion
theorem chain_succ_range' (s n step : Nat) :
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain (fun a b => b = a + step) (s :: range' (s + step) n step) := isChain_range'  _ (n + 1) _
+
+theorem isChain_lt_range' (s n : Nat) {step} (h : 0 < step) :
+    IsChain (· < ·) (range' s n step) :=
+  (isChain_range' s n step).imp fun _ _ e => e.symm ▸ Nat.lt_add_of_pos_right h
 
+@[deprecated isChain_lt_range' (since := "2025-09-19")]
 theorem chain_lt_range' (s n : Nat) {step} (h : 0 < step) :
```
fgdorais:
```suggestion
theorem chain_lt_range' (s n : Nat) (h : 0 < step) :
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain (fun a b => b = a + step) (range' s n step)
+  | 0, _ => .nil
+  | 1, _ => .singleton _
+  | n + 2, step => (isChain_range' (s + step) (n + 1) step).cons_cons rfl
+
+@[deprecated isChain_range' (since := "2025-09-19")]
+theorem chain_succ_range' (s : Nat) (n : Nat) (step: Nat) :
+    IsChain (fun a b => b = a + step) (s :: range' (s + step) n step) := isChain_range'  _ (n + 1) _
+
+theorem isChain_lt_range' (s n : Nat) {step} (h : 0 < step) :
```
fgdorais:
```suggestion
theorem isChain_lt_range' (s n : Nat) (h : 0 < step) :
```

fgdorais (changes requested):
Cosmetic changes.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Yes - I found when adapting Mathlib that it was convenient to still have these even though they are technically not needed, at least for now.

Batteries/Data/List/Basic.lean
```diff
@@ -555,29 +555,55 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
 def pwFilter (R : α → α → Prop) [DecidableRel R] (l : List α) : List α :=
   l.foldr (fun x IH => if ∀ y ∈ IH, R x y then x :: IH else IH) []
 
-section Chain
+/-- `IsChain R l` means that `R` holds between adjacent elements of `l`.
+```
+IsChain R [a, b, c, d] ↔ R a b ∧ R b c ∧ R c d
+``` -/
```
Rob23oba, line 561:
````suggestion
/--
`IsChain R l` means that `R` holds between adjacent elements of `l`. Example:
```
IsChain R [a, b, c, d] ↔ R a b ∧ R b c ∧ R c d
```
-/
````

File: Batteries/Data/List/Basic.lean [hunk 23]
```diff
@@ -555,29 +555,55 @@ pwFilter (·<·) [0, 1, 5, 2, 6, 3, 4] = [0, 1, 2, 3, 4]
...
+  ⟨fun | .cons_cons hr h => ⟨hr, h⟩, fun ⟨hr, h⟩ => .cons_cons hr h⟩
 
-variable (R : α → α → Prop)
+instance instDecidableIsChain {R : α → α → Prop} [h : DecidableRel R] (l : List α) :
+    Decidable (l.IsChain R) := match l with | [] => isTrue .nil | a :: l => go a l
+  where
+    go (a : α) (l : List α) : Decidable ((a :: l).IsChain R) :=
+      match l with
+      | [] => isTrue <| .singleton a
+      | b :: l => haveI := (go b l); decidable_of_iff' _ isChain_cons_cons
```
Rob23oba, line 582:
```suggestion
instance {R : α → α → Prop} [h : DecidableRel R] : (l : List α) → Decidable (l.IsChain R)
  | [] => isTrue .nil | a :: l => go a l
where
  go (a : α) (l : List α) : Decidable ((a :: l).IsChain R) :=
    match l with
    | [] => isTrue <| .singleton a
    | b :: l => haveI := (go b l); decidable_of_iff' _ isChain_cons_cons
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+theorem IsChain.imp (H : ∀ ⦃a b : α⦄, R a b → S a b)
+    (p : IsChain R l) : IsChain S l := by induction p with grind
+
+@[deprecated (since := "2025-09-19")]
+alias Chain.imp := IsChain.imp
+
+theorem IsChain.cons_of_imp_of_cons (h : ∀ c, R a c → R b c) :
+    IsChain R (a :: l) → IsChain R (b :: l) := by cases l <;> grind
+
+@[deprecated "Use IsChain.imp and IsChain.change_head" (since := "2025-09-19")]
```
Rob23oba, line 385:
```suggestion
@[deprecated "Use IsChain.imp and IsChain.cons_of_imp_of_cons" (since := "2025-09-19")]
```

Batteries/Data/List/Lemmas.lean
```diff
@@ -330,43 +330,105 @@ theorem disjoint_take_drop : ∀ {l : List α}, l.Nodup → m ≤ n → Disjoint
...
+    IsChain R (b :: l) := (isChain_cons_cons.1 p).2
+
+@[deprecated (since := "2025-09-19")]
+alias chain_of_chain_cons := isChain_cons_of_isChain_cons_cons
+
+theorem isChain_of_isChain_cons (p : IsChain R (b :: l)) :
+    IsChain R l := by
+  cases l
+  · exact .nil
+  · exact isChain_cons_of_isChain_cons_cons p
```
Rob23oba, line 371: Thinking about this again, it feels a bit pointless to define `isChain_cons_of_isChain_cons_cons` when `isChain_of_isChain_cons` is strictly more general and can always replace `isChain_cons_of_isChain_cons_cons`. So maybe something like this might be better:
```suggestion
theorem isChain_of_isChain_cons (p : IsChain R (b :: l)) : IsChain R l := by
  cases l
  · exact .nil
  · exact (isChain_cons_cons.1 p).2

@[deprecated isChain_of_isChain_cons (since := "2025-09-19")]
theorem chain_of_chain_cons (p : IsChain R (a :: b :: l)) :
    IsChain R (b :: l) := isChain_of_isChain_cons p
```

PR #1433, opened 2025-09-28
chore: forward port of lean4#10627

fgdorais:
Surprisingly, I couldn't find these trivial lemmas anywhere.

### DIFF
--- a/Batteries/Data/Fin/Lemmas.lean
+++ b/Batteries/Data/Fin/Lemmas.lean
@@ -10,6 +10,24 @@ namespace Fin
 
 attribute [norm_cast] val_last
 
+/-! ### trivialities -/
+
+theorem not_exists_zero (P : Fin 0 → Prop) : ¬ ∃ i, P i := by
+  intro ⟨⟨_, _⟩, _⟩; contradiction
+
+theorem forall_zero (P : Fin 0 → Prop) : ∀ i, P i := by
+  intro ⟨_, _⟩; contradiction
+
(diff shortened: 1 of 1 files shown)


Rob23oba:
https://leanprover-community.github.io/mathlib4_docs//Init/Data/Fin/Lemmas.html#Fin.forall_fin_one
but it is surpringly not `simp` and `Fin 0` is missing.

Rob23oba:
Oh, right
```
@[simp]
theorem forall_fin_zero (P : Fin 0 → Prop) : (∀ i, P i) ↔ True := by
  apply iff_true_intro
  intro ⟨_, _⟩; contradiction
```
and add the `not_exists_fin_zero` one as well

fgdorais:
Thanks! I thought I had tried that but I probably did something silly like forgetting the parentheses...

linesthatinterlace:
In Mathlib, the zero case is generally handled with [IsEmpty.forall_iff](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/IsEmpty.html#IsEmpty.forall_iff) and [IsEmpty.exists_iff](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/IsEmpty.html#IsEmpty.exists_iff) (which is why they are missing). I don't disagree it would be good to have them though!

linesthatinterlace:
There is also an obvious instance of NeZero n that comes from "\exists i : Fin n, f i" that we don't quite have: I am not sure if there is a slick way to handle it but in such cases you basically want to immediately move to the n + 1 case. Possible you could make a simple recursor to do it?

PR #1435, opened 2025-09-29
feat: add `Fin.any` and `Fin.all`

fgdorais:


Batteries/Data/Fin/Lemmas.lean
```diff
@@ -191,3 +192,21 @@ theorem exists_iff_exists_minimal (p : Fin n → Prop) [DecidablePred p] :
...
+/-! ### any/all -/
+
+theorem any_eq_true_iff {p : Fin n → Bool} : Fin.any p = true ↔ ∃ i, p i = true := by simp
+
+theorem any_eq_false_iff {p : Fin n → Bool} : Fin.any p = false ↔ ∀ i, p i = false := by simp
+
+theorem all_eq_true_iff {p : Fin n → Bool} : Fin.all p = true ↔ ∀ i, p i = true := by simp
+
+theorem all_eq_false_iff {p : Fin n → Bool} : Fin.all p = false ↔ ∃ i, p i = false := by simp
+
```
wrenna-robson, line 421: So arguably we are missing some lemmas but you could argue they are covered by the fact this is just an abbrev.

I would expect `any_zero`, `all_zero`, `any_succ` and `all_succ`, though.

I also think you want something like `not_any_eq_all_not` and `not_all_eq_any_not` which are not hard proofs but I think are probably worth having. You could also have `all_eq_not_any_not` in the same way. In general I would consider what lemmas from here: https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Lemmas.html#List.not_any_eq_all_not are wanted by analogy.

Batteries/Data/Fin/Basic.lean
```diff
@@ -96,3 +96,17 @@ The function `p` is not evaluated on further inputs after the first `i` is found
...
+Short-circuits upon encountering the first true.
+-/
+protected abbrev any (p : Fin n → Bool) := find? p |>.isSome
+
+/--
+Returns true if `p i` is true for every `i`.
+
+Short-circuits upon encountering the first false.
+-/
+protected abbrev all (p : Fin n → Bool) := find? (! p ·) |>.isNone
```
wrenna-robson, line 144: Could also consider simply `! any (! p ·)`.

linesthatinterlace:
Some good work here! I think we have slightly different tastes for number of lemmas etc. in some ways - but I absolutely agree we should have this. (We're also missing, say, `findIdx?` - if you want another thing to think about after this...)

Batteries/Data/Fin/Lemmas.lean
fgdorais: How about going a step further and have classes to standardize the basic API for `all` and `any`?

Batteries/Data/Fin/Basic.lean
fgdorais: Potatoes/Tomatoes? I don't have a preference either way.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -191,3 +192,30 @@ theorem exists_iff_exists_minimal (p : Fin n → Prop) [DecidablePred p] :
...
   simpa only [decide_eq_true_iff, decide_eq_false_iff_not] using
     exists_eq_true_iff_exists_minimal_eq_true (p ·)
+
+/-! ### any/all -/
+
+theorem any_eq_true_iff {p : Fin n → Bool} : Fin.any p = true ↔ ∃ i, p i = true := by simp
+
+theorem any_eq_false_iff {p : Fin n → Bool} : Fin.any p = false ↔ ∀ i, p i = false := by simp
+
+theorem any_eq_all_finRange {p : Fin n → Bool} : Fin.any p = (List.finRange n).any p := by
```
fgdorais:
```suggestion
theorem any_eq_any_finRange {p : Fin n → Bool} : Fin.any p = (List.finRange n).any p := by
```

Batteries/Data/Fin/Basic.lean
wrenna-robson: I think I prefer using `all`.

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: That is an interesting idea. Arguably could do the same for things like `findSome?`...

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -203,6 +204,33 @@ theorem exists_iff_exists_minimal (p : Fin n → Prop) [DecidablePred p] :
...
+  simp only [Fin.all, find?_eq_find?_finRange, Option.isNone_iff_eq_none, List.find?_eq_none,
+    Bool.not_eq_eq_eq_not, Bool.not_true, Bool.not_eq_false, List.all_eq, decide_eq_true_eq]
+
+-- The instance in Lean is not tail recursive and leads to stack overflow.
+instance (p : Fin n → Prop) [DecidablePred p] : Decidable (∃ i, p i) :=
+  decidable_of_iff (Fin.any (p ·) = true) (by simp)
+
+-- The instance in Lean is not tail recursive and leads to stack overflow.
+instance (p : Fin n → Prop) [DecidablePred p] : Decidable (∀ i, p i) :=
+  decidable_of_iff (Fin.all (p ·) = true) (by simp)
```
eric-wieser: Don't these form instance diamonds with instances in mathlib?

kim-em: I think https://github.com/leanprover/lean4/pull/13938 will fix the underlying problem here.

fgdorais: Thanks Kim! I will adapt soon.

wrenna-robson: Does that mean this PR is dependent on leanprover/lean4#13938?

wrenna-robson: Ah I see this has now merged.

fgdorais:
awaiting-review

PR #1436, opened 2025-09-29
fix: Remove proof wanted

linesthatinterlace:
We have two proof_wanted statements that appear to have already been completed in batteries#1109. This commit removes them (and makes a minor edit to some hypotheses which don't need to be named).

### DIFF
--- a/Batteries/Data/Array/Monadic.lean
+++ b/Batteries/Data/Array/Monadic.lean
@@ -127,9 +127,9 @@ theorem SatisfiesM_foldrM [Monad m] [LawfulMonad m]
   · next h => exact .pure (Nat.eq_zero_of_not_pos h ▸ h0)
 
 theorem SatisfiesM_mapFinIdxM [Monad m] [LawfulMonad m] (as : Array α)
-    (f : (i : Nat) → α → (h : i < as.size) → m β)
+    (f : (i : Nat) → α → i < as.size → m β)
     (motive : Nat → Prop) (h0 : motive 0)
-    (p : (i : Nat) → β → (h : i < as.size) → Prop)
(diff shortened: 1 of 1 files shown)


fgdorais:
Let's fix some of the signatures at the same time: several arguments should be made implicit.

fgdorais:
Any parameter that can be inferred by a later parameter should be implicit. 

For example:
```
theorem SatisfiesM_mapFinIdxM [Monad m] [LawfulMonad m]
    {f : (i : Nat) → α → i < as.size → m β} {p : (i : Nat) → β → i < as.size → Prop} 
    {motive : Nat → Prop} (h0 : motive 0)
    (hs : ∀ i h, motive i → SatisfiesM (p i · h ∧ motive (i + 1)) (f i as[i] h)) :
    SatisfiesM
      (fun arr => motive as.size ∧ ∃ eq : arr.size = as.size, ∀ i h, p i arr[i] h)
      (Array.mapFinIdxM as f) := ...
```

```
theorem SatisfiesM_mapIdxM [Monad m] [LawfulMonad m] 
    {as : Array α} {f : Nat → α → m β} {p : (i : Nat) → β → i < as.size → Prop}
    {motive : Nat → Prop} (h0 : motive 0)
    (hs : ∀ i h, motive i → SatisfiesM (p i · h ∧ motive (i + 1)) (f i as[i])) :
    SatisfiesM
      (fun arr => motive as.size ∧ ∃ eq : arr.size = as.size, ∀ i h, p i arr[i] h)
      (as.mapIdxM f) :=
  SatisfiesM_mapFinIdxM h0 hs
```

fgdorais:
I fixed the implicit arguments in the file. If you agree with my changes then we can merge.

PR #1437, opened 2025-09-29
feat: Add `Fin.findSomeRev?` and `Fin.findRev?`

linesthatinterlace:
`Fin.findSomeRev?` and `Fin.findRev?`are obvious functions with a real, practical meaning which we do not have. This adds them.

This should not affect Mathlib except that there is some minor deprecation to do.

### DIFF
--- a/Batteries/Data/Fin/Lemmas.lean
+++ b/Batteries/Data/Fin/Lemmas.lean
@@ -11,6 +11,23 @@ namespace Fin
 
 attribute [norm_cast] val_last
 
+/-! ### last -/
+
+@[simp] theorem not_last_lt (a : Fin (n + 1)) : ¬(last n) < a := Fin.not_lt.mpr (le_last a)
+
+/-! ### exists, forall -/
+
+-- TODO: deprecate duplicates in Mathlib
+
+theorem forall_fin_succ_last {P : Fin (n + 1) → Prop} :
+    (∀ i, P i) ↔ P (.last _) ∧ (∀ i : Fin n, P i.castSucc) :=
+  ⟨fun H => ⟨H _, fun _ => H _⟩, fun ⟨H0, H1⟩ i => i.lastCases H0 H1⟩
+
+theorem exists_fin_succ_last {P : Fin (n + 1) → Prop} :
+    (∃ i, P i) ↔ P (.last _) ∨ (∃ i : Fin n, P i.castSucc) :=
+  ⟨fun ⟨i, h⟩ => i.lastCases Or.inl (fun i hi => Or.inr ⟨i, hi⟩) h,
+    fun h => h.elim (fun h => ⟨_, h⟩) (fun ⟨_, hi⟩ => ⟨_, hi⟩)⟩
+
 /-! ### foldl/foldr -/
 
 theorem foldl_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n → α} {a₁ a₂} :
@@ -21,9 +38,17 @@ theorem foldl_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
 
 theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n → α} {a₁ a₂} :
     foldr n (fun i x => op (f i) x) (op a₁ a₂) = op (foldr n (fun i x => op (f i) x) a₁) a₂ := by
-  induction n generalizing a₂ with
-  | zero => rfl
-  | succ n ih => simp only [foldr_succ, ha.assoc, ih]
+  simp only [← Fin.foldl_rev]
+  haveI : Std.Associative (flip op) := ⟨fun a b c => (ha.assoc c b a).symm⟩
+  exact foldl_assoc (op := flip op)
+
+theorem foldr_assoc_flip {op : α → α → α} [ha : Std.Associative op] {f : Fin n → α} {a₁ a₂} :
+    foldr n (fun i x => op x (f i)) (op a₁ a₂) = op a₁ (foldr n (fun i x => op x (f i)) a₂) := by
+  simp only [← Fin.foldl_rev, foldl_assoc]
+
+theorem foldl_assoc_flip {op : α → α → α} [ha : Std.Associative op] {f : Fin n → α} {a₁ a₂} :
+    foldl n (fun x i => op (f i) x) (op a₁ a₂) = op (foldl n (fun x i => op (f i) x) a₁) a₂ := by
+  simp only [← Fin.foldr_rev, foldr_assoc]
 
 /-! ### clamp -/
 
--- a/Batteries/Data/Fin/Lemmas.lean
+++ b/Batteries/Data/Fin/Lemmas.lean
@@ -83,20 +83,18 @@ theorem findSome?_eq_some_iff {f : Fin n → Option α} :
   induction n with
   | zero =>
     simp only [findSome?_zero, (Option.some_ne_none _).symm, false_iff]
-    exact fun  ⟨i, _⟩ => i.elim0
+    exact fun ⟨i, _⟩ => i.elim0
   | succ n ih =>
-    simp only [findSome?_succ, Option.or_eq_some_iff, Fin.exists_fin_succ, Fin.forall_fin_succ,
-      not_lt_zero, false_implies, implies_true, and_true, succ_lt_succ_iff, succ_pos,
-      forall_const, ih, and_left_comm (b := f 0 = none), exists_and_left]
+    simp only [findSome?_succ, Option.or_eq_some_iff, exists_fin_succ, forall_fin_succ,
+      succ_lt_succ_iff, succ_pos, not_lt_zero, ih]
+    grind
 
 @[simp, grind =] theorem findSome?_eq_none_iff {f : Fin n → Option α} :
     findSome? f = none ↔ ∀ i, f i = none := by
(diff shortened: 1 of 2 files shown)


fgdorais:
Interesting. I liked the `rev` lemmas since the `rev` API is a bit lacking. Maybe we can still have a separate `rev` PR? Or do you think they were really useless?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -12,6 +12,10 @@ namespace Fin
 
 attribute [norm_cast] val_last
 
+/-! ### rev -/
+
+attribute [grind] rev_lt_rev rev_le_rev rev_rev
```
fgdorais: Should we make a quick PR to core?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -47,42 +51,42 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
 @[simp] theorem findSome?_one {f : Fin 1 → Option α} : findSome? f = f 0 := rfl
 
 theorem findSome?_succ {f : Fin (n+1) → Option α} :
-    findSome? f = (f 0).or (findSome? fun i => f i.succ) := by
+    findSome? f = (f 0).or (findSome? (f ·.succ)) := by
   simp only [findSome?, foldl_succ, Option.orElse_eq_orElse, Option.orElse_eq_or]
   exact Eq.trans (by cases (f 0) <;> rfl) foldl_assoc
 
+@[deprecated findSome?_succ (since := "2025-10-06")]
```
fgdorais: The replacement is not an exact replacement so I don't see the need to deprecate.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -47,42 +51,42 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
 theorem findSome?_succ_of_some {f : Fin (n+1) → Option α} (h : f 0 = some x) :
-    findSome? f = some x := by simp [findSome?_succ, h]
+    findSome? f = some x := findSome?_succ.trans (h ▸ Option.some_or)
 
+@[deprecated findSome?_succ (since := "2025-10-06")]
```
fgdorais: The replacement is not an exact replacement so I don't see the need to deprecate.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -47,42 +51,42 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
 theorem findSome?_succ_of_isSome {f : Fin (n+1) → Option α} (h : (f 0).isSome) :
-    findSome? f = f 0 := by cases _h : f 0 <;> simp_all [findSome?_succ_of_some]
+    findSome? f = f 0 := findSome?_succ.trans (Option.or_of_isSome h)
 
+@[deprecated findSome?_succ (since := "2025-10-06")]
```
fgdorais: The replacement is not an exact replacement so I don't see the need to deprecate.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -47,42 +51,42 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
 theorem findSome?_succ_of_none {f : Fin (n+1) → Option α} (h : f 0 = none) :
-    findSome? f = findSome? fun i => f i.succ := by simp [findSome?_succ, h]
+    findSome? f = findSome? (f ·.succ) := findSome?_succ.trans (Option.or_eq_right_of_none h)
 
+@[deprecated findSome?_succ (since := "2025-10-06")]
```
fgdorais: The replacement is not an exact replacement so I don't see the need to deprecate.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -176,32 +188,185 @@ theorem eq_false_of_find?_eq_none {p : Fin n → Bool} (h : find? p = none) (i)
...
+theorem isSome_findSomeRev?_of_isSome {f : Fin n → Option α} (h : (f i).isSome) :
+    (findSomeRev? f).isSome := isSome_findSomeRev?_iff.2 ⟨_, h⟩
+
+theorem map_findSomeRev? (f : Fin n → Option α) (g : α → β) :
+    (findSomeRev? f).map g = findSomeRev? (Option.map g <| f ·) := by
+  induction n with
+  | zero => rfl
+  | succ n ih => simp [findSomeRev?_succ, Option.map_or, ih]
+
+theorem findSomeRev?_guard {p : Fin n → Bool} : findSomeRev? (Option.guard p) = findRev? p := rfl
```
fgdorais, line 206: simp?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -127,6 +131,14 @@ theorem map_findSome? (f : Fin n → Option α) (g : α → β) :
 
 theorem findSome?_guard {p : Fin n → Bool} : findSome? (Option.guard p) = find? p := rfl
```
fgdorais, line 124: simp?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -176,32 +188,185 @@ theorem eq_false_of_find?_eq_none {p : Fin n → Bool} (h : find? p = none) (i)
...
+    findRev? p = if p (last n) then some (last n)
+    else (findRev? fun i => p i.castSucc).map Fin.castSucc := by
+  simp only [findRev?, findSomeRev?_succ, Option.guard, fun a => apply_ite (Option.or · a),
+    Option.some_or, Option.none_or, map_findSomeRev?, Option.map_if]
+
+@[simp, grind =]
+theorem findRev?_eq_some_iff {p : Fin n → Bool} :
+    findRev? p = some i ↔ p i ∧ ∀ j, i < j → p j = false := by simp [findRev?, and_assoc]
+
+theorem isSome_findRev?_iff {p : Fin n → Bool} :
```
fgdorais: simp?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -176,32 +188,185 @@ theorem eq_false_of_find?_eq_none {p : Fin n → Bool} (h : find? p = none) (i)
...
+    (findRev? p).isSome ↔ ∃ i, p i := by simp [findRev?]
+
+@[simp, grind =]
+theorem findRev?_eq_none_iff {p : Fin n → Bool} : findRev? p = none ↔ ∀ i, p i = false := by
+  simp [findRev?]
+
+theorem isNone_findRev?_iff {p : Fin n → Bool} : (findRev? p).isNone ↔ ∀ i, p i = false := by
+  simp [findRev?]
```
fgdorais: Why doesn't `simp` just work here?

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: Yes, I think that would be good.

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: Where?

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: It does.

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: Is the issue that `findRev` should be an abbrev?

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: ?

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: (Do you want to or should I?)

fgdorais: Go ahead!

Batteries/Data/Fin/Lemmas.lean
fgdorais: Same with `find?`?

fgdorais:
Looks good. Please fix conflicts and decide whether `find?` and `findRev?` should be `abbrev`. That makes sense to me but I'm curious about downstream side-effects.

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: Hmm - I think one still needs to apply associativity, is the issue.

Batteries/Data/Fin/Lemmas.lean
wrenna-robson: lean4#11065

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -16,6 +16,10 @@ namespace Fin
 
 attribute [norm_cast] val_last
 
+/-! ### rev -/
+
+attribute [grind =] rev_lt_rev rev_le_rev rev_rev
```
fgdorais, line 22:
```suggestion
-- Forward port of lean4#11065
attribute [grind =] rev_lt_rev rev_le_rev rev_rev
```

PR #1460, opened 2025-10-12
feat: add `Fin.sum`, `Fin.prod`, `Fin.countP`

fgdorais:


Batteries/Data/Fin/Lemmas.lean
```diff
@@ -40,6 +40,54 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
+/-! ### sum -/
+
+@[simp, grind =]
+theorem sum_zero [Zero α] [Add α] (x : Fin 0 → α) :
+    Fin.sum x = 0 := by
+  simp [Fin.sum]
+
+theorem sum_succ [Zero α] [Add α] (x : Fin (n + 1) → α) :
+    Fin.sum x = x 0 + Fin.sum (x ·.succ) := by
+  simp [Fin.sum, foldr_succ]
```
wrenna-robson, line 52: You are almost certainly going to want this but in the other direction, that is, `Fin.sum (x ·.castSucc) + x (last n)` or similar. Probably call it `sum_succ_last`, in theme with `foldr_succ_last`. Except there will be associativity considerations. Indeed, you might wish to consider if another construct is needed for non-associative addition where you use `foldl` instead of `foldr` (and probably add the other way round).

wrenna-robson: (Obviously the same concern exists with `prod`.)

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -40,6 +40,54 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
 
 @[simp] theorem coe_clamp (n m : Nat) : (clamp n m : Nat) = min n m := rfl
 
+/-! ### sum -/
+
+@[simp, grind =]
+theorem sum_zero [Zero α] [Add α] (x : Fin 0 → α) :
+    Fin.sum x = 0 := by
+  simp [Fin.sum]
+
```
wrenna-robson, line 49: Why not have `sum_one` (and `prod_one`)?

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -40,6 +40,54 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
+@[simp, grind =] theorem count_zero (p : Fin 0 → Bool) : Fin.count p = 0 := by
+  simp [Fin.count]
+
+@[simp, grind =] theorem count_one (p : Fin 1 → Bool) : Fin.count p = (p 0).toNat := by
+  simp [Fin.count, Fin.sum_succ]
+
+theorem count_succ (p : Fin (n + 1) → Bool) :
+    Fin.count p = (p 0).toNat + Fin.count (p ·.succ) := by
+  simp [Fin.count, Fin.sum_succ]
```
wrenna-robson: I wonder if `Fin.count p = bif p 0 then Fin.count (p ·.succ) else Fin.count (p ·.succ) + 1` is a better form of this. Which might suggest that `count` could have a better definition... also, especially here, you want a `last` version, and you could even have a version that uses `succAbove` or something exotic like that, I think.

wrenna-robson: Oh except `succAbove` is a Mathlib thing (which feels weird - it is definitely batteriesable). Leave that one.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -40,6 +40,54 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
+
+@[grind]
+theorem count_le (p : Fin n → Bool) : Fin.count p ≤ n := by
+  induction n with simp only [count_zero, count_succ, Nat.le_refl]
+  | succ _ ih =>
+    rw [Nat.add_comm]
```
wrenna-robson, line 99: This `add_comm` is precisely because I think that `count_succ` is of a sub-optimal form.

Batteries/Data/Fin/Basic.lean
```diff
@@ -84,6 +84,18 @@ This is the dependent version of `Fin.foldl`. -/
...
+protected def sum [Zero α] [Add α] (x : Fin n → α) : α :=
+  foldr n (x · + ·) 0
+
+/-- Product of a tuple indexed by `Fin n`. -/
+protected def prod [One α] [Mul α] (x : Fin n → α) : α :=
+  foldr n (x · * ·) 1
+
+/-- Count the number of true values of a decidable predicate on `Fin n`. -/
+protected def count (p : Fin n → Bool) : Nat :=
+  Fin.sum (p · |>.toNat)
```
wrenna-robson, line 101: `foldr n (fun i k => bif p i then k + 1 else k) 0` would be my suggestion.

Also, this should be called `countP`, analogously to `List.countP`. And you want a `Fin.count` as well.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -40,6 +40,54 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
+    apply Nat.add_le_add
+    · exact ih ..
+    · exact Bool.toNat_le ..
+
```
wrenna-robson, line 103: Want something linking this to `List.countP`.

linesthatinterlace:
Nice PR, but I do have some thoughts.

Batteries/Data/Fin/Lemmas.lean
fgdorais: I used `Bool.toNat` because I thought it was basically a no-op. It turns out it is not and I made this RFC [lean4#10835](https://github.com/leanprover/lean4/issues/10835).

linesthatinterlace:
I think that is reasonable. Though you might also want a theory for `Vector n` and that is even more analogous to this `Fin` case: I suspect in both `Fin` and `Vector` cases you will want theorems that are a thin layer onto `List`, but it's `List` where you should generally do the proofs (that is what it is for!)

linesthatinterlace:
I think for me the key distinction between `Fin` and `List` is that that `List` privileges `a :: l` over `l ++ [a]`, and quite rightly so, but `Fin` should not privilege `(f 0, f (x.succ))` over `(f (x.castSucc), f (last _)`.

Batteries/Data/Fin/Basic.lean
fgdorais: Fixed `countP`. [lean4#11035](https://github.com/leanprover/lean4/pull/11035) will resolve the rest.

Batteries/Data/Fin/Lemmas.lean
fgdorais: The `List.sum`/`List.prod` API has better tools for this, so I prefer users to use that rather than duplicating everything.

Batteries/Data/Fin/Lemmas.lean
fgdorais: This is better in the `List.sum`/`List.prod` API.

Batteries/Data/Fin/Lemmas.lean
```diff
@@ -39,6 +40,67 @@ theorem foldr_assoc {op : α → α → α} [ha : Std.Associative op] {f : Fin n
...
+  simp [Fin.countP]
+
+@[simp, grind =] theorem countP_one (p : Fin 1 → Bool) : Fin.countP p = (p 0).toNat := by
+  simp [Fin.countP, List.finRange_succ]
+
+theorem countP_succ (p : Fin (n + 1) → Bool) :
+    Fin.countP p = (p 0).toNat + Fin.countP (p ·.succ) := by
+  simp [Fin.countP, List.finRange_succ]; rfl
+
+@[simp, grind .]
```
fgdorais:
```suggestion
@[simp, grind =]
```

PR #1491, opened 2025-11-03
fix: Add missing csimp lemma

### LINKED ISSUE #307: Upstream various lemmas about sublists to Std
As of #301, we will have both `List.sublists` and `List.sublistsFast`, where the former is `implemented_by` the latter.

The proof that this is safe, `sublists_eq_sublistsFast`, as of https://github.com/leanprover-community/mathlib4/pull/7746 lives in mathlib.
We should upstream the results needed by this lemma (such as `List.reverseRecOn`) such that `sublists_eq_sublistsFast` can be in Std instead.

linesthatinterlace:
Addresses issue [batteries#307](https://github.com/leanprover-community/batteries/issues/307).

Closes #307


linesthatinterlace:
I managed to do it without needing the new lemma and now the proof is very smooth.

PR #1500, opened 2025-11-07
feat: Extend the API of `findIdxs` and `idxsOf`

linesthatinterlace:
This PR improves the API for `List.indexesOf` and `List.findIdxs`.

This depends on #1506. This is depended on by #1507

### DIFF
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -219,7 +219,9 @@ def scanr (f : α → β → β) (b : β) (l : List α) : List β :=
 
 /--
 Fold a list from left to right as with `foldl`, but the combining function
-also receives each element's index.
+also receives each element's index added to an optional parameter `start`
+(i.e. the numbers that `f` takes as its first argument will be greater than or equal to `start` and
+less than `start + l.length`).
 -/
 @[specialize] def foldlIdx (f : Nat → α → β → α) (init : α) : List β → (start : Nat := 0) → α
   | [], _ => init
@@ -227,9 +229,11 @@ also receives each element's index.
 
 /--
 Fold a list from right to left as with `foldr`, but the combining function
-also receives each element's index.
+also receives each element's index added to an optional parameter `start`
+(i.e. the numbers that `f` takes as its first argument will be greater than or equal to `start` and
+less than `start + l.length`).
 -/
-@[specialize] def foldrIdx {α : Type u} {β : Type v} (f : Nat → α → β → β) (init : β) :
+def foldrIdx {α : Type u} {β : Type v} (f : Nat → α → β → β) (init : β) :
     (l : List α) → (start : Nat := 0) → β
   | [], _ => init
   | a :: l, s => f s a (foldrIdx f init l (s + 1))
@@ -244,13 +248,16 @@ also receives each element's index.
     (foldrIdx f i xs s, s) := by induction xs generalizing s <;> grind [foldrIdx]
   grind [foldrIdxTR]
 
-/-- `findIdxs p l` is the list of indexes of elements of `l` that satisfy `p`. -/
+/-- `findIdxs p l` is the list of indexes of elements of `l` that satisfy `p`, added to an
+optional parameter `start` (so that the members of `findIdxs p l` will be greater than or
+equal to `start` and less than `l.length + start`).  -/
 @[inline] def findIdxs (p : α → Bool) (l : List α) (start : Nat := 0) : List Nat :=
   foldrIdx (fun i a is => bif p a then i :: is else is) [] l start
 
 /--
 Returns the elements of `l` that satisfy `p` together with their indexes in
-`l`. The returned list is ordered by index.
+`l` added to an optional parameter `start`. The returned list is ordered by index.
+We have `l.findIdxsValues p s = (l.findIdxs p s).zip (l.filter p)`.
 -/
 @[inline] def findIdxsValues (p : α → Bool) (l : List α) (start : Nat := 0) : List (Nat × α) :=
   foldrIdx (fun i a l => if p a then (i, a) :: l else l) [] l start
@@ -259,9 +266,11 @@ Returns the elements of `l` that satisfy `p` together with their indexes in
 alias indexsValues := findIdxsValues
 
 /--
-`idxsOf a l` is the list of all indexes of `a` in `l`. For example:
+`idxsOf a l` is the list of all indexes of `a` in `l`,  added to an
+optional parameter `start`. For example:
 ```
-idxsOf a [a, b, a, a] = [0, 2, 3]
+idxsOf b [a, b, a, a] = [1]
+idxsOf a [a, b, a, a] 5 = [5, 7, 8]
 ```
 -/
 @[inline] def idxsOf [BEq α] (a : α) (xs : List α) (start : Nat := 0) : List Nat :=
--- a/Batteries/Data/List/Lemmas.lean
+++ b/Batteries/Data/List/Lemmas.lean
@@ -20,7 +20,6 @@ attribute [grind =] zip_nil_left zip_nil_right zip_cons_cons
 
 attribute [grind =] zipIdx_nil zipIdx_cons
 
-
 /-! ### toArray-/
 
 @[deprecated List.getElem_toArray (since := "2025-09-11")]


Batteries/Data/List/Basic.lean
```diff
@@ -220,38 +221,45 @@ def scanr (f : α → β → β) (b : β) (l : List α) : List β :=
 Fold a list from left to right as with `foldl`, but the combining function
 also receives each element's index.
 -/
-@[simp, specialize] def foldlIdx (f : Nat → α → β → α) (init : α) : List β → (start : _ := 0) → α
+@[specialize] def foldlIdx (f : Nat → α → β → α) (init : α) : List β → (s : _ := 0) → α
```
eric-wieser: This is a breaking  API change; why are you renaming it?

Batteries/Data/List/Count.lean
```diff
@@ -25,8 +26,44 @@ namespace List
 
 section count
 
-variable [DecidableEq α]
+theorem count_singleton' [DecidableEq α] (a b : α) : count a [b] = if b = a then 1 else 0 :=
+  count_singleton.trans (by grind)
 
-theorem count_singleton' (a b : α) : count a [b] = if b = a then 1 else 0 := by simp [count_cons]
+variable [BEq α]
```
fgdorais: `variable` is preferably avoided in Batteries. This one is not necessary.

Batteries/Data/List/Basic.lean
fgdorais: I'm not sure it's useful to keep it secretive that `s` is a `Nat`.

Batteries/Data/List/Basic.lean
```diff
@@ -220,38 +221,45 @@ def scanr (f : α → β → β) (b : β) (l : List α) : List β :=
...
   | [], _ => init
-  | b :: l, i => foldlIdx f (f i init b) l (i+1)
+  | b :: l, s => foldlIdx f (f s init b) l (s + 1)
 
 /--
 Fold a list from right to left as with `foldr`, but the combining function
 also receives each element's index.
 -/
 -- TODO(Mario): tail recursive / array-based implementation
```
fgdorais: Let's add the TR version.

Batteries/Data/List/Lemmas.lean
```diff
@@ -429,47 +431,273 @@ theorem isChain_lt_range' (s n : Nat) (h : 0 < step) :
...
+
+@[simp, grind =]
+theorem Nodup.idxOf_getElem [BEq α] [LawfulBEq α] {xs : List α} (H : Nodup xs)
+    (i : Nat) (h : i < xs.length) : idxOf xs[i] xs = i := by induction xs generalizing i <;> grind
+
+/-! ### idxsOf -/
+
+section
+
+variable [BEq α]
```
fgdorais: No need for this `variable`.

Batteries/Data/List/Basic.lean
fgdorais: This is probably the easiest:
```lean
@[inline] def foldrIdxTR (f : Nat → α → β → β) (init : β) (l : List α) (s : Nat) : β :=
  l.foldr (fun a (b, i) => (f i a b, i + 1)) (init, s) |>.1
```
Or maybe this should be the actual definition?

Batteries/Data/List/Basic.lean
eric-wieser: I was referring to the choice to change `foldlIdx (start := 1)` to `foldlIdx (s := 1)`

Batteries/Data/List/Count.lean
```diff
@@ -25,8 +26,42 @@ namespace List
...
 
-theorem count_concat (a : α) (l : List α) : count a (concat l a) = succ (count a l) := by simp
+theorem count_concat [LawfulBEq α] (a : α) (l : List α) :
+    count a (concat l a) = succ (count a l) := by simp
+
+/-- For a fixed list, given the index of an element,
+  find the corresponding element and multiplicity. -/
```
eric-wieser: This is the multiplicity _so far_, right? I think this docstring needs an example.

Batteries/Data/List/Count.lean
```diff
@@ -25,8 +26,42 @@ namespace List
...
+def idxToCountElem [EquivBEq α] (xs : List α) (i : Fin xs.length) :
+    (x : α) × Fin (xs.count x) := ⟨xs[i], (xs.idxsOf xs[i]).idxOf ↑i, by grind⟩
```
eric-wieser: Computing `idxsOf` only to truncate is wasteful here. You should count the number that occur before position `i` instead.

Batteries/Data/List/Basic.lean
wrenna-robson: Is this an API change? I suppose it breaks references to that variable.

Batteries/Data/List/Count.lean
wrenna-robson: I agree but I was not attempting to create an efficient implementation. Arguably this would be good though! In a sense I do think there's a case for, in general, "folding with fuel", where you fold over a list but limit how far you'll go.

Batteries/Data/List/Count.lean
wrenna-robson: Oh - yes. Agreed.

Batteries/Data/List/Count.lean
wrenna-robson: Why isn't it?

Batteries/Data/List/Lemmas.lean
wrenna-robson: I don't understand why not? Do you mean I should just put this in every theorem?

Batteries/Data/List/Basic.lean
wrenna-robson: It would make a lot of sense to actually define these folds in terms of `List.foldr` and `List.foldl`. Shall I do that?

Batteries/Data/List/Basic.lean
wrenna-robson: Have changed it back.

Batteries/Data/List/Count.lean
wrenna-robson: I've moved this to #1507. Could you leave a similar comment there, by any chance? So I remember to get to it.

Batteries/Data/List/Count.lean
wrenna-robson: I think there's still a discussion to have here as I do see your point, but let's have that as part of #1507.

Batteries/Data/List/Basic.lean
eric-wieser: It breaks callers who pass the argument by name.

Batteries/Data/List/Basic.lean
wrenna-robson: Note that I think this definition does not work (or at least proving it seems a little hard).

wrenna-robson: Yeah this is not the correction definition I believe - there's an element of it being backwards.

Batteries/Data/List/Count.lean
wrenna-robson: Note that we use this elsewhere in the file!

Batteries/Data/List/Basic.lean
wrenna-robson:
```
@[inline] def foldrIdxTR (f : Nat → α → β → β) (init : β) (l : List α) (start : Nat := 0) : β :=
  l.foldr (fun a (b, i) => (f i a b, i - 1)) (init, start + l.length - 1) |>.1
```
This would work - but is not suitable because you need to traverse the list twice.

Batteries/Data/List/Lemmas.lean
```diff
@@ -92,17 +96,15 @@ theorem erase_eq_self_iff_forall_bne [BEq α] (a : α) (xs : List α) :
 
 /-! ### findIdx? -/
 
-theorem findIdx_eq_findIdx? (p : α → Bool) (l : List α) :
-    l.findIdx p = (match l.findIdx? p with | some i => i | none => l.length) := by grind
+@[deprecated (since := "2025-11-06")]
+alias findIdx_eq_findIdx? := findIdx_eq_getD_findIdx?
```
eric-wieser: You shouldn't use `alias` here, since that changes the statement. Better to leave the statement alone, deprecate it, and prove it with the preferred lemma to prove it's a replacement.

Batteries/Data/List/Lemmas.lean
```diff
@@ -92,17 +96,15 @@ theorem erase_eq_self_iff_forall_bne [BEq α] (a : α) (xs : List α) :
...
 
 /-! ### replaceF -/
 
 theorem replaceF_nil : [].replaceF p = [] := rfl
 
 theorem replaceF_cons (a : α) (l : List α) :
-    (a :: l).replaceF p = match p a with
-      | none => a :: replaceF p l
-      | some a' => a' :: l := rfl
+    (a :: l).replaceF p = (p a).elim (a :: replaceF p l) (· :: l) := by grind [replaceF]
```
eric-wieser: Is `elim` preferred over `match`? My inclination would be to leave this alone in the absence of a zulip discussion.

Batteries/Data/List/Lemmas.lean
```diff
@@ -429,47 +431,273 @@ theorem isChain_lt_range' (s n : Nat) (h : 0 < step) :
...
 
-/-! ### indexOf and indexesOf -/
+/-! ### foldrIdx -/
 
-theorem foldrIdx_start :
-    (xs : List α).foldrIdx f i s = (xs : List α).foldrIdx (fun i => f (i + s)) i := by
-  induction xs generalizing f i s with grind [foldrIdx]
+@[grind =] theorem foldrIdx_nil : ([] : List α).foldrIdx f i s = i := rfl
 
-@[simp] theorem foldrIdx_cons :
```
eric-wieser, line 438: why the removal of `simp`?

Batteries/Data/List/Lemmas.lean
```diff
@@ -429,47 +431,273 @@ theorem isChain_lt_range' (s n : Nat) (h : 0 < step) :
...
+theorem mem_findIdxs_iff_getElem_sub_pos :
+    i ∈ (xs : List α).findIdxs p s ↔ s ≤ i ∧
+    ∃ (hix : i - s < xs.length), p xs[i - s] := by
+  induction xs generalizing s <;> grind
+
+theorem mem_findIdxs_iff_getElem_pos :
+    i ∈ (xs : List α).findIdxs p ↔ ∃ (hix : i < xs.length), p xs[i] := by grind
+
+theorem pos_getElem_iff_mem_findIdxs (hi : i < (xs : List α).length) :
+    p xs[i] ↔ i ∈ xs.findIdxs p := by grind
```
eric-wieser: This should surely be flipped around to match `mem_findIdxs_iff_getElem_pos`

Batteries/Data/List/Lemmas.lean
```diff
@@ -429,47 +431,273 @@ theorem isChain_lt_range' (s n : Nat) (h : 0 < step) :
...
+
+theorem getElem_getElem_findIdxs_sub (s : Nat)
+    (h : i < ((xs : List α).findIdxs p s).length)
+    (h' : (findIdxs p xs s)[i] - s < xs.length := by grind) :
+    p xs[(xs.findIdxs p s)[i] - s] := by
+  rw [← getElem_filter_eq_getElem_getElem_findIdxs_sub s] <;> grind
+
+theorem getElem_getElem_findIdxs
+    (h : i < ((xs : List α).findIdxs p).length)
+    (h' : (findIdxs p xs)[i] < xs.length := by grind) :
```
eric-wieser: This shouldn't be an assumption, but a `haveI`

Batteries/Data/List/Basic.lean
wrenna-robson: Oh: the problem with this (and your one above): the current definition of `foldrIdx` allows `β` to be a Sort - this does not.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Sure - it seems so obviously worse to me that I was I think over-zealous.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Oh - this is possibly an error. I don't understand why it currently has it because `foldrIdx` is simp - I think I removed this at one point on that basis and then didn't add it back in. I agree it should have it.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Sure.

Batteries/Data/List/Lemmas.lean
wrenna-robson: What do you think it should be called? It seemed useful to have it this way for rewriting purposes.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Thanks - we seem to do this kind of thing in all sorts of ways and I am never sure what the best way is. Sometimes the wrong choice means lemmas don't fire when they should.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Added back in.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Have made a choice.

Batteries/Data/List/Lemmas.lean
fgdorais: Yes.

Batteries/Data/List/Basic.lean
wrenna-robson: Have added one.

Batteries/Data/List/Lemmas.lean
wrenna-robson: Done.

Batteries/Data/List/Basic.lean
```diff
@@ -221,30 +221,40 @@ def scanr (f : α → β → β) (b : β) (l : List α) : List β :=
...
+@[specialize] def foldrIdx {α : Type u} {β : Type v} (f : Nat → α → β → β) (init : β) :
+    (l : List α) → (start : Nat := 0) → β
   | [], _ => init
-  | a :: l, i => f i a (foldrIdx f init l (i+1))
+  | a :: l, s => f s a (foldrIdx f init l (s + 1))
+
+/-- A tail-recursive version of `foldrIdx`. -/
+@[inline] def foldrIdxTR (f : Nat → α → β → β) (init : β) (l : List α) (start : Nat := 0) : β :=
+  let as := l.toArray
+  (as.foldr (fun a (acc, n) => (f (n - 1) a acc, n - 1)) (init, start + as.size)).1
```
fgdorais: A quick benchmark test showed no significant different between the current version and this version.
```suggestion
@[inline] def foldrIdxTR (f : Nat → α → β → β) (init : β) (l : List α) (start : Nat := 0) : β :=
  l.foldr (fun a (acc, n) => (f (n - 1) a acc, n - 1)) (init, start + l.length) |>.1
```
I prefer this list-only version since it avoids allocating an array.

Please run your own benchmarks and report here.

wrenna-robson: I am unsure how to run benchmarks. I do note that `List.foldrTR` is allocating an array under the hood anyway - but I agree that it's better to stay in list-land if we can.

wrenna-robson: Incidentally, this works perfectly well as an initial definition:
```
@[specialize] def foldrIdx {α : Type u} {β : Type v} (f : Nat → α → β → β) (init : β)
    (l : List α) (start : Nat := 0) : β := (l.zipIdx start).foldr (flip f).uncurry init
````

Clearly it isn't efficient but if we're making an efficient runtime version anyway... (as an aside: it's a shame foldrIdx/foldlIdx are defined as they are, i.e. with Nat and α that way round in the definition of `f` - it would match zipIdx better were it not so).

Having said this, `zipIdxTR` and `foldrTR` do both exist - and `l.length` requires us to traverse the list twice anyway. I wonder if this is actually of comparable performance.

Comment (wrenna-robson) [reply Batteries/Data/List/Basic.lean]: I stuck some lemmas in about zipIdx anyway. I have changed the definition of `foldrIdx` to your suggested one but I remain unsure how to meaningfully produce benchmarks.

Comment (wrenna-robson) [reply Batteries/Data/List/Basic.lean]: Alright, I ran some tests myself:
```
@[noinline] def dummy (f : Nat → Nat → Nat → Nat) (init : Nat) (l : List Nat) (start : Nat := 0) : IO Nat :=
  pure <| l.foldr (fun a (acc, n) => (0, 0)) (init, start + l.length) |>.1

@[noinline] def foldrIdxTR1 (f : Nat → Nat → Nat → Nat) (init : Nat) (l : List Nat) (start : Nat := 0) : IO Nat :=
  pure <| l.foldr (fun a (acc, n) => (f (n - 1) a acc, n - 1)) (init, start + l.length) |>.1

@[noinline] def foldrIdxTR2 (f : Nat → Nat → Nat → Nat) (init : Nat) (l : List Nat) (start : Nat := 0) : IO Nat :=
  let as := l.toArray
  pure <| (as.foldr (fun a (acc, n) => (f (n - 1) a acc, n - 1)) (init, start + as.size)).1

@[noinline]
def n [RandomGen G] (size : Nat) : Option G → IO (Nat × Option G)
  | none => return (0, none)
  | some g => do
      let p := randNat g 0 size
      let n := p.1
      let g := p.2
      return (n, g)

@[noinline]
def xs [RandomGen G] (size : Nat) : Option G → IO (List Nat × Option G)
  | none => return ([], none)
  | some g => do
      let mut x : Array Nat := #[]
      let mut g : G := g
      for _ in [:size] do
        let p := randBool g
        let b := p.1
        g := p.2
        if b then let p := randNat g 0 size; x := x.push p.1; g := p.2
      return (x.toList, g)

def main (args : List String) : IO UInt32 := do
  match args[0]? >>= String.toNat?, args[1]? >>= String.toNat? with
  | none, _ => return 1
  | some size, seed =>
    let g := (mkStdGen ·) <$> seed
    let (a, g) ← n size g
    let (init, g) ← n size g
    let (xs, _) ← xs size g
    let _  ← timeit "dummy:" <| dummy (fun a b c => a + b + c) init xs a
    let r1  ← timeit "TR1:" <| foldrIdxTR1 (fun a b c => a + b + c) init xs a
    let r2  ← timeit "TR2:" <| foldrIdxTR2 (fun a b c => a + b + c) init xs a
    return (r1 != r2).toUInt32

#eval main ["800", "804350"]
```

I think your version is faster.

Comment (linesthatinterlace) [reply Batteries/Data/List/Basic.lean]: I think we should just use this as the definition of foldrIdx though - and do something similar for foldlIdx.

Comment (fgdorais) [reply Batteries/Data/List/Basic.lean]: Yep. I still don't see a significant difference on my end.

File: Batteries/Data/List/Basic.lean [hunk 13]
```diff
@@ -221,30 +221,39 @@ def scanr (f : α → β → β) (b : β) (l : List α) : List β :=
...
+  | b :: l, s => foldlIdx f (f s init b) l (s + 1)
 
 /--
 Fold a list from right to left as with `foldr`, but the combining function
 also receives each element's index.
 -/
--- TODO(Mario): tail recursive / array-based implementation
-@[simp, specialize] def foldrIdx (f : Nat → α → β → β) (init : β) :
-    (l : List α) → (start : _ := 0) → β
+@[specialize] def foldrIdx {α : Type u} {β : Type v} (f : Nat → α → β → β) (init : β) :
```
fgdorais:
```suggestion
def foldrIdx {α : Type u} {β : Type v} (f : Nat → α → β → β) (init : β) :
```

wrenna-robson: Why is specialize not needed here?

fgdorais: This code is never compiled because of the `csimp`.

PR #1504, opened 2025-11-10
chore(List/Perm): use `BEq`/`LawfulBEq`

urkud:
... instead of `DecidableEq`.

Also add a version of `Perm.insert` that doesn't need `LawfulBEq`.

### DIFF
--- a/Batteries/Data/Char/Basic.lean
+++ b/Batteries/Data/Char/Basic.lean
@@ -9,8 +9,8 @@ public import Batteries.Classes.Order
 
 @[expose] public section
 
--- Forward port of lean4#9515
-@[grind ←]
+-- Forward port of https://github.com/leanprover/lean4/pull/9515
+@[simp, grind ←]
 theorem List.mem_finRange (x : Fin n) : x ∈ finRange n := by
   simp [finRange]
 
--- a/Batteries/Data/List/Basic.lean
+++ b/Batteries/Data/List/Basic.lean
@@ -5,6 +5,7 @@ Authors: Leonardo de Moura
 -/
 
 module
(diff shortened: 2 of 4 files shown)


Batteries/Data/List/Perm.lean
```diff
@@ -168,16 +168,17 @@ theorem Nodup.perm_iff_eq_of_sublist {l₁ l₂ l : List α} (d : Nodup l)
       exact (d.1 this).elim
     | .cons₂ _ s₁ => rw [IH d.2 s₁ h.cons_inv]
 
-section DecidableEq
+section LawfulBEq
 
-variable [DecidableEq α]
+variable [BEq α] [LawfulBEq α]
```
fgdorais: Batteries likes to avoid `variable` and `section`. This one was a historical artifact, let's remove it.

Batteries/Data/List/Perm.lean
```diff
@@ -279,6 +280,13 @@ theorem Subperm.cons_left {l₁ l₂ : List α} (h : l₁ <+~ l₂) (x : α) (hx
     refine h y ?_
     simpa [hy'] using hy
 
+omit [LawfulBEq α] in -- TODO: replace `Perm.insert` in core
```
fgdorais: Please PR to core and then add a note to deprecate once that PR is merged.

fgdorais:
awaiting-author

Batteries/Data/List/Perm.lean
urkud: Removed for now. I'll open a PR to core later.

urkud:
-awaiting-author

PR #1511, opened 2025-11-12
feat: List.getElem_flatten

kim-em:
- [x] Depends on #1515


fgdorais:
I've made #1515 to add general partial sums and products. I will update this PR to build upon it when I have a chance. There's a chance I can pick-up `getElem_flatten` sooner than I thought, but if anyone else wants to do it then go ahead!

PR #1535, opened 2025-11-28
feat(by_contra): allow `rintro` patterns

JovanGerb:
This PR upgrades `by_contra` so that it cas use `rintro` patterns. The plan is to then do the same for `by_contra!` in mathlib.

It is now also possible to write a type annotation for the introduced variable (even when not naming the variable).

This PR also changes the behaviour of `by_contra` without any arguments. The introduced variable is now called `this`, instead of an inaccessible `a✝`.

I was surprised to learn that `by_contra` without arguments had hover information with the type of the introduced variable. I found this surprising and confusing, but I also couldn't figure out how to keep this feature with the new implementation, so this feature has been removed.

### DIFF
--- a/BatteriesTest/by_contra.lean
+++ b/BatteriesTest/by_contra.lean
@@ -1,5 +1,5 @@
 import Batteries.Tactic.Basic
-open Batteries.Tactic
+
 private def nonDecid (P : Prop) (x : P) : P := by
   by_contra h
   guard_hyp h : ¬P


fgdorais (approved):

Batteries/Tactic/Init.lean
```diff
@@ -31,23 +31,31 @@ elab (name := exacts) "exacts " "[" hs:term,* "]" : tactic => do
...
 * If `p` is decidable, it uses `Decidable.byContradiction` instead of `Classical.byContradiction`.
-* If `h` is omitted, the introduced variable `_: ¬p` will be anonymous.
+* If `h` is omitted, the introduced variable will be called `this`.
 -/
-macro (name := byContra) tk:"by_contra" e?:(ppSpace colGt binderIdent)? : tactic => do
-  let e := match e? with
-    | some e => match e with
-      | `(binderIdent| $e:ident) => e
-      | e => Unhygienic.run `(_%$e) -- HACK: hover fails without Unhygienic here
-    | none => Unhygienic.run `(_%$tk)
```
digama0, line 46: @fgdorais 

> I was surprised to learn that `by_contra` without arguments had hover information with the type of the introduced variable. I found this surprising and confusing, but I also couldn't figure out how to keep this feature with the new implementation, so this feature has been removed.

The way this works is that the above line transfers the span of `tk` which is the `by_contra` token to the `_` which is then elaborated and acquires the type hover. To replicate this in the new implementation you should replace ``mkIdent `this`` by ``mkIdentFrom tk `this`` where `tk` is bound by `by_contra%$tk` in the macro_rules pattern.

JovanGerb: I tried that, but it didn't work. I also tried to use ``mkIdentFrom tk `this (canonical := true)``. Then it did work, but then the doc-string of the tactic did not show up anymore, which I is quite bad.

fgdorais: Tentative PR at #1537

PR #1537, opened 2025-11-28
fix: patch #1535

fgdorais:


JovanGerb:
In the non-anonymous case the hover should only be on the variable itself, not on the tactic syntax. The tactic should just have its doc-string.

PR #1545, opened 2025-12-03
feat: upstream the `library_note` command from Mathlib

Vierkantor:
This PR moves Mathlib's `library_note2` command to Batteries, replacing `library_note`. The main difference, apart from a slightly changed syntax, is `library_note` now adds a declaration to the `LibraryNote` namespace.

As a consequence, it is no longer possible to have two library notes with the same name. (Was this feature ever used?)

### DIFF
--- a/Batteries/Util/LibraryNote.lean
+++ b/Batteries/Util/LibraryNote.lean
@@ -13,7 +13,7 @@ public meta section
 # Define the `library_note` command.
 -/
 
-namespace Batteries.Util.LibraryNote
+namespace Batteries.Util
 
 open Lean
 
@@ -21,6 +21,8 @@ open Lean
 in its doc-string. -/
 @[expose] def LibraryNote := Unit
 deriving Inhabited
+
+namespace LibraryNote
 
 /-- Entry for library notes in the environment extension.
 


Batteries/Util/LibraryNote.lean
```diff
@@ -28,20 +38,48 @@ initialize libraryNoteExt : SimplePersistentEnvExtension LibraryNoteEntry (Array
...
 Use `#help note "some tag"` to display all notes with the tag `"some tag"` in the infoview.
 This command can be imported from Batteries.Tactic.HelpCmd .
 -/
-elab "library_note " title:strLit ppSpace text:docComment : command => do
-  modifyEnv (libraryNoteExt.addEntry · (title.getString, text.getDocString))
+elab "library_note " name:ident ppSpace dc:docComment : command => do
+  modifyEnv (libraryNoteExt.addEntry · name.getId)
+  let stx ← `(
+    $dc:docComment
+    def $(mkIdent (`_root_.LibraryNote ++ name.getId)) : LibraryNote :=
```
grunweg, line 57: This puts all library notes in a global `LibraryNote` namespace. How hard is it to make this configurable, so e.g. mathlib library notes go into Mathlib.LibraryNote, etc? (This can be a follow-up!)

Batteries/Util/LibraryNote.lean
```diff
@@ -28,20 +38,48 @@ initialize libraryNoteExt : SimplePersistentEnvExtension LibraryNoteEntry (Array
...
+      default)
+  elabCommandTopLevel stx
```
grunweg, line 59: Elaborating this command ensures the library note is actually available in the environment (as opposed to only available to the environment extension), right?

Batteries/Util/LibraryNote.lean
```diff
@@ -28,20 +38,48 @@ initialize libraryNoteExt : SimplePersistentEnvExtension LibraryNoteEntry (Array
...
+
+open Elab Command in
+/-- Support the old `library_note "foo"` syntax, with a deprecation warning. -/
+elab "library_note " name:str ppSpace dc:docComment : command => do
+  logWarningAt name <|
+    "deprecation warning: library_note now takes an identifier instead of a string.\n" ++
+    "Hint: replace the double quotes with «french quotes»."
+  let name := Name.mkSimple name.getString
+  let stx ← `(library_note $(mkIdent name):ident $dc:docComment)
+  elabCommandTopLevel stx
```
grunweg, line 69: You're elaborating this as we still want the only library note to affect the environment.. makes sense.

Batteries/Util/LibraryNote.lean
Vierkantor: I expect this can be quite easy, for example `set_option libraryNote.namespace `Mathlib.LibraryNote`. But I don't see the usecase, so I would prefer to do this in a follow-up.

Batteries/Util/LibraryNote.lean
Vierkantor: Exactly!

PR #1581, opened 2025-12-28
feat: add List.{scanlM, scanrM}, Array.{scanlM, scanrM, scanl, scanr}

cmlsharp:
Per [this](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/scan.20combinator.20for.20iterators/near/564913979) comment on Zulip, there is some interest in expanding the scan{l,r,lM,rM} combinators in Batteries to {List, Array, Vector, Iterator}. This PR (I hope) represents the first  step towards this goal.

This PR adds `List.scanlM`, `List.scanrM`, `Array.scanl`, `Array.scanr`, `Array.scanrM`, `Array.scanlM` as well as a (yet incomplete) set of theorems relating these functions to each other and the existing `List.scanl`, `List.scanr`.

Some notes:
* Like `List.mapM`, the "nicely structured" version of `List.scanlM` cannot be proven equivalent to the tail recursive version without requiring `LawfulMonad`. Hence this PR adopts the same convention as the standard library does with `mapM`. The main `List.scanlM` is the tail-recursive implementation, while `List.scanlM'` is the non-tail recursive implementation with the nice structure. A theorem that proves these functions are equivalent when `m` is Lawful  is included.
* While the existing `List.scanl` and `List.scanr` could have been redefined in terms of `List.scanlM` and `List.scanrM`, I chose to do what the standard library does with `foldl/foldlM` and leave their implementations untouched.
* Similarly for the Array variants, there are reference implementations, as well as efficient versions for each (tied together by `@[implemented_by]`). Unlike in the standard library implementations of `Array.foldlM` and `Array.foldrM`,  the efficient versions make minimal use of unsafe. (only the efficient version for `Array.scanrM` presently requires it, just to assert that an array length fits in a usize). Bounds checks/termination is proven.
  * I believe it should be possible to prove that 'array length fits in usize' implies the equivalence of the reference and efficient implementations, but I do not have these proofs at the moment. 
* ~~What is definitely missing at the moment are more theorems for `Array.scan*`. I have proven minimally that the reference implementations of `Array.scanlM` and `List.scanlM` are equivalent (resp. `scanrM`) (assuming lawful monad). So many other theorems should be derivable, I just haven't done it yet.~~ Also these equivalence proofs (particularly for `scanrM`) could use simplification.

### DIFF
--- a/Batteries/Data/Array/Basic.lean
+++ b/Batteries/Data/Array/Basic.lean
@@ -209,30 +209,30 @@ private theorem scanlM_loop_eq_scanlMFast_loop [Monad m]
     {h_start : start ≤ as.size}
     {h_stop : stop ≤ as.size}
     {acc : Array β}
-    : scanlM.loop f init as start stop h_stop acc
-    = scanlMFast.loop f init as
-        (USize.ofNat start)
-        (USize.ofNat stop)
-        (by rw [USize.toNat_ofNat_of_lt' (Nat.lt_of_le_of_lt h_stop h_size)]; exact h_stop)
-        acc
-    := by
-      generalize h_n : stop - start = n
-      induction n using Nat.strongRecOn generalizing start acc init
-      rename_i n ih
-      rw [scanlM.loop, scanlMFast.loop]
-      have h_stop_usize : (USize.ofNat stop).toNat = stop := nat_index_eq_usize_index (h := h_size) (hn := h_stop)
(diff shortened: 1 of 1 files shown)


cmlsharp:
WIP

fgdorais:
Can you split list/array in two PRs? It's a good idea to focus PRs on one topic and, as a side bonus, once the first PR gets merged, you won't need approvals to run CI.

PR #1585, opened 2025-12-31
feat: add Iter.Scan/Iter.ScanM/IterM.scan/IterM.scanM

cmlsharp:
This PR builds off of #1581 and adds a Scan combinator to Iterators as well (the dependence on #1581 is just for the lemmas that prove correspondence between the iterator combinators and their list/array equivalents).

The primary contribution of this PR is `IterM.scanM` which, given an effectful iterator, an inital value, and an effectful accumulation function, returns an iterator over the partial results of folding that function over the values of the inner iterator. The remaining `Iter.Scan`/`Iter.ScanM`/`IterM.scan` are just implemented in terms of this in the obvious way.

This PR includes instances for both Productive and Finite (if the underlying iterator is Productive, resp. Finite) and includes lemmas showing that (at least for pure iterators / iterators in the identity monad) `Iter.scan*` is equivalent to accumulating the values into a list (resp. Array) and then calling List.scanl/List.scanlM (resp. Array.scanl/Array.scanlM). I would be happy to continue plugging away at more theorems, but this seems like a minimally useful set.

### DIFF
--- a/Batteries/Data/Iterators/Lemmas/Scan.lean
+++ b/Batteries/Data/Iterators/Lemmas/Scan.lean
@@ -1,3 +1,9 @@
+/-
+Copyright (c) 2025 Chad Sharp. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chad Sharp
+-/
+
 module
 public import Batteries.Data.Iterators.Scan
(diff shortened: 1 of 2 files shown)


cmlsharp:
Closed in favor of #1590

PR #1588, opened 2025-12-31
feat: add List.scanlM/scanrM

cmlsharp:
Per [this](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/scan.20combinator.20for.20iterators/near/564913979) Zulip thread. There is interest in improving the state of the `scan{l,m}` combinator and making it consistent across the various collections (iterator, list, array, vector). This PR is the first of several (and was originally a component of #1581).

It adds `List.scanrM` and `List.scanrM` which are the monadic equivalents of `List.scanl` and `List.scanr`. (analogous to `List.foldlM`/`List.foldrM`). A collection of theorems are also proved about these.

Like `List.mapM` the "nice" version of `List.scanlM` cannot be proven equivalent to its tail-recursive version without a `LawfulMonad` instance. This PR handles this the same way as in the standard library, where we define the tail recursive `List.scanlM` and the "nicely structured" non-tail-recursive `List.scanlM'` and prove they are equivalent given a LawfulMonad instance.

PR #1589, opened 2025-12-31
feat: Add Array.scan{l,r,lM,rM}

cmlsharp:
* Depends on: #1588

This PR adds the Array equivalents of List.scan{l,r,L,R}. It was originally part of #1581.

* This PR provides both reference and efficient versions of Array.scanlM and Array.scanrM, and then defines `Array.scanl` and `Array.scanr` in terms of them. This is identical to how the standard library currently handles `Array.fold*`.
* Unlike the standard library, this PR attempts to make minimal use of `unsafe` in the implementations of the efficient versions. It defines an assumption (as `Array.size_fits_usize`) as well as an `unsafe` proof of this assumption (as `Array.unsafe_size_fits_usize`). ~~These are currently public definitions. I could make them private, but they seem useful for writing other 'efficient' array functions.~~
*  `Array.scanlM` is proven to be equivalent to its efficient implementation under the assumption that `Array.size_fits_usize` holds. The efficient version of `Array.scanrM` additionally uses this assumption when constructing the proofs for the bounds checks.
* I do not have a proof that the efficient version of `Array.scanrM`  is equivalent to its efficient version under `Array.size_fits_usize`. It should be possible, but the structure of these two implementations differ more, so the proof is more difficult. With `Array.scanlM` the only difference between the implementations was the use of `USize` or `Nat` for the index, whereas the `Array.scanrM` reference implementation builds the array, then reverses it and the efficient implementation iterates over the array in reverse.

### DIFF
--- a/Batteries/Data/List/Lemmas.lean
+++ b/Batteries/Data/List/Lemmas.lean
@@ -1052,4 +1052,5 @@ theorem take_succ_drop {l : List α} {n stop : Nat}
   rw [← List.take_append_getElem (by simpa [← List.length_drop] using h)]
   simp [List.getElem_drop]
 
+@[simp]
 theorem reverse_singleton {a : α} : [a].reverse = [a] := rfl


cmlsharp:
awaiting-review

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,250 @@ This will perform the update destructively provided that `a` has a reference cou
...
+-/
+private unsafe def unsafe_size_fits_usize {a : Array α} : Array.size_fits_usize (a := a) :=
+  lcProof
+
+@[inline]
+private def scanlMFast [Monad m] (f : β → α → m β) (init : β) (as : Array α)
+    (start := 0) (stop := as.size) : m (Array β) :=
+  let stop := min stop as.size
+  let start := min start as.size
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,250 @@ This will perform the update destructively provided that `a` has a reference cou
...
+  decreasing_by
+      have : start < (start + 1) := by grind only [USize.size_eq]
+      grind only [Nat.min_def, USize.lt_iff_toNat_lt]
+
+/--
+Fold an effectful function `f` over the array from the left, returning the list of partial results.
+-/
+@[implemented_by scanlMFast]
+def scanlM [Monad m] (f : β → α → m β) (init : β) (as : Array α) (start := 0)
+  (stop := as.size) : m (Array β) :=
```
fgdorais:
```suggestion
def scanlM [Monad m] (f : β → α → m β) (init : β) (as : Array α) (start := 0) 
    (stop := as.size) : m (Array β) :=
```

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,250 @@ This will perform the update destructively provided that `a` has a reference cou
...
+      pure <| acc.push init
+
+private theorem scanlM_loop_eq_scanlMFast_loop [Monad m]
+    {f : β → α → m β} {init : β} {as : Array α} {h_size : as.size_fits_usize}
+    {start stop : Nat} {h_start : start ≤ as.size}
+    {h_stop : stop ≤ as.size} {acc : Array β} :
+    scanlM.loop f init as start stop h_stop acc
+      = scanlMFast.loop f init as (USize.ofNat start) (USize.ofNat stop)
+      (by rw [USize.toNat_ofNat_of_lt' (Nat.lt_of_le_of_lt h_stop h_size)]; exact h_stop) acc := by
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,250 @@ This will perform the update destructively provided that `a` has a reference cou
...
+  apply scanlM_loop_eq_scanlMFast_loop
+  simp_all only [gt_iff_lt]
+  apply Nat.min_le_right
+
+@[inline]
+private def scanrMFast [Monad m] (f : α → β → m β) (init : β) (as : Array α)
+    (h_size : as.size_fits_usize) (start := as.size) (stop := 0) : m (Array β) :=
+  let start := min start as.size
+  let stop := min stop start
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,250 @@ This will perform the update destructively provided that `a` has a reference cou
...
+     : m (Array β) := do
+    if h_gt : stop < start then
+      let startM1 := start - 1
+      have : startM1 < start := by grind only [!USize.sub_add_cancel, USize.lt_iff_le_and_ne,
+        USize.lt_add_one, USize.le_zero_iff]
+      have : startM1.toNat < as.size := Nat.lt_of_lt_of_le ‹_› ‹_›
+      have : (startM1 - stop) < (start - stop) := by grind only
+        [!USize.sub_add_cancel, USize.sub_right_inj, USize.add_comm, USize.lt_add_one,
+          USize.add_assoc, USize.add_right_inj]
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,250 @@ This will perform the update destructively provided that `a` has a reference cou
...
+    (f : α → β → m β) (init : β) (as : Array α) (start := as.size) (stop := 0) : m (Array β) :=
+  let start := min start as.size
+  loop f init as start stop (Nat.min_le_right _ _) #[]
+where
+  /-- auxiliary tail-recursive function for scanrM -/
+  loop (f : α → β → m β) (init : β) (as : Array α)
+       (start stop : Nat)
+       (h_start : start ≤ as.size)
+       (acc : Array β)
+     : m (Array β) := do
```
fgdorais:
```suggestion
       (acc : Array β) :
       m (Array β) := do
```

Batteries/Data/Array/Lemmas.lean
```diff
@@ -41,6 +41,10 @@ theorem idxOf?_toList [BEq α] {a : α} {l : Array α} :
     (a.eraseIdxIfInBounds i).size = if i < a.size then a.size-1 else a.size := by
   grind
 
+theorem toList_drop {as: Array α} {n : Nat}
+  : (as.drop n).toList = as.toList.drop n
+  := by simp only [drop, toList_extract, size_eq_length_toList, List.drop_eq_extract]
```
fgdorais:
```suggestion
theorem toList_drop {as: Array α} {n : Nat} : 
    (as.drop n).toList = as.toList.drop n := by 
  simp only [drop, toList_extract, size_eq_length_toList, List.drop_eq_extract]
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,364 @@
...
+                  |>.scanrM f init)
+                ++ acc.toList.reverse
+                |>.toArray := by
+  induction h_ind : start - stop generalizing stop acc init start with
+    | zero =>
+      grind [scanrM.loop, append_eq_toArray_iff, toList_reverse]
+    | succ n ih =>
+      unfold scanrM.loop
+      simp_all only [bind_pure_comp, show stop < start by omega, ↓reduceDIte]
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,364 @@
...
+      conv =>
+        lhs
+        arg 2
+        ext a
+        rw [ih (start := start - 1) (stop := stop) (acc := acc.push init) (by omega)]
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,364 @@
...
+      have h_list : List.take (n + 1) (List.drop stop as.toList)
+        = as[stop] :: List.take n (List.drop (stop + 1) as.toList)
+        := by
+          rw [List.drop_eq_getElem_cons (by simp; omega)]
+          simp
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,364 @@
...
+      have h_rev_list : (List.take (n + 1) (List.drop stop as.toList)).reverse
+        = as[start - 1] :: (List.take n (List.drop stop as.toList)).reverse
+        := by
+          have h_eq : start - 1 = stop + n := by omega
+          rw [← List.take_append_getElem
+            (by simp; omega : n < (List.drop stop as.toList).length)]
+          simp [List.reverse_append, List.getElem_drop, h_eq]
+
```
fgdorais:
```suggestion
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,364 @@
...
+theorem scanl_eq_scanlM {f : β → α → β} {as: Array α} :
+    as.scanl f init = (as.scanlM (m := Id) (pure <| f · ·) init).run := by
+  simp
+
+theorem scanl_toList {f: β → α → β} {as : Array α} :
+    (as.toList.scanl f init).toArray = as.scanl f init := by
+  rw [scanl_eq_scanlM, ← scanlM_toList]
+  simp [Id.run_map, pure, List.scanl]
+
```
fgdorais, line 156:
```suggestion
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,364 @@
...
+theorem getElem_scanl {f : β → α → β} {as: Array α} (h : i < (as.scanl f init).size) :
+    (as.scanl f init)[i]'h = foldl f init (as.take i) := by
+  simp only [←foldl_toList, ←scanl_toList]
+  simp
+
+@[grind =]
+theorem getElem?_scanl {f : β → α → β} :
+    (scanl f a l)[i]? = if i ≤ l.size then some (foldl f a (l.take i)) else none := by
+  grind
+
```
fgdorais, line 199:
```suggestion
```

fgdorais (commented):
More style stuff...

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,246 @@ This will perform the update destructively provided that `a` has a reference cou
 abbrev setN (a : Array α) (i : Nat) (x : α) (h : i < a.size := by get_elem_tactic) : Array α :=
   a.set i x
 
-end Array
+/--
+  This is guaranteed by the Array docs but it is unprovable.
+  May be asserted to be true in an unsafe context via `Array.unsafe_size_fits_usize
+-/
+private abbrev size_fits_usize {a : Array α}: Prop := a.size < USize.size
```
eric-wieser:
```suggestion
private abbrev SizeFitsUSize (a : Array α) : Prop := a.size < USize.size
```
is the usual naming convention.

cmlsharp: What is the general convention on naming? It seemed to me that defs usually used camel case and theorems usually used snake case except to the extent that they included names of defs that were camel case (e.g. size_toArray). 

Are props usually PascalCase?

fgdorais: Props are types, so they use CamelCase. Proofs, i.e. elements of props, use snake_case. Elements of higher types use dromedaryCase.

cmlsharp: Gotcha, ok I'll change this!

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,246 @@ This will perform the update destructively provided that `a` has a reference cou
...
 
-end Array
+/--
+  This is guaranteed by the Array docs but it is unprovable.
+  May be asserted to be true in an unsafe context via `Array.unsafe_size_fits_usize
+-/
+private abbrev SizeFitsUSize {a : Array α}: Prop := a.size < USize.size
+
+@[grind .]
+private theorem nat_index_eq_usize_index {n : Nat} {a : Array α}
```
fgdorais: Same.

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,246 @@ This will perform the update destructively provided that `a` has a reference cou
...
+    have h_start_succ : USize.ofNat start + 1 = USize.ofNat (start + 1) := by
+      simp_all only [← USize.toNat_inj, USize.toNat_add]
+      grind only [USize.size_eq, nat_index_eq_usize_index]
+    rw [h_start_succ]
+    apply ih (stop - (start + 1)) <;> omega
+  case isFalse h_nlt => grind [USize.lt_iff_toNat_lt]
+
+-- this theorem establishes that given the (unprovable) assumption that as.size < USize.size,
+-- the scanlMFast and scanlM are equivalent
+private theorem scanlM_eq_scanlMFast [Monad m]
```
fgdorais: Same.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,358 @@
...
+    | zero =>
+      unfold scanlM.loop
+      simp [show ¬(start < stop) by omega, ← Array.toList_push]
+    | succ n ih =>
+      unfold scanlM.loop
+      rw [List.drop_eq_getElem_cons (by simp; omega)]
+      simp [ show start < stop by omega
+           , show stop - (start + 1) = n by omega
+           , ih
+           ]
```
fgdorais: Lean supports terminal commas in lists and arrays, so there is no need for this syntax.
```suggestion
      simp [show start < stop by omega, show stop - (start + 1) = n by omega, ih]
```
FWIW, I personally like this syntax but it's not standard.

fgdorais (changes requested):

Batteries/Data/Array/Basic.lean
```diff
@@ -136,11 +136,246 @@ This will perform the update destructively provided that `a` has a reference cou
...
+    (acc := Array.replicate (start - stop + 1) init)
+    (by grind only [!Array.size_replicate, = Nat.min_def, Array.nat_index_eq_usize_index])
+where
+  @[specialize]
+  loop (f : α → β → m β) (init : β) (as : Array α)
+       (start stop : USize)
+       (h_start : start.toNat ≤ as.size)
+       (acc : Array β)
+       (h_bound : start.toNat - stop.toNat  < acc.size)
+     : m (Array β) := do
```
fgdorais:
```suggestion
       (h_bound : start.toNat - stop.toNat  < acc.size) :
       m (Array β) := do
```

fgdorais:
You did a lot of excellent work, so I did the final cleanup on your behalf. Please have a look at my changes and I will merge.

fgdorais (approved):

cmlsharp:
Great thanks! I took a look. I will keep these in mind for future PRs. Thanks for being patient with me!

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
-valid for the array. But this parameter has a default argument `by get_elem_tactic` which
-should prove the index bound.
+@[deprecated set (since := "2026-02-02")]
+alias setN := set
+
+/-
+This is guaranteed by the Array docs but it is unprovable.
+May be asserted to be true in an unsafe context via `Array.unsafe_size_fits_usize
+-/
+private abbrev SizeFitsUSize {a : Array α}: Prop := a.size < USize.size
```
eric-wieser:
```suggestion
private abbrev SizeFitsUSize (a : Array α) : Prop := a.size < USize.size
```
having this be implicit seems strange to me.

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+    {h : a.SizeFitsUSize} {hn : n ≤ a.size} :
+    (USize.ofNat n).toNat = n :=
+  USize.toNat_ofNat_of_lt' (Nat.lt_of_le_of_lt ‹_› ‹_›)
+
+/-
+This is guaranteed by the Array docs but it is unprovable.
+Can be used in unsafe functions to write more efficient implementations
+that avoid arbitrary precision integer arithmetic.
 -/
+private unsafe def unsafe_size_fits_usize {a : Array α} : SizeFitsUSize (a := a) :=
```
eric-wieser:
```suggestion
private unsafe def unsafe_sizeFitsUSize {a : Array α} : SizeFitsUSize a :=
```
is the naming scheme

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+private theorem scanlM_loop_eq_scanlMFast_loop [Monad m]
+    {f : β → α → m β} {init : β} {as : Array α} {h_size : as.SizeFitsUSize}
+    {start stop : Nat} {h_start : start ≤ as.size}
+    {h_stop : stop ≤ as.size} {acc : Array β} :
+    scanlM.loop f init as start stop h_stop acc
+      = scanlMFast.loop f init as (USize.ofNat start) (USize.ofNat stop)
+      (by rw [USize.toNat_ofNat_of_lt' (Nat.lt_of_le_of_lt h_stop h_size)]; exact h_stop) acc := by
+  generalize h_n : stop - start = n
+  induction n using Nat.strongRecOn generalizing start acc init
+  rename_i n ih
```
eric-wieser, line 220: instead of `rename_i`, why not se `with | _ n ih` on the line above?

Batteries/Data/Array/Lemmas.lean
```diff
@@ -41,6 +41,10 @@ theorem idxOf?_toList [BEq α] {a : α} {l : Array α} :
     (a.eraseIdxIfInBounds i).size = if i < a.size then a.size-1 else a.size := by
   grind
 
+theorem toList_drop {as: Array α} {n : Nat} :
```
eric-wieser:
```suggestion
theorem toList_drop (as : Array α) (n : Nat) :
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,334 @@
...
+                  |>.scanlM f init)
+               |>.toArray := by
+  induction h_ind : stop - start generalizing start acc init with
+    | zero =>
+      unfold scanlM.loop
+      simp [show ¬(start < stop) by omega, ← Array.toList_push]
+    | succ n ih =>
+      unfold scanlM.loop
+      rw [List.drop_eq_getElem_cons (by simp; omega)]
+      simp [show start < stop by omega, show stop - (start + 1) = n by omega, ih]
```
eric-wieser: style nit: I think indenting `|` is non-standard

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,334 @@
...
+public section
+
+/-!
+# Array
+
+Prove basic results about `Array.scanl`, `Array.scanr`, `Array.scanlM` and `Array.scanrM`.
+-/
+namespace Array
+
+theorem scanlM_loop_toList [Monad m] [LawfulMonad m]
```
eric-wieser:
```suggestion
theorem scanlMLoop_toList [Monad m] [LawfulMonad m]
```
is the naming convention

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,334 @@
...
+    | zero =>
+      unfold scanlM.loop
+      simp [show ¬(start < stop) by omega, ← Array.toList_push]
+    | succ n ih =>
+      unfold scanlM.loop
+      rw [List.drop_eq_getElem_cons (by simp; omega)]
+      simp [show start < stop by omega, show stop - (start + 1) = n by omega, ih]
+
+theorem scanlM_toList [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
+    List.toArray <$> as.toList.scanlM f init = as.scanlM f init := by
```
eric-wieser: This theorem statement looks weird to me; from the name I'd expect
```suggestion
theorem scanlM_toList [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
    as.toList.scanlM f init = Array.toList <$> as.scanlM f init := by
```

Batteries/Data/Array/Scan.lean
fgdorais: I prefer:
```suggestion
theorem scanlM.loop_toList [Monad m] [LawfulMonad m]
```

Batteries/Data/Array/Scan.lean
eric-wieser: Ah, I see that's the lemma below. In which case, perhaps
```suggestion
theorem toArray_scanlM_toList [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
    List.toArray <$> as.toList.scanlM f init = as.scanlM f init := by
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,334 @@
...
+  simp [scanlM, Array.scanlM_loop_toList, ←Array.length_toList]
+
+@[simp, grind =]
+theorem toList_scanlM [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
+    toList <$> as.scanlM f init = as.toList.scanlM f init := by
+  simp [← scanlM_toList]
```
eric-wieser: Can you add the `toArray <$> as.scanlM f init = as.toArray.scanlM f init` version too?

BatteriesTest/array_scan.lean
```diff
@@ -0,0 +1,39 @@
...
+
+/- Docstring examples for `Array.scan{l,r,lM,rM}` -/
+
+example [Monad m] (f : α → β → m α) :
+    Array.scanlM f x₀ #[a, b, c] = (do
+      let x₁ ← f x₀ a
+      let x₂ ← f x₁ b
+      let x₃ ← f x₂ c
+      pure #[x₀, x₁, x₂, x₃])
+    := by simp [scanlM, scanlM.loop]
```
eric-wieser: style:
```suggestion
example [Monad m] (f : α → β → m α) :
    Array.scanlM f x₀ #[a, b, c] = (do
      let x₁ ← f x₀ a
      let x₂ ← f x₁ b
      let x₃ ← f x₂ c
      pure #[x₀, x₁, x₂, x₃]) := by
  simp [scanlM, scanlM.loop]
```
(etc)

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,334 @@
...
+    scanl f init as.reverse = reverse (scanr (flip f) init as) := by
+  apply toList_inj.mp
+  simp only [← scanl_toList, ← scanr_toList, toList_reverse, List.scanl_reverse,
+    List.reverse_toArray]
+
+ end Array
+
+namespace Subarray
+theorem scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Subarray α} :
+    as.scanlM f init = (as.array.extract as.start as.stop).scanlM f init := by
```
eric-wieser: This theorem name looks wrong, `scanLM_extract` would have `(extract _ _).scanlM` on the left, not on the right. Perhaps
```suggestion
theorem scanlM_eq_scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Subarray α} :
    as.scanlM f init = (as.array.extract as.start as.stop).scanlM f init := by
```
or just `scanlM_eq` for short.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,334 @@
...
+@[simp, grind =]
+theorem toList_scanrM [Monad m] [LawfulMonad m] {f : α → β → m β} {as : Array α} :
+    toList <$> as.scanrM f init = as.toList.scanrM f init := by
+  simp [← scanrM_toList]
+
+-- TODO (cmlsharp): rewrite this and scanrM_extract without requiring LawfulMonad?
+-- In principle, it should be possible to flip the dependency order between this and
+-- scanlM_loop_toList, and this lemma should not require LawfulMonad
+theorem scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
+    as.scanlM f init start stop = (as.extract start stop).scanlM f init := by
```
eric-wieser: This is named backwards,
```suggestion
theorem scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
    (as.extract start stop).scanlM f init = as.scanlM f init start stop := by
```
and probably a good `simp` lemma in this direction.

Batteries/Data/Array/Scan.lean
eric-wieser: I guess that's fine too, the naming conventions don't really say anything about how to handle the nested names generated by `def ... := ... where ...`.

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+```lean example
+example [Monad m] (f : α → β → m β) :
+    Array.scanrM f x₀ #[a, b, c] (start := 3) (stop := 1) = (do
+      let x₁ ← f c x₀
+      let x₂ ← f b x₁
+      pure #[x₂, x₁, x₀])
+    := by simp [scanrM, scanrM.loop]
+```
+-/
+@[implemented_by scanrMUnsafe]
```
eric-wieser, line 324: Can we use `csimp` here instead?

Batteries/Data/Array/Scan.lean
eric-wieser: Shouldn't this involve `start` and `stop` too?
```suggestion
    List.toArray <$> as.toList.scanlM f init start stop = as.scanlM f init start stop := by
```

cmlsharp: The list methods do not have start and stop parameters.

Batteries/Data/Array/Basic.lean
cmlsharp: I don't see how? These cannot be proven to be equivalent. Their equivalence relies on a necessarily unprovable assumption.

Batteries/Data/Array/Scan.lean
cmlsharp: The only reasons the Array ones do is so Subarray implementations can exist (this mirrors how e.g. Array.fold* works)

eric-wieser: Ah, I missed that detail, thanks!

Batteries/Data/Array/Basic.lean
eric-wieser: You can use
```
@[csimp]
unsafe def bar_eq_foo : bar = foo := by
```
and then invoke your theorem with the unsafe def to emphasize where the cheating happens.

Batteries/Data/Array/Basic.lean
eric-wieser: Unresolving, it seems this wasn't actually addressed by your most recent push.

Batteries/Data/Array/Basic.lean
eric-wieser: Unresolving, it seems this wasn't actually addressed by your most recent push.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,336 @@
...
+@[grind =]
+theorem scanrM_map [Monad m] [LawfulMonad m] {f : α₁ → α₂ } {g: α₂ → β → m β} {as : Array α₁} :
+    (as.map f).scanrM g init = as.scanrM (fun a b => g (f a) b) init := by
+  simp only [← toArray_scanrM_toList, toList_map, List.scanrM_map]
+
+/-- ### Array.scanl -/
+
+theorem scanl_eq_scanlM {f : β → α → β} {as: Array α} :
+    as.scanl f init = (as.scanlM (m := Id) (pure <| f · ·) init).run := by
+  simp
```
eric-wieser, line 151: I think a more useful version of this is:
```suggestion
@[simp]
theorem idRun_scanlM {f : β → α → Id β} {as: Array α} :
    (as.scanlM f init).run = as.scanl (f · |>.run) init := by
  sorry
```

Batteries/Data/Array/Basic.lean
cmlsharp: Oh I see. What is the advantage of doing this over implemented_by? (In either case right now I don't have a theorem for scanrM but I could do this for scanlM)

fgdorais: I disagree with Eric's proposal: `csimp` with an unsafe def is pointless.

Batteries/Data/Array/Scan.lean
cmlsharp: I believe I use this lemma if not in this PR then in my iterator PR. But I can add this one as well.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,336 @@
...
+theorem scanlM_eq_scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Subarray α} :
+    as.scanlM f init = (as.array.extract as.start as.stop).scanlM f init := by
+  unfold scanlM
+  simp only [Array.extract_scanlM]
+
+theorem scanrM_eq_scanrM_extract [Monad m] [LawfulMonad m] {f : α → β → m β} {as : Subarray α} :
+    as.scanrM f init = (as.array.extract as.stop as.start).scanrM f init := by
+  unfold scanrM
+  simp only [Array.extract_scanrM]
+end Subarray
```
fgdorais, line 372:
```suggestion
namespace Subarray

theorem scanlM_eq_scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Subarray α} :
    as.scanlM f init = (as.array.extract as.start as.stop).scanlM f init := by
  simp only [scanlM, Array.extract_scanlM]

theorem scanrM_eq_scanrM_extract [Monad m] [LawfulMonad m] {f : α → β → m β} {as : Subarray α} :
    as.scanrM f init = (as.array.extract as.stop as.start).scanrM f init := by
  simp only [scanrM, Array.extract_scanrM]

end Subarray
```

Batteries/Data/Array/Scan.lean
cmlsharp: Oh wait, sorry this lemma already exists, but I added simp annotations to it per your suggestion.

Batteries/Data/Array/Scan.lean
cmlsharp: I added these as List.toArray_scanlM and List.toArray_scanrM. Unfortunately, to avoid an import cycle, these theorems are in Batteries.Data.Array.Scan rather than in Batteries.Data.Array.List

Batteries/Data/Array/Scan.lean
```diff
@@ -120,23 +113,18 @@ theorem scanrM_empty [Monad m] {f : α → β → m β} {start stop : Nat} :
 
 theorem scanlM_reverse [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
     as.reverse.scanlM f init = Array.reverse <$> (as.scanrM (flip f) init) := by
-  rw [← scanlM_toList, ← scanrM_toList]
-  simp
+  simp only [← scanlM_toList, ← scanrM_toList, toList_reverse, List.scanlM_reverse, Functor.map_map, List.reverse_toArray]
```
cmlsharp: I noticed several changes like this. just so I understand,  why is this change preferable?

to me, the original version reads much clearer than the long list of simp lemmas (even if the replacement is technically a one-liner)

fgdorais: `rw; simp` and `simp; simp` is always suspicious, so I had to investigate but I didn't bother restoring to the original form.

cmlsharp: Ah that makes sense. I changed some of them back to the original form.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,360 @@
...
+    simp [show start < stop by omega, show stop - (start + 1) = n by omega, ih]
+
+theorem toArray_scanlM_toList [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
+    List.toArray <$> as.toList.scanlM f init = as.scanlM f init := by
+  simp [scanlM, Array.scanlM.loop_toList, ←Array.length_toList]
+
+@[simp, grind =]
+theorem toList_scanlM [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
+    toList <$> as.scanlM f init = as.toList.scanlM f init := by
+  simp [← toArray_scanlM_toList]
```
eric-wieser: Apologies if I said this before; I'm of the opinion that you should drop this first lemma, and keep only the second one (and similarly throughout); I think it's much easier to reason about lemmas of the form `f <$> g x = g (f x)` than it is to have ones of the form `f <$> g (f_inv x) = g x`.

So generally I'd encourage you to remove all the lemmas that mention both `toArray` and `toList` at the same time. I suspect the proofs also end up easier too.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,360 @@
...
+
+theorem scanrM.loop_toList [Monad m] [LawfulMonad m] {f : α → β → m β}
+    {start : Nat} {h : start ≤ as.size} :
+    scanrM.loop f init as start stop h acc =
+      return (← as.toList.drop stop
+                  |>.take (start - stop)
+                  |>.scanrM f init)
+                ++ acc.toList.reverse
+                |>.toArray := by
```
eric-wieser: To me this looks like it would be more natural as
```suggestion
    (scanrM.loop f init as start stop h acc).toList =
      return (← as.toList.drop stop
                  |>.take (start - stop)
                  |>.scanrM f init)
                ++ acc.toList.reverse := by
```

Batteries/Data/Array/Basic.lean
fgdorais: I think there is a legitimate way to use csimp, but I will let Chad decide whether it's worth the effort.

* Define `scanlM` as currently.
* Define `scanlM.loopFast` as `scanlMFast.loop`.
* Prove `scanlM.loop_eq_loopFast` assuming `as.SizeFitsUSize`.
* Define `scanlMImpl` as a switch `if as.SizeFitsUSize then <use loopFast> else <use loop>`.
* Prove the csimp theorem `scanl_eq_scanlMImpl`.

For emphasis, the `else` alternative of `scanlMImpl` can be defined as:
```lean
  letI inst : Inhabited (m (Array β)) := ⟨scanlM.loop ...⟩
  panic! "out of memory"
```
Recall that to the kernel `panic!` is just sugar for `default`, so this is the same as using `scanlM.loop`. Hopefully, the compiler will know that it is not actually necessary to calculate `default` in this case.

Batteries/Data/Array/Scan.lean
cmlsharp: At least for the internal proofs, the proofs do not seem easier to me without this theorem. Usually the thing i want to do is take some statement about `Array.scanlM` and replace instances of `Array.scanlM` with `List.scanlM` so I can apply the corresponding List.scanlM lemma (so really what I end up using more is `← toArray_scanrM_toList`, maybe I should flip the equality around).  Then often times, simp can just take care of the rest.

Batteries/Data/Array/Scan.lean
cmlsharp: This doesn't typecheck:

    (scanrM.loop f init as start stop h acc).toList
    
 It would have to be `Array.toList <$> scanrM.loop f init as start stop h acc`
 
 For me it seemed simpler just to keep all the "inside-the-monad manipulation" on one side of the equation, but if you feel strongly, I can rewrite it.

Batteries/Data/Array/Basic.lean
fgdorais: I just remembered this is about `scanrM` not `scanlM`.

Batteries/Data/Array/Scan.lean
cmlsharp: I think the lemma this is most similar to is `Array.mapM_eq_mapM_toList`. With the equality flipped, these lemmas are directly analogous. I have flipped the equality and renamed it with this naming scheme.

Batteries/Data/Array/Basic.lean
eric-wieser: I don't feel strongly about this; fine to skip the `csimp` approach, indeed the lemma that's already here is probably most important.

Batteries/Data/Array/Scan.lean
eric-wieser: Ok, I'll accept analogy with lemmas in Lean core as an argument!

Regarding the proof strategy, the one I had in mind was applying a lemma that says `toList <$> _` is injective, and then running `simp` to push the `toList`s inside in order to cancel out the `Array.scan`s.

Batteries/Data/Array/Scan.lean
eric-wieser: Indeed I meant this with `<$>`

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+
+/-
+This is guaranteed by the Array docs but it is unprovable.
+May be asserted to be true in an unsafe context via `Array.unsafe_sizeFitsUsize`
+-/
+private abbrev SizeFitsUSize (a : Array α) : Prop := a.size < USize.size
+
+@[grind .]
+private theorem nat_index_eq_usize_index {n : Nat} {a : Array α}
+    {h : a.SizeFitsUSize} {hn : n ≤ a.size} :
```
eric-wieser:
```suggestion
private theorem SizeFitsUSize.toNat_ofNat_eq {n : Nat} {a : Array α}
    (h : a.SizeFitsUSize) (hn : n ≤ a.size) :
```
lets you use dot notation below

Batteries/Data/Array/Scan.lean
cmlsharp: also, when I use this lemma, the thing I want to replace in my goal looks like `scanrM.loop f init as start stop h acc`. I can of course first `apply map_toList_inj.mp` first, but its not clear this is ultimately a simplification

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+    (h_start := by grind only [USize.size_eq, USize.ofNat_eq_iff_mod_eq_toNat, = Nat.min_def])
+    (acc := Array.replicate (start - stop + 1) init)
+    (by grind only [!Array.size_replicate, = Nat.min_def, Array.nat_index_eq_usize_index])
+where
+  @[specialize]
+  loop (f : α → β → m β) (init : β) (as : Array α)
+       (start stop : USize)
+       (h_start : start.toNat ≤ as.size)
+       (acc : Array β)
+       (h_bound : start.toNat - stop.toNat  < acc.size) :
```
eric-wieser: nit:
```suggestion
       (h_bound : start.toNat - stop.toNat < acc.size) :
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,360 @@
...
+    (as.extract stop start).scanrM f init = as.scanrM f init start stop := by
+  rw (occs := [2]) [scanrM]
+  rw [scanrM.loop_toList, scanrM_eq_scanrM_toList, bind_pure_comp]
+  simp_all only [toList_extract]
+  grind [List.take_eq_take_iff, List.drop_eq_drop_iff]
+
+@[simp, grind =]
+theorem scanlM_empty [Monad m] {f : β → α → m β} {start stop : Nat} :
+    #[].scanlM f init start stop = pure #[init] := by
+  simp [scanlM, scanlM.loop]
```
eric-wieser: The naming seems inconsistent between these; I would guess that the first one should be renamed `scanrM_extract` to match most of the other theorems.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,360 @@
...
+  apply toList_inj.mp
+  grind
+
+@[simp]
+theorem scanl_ne_empty {f : β → α → β} : scanl f init as ≠ #[] := by grind
+
+-- This pattern can be removed after moving to a lean version containing
+-- https://github.com/leanprover/lean4/pull/11760
+local grind_pattern Array.eq_empty_of_size_eq_zero => xs.size where
+  guard xs.size = 0
```
eric-wieser: I believe we are now in such a version (at least, if you merge main)

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,360 @@
...
+
+@[simp]
+theorem scanl_iff_empty {f : β → α → β} (c : β) :
+    scanl f init as = #[c] ↔ c = init ∧ as = #[] := by
```
eric-wieser: This looks like it should be called `scanl_eq_singleton_iff` to me.

Batteries/Data/Array/Lemmas.lean
eric-wieser: Generally arguments should be explicit to equality lemmas; and certainly there should be a space before all `:`s

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+Fold a function `f` over the subarray from the left, returning the list of partial results.
+-/
+@[inline]
+def scanl (f : β → α → β) (init : β) (as : Subarray α) : Array β :=
+  as.array.scanl f init (start := as.start) (stop := as.stop)
+
+/--
+Fold a function `f` over the subarray from the left, returning the list of partial results.
+-/
+def scanr (f : α → β → β) (init : β) (as : Subarray α) : Array β :=
```
eric-wieser, line 392:
```suggestion
@[inline]
def scanl (f : β → α → β) (init : β) (as : Subarray α) : Array β :=
  as.array.scanl f init (start := as.start) (stop := as.stop)

/--
Fold a function `f` over the subarray from the left, returning the list of partial results.
-/
@[inline]
def scanr (f : α → β → β) (init : β) (as : Subarray α) : Array β :=
```
unless there's a good reason for this asymmetry.

Batteries/Data/Array/Basic.lean
```diff
@@ -120,27 +121,276 @@ protected def maxI [ord : Ord α] [Inhabited α]
...
+
+/--
+Fold a function `f` over the subarray from the left, returning the list of partial results.
+-/
+@[inline]
+def scanl (f : β → α → β) (init : β) (as : Subarray α) : Array β :=
+  as.array.scanl f init (start := as.start) (stop := as.stop)
+
+/--
+Fold a function `f` over the subarray from the left, returning the list of partial results.
```
eric-wieser:
```suggestion
Fold a function `f` over the subarray from the right, returning the list of partial results.
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,360 @@
+/-
+Copyright (c) 2014 Parikshit Khanna. All rights reserved.
```
eric-wieser: Is this copyright really correct?

fgdorais: The original source was in Lean 3. Probably got ported manually to Std before the great Mathlib port. This copyright notice has important significance in only a few countries anyway.

cmlsharp: no, fixed

Batteries/Data/Array/Scan.lean
cmlsharp: Update I went ahead and just changed it. I don't think it is a big deal, and when doing this I ended up making other simplifications to the proof that I'd have to port if I changed the theorem statement back.

Batteries/Data/Array/Lemmas.lean
cmlsharp: FWIW Array.toList_extract uses implicit args here. What is the general rule?

```
Array.toList_extract.{u_1} {α : Type u_1} {xs : Array α} {start stop : Nat} :
  (xs.extract start stop).toList = xs.toList.extract start stop
```

Batteries/Data/Array/Basic.lean
fgdorais: I feel the same. Feel free to ignore this Chad.

Batteries/Data/Array/Lemmas.lean
fgdorais: For `Eq` lemmas, arguments that can't be inferred are generally explicit. For, `Iff` lemmas all arguments are explicit. (The idea there is to facilitate the use of `.mp`/`.mpr`. I never figured out exactly why the same reasoning doesn't apply to `Eq`. I don't think it makes much difference now that the `(a := x)` notation is available to resolve ambiguities. Before, you had to use `@thm _ _ _ _ ...` which was very annoying.)

Batteries/Data/Array/Scan.lean
eric-wieser: Oh, are these scan functions ported from, inspired by, or otherwise present in Lean 3? If so, the PR description should probably mention that.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,353 @@
...
+  apply map_toList_inj.mp
+  simp [scanrM, Array.scanrM.loop_toList, ← Array.length_toList]
+
+@[simp, grind =]
+theorem toList_scanrM [Monad m] [LawfulMonad m] {f : α → β → m β} {as : Array α} :
+    toList <$> as.scanrM f init = as.toList.scanrM f init := by
+  simp [scanrM_eq_scanrM_toList]
+
+theorem scanlM_extract [Monad m] [LawfulMonad m] {f : β → α → m β} {as : Array α} :
+    (as.extract start stop).scanlM f init  = as.scanlM f init start stop := by
```
eric-wieser:
```suggestion
    (as.extract start stop).scanlM f init = as.scanlM f init start stop := by
```

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,353 @@
...
+    (as.scanl f init)[i]'h = foldl f init (as.take i) := by
+  simp only [scanl_eq_scanl_toList, ← foldl_toList]
+  simp
+
+@[grind =]
+theorem getElem?_scanl {f : β → α → β} :
+    (scanl f a l)[i]? = if i ≤ l.size then some (foldl f a (l.take i)) else none := by
+  grind
+
+@[grind _=_]
```
eric-wieser: Is this different to `grind =`?

Batteries/Data/Array/Lemmas.lean
```diff
@@ -41,6 +41,10 @@ theorem idxOf?_toList [BEq α] {a : α} {l : Array α} :
     (a.eraseIdxIfInBounds i).size = if i < a.size then a.size-1 else a.size := by
   grind
 
+theorem toList_drop (as: Array α) (n : Nat) :
```
eric-wieser, line 44:
```suggestion
@[simp, grind =]
theorem toList_drop (as: Array α) (n : Nat) :
```
to match `toList_erase` above.

cmlsharp: This doesn't work as a simp lemma because simp simplifies '(as.drop n).toList' to 'List.take (as.size - n) (List.drop n as.toList)`

Batteries/Data/Array/Scan.lean
cmlsharp: I don't believe so, changed.

Batteries/Data/Array/Lemmas.lean
cmlsharp: Fwiw I think this is very frustrating. Array.drop is automatically simplified to Array.extract. This over-eager simplification is why I wanted a toList_drop method to begin with.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,353 @@
...
+  rw [← Array.toList_scanrM]
+  simp
+
+theorem toArray_scanl {f : β → α → β} {as : List α} :
+    (as.scanl f init).toArray = as.toArray.scanl f init := by
+  rw [← Array.toList_scanl]
+
+theorem toArray_scanr {f : α → β → β} {as : List α} :
+    (as.scanr f init).toArray = as.toArray.scanr f init := by
+  rw [← Array.toList_scanr]
```
eric-wieser, line 347: I was going to claim that these should be simp, but having found [#lean4 > Direction of List.map_toArray vs Array.toList_map @ 💬](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Direction.20of.20List.2Emap_toArray.20vs.20Array.2EtoList_map/near/571779970) I'm no longer sure.

Batteries/Data/Array/Lemmas.lean
fgdorais: There are way too many simp lemmas for my taste. Eric is very fond of them though and I respect his opinion. We (in general, including Eric and I) need a common language to talk about simp-normal-form. The Mathlib/Batteries concerns are a bit different.

eric-wieser: Thanks for starting [the Zulip thread](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Over.20eager.20simplification.20of.20Array.2Edrop/near/571782698); indeed this can't be `simp` right now.

Batteries/Data/Array/Scan.lean
```diff
@@ -0,0 +1,372 @@
...
+
+theorem scanl_extract {f : β → α → β} {as : Array α} :
+    (as.extract start stop).scanl f init  = as.scanl f init start stop := by
+  unfold scanl
+  rw [scanlM_extract]
+
+theorem scanr_extract {f : α → β → β} {as : Array α} :
+    (as.extract stop start).scanr f init  = as.scanr f init start stop := by
+  unfold scanr
+  rw [scanrM_extract]
```
eric-wieser, line 325: nit: whitespace
```suggestion
theorem scanl_extract {f : β → α → β} {as : Array α} :
    (as.extract start stop).scanl f init = as.scanl f init start stop := by
  unfold scanl
  rw [scanlM_extract]

theorem scanr_extract {f : α → β → β} {as : Array α} :
    (as.extract stop start).scanr f init = as.scanr f init start stop := by
  unfold scanr
  rw [scanrM_extract]
```

Batteries/Data/Array/Lemmas.lean
fgdorais: Can we resolve this or are we hoping to fix the simp issues upstream?

Batteries/Data/Array/Scan.lean
fgdorais: Can we proceed without these simps?

cmlsharp:
leanprover-community/mathlib4#35228

I opened a PR that fixes the issue. The problem was that I added List.reverse_singleton, but that was something that already existed in mathlib. The simple solution is to replace it.

fgdorais:
Thanks! This is not the preferred way to make a Mathlib adaptation PR but I think we can work with it.

The preferred way is to edit the [Mathlib Testing branch](https://github.com/leanprover-community/mathlib4-nightly-testing/tree/batteries-pr-testing-1589) until CI works and then make a PR from that branch into Mathlib. I've added your fix to that branch and we will see what CI says, in case we need to add some additional fixes. If CI clears, then we can just use your Mathlib PR, otherwise we will revert to the preferred route.

@kim-em Could you give @cmlsharp write access to Mathlib Testing. (Is there a way to automate this?)

cmlsharp:
Oh sorry about that! I will do that in the future. I just realized there was a commit I meant to push prior to making the PR (marking this lemma simp since it is marked that way in mathlib). I just pushed it.

PR #1590, opened 2025-12-31
feat: add scan combinator for iterators

cmlsharp:
This PR depends on:
* #1589
* #1588

This PR was originally #1585, but I split the PR it was based on into multiple PRs and wanted to avoid merge conflicts.

This PR builds off of my earlier PRs  and adds a Scan combinator to Iterators as well (the dependence on the earlier PRs is just for the lemmas that prove correspondence between the iterator combinators and their list/array equivalents).

The primary contribution of this PR is IterM.scanM which, given an effectful iterator, an inital value, and an effectful accumulation function, returns an iterator over the partial results of folding that function over the values of the inner iterator. The remaining Iter.Scan/Iter.ScanM/IterM.scan are just implemented in terms of this in the obvious way.

This PR includes instances for both Productive and Finite (if the underlying iterator is Productive, resp. Finite) and includes lemmas showing that (at least for pure iterators / iterators in the identity monad) Iter.scan* is equivalent to accumulating the values into a list (resp. Array) and then calling List.scanl/List.scanlM (resp. Array.scanl/Array.scanlM). I would be happy to continue plugging away at more theorems, but this seems like a minimally useful set.

### DIFF
--- a/Batteries/Data/Iterators/Lemmas/Scan.lean
+++ b/Batteries/Data/Iterators/Lemmas/Scan.lean
@@ -46,7 +46,7 @@ theorem IterM.InternalCombinators.step_scanM
   case false =>
     apply bind_congr
     intro step
-    cases hstep : step.inflate using PlausibleIterStep.casesOn <;> simp
+    cases step.inflate using PlausibleIterStep.casesOn <;> simp
 
 private theorem IterM.toList_scanWithPostCondition_afterInit
     [Monad m] [LawfulMonad m] [Iterator α Id β] [Finite α Id]


Batteries/Data/Iterators/Scan.lean
datokrat, line 1: The iterator and the lemmas look very good to me.

Batteries/Data/Iterators/Scan.lean
```diff
@@ -0,0 +1,266 @@
...
+
+public section
+namespace Std.Iterators.Types
+
+/--
+  Internal state for the ScanM combinator
+-/
+structure ScanM  {β γ : Type w} {n : Type w → Type w''}
+     (α : Type w) (m : Type w → Type w') (f : γ → β → PostconditionT n γ)
+    [Monad m] [Monad n] [MonadLiftT m n] [Iterator α m β] where
```
datokrat: I'm skeptical about parametrizing `ScanM` over the `Iterator` class. Iterator instances usually very canonical, other than (unfortunately) `MonadLiftT`.

It's a trade-off: In the parametrized variant, functions about the iterator don't need to do type class synthesis again because the instances can be obtained by unification. On the other hand, The (already potentially large) type of the iterator becomes even larger.

I think it would be more consistent with the rest of the library if these instances weren't parameters of `ScanM`.

cmlsharp: Thanks for the feedback. Do you think the annotations on e.g. IterM.scanWithPostcondition ought to be there? I notice the standard library iterators have them, though If i remove these annotations from the type they won't be necessary (except for MonadLiftT if thats the only one I leave on the type), and I'll have to explicitly allow unused variables in the batteries linter.

datokrat: Do you mean type annotations or `@[simp]` annotations (or something else)?

Batteries/Data/Iterators/Lemmas/Scan.lean
```diff
@@ -0,0 +1,151 @@
...
+          | .done hp =>
+            return .deflate <|
+              .done (.done (by simp [IterM.InternalCombinators.scanM, h]) hp)) := by
+  cases h: yieldAcc
+  case true => rfl
+  apply bind_congr
+  intro step
+  cases step.inflate using PlausibleIterStep.casesOn <;> rfl
+
+private theorem IterM.toList_scanWithPostCondition_afterInit
```
datokrat, line 51: I just noticed that these lemmas use upper-case C `scanWithPostCondition` while `IterM.scanWithPostcondition` (correctly) uses a lower-case c.

cmlsharp: fixed!

Batteries/Data/Iterators/Scan.lean
cmlsharp: I mean for example an `[Iterator α m β]` constraint on `IterM.scanWithPostcondition` or `[Iterator α m β]  [Monad n]` on `IterM.scanM`. `IterM.filterMapWithPostcondition` and `IterM.filterMapM` respectively have these instance parameters, but they are technically unused (and would be in the scan equivalents if the `ScanM` structure does not have these parameters), so the Batteries linter rejects them.

datokrat: I think these instance parameters aren't necessary and perhaps I should take a look at `FilterMap` again and remove them :-) Another possibility for, say, `Monad n` ist to make it implicit like `{_ : Monad}` that is then inferred based on the iterator's type -- if needed in the declaration. Otherwise, no need to repeat it, I think.

PR #1595, opened 2026-01-03
feat: add Alternative.{many,many1}

414owen:
Closes https://github.com/leanprover-community/batteries/issues/1577

### DIFF
--- a/.github/workflows/nightly_detect_failure.yml
+++ b/.github/workflows/nightly_detect_failure.yml
@@ -74,12 +74,18 @@ jobs:
         import zulip
         client = zulip.Client(email=os.getenv('ZULIP_EMAIL'), api_key=os.getenv('ZULIP_API_KEY'), site=os.getenv('ZULIP_SITE'))
 
-        # Get the last message in the 'status updates' topic
+        # Get the last message from the bot in the 'status updates' topic.
+        # We narrow by sender to ignore human replies in between.
+        bot_email = 'github-mathlib4-bot@leanprover.zulipchat.com'
         request = {
           'anchor': 'newest',
           'num_before': 1,
           'num_after': 0,
-          'narrow': [{'operator': 'stream', 'operand': 'nightly-testing'}, {'operator': 'topic', 'operand': 'Batteries status updates'}],
+          'narrow': [
+            {'operator': 'stream', 'operand': 'nightly-testing'},
+            {'operator': 'topic', 'operand': 'Batteries status updates'},
+            {'operator': 'sender', 'operand': bot_email}
+          ],
           'apply_markdown': False    # Otherwise the content test below fails.
         }
         response = client.get_messages(request)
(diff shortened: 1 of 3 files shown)


fgdorais:
Please add a copyright header and follow style from similar files.

fgdorais:
I made a few minor edits, mostly so that CI works. I hope you don't mind, it was easier to do than to explain.

PS: Avoid force push.

414owen:
> I hope you don't mind, it was easier to do than to explain.

Absolutely fine. Be my guest.

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,31 @@
...
+module
+
+namespace Alternative
+
+/--
+Zero or more Alternatives.
+For example, given a Parser, which implements Alternative, you can use the `many` combinator
+to parse zero or more items.
+-/
+partial def many [Alternative f] [Inhabited (f (List α))]
```
eric-wieser: I'd suggest eliminating this argument with
```lean
private partial def many.aux [Alternative f] [Nonempty (f (List α))]
    (p : f α) : f (List α) :=
  List.cons <$> p <*> many.aux p <|> pure []

/--
Zero or more Alternatives.
For example, given a Parser, which implements Alternative, you can use the `many` combinator
to parse zero or more items.
-/
partial def many [Alternative f] (p : f α) : f (List α) :=
  letI : Nonempty (f (List α)) := ⟨pure []⟩
  many.aux p
```

414owen: That's neat.

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,36 @@
...
+    (p : f α) : f (List α) :=
+  letI : Nonempty (f (List α)) := ⟨pure []⟩
+  List.cons <$> p <*> many.aux p <|> pure []
+
+/--
+One or more Alternatives.
+For example, given a Parser, which implements Alternative, you can use the `many1` combinator
+to parse one or more items.
+-/
+def many1 [Alternative f] [Inhabited (f (List α))]
```
eric-wieser:
```suggestion
def many1 [Alternative f]
```

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,36 @@
...
+to parse one or more items.
+-/
+def many1 [Alternative f]
+    (p : f α) : f (Σ n, Vector α (1 + n)) :=
+  let g x xs := ⟨xs.1, Vector.singleton x ++ xs.2⟩
+  let toVec (l : List α) : Σ n, Vector α n :=
+    let arr := List.toArray l
+    ⟨arr.size, ⟨arr, rfl⟩⟩
+  let manyVec : f (Σ n, (Vector α n)) := toVec <$> Alternative.many p
+  g <$> p <*> manyVec <|> Alternative.failure
```
eric-wieser: Happy to wait for pending Zulip discussion; but the two `many1` functions in Lean already are the much simpler
```suggestion
def many1 [Alternative f] (p : f α) : f (List α) :=
  List.cons <$> p <*> many p
```
in the absence of a concrete use case requiring the complexity, I'd be inclined to pick this much simpler option.

eric-wieser: (note that independently, the `<|> Alternative.failure` looks like a no-op to me)

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,36 @@
...
+
+/--
+Zero or more Alternatives.
+For example, given a Parser, which implements Alternative, you can use the `many` combinator
+to parse zero or more items.
+-/
+partial def many [Alternative f]
+    (p : f α) : f (List α) :=
+  letI : Nonempty (f (List α)) := ⟨pure []⟩
+  List.cons <$> p <*> many.aux p <|> pure []
```
eric-wieser: No need to repeat the logic here:
```suggestion
def many [Alternative f] (p : f α) : f (List α) :=
  letI : Nonempty (f (List α)) := ⟨pure []⟩
  many.aux p
```

Batteries/Control/Alternative.lean
414owen: > def many1 [Alternative f] (p : f α) : f (List α) :=
>  List.cons <$> p <*> many p

I'm not a huge fan of this version to be honest. It means that, even though you know there's at least one element, operations like `head`, and `tail` are partial.

It throws away useful information, and goes against "make error states unrepresentable" / "parse, don't validate", etc.

eric-wieser: True, but my proposal is consistent with [Std.Internal.Parsec.many1](https://leanprover-community.github.io/mathlib4_docs/Std/Internal/Parsec/Basic.html#Std.Internal.Parsec.many1) from the standard library.

414owen: That is something I can't deny :)

fgdorais:
Please edit the `lean-toolchain` file to `leanprover/lean4-pr-releases:pr-release-12041`. That will fix the missing `Nonempty` instance. We can do a forward-port later if necessary.

fgdorais:
I've rebased this PR on `nightly-testing` instead of `main`. This is of no consequence for the most part but there may be a bit more background noise. Ignore any noise that isn't related to your code and I will filter out the rest.

Batteries/Control/Alternative.lean
fgdorais: Please use Eric's proposal for `many1`

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,35 @@
...
+private partial def many.aux [Alternative f] [Nonempty (f (List α))]
+    (p : f α) : f (List α) :=
+  List.cons <$> p <*> many.aux p <|> pure []
+
+/--
+Zero or more Alternatives.
+For example, given a Parser, which implements Alternative, you can use the `many` combinator
+to parse zero or more items.
+-/
+partial def many [Alternative f] (p : f α) : f (List α) :=
```
fgdorais: The missing `Nonempty` instance is fixed upstream. Please replace the definition of `many` by the obsolete `many.aux`.

eric-wieser: Which PR was it fixed in?

fgdorais: This PR is now based on the toolchain for [lean4#12041](https://github.com/leanprover/lean4/pull/12041). We will do a forward port if this PR gets merged before the upstream one.

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,31 @@
...
+module
+
+namespace Alternative
+
+/--
+Zero or more Alternatives.
+For example, given a Parser, which implements Alternative, you can use the `many` combinator
+to parse zero or more items.
+-/
+private partial def many [Alternative f]
```
fgdorais:
```suggestion
partial def many [Alternative f]
```

Batteries/Control/Alternative.lean
```diff
@@ -0,0 +1,31 @@
+/-
+Copyright (c) 2026 Owen Shepherd. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Owen Shepherd
+-/
+module
+
```
fgdorais, line 7:
```suggestion

public section

```

lean-toolchain
```diff
@@ -1 +1 @@
-leanprover/lean4:nightly-2026-01-28
+leanprover/lean4-pr-releases:pr-release-12041
```
eric-wieser, line 1: Instead of doing this and waiting one or two months for https://github.com/leanprover/lean4/issues/12041, why not include

```
private local instance [Pure m] [Nonempty α] : Nonempty (m α) :=
  ‹Nonempty α›.rec (⟨pure ·⟩)
```
in this PR for now, @fgdorais?

fgdorais: Yes, I'm planning on a forward port once this PR is ready to merge.

PR #1600, opened 2026-01-05
fix: correct behavior of `runLinter --update`

thorimur:
This PR changes the behavior of `runLinter --update`. Previously, this would overwrite the nolints.json file with the full list of current linting violations, meaning that all declarations which violated linters were now nolinted.

However, it is intended to only "remove any declarations that no longer need
to be nolinted" (at least according to the docstring). This means that we should just filter out declarations *from the nolints file* which no longer fail (i.e., only include those from the nolints file which *do* still fail).

Since technically we run linters on individual modules, but the nolints.json file may apply to other modules besides the one being linted, we also only filter out those newly non-failing declarations which are from the current module and therefore are actually being linted.

This extra check can introduce some potential (benign) staleness in nolints.json if a nolinted declaration is removed (and thus never appears in any module, and so is never removed by `--update`). However, these ineffective nolints will not cause any problems, while overzealously deleting nolints from other modules might.

In the future, we might want to consider factoring out the `--update` behavior so that it is not part of `runLinterOnModule`, but instead occurs at the top level in `main`.

---

If possible, I'd like someone to verify that this is indeed the intended behavior of `--update`, and that it's not the case that the docstring is wrong. :)

### DIFF
--- a/scripts/runLinter.lean
+++ b/scripts/runLinter.lean
@@ -157,11 +157,18 @@ unsafe def runLinterOnModule (cfg : LinterConfig) (module : Name) : IO Unit := d
     let linters ← getChecks (slow := true) (runAlways := none) (runOnly := none)
     let results ← lintCore decls linters (inIO := true) (currentModule := module)
     if updateNoLints then
-      traceLint s!"Updating nolints file at {nolintsFile}" (inIO := true) (currentModule := module)
-      writeJsonFile (α := NoLints) nolintsFile <|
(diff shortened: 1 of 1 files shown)


thorimur:
On [zulip](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/runLinter.20--update/with/566425581), it was mentioned that the current behavior for adding new violations to the nolint file is also useful (or possibly the primary use case!). (Note that the current behavior still needs a tweak because it overeagerly erases nolints that are not from the current module.)

I think both functionalities have a place; maybe `--cleanup` for this one?

Issue #1442, opened 2025-10-02
Verify Binary Heap


fgdorais:
This is a standard textbook algorithm. It is well implemented in [Batteries.Data.BinaryHeap](https://github.com/leanprover-community/batteries/blob/main/Batteries/Data/BinaryHeap.lean) but not verified. 

If you're looking to get started with formal verification project in Lean, this should be an interesting one to try!

fgdorais:
Yes, this is also needed. It's a common pattern in Batteries to provide a bare-bones and therefore highly flexible implementation, with the expectation that users will use this as part of the internal state of a more sophisticated API. In this case, at least two basic examples for min-heap and max-heap should be provided (using `LE`/`LT` or `Ord` classes, for example).

fgdorais:
> I think I know what you want here, but I think a good first step would be to work out what proofs are actually wanted.

The basic theorems needed are that the functions provided preserve the max-heap property for the provided relation. The difficulty is that the proofs should ideally make minimal use of the expected properties of the relation.

cmlsharp:
I started some initial work on this. It's not nearly ready for a PR, but I have managed to verify the basic heap preserving properties of `heapifyDown` and `mkHeap`.  It relies on `Std.TransOrd`, `Std.ReflOrd` and `Std.OrientedOrd`.

Progress here: https://github.com/cmlsharp/batteries/blob/binomial_heap/Batteries/Data/BinaryHeap/Lemmas.lean

PR #1602, opened 2026-01-06
feat: verification of binary heap

cmlsharp:
This PR is still a work in progress, but I wanted to get it out there.

This PR begins the process of addressing #1442. Thus far I have managed to prove that `heapifyDown`, `heapifyUp`, and `mkHeap` preserve the binary heap correctness property, These should be the lions share of the difficulty, as the remaining operations are largely wrappers around the former two.

As suggested in that thread, this PR also alters the definition of `BinaryHeap` to use an `Ord` instance on the contained type rather than  a comparison function. The proofs additionally rely on `Std.OrientedOrd` and `Std.TransOrd`.  

The core correctness property is encoded as:

```lean
@[expose]
public def WF.children [Ord α] (a : Vector α sz) (i : Fin sz) : Prop :=
  let left := 2 * i.val + 1
  let right := left + 1
  (∀ _ : left < sz, compare a[i] a[left] |>.isGE) ∧
  (∀ _ : right < sz, compare a[i] a[right] |>.isGE)


@[expose]
public def WF [Ord α] (v : Vector α sz) : Prop :=
  ∀ i : Fin sz, WF.children v i
```

However, for proving `heapifyUp` it was more convenient to use a "bottom up" version of a correctness property

```lean
def WF.parent [Ord α] (a : Vector α sz) (i : Fin sz) : Prop :=
  ∀ _ : 0 < i.val, compare a[i] a[(i.val - 1)/2] |>.isLE
  
def WF.bottomUp [Ord α] (v : Vector α sz) : Prop :=
  ∀ i : Fin sz, WF.parent v i
```
I provide a theorem that `WF` is equivalent to `WF.bottomUp`.


cmlsharp:
WIP

Batteries/Data/BinaryHeap/WF.lean
```diff
@@ -0,0 +1,363 @@
...
+@[expose]
+def WF.TopDown [Ord α] (v : Vector α sz) : Prop :=
+  ∀ i : Fin sz, WF.Children v i
+
+/-- The well-formedness property of a binary heap -/
+@[expose]
+def WF [Ord α] (heap : BinaryHeap α) : Prop :=
+  WF.TopDown heap.vector
+
+instance [Ord α] {a : Vector α sz} {i : Fin sz} : Decidable (WF.Children a i) :=
```
fgdorais: `WF.Children` could be defined as `Bool` to avoid using `Decidable`.

Batteries/Data/BinaryHeap/WF.lean
```diff
@@ -0,0 +1,363 @@
...
+  | left : InSubtree root k → InSubtree root (2 * k + 1)
+  | right : InSubtree root k → InSubtree root (2 * k + 2)
+
+namespace InSubtree
+
+@[simp]
+theorem le (ins : InSubtree j k) : j ≤ k := by
+  induction ins with omega
+
+theorem not_of_lt (hlt : k < j): ¬InSubtree j k := by
```
fgdorais:
```suggestion
theorem not_of_lt (hlt : k < j) : ¬InSubtree j k := by
```

Batteries/Data/BinaryHeap/WF.lean
```diff
@@ -0,0 +1,363 @@
...
+      exact this ▸ .right (ih (n / 2) (by omega))
+
+/-- A child index is in the subtree of its parent. -/
+theorem of_child (h : j = 2 * i + 1 ∨ j = 2 * i + 2) : InSubtree i j := by
+  rcases h with h | h <;> (subst h; constructor; exact .refl)
+
+end InSubtree
+
+namespace WF
+variable {α : Type w} {sz : Nat}
```
fgdorais:
```suggestion

```

Batteries/Data/BinaryHeap/WF.lean
```diff
@@ -0,0 +1,363 @@
...
+    (compare a[j] a[k]).isGE := by
+  induction hsub
+  case refl => grind only [Ordering.isGE]
+  all_goals
+    have ⟨hwf_m, _⟩ : WF.Children a ⟨‹_›, by omega⟩ := by
+      grind only [WF.Below, InSubtree.not_of_lt]
+    grind only [= Fin.getElem_fin, !Std.TransOrd.isGE_trans]
+
+/-- Parent dominates all descendants after setting a smaller value -/
+theorem parent_ge_subtree_of_set [Ord α] [Std.TransOrd α] [Std.OrientedOrd α]
```
fgdorais: `Std.TransOrd` implies `Std.OrientedOrd`.
```suggestion
theorem parent_ge_subtree_of_set [Ord α] [Std.TransOrd α]
```

Batteries/Data/BinaryHeap/WF.lean
```diff
@@ -0,0 +1,363 @@
...
+    (compare a[0] a[k]).isGE :=
+  parent_ge_subtree
+    (hwf_at := hwf ⟨0, hne⟩)
+    (hwf_below := fun j _ => hwf j)
+    (hsub := InSubtree.mem_of_zero k.val)
+
+/-- "Dual" correctness property to WF.children. A node should be <= its parent
+    Used when verifying heapifyUp -/
+def Parent [Ord α] (a : Vector α sz) (i : Fin sz) : Prop :=
+  ∀ _ : 0 < i.val, compare a[i] a[(i.val - 1)/2] |>.isLE
```
fgdorais: I would use
```suggestion
def Parent [Ord α] (a : Vector α sz) (i : Fin sz) : Prop :=
  ∀ _ : 0 < i.val, compare a[(i.val - 1)/2] a[i] |>.isGE
```
That will require several other adjustments below.

The purpose is to avoid `Std.OrientedOrd` without `Std.TransOrd`. It's best to think of `Std.OrientedOrd` as if it was an implementation detail for `Std.TransOrd`, but don't be worried using it when it is *exactly* what's needed.

fgdorais (commented):
This is very good work! The review will surely take several steps. This is what I had time for today. 

Key outtakes: 

* In Batteries, `variable` is discouraged but auto-implicits are encouraged. (The reverse is true in Mathlib.)
* In Batteries, `Bool` is preferable over `Prop` + `Decidable`. This is because `Decidable` is a class so it is not possible to fully control which decision algorithm is being used.
* Batteries also prefers comment headers over `section`s.

cmlsharp:
FWIW I do not ever use the fact that WF is decidable. Leaving it a prop makes the theorem statements nicer. Should I just remove the decidable instance?

Batteries/Data/BinaryHeap/WF.lean
cmlsharp: I do not actually use the fact that WF.Children is decidable, so I will just get rid of the instances entirely.

cmlsharp:
I made a bunch of simplifications to a few of the lemmas, and factored out some duplicated shared structure between them.

PR #1608, opened 2026-01-07
feat(Data/BitVec): add lemmas about `pow` on `BitVec`

eric-wieser:


linesthatinterlace:
Do we want the `zpow` versions of this?

Batteries/Data/BitVec/Lemmas.lean
```diff
@@ -15,6 +15,13 @@ public import Batteries.Data.Int
 
 namespace BitVec
 
+@[simp]
+theorem toNat_pow (b : BitVec w) (n : Nat) : (b ^ n).toNat = (b.toNat ^ n) % (2 ^ w) := by
+  induction n <;> simp_all [Lean.Grind.Semiring.pow_succ]
+
+theorem ofNat_pow (n : Nat) (x d : Nat) : BitVec.ofNat n (x ^ d) = BitVec.ofNat n x ^ d := by
```
fgdorais: I wish `BitVec` functions were more consistent using `w` for bit width.
```suggestion
theorem ofNat_pow (w x d : Nat) : BitVec.ofNat w (x ^ d) = BitVec.ofNat w x ^ d := by
```
Why isn't this simp?

eric-wieser: Now `simp`, I don't recall why I didn't.

PR #1619, opened 2026-01-16
feat: linter accepts multiple modules to lint

GZGavinZhao:
Since the linter can lint multiple targets when there are multiple `defaultTargets`, it feels reasonable to also make it accept multiple modules on the command-line.

This is particularly useful in the CI when I want to lint multiple targets that are not part of `defaultTargets`.

### DIFF
--- a/scripts/runLinter.lean
+++ b/scripts/runLinter.lean
@@ -180,7 +180,7 @@ unsafe def runLinterOnModule (cfg : LinterConfig) (module : Name) : IO Unit := d
       IO.println s!"-- Linting passed for {module}."
 
 /--
-Usage: `runLinter [--update] [--trace | -v] [--no-build][Batteries.Data.Nat.Basic]...`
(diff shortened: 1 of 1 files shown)


scripts/runLinter.lean
```diff
@@ -185,9 +180,9 @@ unsafe def runLinterOnModule (cfg : LinterConfig) (module : Name) : IO Unit := d
       IO.println s!"-- Linting passed for {module}."
 
 /--
-Usage: `runLinter [--update] [--trace | -v] [--no-build] [Batteries.Data.Nat.Basic]`
+Usage: `runLinter [--update] [--trace | -v] [--no-build][Batteries.Data.Nat.Basic]...`
```
fgdorais:
```suggestion
Usage: `runLinter [--update] [--trace | -v] [--no-build] [Batteries.Data.Nat.Basic]...`
```

PR #1643, opened 2026-01-30
feat: scan combinators for Vectors

cmlsharp:
This depends on #1589 

In conjunction with #1590 and #1589, this should conclude the project of bringing the 'scan' combinator to parity across List, Array, iterator, and Vector.

This takes a basically analogous approach to Array. In principle Array could be implemented in terms of Vector, except the Vector methods do not have the optional 'start' 'stop' parameters (which also keeps the return type of these functions much simpler). This is inline with the standard library's Vector methods.


cmlsharp:
Done. right now some of the usize logic in Vector is duplicated with that in Array. Maybe there should be some shared Batteries.Data.USize for these theorems?

fgdorais:
Yes, please make a quick PR to add the `USize` lemmas you need to `Batteries.Data.UInt`.

PR #1700, opened 2026-02-25
chore: deprecate String.containsSubstr in favour of String.contains

kim-em:
This PR deprecates `String.containsSubstr` in favour of `String.contains`, which is now available in Lean core as a generic pattern-based function that works with strings, substrings, and chars.

The other Batteries functions in the same file (`String.findSubstr?` and `String.findAllSubstr`) don't have direct Lean core equivalents yet (`String.Slice.find?` returns `Option Pos` rather than `Option Substring.Raw`, and there's no `findAll`), so they are left as-is for now.

🤖 Prepared with Claude Code


fgdorais (commented):
If I understand correctly, the all-default `String.contains` uses (core's implementation of) KMP but that depends on using the appropriate class instances. This file is about using the (batteries' implementation of) KMP. It doesn't matter much which implementation is used but the replacement should always use KMP since that is the point of this file. At least, the deprecation string should indicate how to ensure that KMP is used.

kim-em:
> If I understand correctly, the all-default `String.contains` uses (core's implementation of) KMP but that depends on using the appropriate class instances. This file is about using the (batteries' implementation of) KMP. It doesn't matter much which implementation is used but the replacement should always use KMP since that is the point of this file. At least, the deprecation string should indicate how to ensure that KMP is used.

@fgdorais, could you suggest your preferred update to the deprecation string?

fgdorais:
I'm sorry to answer with a question: is there a simple way to force `String.contains` to use KMP?

kim-em:
Sorry, I'm really confused, @fgdorais, `String.contains` always uses KMP unless you go out of your way to disable instances or something. Do you have an example where `String.contains` is behaving incorrectly?

fgdorais (approved):
Shouldn't there be a `String.Slice.contains` companion to this?

PR #1717, opened 2026-03-14
feat: add a syntax linter version of the impossibleInstance and nonClassInstance linters 

MoritzBeroRoos:
Currently we have the `impossibleInstance` and `nonClassInstance` linters which are environment linters in batteries.
- `impossibleInstance` warns when it detects an instance with arguments that are not inferable for typeclass synthesis.
- `nonClassInstance` warns when a non-class-valued function is declared as `instance`.
 
These linters are especially useful for beginners who don't yet understand how typeclass synthesis works.
But since these linters currently are environment linters they only fire in CI or when manually using `#lint`.
However beginners don't typically have CI set up for their pet projects and mostly don't know about `#lint`.

This PR adresses this issue by adding syntax linter versions of these two linters.
These will automatically check every declaration and thus be of much more use to beginners.


Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+      ts.foldl (init := acc) (go ctx?)
+    | hole _ => acc
+
+
+
+namespace Batteries.Linter
+open Lean Elab Command Linter Std Meta
+
+/-- Option for turning the `impossibleInstance'` linter on and off. -/
+register_option linter.impossibleInstance' : Bool := {
```
fgdorais: Using `'` will cause issues when using from the command line/shell scropts/etc. Please use a clearer name. Same for other names below.

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+instance {α β : Type} [Group α] : Mul α where ...
+```
+This is a syntax linter, i.e. it runs on your declarations as you write them.
+-/
+def impossibleInstance' : Linter where run cmdSyntax := do
+  unless Linter.getLinterValue linter.impossibleInstance' (← Linter.getLinterOptions) do
+    return
+  /- todo use `withSetOptionIn` after `https://github.com/leanprover/lean4/pull/11313` has
+     been resolved, to allow disabling this linter with
+     `set_option linter.impossibleInstance false in`. -/
```
fgdorais: Do you mean `linter.impossibleInstance'` here?

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+    if not (← liftTermElabM (isClass? constInfo.type)).isSome then continue
+    /- Now the actual linting check: -/
+    let some lintmessage ← liftTermElabM (test name) | continue
+    /- Use the range that actually corresponds to the `name` not to the whole mutual block: -/
+    Linter.logLint linter.impossibleInstance' stx lintmessage
+  return
+initialize addLinter impossibleInstance'
+
+/-- Option for turning the `nonClassInstance'` linter on and off. -/
+register_option linter.nonClassInstance' : Bool := {
```
fgdorais: Same issue with `'`.

fgdorais (changes requested):
This is nice, thanks!

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: What if we rename the Environment versions to `impossibleInstanceEnv` and give the nonprime names to the syntax versions?
Any user is much more likely to encounter the syntax Linters and all the names i can come up with are a bit confusing.
E.g. `Note: This linter can be disabled with set_option linter.impossibleInstanceSyntax false` sounds like the user did something syntactically wrong.

fgdorais: Such a change would require a complicated deprecation scheme. Besides, `impossibleInstance` is a poor description of what this linter actually does.

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+end Batteries.Tactic.Lint
+end environmentLinters
+
+
+
+section StandardLinters
+
+/-- This function searches laterally to find the top level names of declarations (along with their
+    syntax). Thanks Thomas Murrills!
```
fgdorais: Does not belong in a docstring.
```suggestion
    syntax).
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
     return none
+
+end Batteries.Tactic.Lint
+end environmentLinters
+
+
+
+section StandardLinters
+
+/-- This function searches laterally to find the top level names of declarations (along with their
```
fgdorais: What does "laterally" mean here?

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+register_option linter.impossibleInstance' : Bool := {
+  defValue := true
+  descr := "Warn when an instance is found that can never be synthesized by typeclass synthesis."
+}
+
+
+/--
+Lints for instances with arguments that cannot be filled in, like
+```
+instance {α β : Type} [Group α] : Mul α where ...
```
fgdorais: Let's give a non-mathlib example. Note that all docstring examples should ideally also be in a test file.

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+This is a syntax linter, i.e. it runs on your declarations as you write them.
+-/
+def impossibleInstance' : Linter where run cmdSyntax := do
+  unless Linter.getLinterValue linter.impossibleInstance' (← Linter.getLinterOptions) do
+    return
+  /- todo use `withSetOptionIn` after `https://github.com/leanprover/lean4/pull/11313` has
+     been resolved, to allow disabling this linter with
+     `set_option linter.impossibleInstance' false in`. -/
+  let errorsFound1 := m!"This instance has at least one argument that is impossible \
+    to infer for typeclass inference. Specifically\n"
```
fgdorais: Do you mean this?
```suggestion
  let errorsFound1 := m!"This instance has at least one argument that cannot be \
    inferred using typeclass synthesis. Specifically\n" 
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  let errorsFound2 := m!"\nThese are arguments that are not instance-implicit and \
+    appear neither in another instance-implicit argument nor the return type, so they can't \
+    be filled in by typeclass inference."
```
fgdorais:
```suggestion
  let errorsFound2 := m!"\nThese are arguments that are not instance-implicit and \
    appear neither in another instance-implicit argument nor the return type, so they cannot \
    be inferred using typeclass synthesis."
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,141 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+
+/--
+Lints for instances with arguments that cannot be filled in, like
+```
+instance {α β : Type} [Group α] : Mul α where ...
+```
+This is a syntax linter, i.e. it runs on your declarations as you write them.
+-/
+def impossibleInstance' : Linter where run cmdSyntax := do
```
fgdorais: This linter does not do exactly the same as the environment linter. Please align them.

Batteries/Tactic/Lint/TypeClass.lean
fgdorais: My personal preference would be `linter.environment.*` and `linter.syntax.*` to distinguish the two types of linters.

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: i added a few sentences to the doc to make it more clear and removed the word laterally

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: I changed it to the example already in `BatteriesTest/lintTC.lean`.
I ommitted the `local` in the doc version, since it has nothing to do with the example.
I presume the local is in the test file to not pollute the testing environment?

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: i added `.syntax` to the syntax version. Since they live in different namespaces already i realize we can have them both be `impossibleInstance`.
So now we would have:
`Batteries.Tactic.Lint.impossibleInstance` (env)
and 
`Batteries.Linter.syntax.impossibleInstance` (syntax).
Not sure why the environment version is in `Tactic`?

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: The syntax version is in my opinion as close as possible to the code of the environment linter.
The reported error messages are different because they are viewed from entirely different contexts.

All other additional code in the syntax linter is only there to obtain the correct name (multiple names in case of mutual blocks), which is quite tricky to do for a syntax linter since the input is only a `Syntax`.
Since the syntax version fires on the syntax of the mutual block instead of the syntax of each declaration in the mutual block the environment version does a loop on all found names additionally.

The environment linter already gets as input a single `Name` and runs its code for that declaration.

The run code on the obtained name is already identical (modulo the exact error messages) and im not sure what more can be done

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  /- We do the check for each (different) top level instance name we can get from the infotrees.
+     Mostly this will only be one name, but for `mutual` blocks this will be more. -/
+  let mut names : List (Name × Syntax) := []
+  for t in ←getInfoTrees do
+    names := t.getTopLevelDeclsByBody ++ names
+  /- Each name shall only be checked once, so we remove duplicates. -/
+  names := names.pwFilter (fun (a, _) (b, _) => a != b)
+  /- the `getTopLevelDeclsByBody` function picks up some internal names from `examples` that it
+     probably shouldn't. We filter these here. -/
+  names := (← names.filterM (fun (name, _) => return (← getEnv).contains name))
```
Vierkantor: The `for` loop can become a `flatMap`, and then we can make the variable immutable.

Also, we should do the `getEnv` outside of the `filter` loop:

```suggestion
  let names := (← getInfoTrees).toList.flatMap (·.getTopLevelDeclsByBody)
  /- Each name shall only be checked once, so we remove duplicates. -/
  let names := names.pwFilter (fun (a, _) (b, _) => a != b)
  /- the `getTopLevelDeclsByBody` function picks up some internal names from `examples` that it
     probably shouldn't. We filter these here. -/
  let env ← getEnv
  let names := names.filter fun (name, _) => env.contains name
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
```
Vierkantor:
```suggestion
  let names := (← getInfoTrees).toList.flatMap (·.getTopLevelDeclsByBody)
  /- Each name shall only be checked once, so we remove duplicates. -/
  let names := names.pwFilter (fun (a, _) (b, _) => a != b)
  /- the `getTopLevelDeclsByBody` function picks up some internal names from `examples` that it
     probably shouldn't. We filter these here. -/
  let names := (← names.filterM (fun (name, _) => return (← getEnv).contains name))
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+/-- This function searches the infotree to find the top level names of declarations
+    (along with their syntax).
+    If it encounters a `.ofCustomInfo` node carrying a `BodyInfo` value it tries to adds the
+    contained name and syntax to the result and stops searching this subtree.
+    If it encounters any other node, the search continues in each subtree and results are merged.
+    Note: This function picks up some internal names from e.g. `examples` (like `bla._example`)
+    that it probably shouldn't. You should probably filter these using `(← getEnv).contains name`.
+-/
+partial def Lean.Elab.InfoTree.getTopLevelDeclsByBody : InfoTree → List (Name × Syntax) :=
```
Vierkantor: How about turning this into a `NameMap`, so that duplicates are already automatically erased?
```suggestion
partial def Lean.Elab.InfoTree.getTopLevelDeclsByBody : InfoTree → NameMap Syntax :=
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+section StandardLinters
+
+/-- This function searches the infotree to find the top level names of declarations
+    (along with their syntax).
+    If it encounters a `.ofCustomInfo` node carrying a `BodyInfo` value it tries to adds the
+    contained name and syntax to the result and stops searching this subtree.
+    If it encounters any other node, the search continues in each subtree and results are merged.
+    Note: This function picks up some internal names from e.g. `examples` (like `bla._example`)
+    that it probably shouldn't. You should probably filter these using `(← getEnv).contains name`.
```
Vierkantor: I am okay with this function not doing the filtering (so that it can avoid taking a `Environment`/living in a `MonadEnv`). But we do need to do it twice downstream. So let's make a function that does this, along the lines of:

```suggestion
/-- `getInfoTreesDecls` returns the top-level names and syntax declarations made by the current command,
based on `BodyInfo` nodes in the infotree data.

Each name is only returned once.
This function filters out declarations appearing in the infotree that do not appear in the environment.
-/
partial def Lean.Elab.getInfoTreesDecls : Command.CommandElabM (List (Name × Syntax)) := do
  let names := (← getInfoTrees).toList.flatMap (·.getTopLevelDeclsByBody)
  /- Each name shall only be checked once, so we remove duplicates. -/
  let names := names.pwFilter (fun (a, _) (b, _) => a != b)
  /- the `getTopLevelDeclsByBody` function picks up some internal names from `examples` that it
     probably shouldn't. We filter these here. -/
  let env ← getEnv
  return names.filter fun (name, _) => env.contains name
```

(But returning a `NameMap` if you agree with the suggestion below.)

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+
+
+section StandardLinters
+
+/-- This function searches the infotree to find the top level names of declarations
+    (along with their syntax).
+    If it encounters a `.ofCustomInfo` node carrying a `BodyInfo` value it tries to adds the
+    contained name and syntax to the result and stops searching this subtree.
+    If it encounters any other node, the search continues in each subtree and results are merged.
```
Vierkantor: I have rewritten the docstring a bit, since that is my main job currently :) Generally we want the first line of the docstring to be a single sentence starting with the declaration name. We should focus on the result of the declaration, not the method by which we achieve it. (See also [the Lean documentation style guide](https://github.com/leanprover/lean4/blob/master/doc/style.md).)

```suggestion
/-- `getTopLevelDeclsByBody tree` returns the top level names of declarations (along with their syntax) which have been logged in the infotree `tree`.

Specifically, this function collects the contained names and syntax for all nodes with a `BodyInfo` value.
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+where
+  go (ctx? : Option ContextInfo) (acc : List (Name × Syntax)) : InfoTree → List (Name × Syntax)
+    | context ctx t => go (ctx.mergeIntoOuter? ctx?) acc t
+    | node i ts => Id.run do
+      if let .ofCustomInfo i := i then
+        if i.value.typeName == ``Lean.Elab.Term.BodyInfo then
+          if let some decl := ctx?.bind (·.parentDecl?) then
+            return (decl, i.stx) :: acc -- don't descend into `ts`
+      ts.foldl (init := acc) (go ctx?)
+    | hole _ => acc
```
Vierkantor: More for future reference: we could also use `foldInfo` here, although it doesn't short-circuit like your version does. So I think your approach is better.
```lean
  t.foldInfo (init := []) fun ctx i ts => Id.run do
    if let .ofCustomInfo i := i then
      if i.value.typeName == ``Lean.Elab.Term.BodyInfo then
        if let some decl := ctx.parentDecl? then
          return [(decl, i.stx)] -- don't descend into `ts`
    return ts
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+/--
+Lints for instances with arguments that cannot be filled in, like
+```
+instance impossible {α β : Type} [Inhabited α] : Nonempty α := ⟨default⟩
+```
+This is a syntax linter, i.e. it runs on your declarations as you write them.
+-/
+def syntax.impossibleInstance : Linter where run cmdSyntax := do
+  unless Linter.getLinterValue linter.syntax.impossibleInstance (← Linter.getLinterOptions) do
+    return
```
Vierkantor: We should not lint either on declarations with errors, since that would make it harder to diagnose the real error.
```suggestion
  unless Linter.getLinterValue linter.syntax.impossibleInstance (← Linter.getLinterOptions) do
    return
   if (← get).messages.hasErrors then
    return
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  defValue := true
+  descr := "Warn when a declaration is found whose type is not a class but is marked as instance. "
+}
+
+/--
+A linter for checking if any declaration whose type is not a class is marked as an instance.
+-/
+def syntax.nonClassInstance : Linter where run cmdSyntax := do
+  unless Linter.getLinterValue linter.syntax.nonClassInstance (← Linter.getLinterOptions) do
+    return
```
Vierkantor, line 165:
```suggestion
    return
  if (← get).messages.hasErrors then
    return
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  names := names.pwFilter (fun (a, _) (b, _) => a != b)
+  /- the `getTopLevelDeclsByBody` function picks up some internal names from `examples` that it
+     probably shouldn't. We filter these here. -/
+  names := (← names.filterM (fun (name, _) => return (← getEnv).contains name))
+  -- names := names.eraseDups -- todo only first elements duplicate remove
+  for (name, stx) in names do
+    /- check if the return type is class-valued,
+       else the `nonClassInstance'` linter will lint against this. -/
+    let constInfo ← (Lean.getConstInfo name)
+    if not (← liftTermElabM (isClass? constInfo.type)).isSome then continue
```
Vierkantor: I don't think the logic of this check is correct: if we have a `def` which is class-valued but can't be an instance for exactly this reason, the linter would still flag it. But having such a `def` is a valid design pattern, for example [Module.addCommMonoidToAddCommGroup](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Module/NatInt.html#Module.addCommMonoidToAddCommGroup).

Instead we should be checking for `isInstance name`, like below.

BatteriesTest/lintTC.lean
```diff
@@ -9,6 +9,12 @@ namespace A
...
 Note: This linter can be disabled with `set_option linter.unusedVariables false`
+---
+warning: This instance has at least one argument that cannot be inferred using typeclass synthesis. Specifically
+    argument 2: `{β : Type}`
+These are arguments that are not instance-implicit and appear neither in another instance-implicit argument nor the return type, so they cannot be inferred using typeclass synthesis.
+
+Note: This linter can be disabled with `set_option linter.syntax.impossibleInstance false`
 -/
 #guard_msgs in
 local instance impossible {α β : Type} [Inhabited α] : Nonempty α := ⟨default⟩
```
Vierkantor, line 30: Could you also add a test that ensures there are no false positives, something like:

```lean4
/--
warning: unused variable `β`

Note: This linter can be disabled with `set_option linter.unusedVariables false`
-/
#guard_msgs in
theorem okayAsThm {α β : Type} [Inhabited α] : Nonempty α := ⟨default⟩
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,144 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+      if (← fv.getDecl).binderInfo.isInstImplicit then return none
+      if ty.containsFVar fv then return none
+      if argTys[i+1:].any (·.containsFVar fv) then return none
+      let brackets : String × String := match (← fv.getDecl).binderInfo with
+        | .implicit        => ("{", "}")
+        | .strictImplicit  => ("⦃", "⦄")
+        | .instImplicit    => ("[", "]")
+        | .default         => ("(", ")")
+      return some (m!"    argument {i+1}: "
+        ++ s!"`{brackets.1}{←fv.getUserName} : {← inferType arg}{brackets.2}`")
```
Vierkantor: Could you extract this logic into its own function, for formatting a free variable?

Vierkantor (commented):
Thanks! I have some suggestions on restructuring the code, but I like your approach overall and it should work well.

Batteries/Tactic/Lint/TypeClass.lean
Vierkantor: I'm not sure I like that we use `.syntax.` in the option name: it reflects their implementation rather than the goals. The syntax version is, IMO, a strict improvement since it gives instant feedback rather than having to wait for an external `runLinter` step. And (assuming both work the same) I would like to deprecate the environment linter version eventually. Having `.syntax.` in the name would then be meaningless. (Although for the `def ... : Linter` this is less relevant since it is more of an implementation detail.)

In any case there should be no issue with conflicting name in the option, since environment linters use a different mechanism to enable/disable them. So I would call the option simply `linter.impossibleInstance`.

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: oh i see, the comment i left there is misleading since it doesn't include the `not` that is present in the check - i will improve the comment there.
We check for isInstance in the `test` function.
For a 
```lean4
def classvalued ...
```
the impossibleInstance linter won't fire (and we don't want it to) since it won't pass through the isInstance check in `test`.

For 
```lean4
instance notClassValued ...
```
we do not want `impossibleInstance` to fire (and it won't because of this check) since we want the 
`nonClassInstance` linter to fire (and it will)

I will have a look at the rest later today!

Vierkantor: Ah I see, this is to avoid duplicate messages. Then the check makes sense. Then I'd write the comment something like:

```suggestion
    /- Check if the return type is class-valued, otherwise the `nonClassInstance'` linter will already
    have put a message on this declaration. -/
    let constInfo ← (Lean.getConstInfo name)
    if not (← liftTermElabM (isClass? constInfo.type)).isSome then continue
```

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: i have extracted it into `ppFVar` but im not sure in which namespace (or even file) it should live in

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: i have now removed `.syntax` from the option names

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: I did the switch to Namemap, however couldn't find a merge function for NameMap specifically, so i just used the one from the underlying `Std.TreeMap`. Is there just API missing or is it intended to use `TreeMap` api on `NameMap`?

Vierkantor: Using `TreeMap` definitions on `NameMap` is okay. At least, that's what I do too :)

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: woops, i accidentally marked this as resolved.
Currently `ppFVar` has no Namespace at all, so that is definitely wrong

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -1,22 +1,28 @@
...
 public meta import Lean.Meta.Instances
 public meta import Batteries.Tactic.Lint.Basic
+public meta import Lean.Elab.Command
+public meta import Lean.Linter.Basic
+public meta import Lean.Server.InfoUtils
+public meta import Batteries.Data.List.Basic
 
 public meta section
 
+section environmentLinters
```
fgdorais: Don't use `section` just add a comment.

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
     let info ← getConstInfo declName
     if !(← isClass? info.type).isSome then return "should not be an instance"
     return none
+
+end Batteries.Tactic.Lint
+end environmentLinters
+
+
+
+section StandardLinters
```
fgdorais: Don't use `section` just add a comment.

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  let env ← getEnv
+  return names.filter (fun name _ => env.contains name)
+
+
+
+open Lean Meta in
+/-- Returns the pretty-printed form of a free variable with its type,
+    using the appropriate brackets for its `BinderInfo`.
+    Example output: `{α : Type}`. -/
+def ppFVar (fv : FVarId) : MetaM MessageData := do
```
fgdorais: Why is this in the root namespace?

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  let decl ← fv.getDecl
+  let (lBracket, rBracket) : String × String := match decl.binderInfo with
+    | .implicit       => ("{", "}")
+    | .strictImplicit => ("⦃", "⦄")
+    | .instImplicit   => ("[", "]")
+    | .default        => ("(", ")")
+  return s!"`{lBracket}{decl.userName} : {← inferType (mkFVar fv)}{rBracket}`"
```
fgdorais:
```suggestion
  return m!"`{lBracket}{decl.userName} : {← inferType (mkFVar fv)}{rBracket}`"
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+       linter will already put a message on this declaration, so we skip it here in that case.
+       If the declaration is not an instance it is skipped in `test`.  -/
+    let constInfo ← (Lean.getConstInfo name)
+    if not (← liftTermElabM (isClass? constInfo.type)).isSome then continue
+    /- Now the actual linting check: -/
+    let some lintmessage ← liftTermElabM (test name) | continue
+    /- Use the range that actually corresponds to the `name` not to the whole mutual block: -/
+    Linter.logLint linter.impossibleInstance stx lintmessage
+  return
+initialize addLinter syntax.impossibleInstance
```
fgdorais: No need for final return.
```suggestion

initialize addLinter impossibleInstance
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+    if !(← isClass? info.type).isSome then
+      return "This declaration should not be an instance as it is not class-valued."
+    return none
+  /- We do the check for each (different) top level instance name we can get from the infotrees.
+     Mostly this will only be one name, but for `mutual` blocks this will be more. -/
+  let names ← Lean.Elab.getInfoTreesDecls
+  for (name, stx) in names do
+    let some lintmessage ← liftTermElabM (test name) | continue
+    Linter.logLint linter.nonClassInstance stx lintmessage
+  return
```
fgdorais: No need for final return.
```suggestion
 ```

File: Batteries/Tactic/Lint/TypeClass.lean [hunk 27]
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+initialize addLinter syntax.nonClassInstance
```
fgdorais:
```suggestion
initialize addLinter nonClassInstance
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+/-- Option for turning the `nonClassInstance` linter on and off. -/
+register_option linter.nonClassInstance : Bool := {
+  defValue := true
+  descr := "Warn when a declaration is found whose type is not a class but is marked as instance. "
+}
+
+/--
+A linter for checking if any declaration whose type is not a class is marked as an instance.
+-/
+def syntax.nonClassInstance : Linter where run cmdSyntax := do
```
fgdorais:
```suggestion
def nonClassInstance : Linter where run cmdSyntax := do
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,156 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+
+/--
+Lints for instances with arguments that cannot be filled in, like
+```
+instance impossible {α β : Type} [Inhabited α] : Nonempty α := ⟨default⟩
+```
+This is a syntax linter, i.e. it runs on your declarations as you write them.
+-/
+def syntax.impossibleInstance : Linter where run cmdSyntax := do
```
fgdorais:
```suggestion
def impossibleInstance : Linter where run cmdSyntax := do
```

fgdorais (changes requested):
Looks good! Just a few small changes.

I changed my mind about the `syntax` prefix, it is not necessary after all.

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: see the comment [above](https://github.com/leanprover-community/batteries/pull/1717#discussion_r2952349607)
I wasn't sure if this should go into `Batteries.Linter` or `Lean.Meta`. Any suggestion?

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: i have added module docstrings like `/-! ### Environment Linters -/` instead

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: i have now chosen `Batteries.Linter.ppFVar`. If you think something else is more appropriate let me know!

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  let env ← getEnv
+  return names.filter (fun name _ => env.contains name)
+
+
+
+open Lean Meta in
+/-- Returns the pretty-printed form of a free variable with its type,
+    using the appropriate brackets for its `BinderInfo`.
+    Example output: `{α : Type}`. -/
+def Batteries.Linter.ppFVar (fv : FVarId) : MetaM MessageData := do
```
thorimur: A few points on this declaration:
- Docstring formatting :)
- We should probably use the word "binder" in the name of this function. My suggestion here is to call it `ppAsBinder` and put it in the `FVarId` namespace, so that we can use dot notation with free variables: `← fvarId.ppAsBinder`
- The `.type` is present within the `decl` we've obtained; we don't need to `inferType`.
- In fact, since the only use of `MetaM` is to `getDecl`, we may as well define `LocalDecl.ppAsBinder` as a pure function, then define `FVarId.ppAsBinder` in terms of it.
- We shouldn't use the `userName`, as that doesn't give us hovers. Rather, use `Expr`s when possible in message data; `decl.toExpr`, which is ultimately `mkFVar fvarId`/`.fvar fvarId`, is probably correct here.
- We need to handle let decls. It doesn't really make sense to print those as a binder, so let's make this `Option MessageData` and return `none` in that case.
- Let's not include code fencing in the message. We may want to concatenate it with other binders, and that closes off possibilities of use.
- There's an obscure definition `MessageData.bracket lbracket msg rbracket` which does some grouping/nesting we want here.

Putting that all together we get:
```lean4
/-- Pretty-prints a local declaration and its type as a binder,
using the appropriate brackets given its `BinderInfo`. Example output: `{α : Type}`.

Returns `none` for let decls. -/
protected def _root_.Lean.LocalDecl.ppAsBinder : LocalDecl → Option MessageData
  | .ldecl .. => none
  | .cdecl _ fvarId _ type binderInfo _ =>
    let (lBracket, rBracket) : String × String := match binderInfo with
      | .implicit       => ("{", "}")
      | .strictImplicit => ("⦃", "⦄")
      | .instImplicit   => ("[", "]")
      | .default        => ("(", ")")
    some <| .bracket lBracket m!"{mkFVar fvarId} : {type}" rBracket

/-- Pretty-prints a free variable and its type as a binder,
using the appropriate brackets given its `BinderInfo`. Example output: `{α : Type}`.

Returns `none` for let decls. -/
@[inline]
protected def _root_.Lean.FVarId.ppAsBinder (fvarId : FVarId) : MetaM (Option MessageData) :=
  return (← fvarId.getDecl).ppAsBinder
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+
+/-! ### Syntax Linters -/
+
+/-- `getTopLevelDeclsByBody tree` returns the top level names of declarations
+    (along with their syntax) which have been logged in the infotree `tree`.
+
+    Specifically, this function collects the contained names and syntax for all nodes
+    with a `BodyInfo` value.
+    Since we return a `NameMap Syntax`, each name is only returned once.-/
```
thorimur: I believe standard docstring style is to not indent the second lines, and for the delimiters to either be both inline or both on their own line. Also, I think we should show this being used with dot notation as it would be typically, i.e. `t.getTopLevelDeclsByBody`

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+    using the appropriate brackets for its `BinderInfo`.
+    Example output: `{α : Type}`. -/
+def Batteries.Linter.ppFVar (fv : FVarId) : MetaM MessageData := do
+  let decl ← fv.getDecl
+  let (lBracket, rBracket) : String × String := match decl.binderInfo with
+    | .implicit       => ("{", "}")
+    | .strictImplicit => ("⦃", "⦄")
+    | .instImplicit   => ("[", "]")
+    | .default        => ("(", ")")
+  return m!"`{lBracket}{decl.userName} : {← inferType (mkFVar fv)}{rBracket}`"
```
thorimur: Using `decl.userName` by itself means we don't get hovers. We should instead just print the expression; also, we already have the type! :)
```suggestion
  return m!"`{lBracket}{decl.toExpr} : {decl.type})}{rBracket}`"
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+    forallTelescopeReducing (← inferType (← mkConstWithLevelParams declName)) fun args ty => do
+    let argTys ← args.mapM inferType
+    let impossibleArgs ← args.zipIdx.filterMapM fun (arg, i) => do
+      let fv := arg.fvarId!
+      if (← fv.getDecl).binderInfo.isInstImplicit then return none
+      if ty.containsFVar fv then return none
+      if argTys[i+1:].any (·.containsFVar fv) then return none
+      return some (m!"    argument {i+1}: " ++ (← ppFVar fv))
+    if impossibleArgs.isEmpty then return none
+    return errorsFound1 ++ (Lean.MessageData.joinSep impossibleArgs.toList ", ") ++ errorsFound2
```
thorimur: We only use this function once, and likewise for `errorsFound1` and `errorsFound2`. Let's inline them at their use site (and use `m!` syntax for the final message).

Further, instead of returning `none` then destructuring it, let's just `continue`, and `log` the message directly at the bottom of the code block. Generally, it's good to avoid defining `let` functions :)

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  /- We do the check for each (different) top level instance name we can get from the infotrees.
+     Mostly this will only be one name, but for `mutual` blocks this will be more. -/
+  let names ← Lean.Elab.getInfoTreesDecls
+  for (name, stx) in names do
+    /- If the return type is not class valued (but an instance), the `nonClassInstance'`
+       linter will already put a message on this declaration, so we skip it here in that case.
+       If the declaration is not an instance it is skipped in `test`.  -/
+    let constInfo ← (Lean.getConstInfo name)
+    if not (← liftTermElabM (isClass? constInfo.type)).isSome then continue
+    /- Now the actual linting check: -/
```
thorimur: After inlining `test` (see above), we should just move this within the `TermElabM`. Note also that we have two ways of getting the type; we may as well instead just use `cinfo.type` and abandon `mkConstWithLevelParams` in this case. :)

Also note that `isNone` exists :)

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+  descr := "Warn when a declaration is found whose type is not a class but is marked as instance. "
+}
+
+/--
+A linter for checking if any declaration whose type is not a class is marked as an instance.
+-/
+def nonClassInstance : Linter where run cmdSyntax := do
+  unless Linter.getLinterValue linter.nonClassInstance (← Linter.getLinterOptions) do
+    return
+  if (← get).messages.hasErrors then
```
thorimur: These are the same; this is essentially a style nitpick :)
```suggestion
  if ← MonadLog.hasErrors then
```

fgdorais:
Go ahead @thorimur! I forgot to thank @Vierkantor earlier! Once you are done I will do a final review, which will presumably be very quick with all this help!

MoritzBeroRoos:
I have not benched it against mathlib, but that is a good idea. Is there an easy way to do this? I think i have seen something like `!bench` on mathlib pr's

Batteries/Tactic/Lint/TypeClass.lean
fgdorais: Good enough for now. Maybe I or someone will see the right place for it soon. If not, it's easy to fix so long as it's not in the root space.

Batteries/Tactic/Lint/TypeClass.lean
Vierkantor: Not sure about our specific Markdown parser, but some render text indented with 4 characters as a literal block of code. For safety I always write:
```suggestion
/--
`getTopLevelDeclsByBody tree` returns the top level names of declarations
(along with their syntax) which have been logged in the infotree `tree`.

Specifically, this function collects the contained names and syntax for all nodes
with a `BodyInfo` value.
Since we return a `NameMap Syntax`, each name is only returned once.
-/
```

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +54,154 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+
+/-! ### Syntax Linters -/
+
+/-- `getTopLevelDeclsByBody tree` returns the top level names of declarations
+    (along with their syntax) which have been logged in the infotree `tree`.
+
+    Specifically, this function collects the contained names and syntax for all nodes
+    with a `BodyInfo` value.
+    Since we return a `NameMap Syntax`, each name is only returned once.-/
+partial def Lean.Elab.InfoTree.getTopLevelDeclsByBody : InfoTree → NameMap Syntax :=
```
Vierkantor: I just discovered that Mathlib also has [getDeclsByBody](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Lean/Elab/InfoTree.html#Lean.Elab.InfoTree.getDeclsByBody), by @thorimur. So I'd like to ask Thomas to check if the two versions can coexist without duplication :)

Vierkantor (approved):
Thanks for the changes! 🎉 

(Style remark, I am unsure Batteries follows this as strictly as Mathlib so feel free to ignore if I am too strict: we only have one line of whitespace in a file.)

thorimur (commented):
Okay, so, good news and bad news.

Bad news first (and let me preface this by saying this is **not** a fault of the code you contributed!): as-is, the approach used by the original `env_linters`, which is being inherited by the syntax linters, is just not performant enough. Syntax linters can be a lot more sensitive to performance issues. Some parts of this PR therefore need major changes (not through any fault of your own, I should emphasize! It's just the nature of doing this sort of thing in a syntax linter.)

Good news: I've done a bunch of benching of different strategies in isolation, and I've figured out a way to make the underlying infrastructure performant enough. Part of this is #1754, which I've split out so it can be reviewed separately and more easily :) The other part involves upstreaming and modifying a more-performant mathlib definition to create `hasUnusedForallBindersWhere`.

The review suggestions are in [the linked PR](https://github.com/MoritzBeroRoos/batteries/pull/1)--you **don't** need to apply any suggestions made here (and shouldn't, lest they create a merge conflict :) ), they're just for the diff. But let me (1) summarize the changes that needed to happen and why, and what now needs to be reviewed (2) try to communicate the motivation behind the changes as much as I can.

---

**1:** There were two factors that contributed to the performance hit: using infotrees (my suggestion! 🙃) and the expensive expression traversals.

Moving away from infotrees entirely can be done by strategy I recently saw in mathlib: look at the declaration ranges. This picks up more declarations and is faster as well. However, that strategy (in mathlib as `getNamesFrom`) needed some polishing to make it more performant (and also constructed syntax, which we don't need during normal operation until we log).

But since we don't need infotrees, we don't need `getTopLevelDeclsByBody` at all. Which is both happy and sad...but if we ever need richer data than the declaration names, now we know that that's what to do :)

Now the declarations in `Batteries.Lean.Positions` provide a better way of getting the variable names.

Re: expression traversals: the old linters were evaluating `containsFVar` a quadratic number of times: for each argument, on all subsequent argument types. Further, this means we need to look through all the fvar-containing parts of the expression and identify them, then move to the next fvar locations, etc. Not as bad as finding a generic expression thanks to little optimizations that let you detect if there is *any* fvar in a subexpression in constant time, but not great.

Also, they did this within a `forallTelescopeReducing`. That incurs a little overhead as well.

For dependency detection, I adapted a strategy I used for the UnusedInstancesInType linter based on bvars; here, this is `hasUnusedForallBindersWhere`. To @Vierkantor : it is unfortunately slightly different than the related `getUnusedForallBinderIdxsWhere` in mathlib, but they could be unified in the future!

The key here is that this lets us avoid `MetaM` entirely under typical usage, where no errors are logged. When we do want to log an error, we're allowed to be slow (since these shouldn't persist!), and then we can do a forallTelescope-like computation.

---

**2:** Other changes you might like to be aware of:

- Likewise, when you log some `MessageData`, the logging function actually saves the current context information into the message itself. This means that if you're in one `MetaM` context with some free variables available, and you mention them in your `MessageData`, then log the message, Lean will still know how to interpret them. But if you create the `MessageData` that contains fvars, ferry it *outside* of that `MetaM` context, *then* log it, you might introduce unknown free variables into the infoview. [For example](https://live.lean-lang.org/#codez=JYWwDg9gTgLgBAGQKYEMB2AoUlZwIoCOGGEYSaiqFAskjCnAKIA2KARnAMIQgjoAm+IhiSsOAIgDEMJAGcY4uAC44AYx580ggLwA+OPwgY4cZnTghZAcziAEwlPAAZjAAqSKCBbtqBiHEfQKMzMbmay6mRwBAAUAEIQEMxwgEmEcAByKDAAlP4ArhQoUFaycAD6cHq+xiZwUHS5UBQgAITiAN6FxQC+4tXMEFYAkmgBFtbEALQTcAASEABu7sBoNrIAFhAA7iXLMo1BcABKAAqccO5Q0HD5ANZoWxSOdUhw84XA7GYYAPRTcMsBFQAbVKjjeUAAdAAWACcAEYALpwCbfDCSKy5Qr8UqWYr/TDSOQwYiidhwKRsOh7RQqdS8AQVfSGPpOVzuTxiHyGfyBYKhOQRF4xeKJFLpTI5Rz5OCdErlSrMmqmAbDUYAHgAPhZWh0irIepNpmk/LlZO40CgQEgADRwNi5eAATwguTU6DgG0WUBlWjgZpewBgAHISgwRcwfn8ARBgUD0BA0I6QC7ZAikSi0RisTjrDsCZSYHsgA). So it's good to log inside the forallTelescope, when the free variables exist :)
- I also tweaked some of the logic you inherited for message-making: it didn't handle chains of unused arguments, where one unused argument depends on another, and both are unused, correctly. I used `collectFVars` here, which is pretty useful for things like this, and you might find it useful in other situations too.
- I suggested some style changes; mostly just removing parentheses where possible and collapsing a single-branched `if !...` into `unless`. Also consolidating `open`s, removing multiline whitespace, removing superfluous namespaces where possible, and reformatting comments. All of this is very minor and does not really matter very much.
- Minor tweaks to the message wording; they look very good.
- I tried to move the expensive tests performed by the linter as far inwards as possible for performance reasons, so that they are triggered as infrequently as possible.

---

The rest looks really good! :) I'm quite sorry for the wall of text and massive amount of review suggestions...usually I would not want to do that, but the unusual performance concerns you just so happened to hit here made it necessary.

If you're happy with the suggestions and merge the PR into your branch, please bench again after updating the batteries rev...and we'll see if [my isolated benching experiments](https://github.com/leanprover-community/mathlib4/pull/37237) actually mean what they seem to mean. :)

Please feel free to message me with any questions on any review suggestion I failed to cover! :) And thanks once again for making this PR and bearing through all this review!!

Batteries/Tactic/Lint/TypeClass.lean
thorimur: I now think we ought to avoid introducing this definition altogether and use a different strategy for performance; see below 😅

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +57,129 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+/-! ### Syntax Linters -/
+
+/-- Pretty-prints a local declaration and its type as a binder,
+using the appropriate brackets given its `BinderInfo`. Example output: `{α : Type}`.
+
+Returns `none` for let decls. -/
+protected def Lean.LocalDecl.ppAsBinder : LocalDecl → Option MessageData
+  | .ldecl .. => none
+  | .cdecl _ fvarId _ type binderInfo _ =>
+    let (lBracket, rBracket) : String × String := match binderInfo with
```
adomani, line 69: Note that [docs#Lean.BinderInfo.brackets](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Lean/Expr/Basic.html#Lean.BinderInfo.brackets) seems to do something very similar.

fgdorais:
Please merge main now that #1754 has landed.

Batteries/Tactic/Lint/TypeClass.lean
fgdorais: But it strangely does `{{...}}` instead of `⦃...⦄`. (FWIW, I personally prefer the former but the latter is more prevalent.)

MoritzBeroRoos: Since Lean will show `⦃...⦄` even if the user inputted `{{...}}`, i think we should also go with `⦃...⦄` here

MoritzBeroRoos:
This pr currently breaks a `guard_msgs` of a testing instance in `MathlibTest` [here](https://github.com/leanprover-community/mathlib4/blob/09037d340e5cc2adbcc8428f3c61ec1ebe65d5ad/MathlibTest/LintStyle.lean#L289).

This wasn't found previously, since the environment linters don't seem to be run in `MathlibTest`.
Since the point of that instance is to test a different linter im thinking we just disable the impossibleInstance linter for that declaration.
The Batteries main page says to ask for write access to mathlib-nightly-testing, so i guess this is me asking :)

[nightly-testing-branch](https://github.com/leanprover-community/mathlib4-nightly-testing/tree/batteries-pr-testing-1717)

fgdorais:
That part of the docs needs an update but we haven't yet figured out what the update should be 🙃

The easiest way is for @kim-em to give you access to `mathlib-nightly-testing`. That's up to them....

The harder way is to fork `mathlib-nightly-testing` and make a PR to the `batteries-pr-testing-1717` branch. Such PRs are known to be experimental so there is no review process beyond basic sanity checks. They should be merged  very quickly 🤞

After `batteries-pr-testing-1717` works, there needs to be a PR to mathlib from that branch. I'm happy to do that part. It's unlikely you could do this without access to `mathlib-nightly-testing`.

Vierkantor:
I have added @MoritzBeroRoos to the [nightly-testing team](https://github.com/orgs/leanprover-community/teams/nightly-testing/members). This is the first time I added anyone here, so please confirm that you can now push to the branch!

Batteries/Tactic/Lint/TypeClass.lean
Vierkantor: Let's add a comment explaining the difference with [Lean.BinderInfo.brackets](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Lean/Expr/Basic.html#Lean.BinderInfo.brackets). That should help us understand what to do if the upstream function ever gets fixed.

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -48,3 +57,129 @@ A linter for checking if any declaration whose type is not a class is marked as
...
+          let possibleFVars ← possibleFVars.addDependencies
+          let mut impossibleArgMsgs := #[]
+          for arg in args, i in 1...* do
+            unless possibleFVars.fvarSet.contains arg.fvarId! do
+              let some binder ← arg.fvarId!.ppAsBinder | continue
+              impossibleArgMsgs := impossibleArgMsgs.push <|
+                indentD m!"argument {i}: `{binder}`"
+          if impossibleArgMsgs.isEmpty then return -- Should not be reachable.
+          -- TODO: log on a better location; see #1734
+          Linter.logLint linter.impossibleInstance (← getRef) m!"\
```
Vierkantor: #1734 seems to have been merged: can we solve this TODO now?

Vierkantor (approved):
I am satisfied with this PR. Once you get the Mathlib nightly testing to build, let's run a benchmark and then I'd like to get it merged!

Batteries/Tactic/Lint/TypeClass.lean
MoritzBeroRoos: done!

MoritzBeroRoos:
Thanks @Vierkantor, i have confirmed that i can push to the nightly branch now!

1. I have added a description of what the linters do to the hover of their options, since this might well be the only hover a user might see when the linter is disabled - some other linters also have such explanations at this place.
(Altough i don't foresee why one would ever disable one of those two specific linters)

2. I did the remaining todo (use the new `withDeclRef?` to put the linters warning at the name (if existent) or else the `instance` keyword of the declaration instead of at the beginning of the doc comment.
I Also added/modified some tests which (now) test for the precise position of the linters warning.
These also double as tests for `withDeclRef?`.

3. After disabling the new linter in a few mathlib **tests** the nightly branch 
at [batteries-pr-testing-1717](https://github.com/leanprover-community/mathlib4-nightly-testing/tree/batteries-pr-testing-1717) is successfully building.
Only one of the tests (the changed one for `toDual` [here](https://github.com/leanprover-community/mathlib4-nightly-testing/commit/c4c5845cf955e8f06231f52fb733bc1a2c47706f)) was looking slightly nontrivial - after a quick consultation with
@JovanGerb  i changed it into an implicit reducible def (and thus the `impossibleInstance` linter rightly doesn't complain anymore). (The `set_option` was also from me a few minutes before that, the environment linters never run on that file and as such this instance wasn't reported)

4. I also opened a pr to mathlib from this nightly for benching [here](https://github.com/leanprover-community/mathlib4/pull/38088) and started a bench. The results are pretty much what we had when benching
@thorimur's new approach before (in fact it is very slightly faster now).

    The bench is at `+213.3G (+0.12%)`, `No significant results found`.

    Note that this benching pr fails ci only because mathlib checks that the batteries branch used is an official one (but this isn't true for any of the nightlys).

Thanks again for @thorimur for his massive involvement, making this much better in pretty much every way!

Batteries/Tactic/Lint/TypeClass.lean
```diff
@@ -1,22 +1,31 @@
...
 module
 
 public meta import Lean.Meta.Instances
 public meta import Batteries.Tactic.Lint.Basic
+public meta import Lean.Elab.Command
+public meta import Lean.Linter.Basic
+public meta import Lean.Server.InfoUtils
+public meta import Batteries.Data.List.Basic
+public meta import Batteries.Lean.Position
+public meta import Batteries.Lean.Meta.Expr
```
fgdorais, line 12: Shake doesn't currently run in batteries, can somebody check that these imports are necessary and minimal?

Batteries/Tactic/Lint/TypeClass.lean
fgdorais: Also make a PR to core and link it in the comment.

MoritzBeroRoos:
@fgdorais quick clarification:
The above-mentioned [Lean.BinderInfo.brackets](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Lean/Expr/Basic.html#Lean.BinderInfo.brackets) currently lives downstream in `Mathlib` (not upstream in core!)
and uses a fallback to normal brackets (and returns 2 strings) instead of returning an `Option MessageData`.

The plan is to leave mathlibs `Lean.BinderInfo.brackets` alone for now then and upstream what is used in this PR (namely `Lean.LocalDecl.ppAsBinder`) to core?

MoritzBeroRoos:
I have now minimized the imports by hand in the following way:
1. try removing a full import
2. try removing `public` or `meta` from kept imports.

I have never used `lake shake` before, but running `lake shake Batteries.Tactic.Lint.Typeclass` now gives the following output:

```sh
/Users/moroo/_EigeneDaten/GIT/batteries/Batteries/Tactic/Lint/Basic.lean:
  remove #[public meta import Lean.Elab.Exception]
  add #[public import Lean.Meta.Basic, import Lean.Structure]
/Users/moroo/_EigeneDaten/GIT/batteries/Batteries/Lean/Position.lean:
  remove #[public import Lean.Syntax, public import Lean.Data.Lsp.Utf16]
  add #[public import Lean.DeclarationRange]
/Users/moroo/_EigeneDaten/GIT/batteries/Batteries/Tactic/Lint/Typeclass.lean:
  remove #[public import Lean.Linter.Basic]
  add #[public meta import Lean.Linter.Basic]
  ```
  
  which for `Batteries/Tactic/Lint/Typeclass.lean` boils down to adding the `meta` back to the import?
  Why should we do that?

thorimur:
Thanks for adding these new tests, they look great! :) I'm also glad that `withDeclRef?` is now getting tested! :)

Batteries/Tactic/Lint/TypeClass.lean
fgdorais: My bad! I misremembered that this was from mathlib and not core.

Batteries/Tactic/Lint/TypeClass.lean
fgdorais: I'm satisfied with your manual checks. There will be a global batteries shake eventually which will deal with remaining issues.

fgdorais (approved):

fgdorais:
I am also satisfied. All we need now is approval of the bench results from mathlib. @Vierkantor 

For the mathlib adaptation PR, I recommend using this PR as a template: https://github.com/leanprover-community/mathlib4/pull/35104

MoritzBeroRoos:
Let's use the benching pr as an [adaptation-pr](https://github.com/leanprover-community/mathlib4/pull/38088)

Vierkantor:
I am happy enough with the benchmark results. About 0.1% is not going to be noticeable. Especially if we can also disable the `runLinter` version afterwards.

fgdorais:
Merging. Please update the Mathlib adaptation PR as soon as possible.

wkrozowski:
Sadly, I have seen this right now, as being quite unaware of that, I have independently done and just merged https://github.com/leanprover/lean4/pull/13389, which upstreams the batteries environment linters for impossible instances and non-class instances to core lean4 as errors.

MoritzBeroRoos:
oh well, i guess i'm both sad and happy at the same time!

MoritzBeroRoos:
pinging @fgdorais for a revert

fgdorais:
Sad indeed. Revert at #1773

thorimur:
This is really rough. @MoritzBeroRoos I'm sorry this happened!

It may seem a small consolation, but even with this PR reverted, I would like to emphasize that you've still learned everything that you learned in the course of making this PR. Imo, the effect on the author is actually main benefit of going through the review process. :)

thorimur:
@wkrozowski I'm also wondering: in this batteries PR, there were various improvements to performance and especially to error logic/error message formatting (to account for the transition from CI-based linter to interactive linter) which have already undergone careful review.

Would upstreaming those improvements to core, either in whole or in part, be welcome?

wkrozowski:
@thorimur Yes, of course! I think it would be very good to have the improved error messages in core.

wkrozowski:
@MoritzBeroRoos  We are more than happy to do the upstreaming of the error messages ourselves, but if you prefer you can open your PR to core lean4 and I will help with getting that in, if you prefer to be on the lean4 contributors list with your work.

MoritzBeroRoos:
@wkrozowski This would be a good motivation to finally get a lean-dev environment going for me!
But i will be away the next few days, so i could only open a pr some time in the upcoming week.
If you want this completed sooner just tell me, i'm happy with you doing it then!

wkrozowski:
@MoritzBeroRoos That is a nice coincidence, as I am away early next week as well! I am of course happy to wait a bit. Feel free to ping me on zulip.

PR #1727, opened 2026-03-20
fix: update `defLemma` and related linters correctly

thorimur:
This PR fixes an issue where the `defLemma` linter wrongly ignored `@[implicit_reducible]` defs due to an incorrect adaptation when bumping lean versions. [Zulip](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/defLemma.20linter.20false.20negative.20on.20.60.40.5Bimplicit_reducible.5D.60/with/580732427)

However, instances created through the delta-deriving handler do not properly get registered as `theorem`s (leanprover/lean4#13295) and are instead always `def`s. We ignore definitions satisfying *both* `isInstance` and `isImplicitReducible` as a workaround.

We also take this chance to (1) note that leanprover/lean4#2575 has been fixed, allowing us to remove some workarounds in `defLemma` and other linters (2) clean up the code slightly.


Batteries/Tactic/Lint/Misc.lean
```diff
@@ -73,63 +73,43 @@ We skip all declarations that contain `sorry` in their value. -/
...
     let info ← getConstInfo declName
-    let isThm ← match info with
-      | .defnInfo .. => pure false
-      | .thmInfo .. => pure true
-      | _ => return none
-    match isThm, ← isProp info.type with
-    | true, false => pure "is a lemma/theorem, should be a def"
-    | false, true => pure "is a def, should be lemma/theorem"
+    match info, ← isProp info.type with
+    -- note: first case should be unreachable
```
Rob23oba: You're right, the first case is impossible, all kernels check for this (see https://arena.lean-lang.org/tutorial/#test-011_nonPropThm). Because of this, I'd say we refactor this to just
```lean-4
if ← pure info.isDefinition <&&> isProp info.type then
  return "is a def, should be lemma/theorem"
else
  return none
```

thorimur: Nice, thanks for checking! :) I've accepted the suggestion (and also realized there's been a missing "a" before "lemma/theorem" this whole time! :) )

thorimur:
Hmm, this builds mathlib after including a workaround, but maybe we should instead just wait for leanprover/lean4#13304 to land and percolate through, which fixes the issue I'm working around? Not sure how long that would be.

EDIT: Ok, great, Kim has fixed it, so I suppose let's just wait :)

PR #1754, opened 2026-04-03
feat: `getDeclsAfter` for extracting declarations in syntax linters

thorimur:
This PR adds `Position.quickLt`, `Position.getDeclsAfter`, and `String.Pos.Raw.getDeclsAfter` to provide syntax linters with an efficient means of extracting declarations associated with the current command.

These declarations are needed for making #1717 sufficiently performant, and also can be used to improve the performance of syntax linters currently in Mathlib.

This approach was adapted from `getNamesFrom` in Mathlib. It's undergone optimizations to make it more performant, and also fulfills a slightly different role (`getNamesFrom` constructs `Syntax`).


Batteries/Lean/Position.lean
```diff
@@ -25,6 +25,43 @@ def findIndentAndIsStart (s : String) (pos : String.Pos.Raw) : Nat × Bool :=
   let body := (s.pos! start).find (· ≠ ' ') |>.offset
   (start.byteDistance body, body == pos)
 
+/-- A quicker version of `Position.lt`, without sacrificing semantics. We need the performance
+optimization since we're doing this for many positions on every declaration within linters. -/
+@[inline] protected def Position.quickLt : Position → Position → Bool
+  | ⟨l₁, c₁⟩, ⟨l₂, c₂⟩ => l₁ < l₂ || l₁ = l₂ && c₁ < c₂
```
fgdorais: Any reason not to fix this in core?

Batteries/Lean/Position.lean
```diff
@@ -25,6 +25,43 @@ def findIndentAndIsStart (s : String) (pos : String.Pos.Raw) : Nat × Bool :=
...
+usually smaller than the `range`, we err on the side of including declarations when possible.
+
+By default, this only inspects the local branch of the environment. This is compatible with being
+used to find declarations from the current command in a linter, where we have already waited for
+async tasks/parallel branches to complete. Further, since the environment exposed to linters does
+not include constants added after the elaboration of the current command, it is safe to use this on
+the command's start position without picking up later declarations.
+-/
+@[inline] protected def _root_.String.Pos.Raw.getDeclsAfter (env : Environment) (map : FileMap)
+    (pos : String.Pos.Raw) (asyncMode := EnvExtension.AsyncMode.local) : Array Name :=
```
fgdorais, line 54: Is there a precedent adding this kind of function in `_root_.String...`? Is the reason just to make dot notation work? I'm worried that the name `getDeclsAfter` in this namespace loses too much of the contextual information carried by the `Lean.Position` namespace. I can't think of a specific example off the top of my head but `getDeclsAfter` is generic enough that I can't exclude that another context may have an equally valid claim to the name `String.Pos.Raw.getDeclsAfter`.

Batteries/Lean/Position.lean
```diff
@@ -25,6 +25,43 @@ def findIndentAndIsStart (s : String) (pos : String.Pos.Raw) : Nat × Bool :=
...
+If `pos` is a `String.Pos.Raw`, then `pos.getDeclsAfter` returns the array of names of declarations
+whose selection range begins in position at least `pos`. By using the `selectionRange`, which is
+usually smaller than the `range`, we err on the side of including declarations when possible.
+
+By default, this only inspects the local branch of the environment. This is compatible with being
+used to find declarations from the current command in a linter, where we have already waited for
+async tasks/parallel branches to complete. Further, since the environment exposed to linters does
+not include constants added after the elaboration of the current command, it is safe to use this on
+the command's start position without picking up later declarations.
+-/
```
fgdorais, line 52: Refer to the docstring for `Lean.Position.getDeclsAfter` instead of copying all the details.

Batteries/Lean/Position.lean
thorimur: Just that core PRs can take a while, and there's some discussion to be had :) Namely, it's not clear to me if `Prod.lexLt` (which is what core uses) should be changed so that the produced IR is better, or if just writing this definition would be the right approach.

Batteries/Lean/Position.lean
thorimur: It is primarily for dot notation, and because this is the function metaprogrammers will almost always use over the `Lean.Position` one (since the `String.Pos.Raw` will be available immediately from the syntax, while the `Position` is not as readily available). I'm used to adding dot notation to core Lean types like `Expr`, but I don't know about `String.Pos.Raw` in particular.

But I see what you mean about losing context...hmm. `String.Pos.Raw.getRecordedDeclsAfter`? `String.Pos.Raw.getDeclsAfterByRange` (kind of a mouthful, and we lose consistency with `Position.getDeclsAfter`...unless we change that too, which is an option!)? `Lean.Linter.getDeclsAfter` (though we lose dot notation :/ )? Do you have another name in mind?

Batteries/Lean/Position.lean
thorimur: Actually, now that I think about it a second time: is this approach *worse*, because core's version is compiled while this one isn't? I'm not completely clear on when the compiled version will be used in the interpreter...

fgdorais: Did you notice a performance improvement with the new definition?

thorimur: I thought so, but honestly I've done enough benches now to lose track of what was benched separately from what. :) I'll bench the two options now in a head to head.

thorimur: Actually, I had forgotten how much of a time sink it is to bench as-is, when nothing uses it yet!

Sorry to pivot, but would it be alright to either leave `Position.quickLt` out (or in!), wait for #1717 and performance improvements to the UnusedInstancesInType linters to land, then PR `Position.quickLt` as a potential performance boost (or, if left in, remove the unnecessary code) when it's a bit easier to bench? Which would you prefer? (If left in, I could make it `private`.)

fgdorais: Let's leave it out and come back to it later, if necessary.

thorimur: Sounds good! Removed :)

Batteries/Lean/Position.lean
fgdorais: I've been scratching my head and, after some thought, `getDeclsAfter` is not that bad. At the risk of kicking myself later, I'm happy to leave it.

fgdorais (approved):

PR #1760, opened 2026-04-07
chore: move AlternativeMonad MetaM, to reduce Mathlib imports

kim-em:
Currently Aesop needs to import Batteries.Lean.Meta.Basic, which transitively brings in SatisfiesM, which is otherwise unneeded in Mathlib; I just ran into "needing" to fix SatisfiesM on the lean-pr-testing-13283 branch, and would prefer to avoid this.

### DIFF
--- a/Batteries/Control/AlternativeMonad.lean
+++ b/Batteries/Control/AlternativeMonad.lean
@@ -11,6 +11,7 @@ import all Init.Control.Option
 import all Init.Control.State
 import all Init.Control.Reader
 import all Init.Control.StateRef
+public import Lean.Meta.Basic
 
 @[expose] public section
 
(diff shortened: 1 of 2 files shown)


fgdorais:
I'm confused: `Batteries.Lean.Meta.Basic` does not transitively bring in `SatisfiesM`.

The transitive imports of `Batteries.Lean.Meta.Basic`, omitting core modules, are:

* `Batteries.Control.AlternativeMonad`
* `Batteries.Control.LawfulMonadState`
* `Batteries.Control.Lemmas`
* `Batteries.Control.OptionT`

All of these look harmless to me.

I'm sure you ran into issues with something importing `SatisfiesM` but this wasn't it.

fgdorais:
No action for now. Could be revived with further information.

PR #1779, opened 2026-04-21
fix: `List.prod` has already been declared

srghma:
should be named update to v4.31.0 ?


fgdorais:
I changed the target branch to `nightly-testing` since I think that's what you meat to do. Of course, you need to resolve conflicts for the correct target.

Are you sure this is not redundant since a Lean PR adaptation usually comes in to fix these problems.

srghma:
> Are you sure this is not redundant

yes, it was already done seems like. Tnx. Closing

PR #1792, opened 2026-05-05
chore: deprecate `Batteries.RBMap` declarations

kim-em:
This PR deprecates every public declaration in `Batteries.Data.RBMap` (and the companion `WF`, `Lemmas`, `Alter`, `Depth` modules) in favour of `Std.TreeMap` from the standard library. The header docstring of `Basic.lean` already pointed at `Std.TreeMap` as the recommended replacement; this makes the deprecation visible to users via `@[deprecated]` attributes.

A survey of downstream consumers — Aesop, Mathlib, ProofWidgets4, lean4lean, lean4checker, lean4export, mm0 — found zero remaining uses of `RBMap`/`RBSet`/`RBNode`/`RBColor`, so deprecating now should not disturb the wider ecosystem. Each file gets a `set_option linter.deprecated false` so that internal cross-references between the now-deprecated declarations don't produce noise.

Follows the precedent from #1750 (deprecation of `String.Legacy`).

🤖 Prepared with Claude Code


kim-em:
@fgdorais has an alternative plan here: I'm happy to hold off for a bit until that has been discussed.

BatteriesRecycling/README.md
```diff
@@ -0,0 +1,5 @@
+# Batteries Recycling
+
+This is a collection of modules that have been deprecated due to obsolescence.
+
+These modules are no longer actively maintained but you can still import them if you need to.
```
eric-wieser, line 5: Don't we still want them to be tested?

https://github.com/leanprover-community/batteries/pull/1846 attempts to fix this.

fgdorais: Who is going to fix them if testing fails?

eric-wieser: We can always delete them if the tests fall. There's no point keeping code around that doesn't compile

fgdorais: Not until the modules they came from are fully removed.

fgdorais: (That's also the time they are expected to be deleted.)

eric-wieser: To be clear, we have three options here:
1. We keep these modules compiling until a suitable deprecation period elapses
2. We delete these modules as soon as they break in any way
3. We bury our head in the sand and pretend that the modules continue to compile

Without the tests, we are doing 3. I don't have a strong opinion between 1 and 2, but we can make that decision once things start breaking.

thorimur: I believe that generally deprecated declarations should continue to be available in working form while they still exist, so that downstream users can get a nice error message for a period of time. This seems like the point of deprecation in the first place to me :)

I.e. I'd support 1.

thorimur: Also I think the answer to "who will fix them if they break" is "whoever would have fixed them had they not yet been deprecated". 😁

fgdorais: Unfortunately, this isn't likely to happen in this case.

PR #1793, opened 2026-05-05
chore: deprecate the SatisfiesM / MonadSatisfying API

kim-em:
This PR deprecates the entire `SatisfiesM` / `MonadSatisfying` API and the associated lemmas in `Batteries.Data.{Array,List,Vector}.Monadic`. A downstream audit found no real users — mathlib4 only references `Batteries.Classes.SatisfiesM` as a *forbidden* import in its `DirectoryDependency` linter; lean4 core has two informational comments and nothing else.

The size-preservation lemmas can't be cleanly restated without `SatisfiesM`: the obvious functorial form `Array.size <$> Array.mapM f as = pure as.size` is false in non-trivial monads (when `mapM` short-circuits, `size <$>` propagates the failure). They genuinely express Hoare triples — the modern replacement is `Std.Do.Triple` with the `mvcgen` tactic.

The `Vector.{mapM,mapIdxM,mapFinIdxM}_mk` lemmas redirect to upstream `Vector.toArray_*`, which subsumes the same content without `MonadSatisfying`. Other deprecations carry a generic message pointing to `Std.Do.Triple`. Each file sets `linter.deprecated false` internally so the deprecated declarations can still call each other.

🤖 Prepared with Claude Code


kim-em:
@fgdorais has an alternative plan here: I'm happy to hold off for a bit until that has been discussed.

PR #1811, opened 2026-05-18
feat: add basic byte order support

fgdorais:


wrenna-robson:
This looks basically sensible to me and useful to have. However, I would feel more comfortable if there were some checks on all the definitions you've made - either with appropriately chosen theorems, or I suppose an addition to test suites. The actual endianness check feels a little hacky but so does every endianness check I've ever seen, so probably no issue.

fgdorais:
I have no idea why `precompileModules = true` breaks the `seq_focus` and `simp_trace` tests.

fgdorais:
I've added a bunch of theorems. I probably missed some. I still have no clue why the tests are broken.

kim-em:
Could you say something about what this is for? I wasn't aware that endian-ness was a live issue anymore. :-)

fgdorais:
I meant this to stay WIP. While my primary purpose was to test whether `extern c inline` would even work here, there are some applications. For example, to properly encode/decode a utf16 or utf32 string, one must pay attention to the byte-order mark (BOM) and adapt. It's also important to properly verify some transmission protocols. You're right though, this is not a concern in many higher level languages where low-level stuff is handled behind the scenes by the compiler. This is true for Lean except for that verification is important in order to trust the compiler.

PR #1814, opened 2026-05-22
feat: cross-reference earlier `proof_wanted` via `❰foo❱`

kim-em:
This PR makes `proof_wanted` leave a private placeholder `def name args : ProofWanted T := ⟨⟩` in the environment (previously the declaration was elaborated inside `withoutModifyingEnv` and discarded), and adds `❰foo❱` syntax for referring to earlier `proof_wanted`s in the same file. Inside a `proof_wanted`'s statement, `❰foo❱` desugars to a fresh hypothesis binder of `foo`'s statement (repeated references share one binder), so the recorded type is `foo.Stmt → …`. No axioms or sorries are introduced; `example : 17 = 37 := env_trace` rightly fails to typecheck because `env_trace : ProofWanted (17 = 37)`, not `17 = 37`.

`ProofWanted α` is a Unit-like wrapper and `ProofWanted.Stmt p` is a reducible accessor that exposes the statement as a type (never as an inhabitant). The bracket is only valid inside a `proof_wanted` statement; outside it errors. Brackets may not appear in binder types (they would reference an out-of-scope hypothesis).

Test coverage in [BatteriesTest/proof_wanted.lean](BatteriesTest/proof_wanted.lean): the original "no unused-var warning" and "not a proposition" cases plus forward references, deduplication, distinct hypotheses for distinct brackets, namespacing, the error paths, and a soundness check that a `proof_wanted` can't be used as a proof of its statement.

Parametrised `proof_wanted`s work too: for `foo : (n : Nat) → ProofWanted (P n)`, `❰foo❱ x` desugars to `h_foo x` where the generated hypothesis binder is `h_foo : ∀ (n : Nat), (foo n).Stmt`. The implementation uses `forallTelescope` to strip `foo`'s Π binders and builds the parameter-binder syntax directly (using `@foo.{us}` to keep implicits explicit, so the syntax round-trips). Implicit, strict-implicit, and instance binders all preserve their `BinderInfo`.

🤖 Prepared with Claude Code, reviewed by Codex.

### DIFF
--- a/Batteries/Util/ProofWanted.lean
+++ b/Batteries/Util/ProofWanted.lean
@@ -5,6 +5,7 @@ Authors: David Thrane Christiansen, Kim Morrison
 -/
 module
 
+public import Batteries.Tactic.Lint.Misc
 public meta import Lean.Elab.Command
 public meta import Lean.Elab.Term
 public meta import Batteries.Lean.Syntax
@@ -28,7 +29,7 @@ structure ProofWanted (α : Sort u) : Type where
 
 /-- Reducible accessor so a binder of type `ProofWanted.Stmt foo` reduces to `foo`'s
 statement. Used by the desugaring of `❰foo❱`. -/
-@[reducible]
+@[reducible, nolint unusedArguments]
 def ProofWanted.Stmt {α : Sort u} (_ : ProofWanted α) : Sort u := α
 
 end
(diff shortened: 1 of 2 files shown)


Batteries/Util/ProofWanted.lean
```diff
@@ -1,46 +1,153 @@
...
+placeholder `def name binders : ProofWanted T := ⟨⟩` — a sound, file-local trace. Inside another
+`proof_wanted` statement, the bracket syntax `❰foo❱` lets you assume an earlier `proof_wanted`.
+The brackets desugar to fresh hypothesis binders, so the wanted statement honestly records the
+dependency: "if `foo` holds, then …" — no axioms or sorries are introduced.
+-/
+
+/-- Wrapper recording that a `proof_wanted` of statement `α` has been declared. -/
+structure ProofWanted (α : Sort u) : Type where
+  /-- The trivial constructor; carries no information. -/
+  mk ::
```
fgdorais, line 32: Does this need to be a trivial structure? What goes wrong with `def ProofWanted (α : Sort u) := Unit`?

Batteries/Util/ProofWanted.lean
```diff
@@ -1,46 +1,153 @@
...
+resolve names within the current file, since the placeholders are `private`.
 
 Typical usage:
 ```
-@[simp] proof_wanted empty_find? [BEq α] [Hashable α] {a : α} :
-    (∅ : HashMap α β).find? a = none
+proof_wanted seventeen_eq_thirtyseven : 17 = 37
+
+proof_wanted seventeen_in_fifty :
+    (⟨17, by rw [❰seventeen_eq_thirtyseven❱]; decide⟩ : Fin 50) = 0
```
fgdorais: Shouldn't examples be statements that one would actually want? I doubt anyone would actually want a proof of `17 = 37` in a general context.

Batteries/Util/ProofWanted.lean
```diff
@@ -1,46 +1,153 @@
...
 -/
 module
 
-public meta import Lean.Elab.Exception
+public import Batteries.Tactic.Lint.Misc
 public meta import Lean.Elab.Command
+public meta import Lean.Elab.Term
+public meta import Batteries.Lean.Syntax
+
+@[expose] public section
```
fgdorais: Probably could be refined.

fgdorais (commented):
This is a great improvement! Thanks.

Batteries/Util/ProofWanted.lean
kim-em: I'd have the opposite attitude: what does this have to do with `Unit`? :-) It is not helpful for anyone to use the fact this is the canonical 1-element type

fgdorais: Fair attitude. :-) Maybe rename the constructor `intro`? Actually, with the forthcoming `construction_wanted`, I think `mk` is fine. Unless you have a better idea.

Batteries/Util/ProofWanted.lean
kim-em: I've replaced these examples.

fgdorais (approved):

PR #1817, opened 2026-05-24
feat: allow a body on `proof_wanted` referencing other `proof_wanted`s

kim-em:
This PR (stacked on https://github.com/leanprover-community/batteries/pull/1814) extends `proof_wanted` to optionally take a body: `proof_wanted name binders : T := body`. Bracket references `❰foo❱` in the body are collected alongside those in the statement, deduplicated, and become extra hypothesis binders on the recorded `ProofWanted`. The body is type-checked against the statement via a `have`; the placeholder still carries no proof.

Diff against #1814's tip: https://github.com/leanprover-community/batteries/compare/proof_wanted...proof_wanted_body

A declaration that supplies a body but uses no `❰…❱` reference is rejected; such a body is a complete proof, and the declaration should be a `theorem`. The error includes a `Try this:` suggestion replacing it with a `theorem`.

Test coverage in [BatteriesTest/proof_wanted.lean](BatteriesTest/proof_wanted.lean) under "Bodies": body-only references that become binders; a reference shared between statement and body that is deduplicated; a body whose type check fails; a body using a parametrised `proof_wanted`; and the no-reference case showing the rendered `Try this:`.

🤖 Prepared with Claude Code, reviewed by Codex.

### DIFF
--- a/Batteries/Util/ProofWanted.lean
+++ b/Batteries/Util/ProofWanted.lean
@@ -161,12 +161,17 @@ private def validateProofWantedRef (nm : Name) (stx : Syntax) :
       let isCls : Bool := match αInst.getAppFn with
         | .const n _ => Lean.isClass env n
         | _ => false
-      -- Build syntax `∀ <foo's binders>, ProofWanted.Stmt (@foo.{us} arg1 ... argN)`. The `@`
-      -- keeps every implicit argument explicit so the syntax round-trips; the universe
-      -- annotations bind to `nm`'s own level params (auto-bound for the enclosing `proof_wanted`).
+      -- Build syntax `∀ <foo's binders>, ProofWanted.Stmt (@foo.{_, …} arg1 ... argN)`. The `@`
+      -- keeps every implicit argument explicit so the syntax round-trips. Each universe level is a
+      -- hole `_` rather than `nm`'s own level name: a named level would auto-bind to a fresh,
+      -- distinct param of the enclosing `proof_wanted`, leaving the hypothesis monomorphic at a
+      -- universe that can never match the use site (e.g. referencing `exists_ulift.{w}` at universe
+      -- `u`). A hole instead unifies with whatever universe the reference is used at. The single
+      -- remaining gap is using one reference at two different universes, which no single binder can
+      -- express; see the `ulift`/`TODO` tests in `BatteriesTest/proof_wanted.lean`.
       let fooIdent := mkIdent nm
       let levelStxs : Array (TSyntax `level) ←
-        info.levelParams.toArray.mapM fun u => `(level| $(mkIdent u):ident)
+        info.levelParams.toArray.mapM fun _ => `(level| _)
       let argIdents : Array (TSyntax `term) ← fvars.mapM fun fvar => do
         let decl ← fvar.fvarId!.getDecl
         return ⟨mkIdent decl.userName⟩
--- a/Batteries/Util/ProofWanted.lean
+++ b/Batteries/Util/ProofWanted.lean
@@ -11,8 +11,6 @@ public meta import Lean.Elab.Term
 public meta import Lean.Meta.Tactic.TryThis
 public meta import Batteries.Lean.Syntax
 
-@[expose] public section
-
 /-!
 # The `proof_wanted` command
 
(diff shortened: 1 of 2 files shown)


BatteriesTest/proof_wanted.lean
```diff
@@ -92,6 +92,68 @@ of sort `Type`
 -/
 #guard_msgs in proof_wanted bad_not_a_prop (x : Nat) : Nat
 
+/-! ## Bodies for `proof_wanted`
+
+A `proof_wanted` may be given a body, which must reference at least one earlier `proof_wanted`
+via `❰…❱` (otherwise the body is a complete proof and should be a `theorem`). -/
+
+private def Foo : Nat := 17
+proof_wanted Foo_eq_37 : Foo = 37
```
fgdorais:
```suggestion
private def foo : Nat := 17
proof_wanted foo_eq_37 : Foo = 37
```

BatteriesTest/proof_wanted.lean
```diff
@@ -92,6 +92,68 @@ of sort `Type`
...
+proof_wanted thirtyseven_eq_41 : (37 : Nat) = 41
+
+/-! A body that uses `❰…❱` to refer to earlier `proof_wanted`s works. The body is type-checked
+against the statement and the placeholder still has type `ProofWanted …`. Brackets that appear
+only in the body become extra hypothesis binders just as if they had appeared in the statement. -/
+proof_wanted Foo_eq_41 : Foo = 41 := by
+  rw [❰Foo_eq_37❱]; exact ❰thirtyseven_eq_41❱
```
fgdorais:
```suggestion
proof_wanted foo_eq_41 : foo = 41 := by
  rw [❰foo_eq_37❱]; exact ❰thirtyseven_eq_41❱
```

fgdorais (commented):
Tests should also follow naming conventions.

BatteriesTest/proof_wanted.lean
eric-wieser, line 1: Can you add a test along the lines of

```lean4
theorem exists_ulift.{u} : ∃ x : ULift.{u} Nat, x.down = 1 := sorry

theorem exists_ulift'.{u, v} : ∃ x : ULift.{u} Nat, ∃ y : ULift.{v} Nat, x.down = y.down := by
  obtain ⟨x, hx⟩ := exists_ulift.{u}
  obtain ⟨y, hy⟩ := exists_ulift.{v}
  exact ⟨x, y, hx.trans hy.symm⟩
```
with the sorry replaced with `proof_wanted`? I kind of expect this to fail; if it does, please add a test with a TODO comment to flag it to other reviewers / future contributors.

eric-wieser (changes requested):
Seems to be a bug with universe scoping:

```lean4
proof_wanted exists_ulift.{w} : ∃ x : ULift.{w} Nat, x.down = 1

/--
error: Tactic `rcases` failed: scrutinee has type
  ULift.{w, 0} Nat
of sort `Type w` but is expected to have type
  ULift.{u, 0} Nat
of sort `Type u`
-/
#guard_msgs in
proof_wanted exists_ulift'.{u} : ∃ x : ULift.{u} Nat, x.down = 1 := by
  obtain ⟨x : ULift.{u} Nat, hx⟩ := ❰exists_ulift❱
  exact ⟨x, y, hx.trans hy.symm⟩
```

Batteries/Util/ProofWanted.lean
```diff
@@ -215,10 +243,21 @@ def elabProofWanted : CommandElab
...
+        private def $name $args* $extraBinders* : ProofWanted (($res' : Prop)) := ⟨⟩
+        end)
+    | some body' =>
+      elabCommand <| ← `(
+        section
+        set_option linter.unusedVariables false
+        private def $name $args* $extraBinders* : ProofWanted (($res' : Prop)) :=
+          have : ($res' : Prop) := $body'
+          ⟨⟩
+        end)
```
eric-wieser, line 267: Given we actually have a well-typed body here, why not keep it? This allows downstream code to potentially use the ProofWanted if they locally prove its hypotheses!

eric-wieser: This could look like changing `ProofWanted P` to `Option P`

fgdorais: It seems like you're asking for a vastly improved `sorry` here. That sounds like a great idea but maybe too much for this PR. Perhaps it's better to see how the ship floats before adding this.

fgdorais (approved):
Happy to merge once issues raised by Eric are addressed.

PR #1822, opened 2026-05-28
feat: add disable hint to environment linter output

### LINKED ISSUE #1468: Environment linters should remind users how to disable them
Similarly to how syntax linters in Lean core do it: https://github.com/leanprover/lean4/blob/206eb73cd9de4b58abbb2eae20a41fd9b7af187b/src/Lean/Linter/Basic.lean#L75

The relevant code seems to be at: https://github.com/leanprover-community/batteries/blob/6f7d05f4955eb5af9799d828b697251bb2c9c0b8/Batteries/Tactic/Lint/Frontend.lean#L194

cf. [#FLT > unused argument linter issue @ 💬](https://leanprover.zulipchat.com/#narrow/channel/416277-FLT/topic/unused.20argument.20linter.20issue/near/543106977)

(It might also be good for the documentation of environment linters to mention the term "environment linter" and disucss the distinction with "syntax linters" as well).

yurekami:
## What
Add a note to environment linter output telling users how to suppress warnings with `@[nolint linter_name]`, similar to how syntax linters in Lean core remind users about `set_option`.

## Why
Users encountering linter warnings have no indication of how to disable them. Lean core's syntax linters already do this (via `logLint` in `Lean/Linter/Basic.lean`), but batteries' environment linters do not.

## Scope
Single-line change in `formatLinterResults` — the disable hint appears once per failing linter in the comment block header, not per-warning.

---
Closes #1468.


wrenna-robson:
This is a useful contribution I think, but at minimum I think the error message could do with some tweaking - in particular I prefer

"This linter can be disabled with `set_option ___ false`" to "Use ____ to ignore a declaration."

What is a declaration? And - this is subtle - I think we should avoid using the imperative mood here. "can" is not an instruction. "Use" is.

yurekami:
Good points — updated in a81a85d. The hint now reads:

```
This linter can be disabled with `@[nolint linter_name]`.
```

Matches Lean core's indicative phrasing and drops "declaration".

fgdorais:
Several tests need to be updated. Run `lake test` to see which.

yurekami:
Thanks! In the end only `lint_simpNF_respectTransparency.lean` needed updating — it's the only test that captures a failing linter's report through `#guard_msgs` (the others drop the error or expect no output). Fixed in e34c1f6; `lake test` passes locally.

PR #1830, opened 2026-06-01
refactor: consolidate linters

fgdorais:
This PR proposes to consolidate all linters in `Batteries.Linters`.

Previously, environment linters were in `Batteries.Tactic.Lint` and only syntax linters were in `Batteries.Linters`. The distinction is based on implementation instead of function.

### DIFF
--- a/Batteries.lean
+++ b/Batteries.lean
@@ -108,7 +108,12 @@ public import Batteries.Tactic.HelpCmd
 public import Batteries.Tactic.Init
 public import Batteries.Tactic.Instances
 public import Batteries.Tactic.Lemma
-public import Batteries.Linter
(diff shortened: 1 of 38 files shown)


kim-em:
This seems wrong. This is an environment linter, so belongs in `Batteries.Tactic.Lint` where it is now, not in `Batteries.Linter`, which is the syntax linters. We can reorganize, but we shouldn't do it piece-meal without discussion.

fgdorais:
This was indeed the start of a process to move all (syntax and environment) linters at the same place. The idea is that function takes precedence over implementation. However, I agree, there should be a discussion.

fgdorais:
awaiting-review

PR #1831, opened 2026-06-02
feat: don't consider private names autogenerated in `isAutoDecl`

thorimur:
This PR modifies `isAutoDecl` so that it does not consider private names autogenerated. Linters which still want to also ignore private declarations may call `isPrivateOrAutoDecl` instead.

Whether a linter should apply only to the public interface or to every declaration for overall code quality is a question different linters may answer differently. Linter authors adapting to this PR should consider carefully whether private declarations are indeed out of scope, and should instead first consider (if the linter is name-based) whether `privateToUserName` needs to be run at some point, or whether exemptions on another basis need to be made (e.g. for private declarations created by specific commands, like those created by `proof_wanted`).

This PR also must modify `dupNamespace` slightly by running `privateToUserName` (and adjusting the message for readability), and by ignoring declarations with a return type of `ProofWanted _` or `DefWanted _` in the `unusedArguments` linter.

---

Most of the mathlib adaptations for this PR ~~are present in~~ have already been merged in leanprover-community/mathlib4#41470. The `defsWithUnderscore` linter just needs to use `isPrivateOrAutoDecl`. Ideally, this would not be necessary, but this is left as a TODO since many private declarations have underscores.

- [x] depends on: #1835
- [x] depends on: #1889
- [x] depends on: #1892
- [x] depends on: #1895


Batteries/Tactic/Lint/Misc.lean
```diff
@@ -162,6 +169,7 @@ with rfl when elaboration results in a different term than the user intended. -/
   isFast := true
   test declName := do
     if ← isAutoDecl declName then return none
+    if Linter.isDeprecated (← getEnv) declName then return none
```
eric-wieser, line 161: Was this intended to be in this PR? I don't see it in the description, and it would probably make sense to split it.

thorimur: It's only not in the description because there is no description! XD But I will split it off. I've only just pushed this—I want to see if I can get to green, then I'm going to split as appropriate.

thorimur: Success! Split off into #1892 .

thorimur:
awaiting-review

grunweg:
The diff looks great now. Can you add a test that e.g. the unusedArguments linter checks private declarations also (and perhaps similarly for other linters)?

Batteries/Tactic/Lint/Misc.lean
```diff
@@ -44,6 +44,7 @@ their value, and allow arguments starting with `_` to be unused. -/
       | _ => return none
     if val.hasSorry || ty.hasSorry then return none
     forallTelescope ty fun args ty => do
+      if ty.isAppOfArity' `ProofWanted 1 || ty.isAppOfArity' `DefWanted 1 then return none
```
eric-wieser, line 49: Honestly I think I'd consider this a bug in proof_wanted, and that the `ProofWanted` wrapper should be around the entire term, Then the linter would not fire.

thorimur: It seems like it's at least intentional, since this is the behavior described in the module docstring for (`proof`|`def`)`_wanted` :)

Tbh I'm not really willing to insist on (or suggest) a breaking change for code I'm not familiar with (I can imagine consumers might be relying on this behavior), but I can leave a comment here explaining the situation. If you want to make an issue or PR about this for `proof_wanted` I can link that too :)

thorimur:
@grunweg I've added a test for `unusedArguments` (to show it does lint private declarations) and upgraded the ones for `docBlame` (to show it *doesn't*, as intended)...but I can't find any existing tests for the other linters? 👀

Batteries/Tactic/Lint/Misc.lean
thorimur: (Provisional explanatory comment added, at least!)

eric-wieser: Thanks!

eric-wieser (approved):

grunweg (approved):

thorimur:
Looking at the log, the new errors seem to be completely unrelated.

Would it be alright to ping this PR, @fgdorais ? It's been sitting for a while (green on mathlib for a while before merging the latest `batteries/main`) and has been reviewed. :) There are a few small adaptations on the nightly testing branch; happy to do whatever I have to there, if anything.

grunweg:
@wrenna-robson Would you like to take a look?

wrenna-robson:
I won't have a chance to look today - starting a new job. I can try and remember to take a look tonight.

thorimur:
@wrenna-robson Congrats on the new job, and no rush! This has waited a couple months already, it can also wait a few more days until you're settled if that winds up being more convenient for you :)

wrenna-robson:
LGTM

PR #1834, opened 2026-06-02
feat: allow arguments starting with `_` to be unused in `unusedArguments` linter

thorimur:
This PR makes the `unusedArguments` environment linter consistent with the unused variables syntax linter, and does not lint against arguments starting with `_`.

---

Note: the `unusedArguments` linter does not lint theorems. I've updated the tests to be meaningful, but I'm not sure if this is intentional.

Mentioned on zulip [here](https://leanprover.zulipchat.com/#narrow/channel/348111-batteries/topic/unused.20anonymous.20argument/near/599546336), but no response.


JovanGerb:
Thanks for doing this, the code looks good to me!

fgdorais (approved):

PR #1837, opened 2026-06-03
chore: use core checks/error messages in the `env_linter` attribute

chenson2018:
As pointed out at the Meta Café, these checks have an existing implementation in core we can reuse. This also adds a missing test corresponding to the check that `env_linter` is used as a global attribute.


Batteries/Tactic/Lint/Basic.lean
Rob23oba: I guess we should also use `throwAttrMustBeGlobal` here.

chenson2018: I've added this as well, thanks!

chenson2018: (I also added some tests, as this behavior wasn't previously being checked.)

Batteries/Tactic/Lint/Basic.lean
```diff
@@ -124,12 +124,7 @@ initialize registerBuiltinAttribute {
     `Linter.mk` is already `meta` and thus will likely cause an error anyway, but the explicit
     instruction to mark this declaration `meta` might help the user resolve that and similar
     errors. -/
-    let isPublic := !isPrivateName decl; let isMeta := isMarkedMeta (← getEnv) decl
```
eric-wieser, line 137: WHhat happened to the `public` check? Was it unnecessary, or should we also be calling `Lean.ensureAttrDeclIsPublic`?

chenson2018: `ensureAttrDeclIsMeta` also calls `ensureAttrDeclIsPublic`, so it checks for both.

fgdorais: Let's at least add a comment to that effect in case this behavior changes.

chenson2018: I've added a mention of this in the above comment. If the behavior changes, the updated tests should also fail as well.

Batteries/Tactic/Lint/Basic.lean
```diff
@@ -112,24 +112,20 @@ initialize registerBuiltinAttribute {
...
+    unless kind == .global do throwAttrMustBeGlobal `env_linter kind
     let shortName := decl.updatePrefix .anonymous
     if let some (declName, _) := (batteriesLinterExt.getState (← getEnv)).find? shortName then
       Elab.addConstInfo stx declName
       throwError
         "invalid attribute `env_linter`, linter `{shortName}` has already been declared"
     /- Just as `env_linter`s must be `global`, they also must be accessible from `#lint`, and thus
-    must be `public` and `meta`.
+    must be `public` and `meta`. `ensureAttrDeclIsMeta` checks for both of these, additionally
+    calling `ensureAttrDeclIsMeta`.
```
fgdorais:
```suggestion
    calling `ensureAttrDeclIsPublic`.
```

Batteries/Tactic/Lint/Basic.lean
fgdorais: I fixed a typo in the comment.

fgdorais:
The Mathlib fail is fake, we just need to wait for Mathlib to catch up.

PR #1838, opened 2026-06-04
feat: add `Environment.isAutoDecl` and make `isAutoDecl` monad-generic

thorimur:
This PR adds the pure `Lean.Environment.isAutodecl (env : Environment) (decl : Name)` and makes `Batteries.Tactic.Lint.isAutoDecl` monad-generic, assuming `MonadEnv`.

The motivation here is the upcoming migration of many `env_linter`s to syntax linters, which run in `CommandElabM` (requiring an explicit `liftCoreM`) and will still want to detect if a declaration is autogenerated or not.

The reason we have both of these is that `MonadEnv` also requires the ability to modify the environment, which we don't need. Hence the pure function may be useful in certain contexts, such as in pure fields of environment extensions or e.g. `ImportM` (which only allows reading the environment).

### DIFF
--- a/Batteries/Tactic/Lint/Basic.lean
+++ b/Batteries/Tactic/Lint/Basic.lean
@@ -37,13 +37,12 @@ Returns true if `decl` is an automatically generated declaration.
 Also returns true if `decl` is an internal name or created during macro
 expansion.
 -/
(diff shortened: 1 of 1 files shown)


wrenna-robson:
This seems to me to be sensible with no obvious issues.

PR #1846, opened 2026-06-08
fix: ensure that CI actually runs `BatteriesRecycling`

eric-wieser:
Maybe this was already tested somewhere else, but I didn't see it.

### DIFF
--- a/lakefile.toml
+++ b/lakefile.toml
@@ -1,7 +1,7 @@
 name = "batteries"
 testDriver = "BatteriesTest"
 lintDriver = "runLinter"
-defaultTargets = ["Batteries", "runLinter"]
+defaultTargets = ["Batteries", "BatteriesRecycling", "runLinter"]
 
 [leanOptions]
 linter.missingDocs = true


thorimur:
I think we need to either introduce `BatteriesRecycling.lean` for linting to succeed, or to pass the modules explicitly to `runLinter`. (Ideally `runLinter` should be lake-managed, buuuut...)

Also I'm not sure we want this to be a default target. Is there a way to run it in batteries CI without forcing others to build it?

eric-wieser:
#1847 fixes the linting error.

eric-wieser:
> Also I'm not sure we want this to be a default target. Is there a way to run it in batteries CI without forcing others to build it?

Which "others" are we concerned about here?

thorimur:
> > Also I'm not sure we want this to be a default target. Is there a way to run it in batteries CI without forcing others to build it?
> 
> Which "others" are we concerned about here?

Yeah, this worry was probably misplaced...I was worried that some downstream users might be building all the default targets of upstream things in some circumstances, but this doesn't seem to be the case.

fgdorais:
I'm closing pending further discussion on Zulip. Reopening is always an option!

eric-wieser:
Do you meant the reviewer thread, or were you thinking of https://github.com/leanprover-community/batteries/pull/1792#discussion_r3376522641?

fgdorais:
Nonspecific: let's revive if _any_ relevant discussion discovers this turns out to be as a necessity.

(Note that all these files are currently checked by CI, so there is no urgency.)

eric-wieser:
I'm going to briefly open and close in order to verify that CI currently passes (in the sense of my config working)

PR #1850, opened 2026-06-11
feat: consume lake target specs in runLinter

eric-wieser:
This provides better behavior for packages which do not provide a top-level `Package.lean`.

The semantics of `lake exe runLinter Package` are now "import everything matched by the globs into a single environment, and run the linters there". Previously they were "import each file in isolation".

The linting now also respects the `isLocalModule` filters provided by lake to decide which warnings to attribute to the current project.


eric-wieser:
cc @thorimur, @wkrozowski

wkrozowski:
Hmm, I am interested if these changes would also make sense for the core environment linting architecture that is highly inspired by batteries.

thorimur:
I suspect there is also further lake-ification that could be done (beyond the scope of this PR), which might be easier to do in core (though I'm not sure).

eric-wieser:
You could imagine in the extreme case that `lake lint foo` is really just sugar for `lake build foo:lint`. I think this is the kind of thing that should be designed in core though; the intent of this PR is to provide a reasonable middle ground that has correct behavior for "head"-less projects.

thorimur:
Ideally that sort of thing would also let us skip applying certain linters when their source modules were not out-of-date, too. (Doing it all in one big environment might still be desirable for performance, though, to avoid loading a bunch of separate environments...so to get any real benefit we'd have to tag `env_linter`s as "global" like simpNF or "local" to modules/declarations, and then for local linters just skip the `test decl` if the source module of `decl` were up to date as per lake. Or optionally run the local ones during build, or something?)

I'll mention also that `buildIfNeeded` here is basically just trying to be a lake fetch, and it would be nice if it became one. 😁

eric-wieser:
Probably in the meantime it could quite easily be replaced with `lake --no-build`, assuming that exists. That seems like an orthogonal refactor to this PR though.

Batteries/Tactic/Lint/Frontend.lean
```diff
@@ -256,16 +256,20 @@ def getDeclsInCurrModule : CoreM (Array Name) := do
...
   pure $ (← getEnv).constants.map₁.fold (init := ← getDeclsInCurrModule) fun r k _ => r.push k
 
-/-- Get the list of all declarations in the specified package. -/
-def getDeclsInPackage (pkg : Name) : CoreM (Array Name) := do
+/-- Get the list of all declarations in modules satifying `includeModule`. -/
+def getDeclsFilterModuleName (includeModule : Name → Bool) : CoreM (Array Name) := do
   let env ← getEnv
   let mut decls ← getDeclsInCurrModule
-  let modules := env.header.moduleNames.map (pkg.isPrefixOf ·)
+  let modules := env.header.moduleNames.map (includeModule)
```
fgdorais, line 263:
```suggestion
  let modules := env.header.moduleNames.map includeModule
```

scripts/runLinter.lean
```diff
@@ -59,19 +50,17 @@ config settings are determined by the default values for fields of `LinterConfig
 Throws an exception if unable to parse the arguments.
 Returns `none` for the specified module if no modules are specified.-/
 def parseLinterArgs (args : List String) :
-    Except (List String) (LinterConfig × List Name) :=
+    Except (List String) (LinterConfig × List String) :=
   go {} [] args
 where
   /-- Traverses the list, handling the non-flag elements as modules and erroring if parsing fails. -/
-  go (parsed : LinterConfig) (mods: List Name) : List String → Except (List String) (LinterConfig × List Name)
+  go (parsed : LinterConfig) (specs: List String) : List String → Except (List String) (LinterConfig × List String)
```
fgdorais, line 57:
```suggestion
  go (parsed : LinterConfig) (specs : List String) : List String → Except (List String) (LinterConfig × List String)
```

Batteries/Tactic/Lint/Frontend.lean
```diff
@@ -256,16 +256,20 @@ def getDeclsInCurrModule : CoreM (Array Name) := do
...
   return env.constants.map₁.fold (init := decls) fun decls declName _ =>
     if modules[env.const2ModIdx[declName]?.get!]! then
       decls.push declName
     else decls
 
+/-- Get the list of all declarations in the specified package. -/
+def getDeclsInPackage (pkg : Name) : CoreM (Array Name) :=
+  getDeclsFilterModuleName (pkg.isPrefixOf ·)
```
fgdorais, line 271:
```suggestion
  getDeclsFilterModuleName pkg.isPrefixOf
```

fgdorais (commented):
Style.

PR #1851, opened 2026-06-11
feat: produce better messages from the simpNF linter

eric-wieser:
Sometimes this linter would complain that `X` would simplify to `X`. This makes sure that the difference is visible.

This also improves the line-wrapping of long messages by using `indentD`.


wrenna-robson:
Seems sensible to me. I am a little wary of the fact that this is two separate changes but they are both good ones.

wrenna-robson:
In consideration I'm happy.

PR #1853, opened 2026-06-15
feat(Data): interleaving lists

YaelDillies:
Define interleaving of lists as an operation.

This will be used to define interleaving polynomials, which in turn are a central concept in the line of work that earned June Huh his 2022 Fields medal.

[Zulip](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Interleaving.20lists/with/604506493)

---

Originally PRed as https://github.com/leanprover-community/mathlib4/pull/39870 but moved to batteries at the request of @ocfnash.


wrenna-robson:
"Interleaves two lists `l₁` and `l₂`, starting with an element of `l₂`."

Obviously this choice is somewhat arbitrary. But I have to say this is counter to what I would expect - intuitively I expect to start with `l₁`. Can you say more about this choice?

wrenna-robson:
I would support both the upstreams you mention by the way - indeed, this is a great reason to do them! Would you be willing to prepare them as PRs?

wrenna-robson:
https://www.symmetricfunctions.com/realRootedInterlacing.htm

Is this a good source for the concept you're referring to?

YaelDillies:
> "Interleaves two lists `l₁` and `l₂`, starting with an element of `l₂`."
> 
> Obviously this choice is somewhat arbitrary. But I have to say this is counter to what I would expect - intuitively I expect to start with `l₁`. Can you say more about this choice?

This is the convention that the literature around real-rooted polynomials uses (assuming the lists are sorted decreasingly). This is motivated by the idea that `l₂` should be the longer list.

> https://www.symmetricfunctions.com/realRootedInterlacing.htm
> 
> Is this a good source for the concept you're referring to?

Certainly a great reference! This is the personal website of @PerAlexandersson, whose repo I am upstreaming this from.

YaelDillies:
I have opened #1854 for `@[mk_iff]`. `existsAndEq` is more complicated to upstream because it depends on Qq. Are you happy for batteries to depend on Qq?

wrenna-robson:
It looks a bit like what you have as "interleave" is called "interlace" there, unless I'm mistaken.

I think for batteries the interleaving list operation should not be defined with reference to this theory in particular. I would strongly prefer, I think, a definition like this:
```
def interleave : List α → List α → List α
  | [], [] => []
  | [], b :: l₂ => b :: l₂
  | a :: l₁, [] => a :: l₁
  | a :: l₁, b :: l₂ => a :: b :: interleave l₁ l₂
  ```
  
  Now, I can also see the argument for:
  ```
def interleave : List α → List α → List α
  | [], [] => []
  | [], b :: l₂ => []
  | a :: l₁, [] => []
  | a :: l₁, b :: l₂ => a :: b :: interleave l₁ l₂
  ```
  
 I think this might be more similar to your definition, though you seem to want to take one element in the cons, nil (or nil, cons using your argument order) case.
 
```
def interleave : List α → List α → List α
  | [], [] => []
  | [], b :: l₂ => []
  | a :: l₁, [] => [a]
  | a :: l₁, b :: l₂ => a :: b :: interleave l₁ l₂
  ```
This last definition is the same as yours with the argument order reversed - but it has the advantage that it doesn't need a termination hint.

I might characterise these as:

"take from the first list, then take from the second list, then repeat - if you ever can't take from one, return the other"
"take from the first list, then take from the second list then repeat - if you can't take from both, stop"
"take from the first list, then take from the second list then repeat - always take from the first but stop if you can't take from the second"

Alternatively I can see the argument for your argument order if "l₁.interleave l₂" is taken to mean "interleave `l₁` into `l₂`, a la intersperse. I really don't like your behaviour of taking just the one element from one list if the other is empty - to me that produces an unsatisfying state of affairs.

So: at minimum, change your definition to
```
def interleave : List α → List α → List α
  | _, [] => []
  | [], a :: _ => [a]
  | b :: l₁, a :: l₂ => a :: b :: interleave l₁ l₂
  ```
  
 `interleave_cons` then becomes a theorem but one fairly easily proved.
 
 I think there's not a common consensus which one of these is right but I don't think yours is the one most people would intuitively reach for.

**Comment by YaelDillies** [comment]
Do you understand why the build fails? It looks like a pre-existing issue surfacing now because I modified the file `Batteries.Data.List`.

**Comment by wrenna-robson** [comment]
@YaelDillies Ah - the Qq dependency will need some discussion I think.

**Comment by wrenna-robson** [comment]
> Do you understand why the build fails? It looks like a pre-existing issue surfacing now because I modified the file `Batteries.Data.List`.

Yes - you have changed the first line of it, removing "-- deprecated_module: ignore", which causes that warning to no longer be ignored. But that was deliberate on the part of the PR that wrote that line: reinstate it and the warning will not cause the build to fail.

**Comment by YaelDillies** [comment]
> It looks a bit like what you have as "interleave" is called "interlace" there, unless I'm mistaken.

I am sitting with @PerAlexandersson right now and he confirms that the naming in the literature is a mess. Different experts in the area use both names for different things, and even the same expert will not use the two names consistently. We get to set a consistent naming convention here.

**Comment by YaelDillies** [comment]
> I can see the argument for your argument order if "l₁.interleave l₂" is taken to mean "interleave `l₁` into `l₂`, a la intersperse.

Yes, this is indeed how the name is supposed to be read.

> I really don't like your behaviour of taking just the one element from one list if the other is empty - to me that produces an unsatisfying state of affairs.

I do not care what `l.interleave l'` is if `l'.length` is neither `l.length` nor `l.length + 1` . To me, this is junk. All I want is the characterisation of `Interleaves`. It is however very important that `[].interleave [a] = [a]`, else there is no easy way to state my characterisation.

**Comment by wrenna-robson** [comment]
> > I can see the argument for your argument order if "l₁.interleave l₂" is taken to mean "interleave `l₁` into `l₂`, a la intersperse.
> 
> Yes, this is indeed how the name is supposed to be read.
> 
> > I really don't like your behaviour of taking just the one element from one list if the other is empty - to me that produces an unsatisfying state of affairs.
> 
> I do not care what `l.interleave l'` is if `l'.length` is neither `l.length` nor `l.length + 1` . To me, this is junk. All I want is the characterisation of `Interleaves`. It is however very important that `[].interleave [a] = [a]`, else there is no easy way to state my characterisation.

Understood. I think the examples of `intersperse`/`intercalate` support your argument order. This suggests to me the following definition:

```
def interleave : List α → List α → List α
  | [], [] => []
  | [], b :: l₂ => b :: l₂
  | a :: l₁, [] => a :: l₁
  | a :: l₁, b :: l₂ => b :: a :: interleave l₁ l₂
  ```
  
  I think this has the behaviour that you want while I think giving the more sensible outputs in your "junk values" range. What do you think?

PerAlexandersson:
I personally want to define interlacing so that Theorem 7.2 in [This book chapter](https://arxiv.org/pdf/1410.6601) is true. The empty list should interlace lists of length 0 and 1.  I agree that the definition is not totally intuitive, unless you have this background.

wrenna-robson:
That makes sense to me. I just want to characterise interlacing in a way that fits your usage but which is also suitable for wider consumers.

PerAlexandersson:
@wrenna-robson  Yes, of course. I discussed with @YaelDillies  about this a bit. So, Wagner defines two properties he calls _interleaves_ and _alternates_, see [my references](https://www.symmetricfunctions.com/realRootedInterlacing.htm). This is essentially splitting into two cases: both list have same length, and one list has one more element compared to the other.  These two separate concepts could be useful, but for questions about polynomials with only real roots, the 'either-or' is a useful concept as well. 

Now, the literature here is very inconsistent and **interlaces** and **interleaves** are somewhat synonyms, but for some authors they mean slightly different things. 

The way I'd implement it, would be to have three named concepts:

- Lists of same length:  a<b<a<b<a<b<a<b
- Lists where the 'largest' has one more element: b<a<b<a<b<a<b<a<b
- Either of the above (as `interleave` above is supposed to represent).

The first two concepts are natural from computer-science perspective, while the third is natural in the world of real-rooted polynomials. 

All three concepts are useful, its only the naming that's the issue IMHO.
[Riffle](https://reference.wolfram.com/language/ref/Riffle.html), alternate, interlace, interleave, intermingle, intertwine, braid, lace (this is taken), entwine?

wrenna-robson:
Well, interleave above should, ideally, *create* a list which is the interleaving of two lists. The predicate should also exist (or instead of?)

YaelDillies:
The operation is useful for the characterisation and also to show that interleaving lists are sorted

wrenna-robson:
After discussion with @fgdorais, we'd quite like you to take a discussion of "interleave" to Zulip - we would like to understand what other consumers of this function might want. I can see various options being sensible.

YaelDillies:
See https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Interleaving.20lists/with/604506493

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,264 @@
...
+
+This file defines interleaving of lists, both as an operation and as a relation.
+-/
+
+public section
+
+namespace List
+variable {α : Type u} {r s : α → α → Prop} {l l₁ l₂ : List α} {a b c : α}
+
+/-- Interleaves two lists `l₁` and `l₂`, starting with an element of `l₂`.
```
eric-wieser: I would have expected it to start with the first list!

YaelDillies: As explained in https://github.com/leanprover-community/batteries/pull/1853#issuecomment-4718348271, the name should be read as "interleave `l₁` *into* `l₂`". Does this make sense?

YaelDillies:
What is this PR waiting for? The two upstreaming tasks would be nice to have, but I hope they are not blocking this PR? There are many thousands of lines of code waiting on `List.Interleaves` to be upstreamed.

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,264 @@
...
+
+This operation is well-behaved only when the length of `l₂` is either the length of `l₁`
```
eric-wieser: I think you should still spell out what "not well behaved" means, i.e whether elements are dropped or kept without further interleaving

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,264 @@
...
+    ∀ {l₁ l₂ : List α}, l₁.length + 1 = l₂.length →
+      (l₁.interleave l₂).reverse = l₁.reverse.interleave l₂.reverse
+  | [], [a], _ => by simp
+  | a :: l₁, b :: l₂, _ => by simp_all [reverse_interleave_of_length_add_one_eq_length]
+
+@[simp]
+theorem interleave_ofFn_ofFn :
+    ∀ {n : Nat} {f g : Fin n → α},
+      interleave (ofFn f) (ofFn g) =
+        ofFn (n := 2 * n) (fun i => if i.val % 2 = 0 then g ⟨i / 2, by lia⟩ else f ⟨i / 2, by lia⟩)
```
eric-wieser: Perhaps better to pull through `finProdFinEquiv` or whatever it's called here?

Batteries/Data/List/Interleave.lean
eric-wieser: I think this might be worth a Zulip poll. Certainly when you generalize to the nary version this is clearly just backwards.

fgdorais: There has been a Zulip discussion and most people said "interleave l₁ into l₂" is counterintuitive. While I respect  all opinions here, I'm afraid I have to use my maintainer foot and request for the longer list to come first.

PS: Yaël, I hate making such decisions. I fully respect your and everyone's opinions. This debate is just slowing down progress and a decisive action is the best way to make things move forward. I picked what seemed to be the most peaceful alternative.

YaelDillies: Again, I don't care about `List.interleave`. However I hope you don't ask me to perform the same change with `List.Interleaves ` (where
this matches the reading order and the literature about real-rooted polynomials)

Batteries/Data/List/Interleave.lean
YaelDillies: `finSumFinEquic`? Unfortunately, it stacks the first Fin on top of the second one instead of interleaving them, so it is useless here

Batteries/Data/List/Interleave.lean
fgdorais: I'm not asking for that but please stop mentioning "real rooted polynomials". My mathematician side fully understands that motive, but my Batteries maintainer side doesn't care much.

YaelDillies: I would be very happy for this material to be contributed to mathlib instead for the record, but @ocfnash redirected me here

eric-wieser: Maybe only the computation makes sense in batteries, and the characteristic belongs in mathlib?

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,281 @@
...
+See `interleaves_iff_length_isChain_interleave` for the connection with `List.interleave`. -/
+inductive Interleaves : List α → List α → Prop
+  /-- The empty list interleaves itself. -/
+  | nil_nil : Interleaves [] []
+  /-- The empty list interleaves any singleton list. -/
+  | nil_singleton (a : α) : Interleaves [] [a]
+  /-- If `l₁` interleaves `b :: l₂` and `a` is related to `b`, then `b :: l₂` interleaves
+  `a :: l₁`. -/
+  | cons_symm ⦃l₁ l₂ : List α⦄ ⦃b : α⦄ (hl : Interleaves l₁ (b :: l₂)) ⦃a : α⦄ (hab : r a b) :
+      Interleaves (b :: l₂) (a :: l₁)
```
YaelDillies: As suggested by @eric-wieser in https://github.com/leanprover-community/batteries/pull/1853#discussion_r3487535117, an option would be to drop `Interleaves` from this PR and instead PR that to mathlib. What do people think? Are we expecting applications for `Interleaves` besides real-rooted polynomials?

Batteries/Data/List/Interleave.lean
YaelDillies: Let me resolve this since the original question was addressed, but I have reposted your remark at https://github.com/leanprover-community/batteries/pull/1853/files#r3487590184

Batteries/Data/List/Interleave.lean
PerAlexandersson: If a new application shows up, perhaps then its worth discussing if a more abstract/general interleaves notion should be put in batteries. Side question: Interleaves for real-rooted, can one use the << notation for that? It would look neat.

YaelDillies: I don't think there is a more general notion of interleaving out there. Perhaps you mean the other conventions (about which list starts and which list ends)?

Regarding the notation, yes certainly we can add it, but I would keep that for interleaving *polynomials* instead, and therefore in mathlib.

eric-wieser: The more general notion is `transpose` as discussed in the Zulip thread.

fgdorais: Uhm. Are you confounding `interleave` with `Interleaves`?

fgdorais: @YaelDillies If you want to split `interleave` from `Interleaves`  then I'm happy to merge the first very quickly. It seems to me that the combination is delaying things,

YaelDillies:
See https://github.com/leanprover-community/mathlib4-nightly-testing/pull/249 for the adaptation.

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,144 @@
...
+
+@[simp]
+theorem interleave_ofFn_ofFn :
+    ∀ {n : Nat} {f g : Fin n → α},
+      interleave (ofFn f) (ofFn g) =
+        ofFn (n := 2 * n) (fun i => if i.val % 2 = 0 then f ⟨i / 2, by lia⟩ else g ⟨i / 2, by lia⟩)
+  | 0, f, g  => by simp
+  | n + 1, f, g => by simp_all [interleave_ofFn_ofFn]; grind
+
+theorem interleave_ofFn_ofFn' :
```
eric-wieser: Perhaps

```suggestion
theorem interleave_ofFn_ofFn_odd :
```
And `even` for the other one?

Batteries/Data/List/Interleave.lean
eric-wieser: No, I mean the one for `Fin (m * n)`

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,144 @@
...
+    ∀ {n : Nat} {f : Fin (n + 1) → α} {g : Fin n → α},
+      interleave (ofFn f) (ofFn g) =
+        ofFn (n := 2 * n + 1)
+          (fun i => if hi : i.val % 2 = 0 then f ⟨i / 2, by lia⟩ else g ⟨i / 2, by lia⟩)
+  | 0, f, g  => by simp
+  | n + 1, f, g => by simp_all [interleave_ofFn_ofFn]; grind
+
+@[simp]
+theorem right_sublist_interleave :
+    ∀ {l₁ l₂ : List α}, l₂.length ≤ l₁.length → l₂ <+ l₁.interleave l₂
```
eric-wieser: Why does this need the inequality hypothesis?

Batteries/Data/List/Interleave.lean
YaelDillies: I considered this but there is no `Even`/`Odd` in sight. I guess what is meant is clear enough anyway

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,144 @@
...
+  | _, [], _ => by simp
+  | a :: l₁, b :: l₂, h => by
+    simp only [cons_interleave]
+    exact .cons _ <| .cons_cons _ <| right_sublist_interleave <| by simpa using h
+
+@[simp]
+theorem left_sublist_interleave {l₁ l₂ : List α} (hl : l₁.length ≤ l₂.length + 1) :
```
eric-wieser: Can we also have the one that concludes `a.interleave b <: c.interleave d` from `a <: c` and `b <: d` or however IsPrefix is spelt?

Batteries/Data/List/Interleave.lean
YaelDillies: Good point, they became unnecessary when I switched over to Wrenna's definition

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,143 @@
...
+    (a :: l₁).interleave l₂ = a :: interleave l₂ l₁ := by rw [interleave]
+
+@[simp] theorem interleaves_perm_append : ∀ {l₁ l₂ : List α}, l₁.interleave l₂ ~ l₁ ++ l₂
+  | [], l₂ => by simp
+  | a :: l₁, l₂ => by
+    rw [cons_interleave]
+    exact ((interleaves_perm_append ..).trans perm_append_comm).cons _
+termination_by l₁ l₂ => l₁.length + l₂.length
+
+protected theorem Perm.interleaves (h₁₃ : l₁ ~ l₃) (h₂₄ : l₂ ~ l₄) :
```
eric-wieser:
```suggestion
protected theorem Perm.interleave (h₁₃ : l₁ ~ l₃) (h₂₄ : l₂ ~ l₄) :
```

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,143 @@
...
+@[simp] theorem countP_interleaves (l₁ l₂ : List α) (p : α → Bool) :
+    (l₁.interleave l₂).countP p = l₁.countP p + l₂.countP p := by
+  simp [interleaves_perm_append.countP_eq]
+
+@[simp] theorem count_interleaves [BEq α] (l₁ l₂ : List α) (a : α) :
+    (l₁.interleave l₂).count a = l₁.count a + l₂.count a := by
+  simp [interleaves_perm_append.count_eq]
+
+@[simp]
+theorem interleave_append_append_of_length_eq_length :
```
eric-wieser: The `join` version would be nice to have too

YaelDillies: What's `join`?

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,143 @@
...
+    ∀ {l₁ l₂ : List α} (_h₁₂ : l₁.length = l₂.length) (l₃ l₄ : List α),
+      (l₁ ++ l₃).interleave (l₂ ++ l₄) = l₁.interleave l₂ ++ l₃.interleave l₄
+  | [], [], _, l₃, l₄ => by simp
+  | a :: l₁, b :: l₂, _, l₃, l₄ => by simp_all [interleave_append_append_of_length_eq_length]
+
+@[simp]
+theorem interleave_append_left_of_length_eq_length (h₁₂ : l₁.length = l₂.length) (l₃ : List α) :
+    (l₁ ++ l₃).interleave l₂ = l₁.interleave l₂ ++ l₃.interleave [] := by
```
eric-wieser: Isn't this

```suggestion
    (l₁ ++ l₃).interleave l₂ = l₁.interleave l₂ ++ l₃ := by
```

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,143 @@
...
+  simp [interleaves_perm_append.count_eq]
+
+@[simp]
+theorem interleave_append_append_of_length_eq_length :
+    ∀ {l₁ l₂ : List α} (_h₁₂ : l₁.length = l₂.length) (l₃ l₄ : List α),
+      (l₁ ++ l₃).interleave (l₂ ++ l₄) = l₁.interleave l₂ ++ l₃.interleave l₄
+  | [], [], _, l₃, l₄ => by simp
+  | a :: l₁, b :: l₂, _, l₃, l₄ => by simp_all [interleave_append_append_of_length_eq_length]
+
+@[simp]
```
eric-wieser: I think an inequality suffices here?

Batteries/Data/List/Interleave.lean
YaelDillies: Ah sure. Not sure it is worth it since that would only work for `_even`, not `_odd`, and `finProdFinEquiv` is in mathlib, bundled as an `Equiv`

Batteries/Data/List/Interleave.lean
eric-wieser: Maybe I mean `flatten`?

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,156 @@
...
+/-!
+# Interleaving lists
+
+This file defines interleaving of lists as an operation.
+-/
+
+public section
+
+namespace List
+variable {α : Type u} {r s : α → α → Prop} {l l₁ l₂ l₃ l₄ : List α} {a b c : α}
```
eric-wieser:
```suggestion
variable {α : Type u} {l l₁ l₂ l₃ l₄ : List α} {a b c : α}
```

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,156 @@
...
+#eval interleave [0] [1, 3] -- [0, 1, 3] -- The first list is too short
+```
+-/
+@[expose]
+def interleave : List α → List α → List α
+  | [], l₂ => l₂
+  | a :: l₁, l₂ => a :: interleave l₂ l₁
+termination_by l₁ l₂ => l₁.length + l₂.length
+
+@[simp] theorem nil_interleave (l₂ : List α) : [].interleave l₂ = l₂ := by rw [interleave]
```
eric-wieser: Perhaps worth an `interleave_eq_nil_iff` lemma, and similarly for `cons`?

Batteries/Data/List/Interleave.lean
```diff
@@ -0,0 +1,156 @@
...
+theorem interleave_append_right :
+    ∀ {l₁ l₂ : List α} (_h₁₂ : l₁.length ≤ l₂.length + 1) (l₃ : List α),
+    l₁.interleave (l₂ ++ l₃) = l₁.interleave l₂ ++ l₃
+  | [], _, _, l₃ => by simp
+  | [a], [], _, l₃ => by simp
+  | a :: l₁, b :: l₂, _, l₃ => by simp_all [interleave_append_right]
+
+theorem interleave_flatten_flatten_of_length_eq_length :
+    ∀ {L₁ L₂ : List (List α)} (h₁₂ : L₁.length = L₂.length),
+      (∀ i : Fin L₁.length, length L₁[i] = length L₂[i]) →
```
eric-wieser: Here you only need this to be true for `i + 1 < length`, right?

Batteries/Data/List/Interleave.lean
YaelDillies: What would the `cons` one state? or did you mean `singleton`?

fgdorais: I also don't understand. This leads me to conclude that the lemma, whatever it is, is perhaps not that useful. It's always possible to add more lemmas later, especially if they come with a concrete use case. I'm marking this final issue as resolved in the interest of merging quickly.

fgdorais (approved):

PR #1854, opened 2026-06-15
feat(Tactic): upstream `@[mk_iff]` from mathlib

YaelDillies:
Motivated by #1853.


wrenna-robson:
I would be happy (and it might be desirable) to add e.g. its usage for IsChain in practice (and other inductive props - your examples seem good).

wrenna-robson:
I had a bit of a think after digging into the file and with some Claude aid tried a refactor. What I'd like to do is avoid duplication with Core.

Having said that - in fact the elephant in the room is that Lean.Meta.MkIffOfInductiveProp exists in Core. But it doesn't actually export the command or the attribute - just the essential plumbing. (And it also has some of the redundancy I've got rid of here.)

I think the best strategy here - given that file *already exists in Core* - is that we consider a change to Core. Ideally they'd refactor as I have here - but at minimum mkIffOfInductivePropImpl should be made public, I think. Then we could add the command and attribute here in Batteries.

YaelDillies:
But the version in core does not properly decapitalise the iff lemma name, right?

wrenna-robson:
> But the version in core does not properly decapitalise the iff lemma name, right?

Well, I think the decapitalization happens in the part that Core just doesn't have. I *think* its mkIffOfInductivePropImpl is the same - or nearly. It's a weird state of affairs.

YaelDillies:
I am interested in things moving forward. Although a core change is possibly the better long term solution, is there any downside to just moving the code from mathlib to batteries for now?

Btw I have fixed the relevant batteries-pr-testing branch but I am waiting on someone to merge the fixes: leanprover-community/mathlib4-nightly-testing#247

wrenna-robson:
That is the nightly tests for 1853, not 1854.

YaelDillies:
One more adaptation PR: https://github.com/leanprover-community/mathlib4-nightly-testing/pull/248

Also brought up exposing `mk_iff_of_inductive_prop` in core on Zulip: https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Expose.20mk_iff_of_inductive_prop/with/604472007

wrenna-robson:
I think in principle I feel this is good to merge. I'd hope that if we do expose this in Core we can upstream the improvements this PR makes to there.

wrenna-robson:
I've altered the namespaces to match what Core does and made public the thing we want them to make public - that should make further upstreaming to Core more or less drag and drop.

wrenna-robson:
You look like you're still have issues with Mathlib though I have to assume this is some kind of imports bug?

YaelDillies:
I just checked and the breakage is completely unrelated to this PR. I assume this is because no one has yet adapted mathlib to recent changes in core

wrenna-robson:
Fiddlesticks. I assume it'll catch up but it'll hold this for now. Sorry.

Batteries/Lean/Name.lean
```diff
@@ -0,0 +1,20 @@
...
+public import Lean.Data.Name
+
+/-!
+# Additional functions on `Lean.Name`.
+-/
+
+public section
+
+/-- Decapitalize the last component of a name. -/
+def Lean.Name.decapitalize (n : Name) : Name :=
```
fgdorais, line 17: Can we fix this?
```lean
#eval Lean.Name.decapitalize `Test.Me.That'sLife == `Test.Me.that'sLife -- true
#eval Lean.Name.decapitalize `Test.Me.Ç'estLaVie == `Test.Me.ç'estLaVie -- false
```

fgdorais: Since core hasn't fully embraced Unicode (and related pitfalls), I think we should use a naming scheme that doesn't require decapitalization such as just `*.iff`. There's no point in supporting a broken scheme.

YaelDillies: > There's no point in supporting a broken scheme.

It's only theoretically broken, right? How many people out there name their Prop-valued structure with a leading unicode character?

IMO it is much more important that we follow mathlib's naming convention, ie `lowerCamelCase_iff`.

> Since core hasn't fully embraced Unicode (and related pitfalls)

WDYM here exactly? Lean core specifies a range of valid characters for use in identifiers, so why can't we just ensure that the decapitalisation function does the right thing to these?

fgdorais: > > There's no point in supporting a broken scheme.
> 
> It's only theoretically broken, right? How many people out there name their Prop-valued structure with a leading unicode character?
> 
> IMO it is much more important that we follow mathlib's naming convention, ie `lowerCamelCase_iff`.

Batteries loves Mathlib. However, some Mathlib conventions don't make much sense to plain Batteries users. If Lean core were to use that convention, that would make a significant motivation. Otherwise, I don't feel the need to support something that's already broken.

> > Since core hasn't fully embraced Unicode (and related pitfalls)
> 
> WDYM here exactly? Lean core specifies a range of valid characters for use in identifiers, so why can't we just ensure that the decapitalisation function does the right thing to these?

I don't think I need to add more here. Look at the range of valid identifier characters that Lean supports. Lean core doesn't currently have case support for all of these.

fgdorais: PS: Mathlib could import UnicodeBasic and there is proper casing support there. (Full disclosure: I own and maintain UnicodeBasic.)

YaelDillies:
Since we can't agree on the naming scheme and I don't need this anymore, closing.

PR #1858, opened 2026-06-17
fix: `❰…❱` references resolve at concrete universes

kim-em:
This PR makes a `❰foo❱` reference to a universe-polymorphic `theorem_wanted` or `def_wanted` usable at a concrete universe, not only at a use site that shares `foo`'s universe variable.

The generated hypothesis binder spelled `foo`'s universe parameters as named levels inside the binder types (e.g. `(R : Type u_1)`), because they came straight out of `PrettyPrinter.delab`. Those named levels then auto-bound as fresh rigid universe parameters of the enclosing wanted declaration, pinning the reference to a single universe; `❰foo❱ ℚ` for `foo (R : Type _)` therefore failed with "ℚ should be of higher level". The fix rewrites them to level holes `_`, mirroring the treatment already applied to the application levels (`@foo.{_}`), so they unify with whatever universe the reference is used at. The existing `use_poly_dw` test still prints `Type u_1`, because the hole correctly solves to the shared universe when the use site genuinely shares it.

🤖 Prepared with Claude Code

### DIFF
--- a/BatteriesTest/def_wanted.lean
+++ b/BatteriesTest/def_wanted.lean
@@ -167,7 +167,7 @@ info: use_poly_dw : Type u_1 → {d_poly_dw : (α : Type u_1) → (poly_dw α).V
 -/
 #guard_msgs in #check @use_poly_dw
 
-/-! A universe-polymorphic wanted may equally be referenced at a *concrete* universe: the generated
+/-! A universe-polymorphic `def_wanted` may equally be referenced at a *concrete* universe: the generated
 binder spells `poly_dw`'s universe as a hole, so `❰poly_dw❱ Nat` resolves at `Type` rather than
 pinning the reference to a rigid universe parameter that no concrete type could match. -/
 def_wanted use_poly_concrete : Type := ❰poly_dw❱ Nat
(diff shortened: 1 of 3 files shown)


BatteriesTest/def_wanted.lean
```diff
@@ -167,6 +167,16 @@ info: use_poly_dw : Type u_1 → {d_poly_dw : (α : Type u_1) → (poly_dw α).V
 -/
 #guard_msgs in #check @use_poly_dw
 
+/-! A universe-polymorphic wanted may equally be referenced at a *concrete* universe: the generated
```
faenuccio:
```suggestion
/-! A universe-polymorphic wanted definition may equally be referenced at a *concrete* universe: the generated
```

kim-em:
```suggestion
/-! A universe-polymorphic `def_wanted` may equally be referenced at a *concrete* universe: the generated
```

faenuccio:
@kim-em I ignore the merging procedure for `batteries`, but if there's anything I can do, please let me know.

kim-em:
Superseded by https://github.com/leanprover-community/batteries/pull/1861, which consolidates the universe fix, the transitive-dependency surfacing, and a new transparent (body-ful) `def_wanted` into one reviewable change (they touch the same file and build on one another).

PR #1859, opened 2026-06-17
feat: surface transitive dependencies of `❰…❱` references

kim-em:
This PR makes a `❰foo❱` reference to a `def_wanted`/`theorem_wanted` that was itself declared with a body type-check. Such a declaration carries its own `❰…❱` dependencies as auto-generated parameter binders; referencing it previously re-quantified those binders inside the reference's own parameter, where they became unsolvable implicits (the `.Val`/`.Stmt` accessor discards its argument, so nothing fixes their value), and elaboration failed with "don't know how to synthesize implicit argument".

The reference now surfaces the referenced wanted's non-class dependencies as sibling binders on the enclosing declaration, threading each into the reference's application, so the dependency chain is recorded and the reference type-checks. Instance-valued dependencies are unchanged: they stay quantified and resolve via instance synthesis. A small per-declaration arity registry lets a reference tell the user binders apart from the auto-generated ones; the surfacing flattens through multi-level chains and deduplicates against direct references.

This is the second of two issues Filippo Nuccio reported with `def_wanted`. The universe-polymorphic form of his examples additionally needs https://github.com/leanprover-community/batteries/pull/1858 ("fix: `❰…❱` references resolve at concrete universes"); the two fixes are independent and compose.

🤖 Prepared with Claude Code

### DIFF
--- a/Batteries/Util/ProofWanted.lean
+++ b/Batteries/Util/ProofWanted.lean
@@ -24,7 +24,9 @@ Inside any of these commands, `❰foo❱` references an earlier wanted declarati
 `foo`, write `❰foo❱ x y` to apply it. The brackets desugar to fresh parameter binders, so the
 dependency appears in the recorded type. When the referenced wanted's payload is a typeclass,
 the generated parameter is an instance binder, so Lean's instance synth can pick it up at use
-sites (including via Π-instance synth when chained through another wanted).
+sites (including via Π-instance synth when chained through another wanted). If `foo` was itself
+declared with a body (and so has its own `❰…❱` dependencies), referencing it also surfaces those
+dependencies as binders on the referencing declaration, so the whole chain is recorded.
 
 `instance_wanted name : ClassT` is a variant of `def_wanted` whose payload must be a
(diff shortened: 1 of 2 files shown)


Batteries/Util/ProofWanted.lean
```diff
@@ -138,6 +140,18 @@ private initialize wantedInstancesExt :
     addImportedFn := fun _ => #[]
   }
 
+/-- File-local registry recording, for each wanted declaration, how many of its trailing binders
+are auto-generated (chained `❰…❱` references and ambient `instance_wanted` includes) rather than
```
faenuccio, line 144:
```suggestion
are auto-generated (chained `❰…❱` references and ambient `instance_wanted` included) rather than
```

kim-em:
Superseded by https://github.com/leanprover-community/batteries/pull/1861, which consolidates the universe fix, the transitive-dependency surfacing, and a new transparent (body-ful) `def_wanted` into one reviewable change (they touch the same file and build on one another).

PR #1861, opened 2026-06-18
feat: flesh out def_wanted / ❰…❱ references (universes, transitive deps, transparency)

kim-em:
This PR fleshes out the `theorem_wanted` / `def_wanted` / `instance_wanted` machinery, superseding #1858 and #1859 (which touch the same file and build on each other).

- `❰foo❱` references to a universe-polymorphic wanted now resolve at concrete universes, not just a shared universe variable (Filippo Nuccio's first report).
- Referencing a wanted that has its own `❰…❱` dependencies surfaces them on the referrer, instead of leaving them as unsolvable implicits (Filippo Nuccio's second report).
- A `def_wanted` *with a body* becomes a transparent `@[reducible]` definition (wrapped in `DerivedWanted`), so `❰foo❱` is definitionally its body — letting you build honest accessors over opaque holes. The wrapper keeps it from being used directly as a value of its type, and no `axiom`/`sorry` is introduced.
- `instance_wanted` is now included *on use* (like `variable [inst]`): a later declaration carries an instance binder only when it actually needs it, rather than carrying every prior instance unconditionally. The latter grew super-exponentially down a file, making `def_wanted` unusable for a specification with several wanted instances.

🤖 Prepared with Claude Code


kim-em:
In draft mode while I sort out some performance issues.

kim-em:
Performance issues sorted — ready for review.

The slowdown was super-exponential binder growth down a chain of `instance_wanted`s: every later wanted unconditionally carried every prior instance, and class deps nested rather than flattening. Fixed by surfacing deps flat and making ambient inclusion *on use* (like `variable [inst]`). A Jacobian-scale `def_wanted` file now elaborates in ~3s, down from not terminating.

fgdorais (approved):

PR #1863, opened 2026-06-19
chore: remove the defLemma linter

grunweg:
It has been upstreamed to Lean core (as the `defProp` linter, see leanprover/lean4#13803) and is enabled by default (leanprover/lean4#13955) --- so there is no need for the `defLemma` linter duplicating it.


grunweg:
I'm happy about input for fixing the test. I'm not sure what the first test is meant to test, so I cannot judge well if my fix is the right one.

fgdorais (approved):

PR #1878, opened 2026-06-24
feat: make unused_argument work again

fpvandoorn:
* enable the `unusedArguments` linter for theorems.
* enable the linter for unnamed type-class arguments.
* the linter remained disabled for names starting with `_`.
* add tests for all of these cases.

Zulip discussion: https://leanprover.zulipchat.com/#narrow/channel/348111-batteries/topic/unused.20anonymous.20argument/with/606308593

### DIFF
--- a/Batteries/Tactic/Lint/Misc.lean
+++ b/Batteries/Tactic/Lint/Misc.lean
@@ -47,7 +47,7 @@ their value, and allow arguments starting with `_` to be unused. -/
     if ← isProjectionFn declName then return none
     let info ← getConstInfo declName
     let ty := info.type
-    let some val := info.value? | return none
+    let some val := info.value? (allowOpaque := true) | return none
     if val.hasSorry || ty.hasSorry then return none
     forallTelescope ty fun args ty => do
       let mut e := (mkAppN val args).headBeta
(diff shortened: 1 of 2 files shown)


Batteries/Tactic/Lint/Misc.lean
```diff
@@ -58,7 +58,12 @@ their value, and allow arguments starting with `_` to be unused. -/
         if let some val := ldecl.value? then
           e := mkApp e val
       let unused := ldecls.zipIdx.filter fun (ldecl, _) =>
-        !ldecl.userName.isInternal && !e.containsFVar ldecl.fvarId
+        -- We allow unused names that start with `_`, similar to the syntax linter, *unless* the
+        -- name starts with `inst` and has macro scopes, because then it is an
+        -- automatically generated instance name, and we do want to check those.
+        (!ldecl.userName.isInternal ||
+          (ldecl.userName.hasMacroScopes && ldecl.userName.getRoot == `inst)) &&
+          !e.containsFVar ldecl.fvarId
```
thorimur, line 66: This might be more principled:
```suggestion
        -- We allow unused names that start with `_`, similar to the syntax linter,
        -- but make sure to erase macro scopes to avoid false positives from `_@`.
        -- This ensures that we still e.g. lint unused instance arguments with autogenerated names.
        !ldecl.userName.eraseMacroScopes.isInternal && !e.containsFVar ldecl.fvarId
```

fpvandoorn: This flags `theorem foo6_ok (_ : Nat) : True := trivial` as illegal, but probably something in this direction is indeed better.

thorimur: Aha, I actually didn't anticipate that. Hmm.

fpvandoorn:
superseded by #1879

PR #1880, opened 2026-06-25
perf: make a List.filter grind annotation less aggressive

hargoniX:
This PR makes a grind annotation that connects the theory of `List.filter` and `List.findIdxs` less aggressive. Previously it would eagerly convert from any `List.filter` indexing to reasoning about `List.findIdxs`. Now it only does this if we are already reasoning about this type of indexing from both sides.


hargoniX:
There's exactly one proof in mathlib that fails, looking into it.

hargoniX:
Addressed by https://github.com/leanprover/lean4/pull/14178

b-mehta (approved):

fgdorais (approved):

fgdorais:
Looks good. I guess we have to wait until the next release before merging.

PR #1889, opened 2026-07-07
perf(unusedArguments): collect used fvars once instead of per-arg containsFVar

marcelolynch:
`unusedArguments` checks each argument with `e.containsFVar`, where `e` includes the decl's value (for theorems, the whole proof term). `containsFVar` doesn't cache, so it re-walks `e`'s shared subterms once per argument, which gets very slow when the value is a large, shared proof term. Collect the used fvars once with `collectFVars` instead.

Surfaced by #1831, where linting one private lemma took over 5 minutes, and 1 ms with this change.


thorimur:
Beat me to it 😁 Note, we probably don't need the explanatory comment (or at least it could be condensed to something like "prefer `collectFVars` to `containsFVar` for performance"), `collectFVars` is fairly standard.

But the rest looks good to me. :)

wrenna-robson:
Looks pretty great to me. Let me know when it's good to go.

marcelolynch:
@wrenna-robson thanks! I published, with a blessing from @thorimur

Batteries/Tactic/Lint/Misc.lean
eric-wieser, line 59: This use of `mkApp` is nonsense (in that the result is not well-typed), right? This is just a hack to produce a single Expr to recurse over? Perhaps worth a comment if so, but obviously that's not really in scope for this PR.

wrenna-robson: Right, I think this is existing nonsense?

eric-wieser: Yes, I'm at most suggesting we explain the nonsense.

eric-wieser (approved):

Batteries/Tactic/Lint/Misc.lean
wrenna-robson: I think that would be good but if so it would be worth trying to find everywhere this pattern appears (and possibly find an alternative?)

thorimur:
Note, should the bot have removed `WIP` and replaced it with `awaiting-review`? 👀

fgdorais:
No, the bot marks drafts as WIP but not the other way around. Authors need to write a comment containing only "awaiting-review" to replace the `WIP` tag with `awaiting-review`.

PR #1892, opened 2026-07-07
chore: ignore deprecated declarations in some linters

thorimur:
This PR ensures we ignore deprecated declarations in certain linters. 

The rule followed here for choosing linters to adjust is: "could modifying my declaration's code to satisfy the linter result in deprecating that declaration?" If so, we ignore deprecated declarations in that linter.

This PR also adds a test for the `unusedArguments` linter, but does not add tests for modified linters that do not already have tests.

---

Necessary for linting private declarations in #1831.

### DIFF
--- a/Batteries/Tactic/Lint/Misc.lean
+++ b/Batteries/Tactic/Lint/Misc.lean
@@ -33,7 +33,6 @@ This file defines several small linters.
   test declName := do
     if ← isAutoDecl declName then return none
     if ← isImplicitReducible declName then return none
-    if Linter.isDeprecated (← getEnv) declName then return none
     let nm := declName.components
     let some (dup, _) := nm.zip nm.tail! |>.find? fun (x, y) => x == y
       | return none
--- a/Batteries/Tactic/Lint/Misc.lean
+++ b/Batteries/Tactic/Lint/Misc.lean
@@ -10,6 +10,7 @@ public meta import Lean.Util.ForEachExpr
 public meta import Lean.Meta.Check
 public meta import Lean.Meta.Instances
 public meta import Lean.Util.Recognizers
+public meta import Lean.Linter.Deprecated
 public meta import Lean.DocString
 public meta import Batteries.Tactic.Lint.Basic
 
@@ -32,6 +33,7 @@ This file defines several small linters.
   test declName := do
     if ← isAutoDecl declName then return none
     if ← isImplicitReducible declName then return none
+    if Linter.isDeprecated (← getEnv) declName then return none
     let nm := declName.components
     let some (dup, _) := nm.zip nm.tail! |>.find? fun (x, y) => x == y
       | return none
@@ -45,6 +47,7 @@ their value, and allow arguments starting with `_` to be unused. -/
   test declName := do
     if ← isAutoDecl declName then return none
     if ← isProjectionFn declName then return none
+    if Linter.isDeprecated (← getEnv) declName then return none
     let info ← getConstInfo declName
     let ty := info.type
     -- Don't use `value? (allowOpaque := true)`, which would include `opaque` definition values.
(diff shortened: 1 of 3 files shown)


Batteries/Tactic/Lint/Misc.lean
```diff
@@ -32,6 +33,7 @@ This file defines several small linters.
   test declName := do
     if ← isAutoDecl declName then return none
     if ← isImplicitReducible declName then return none
+    if Linter.isDeprecated (← getEnv) declName then return none
```
grunweg: Agreed to this one --- but since the dupNamespace linter is now in mathlib (with checking for consecutive items) and in core (not in mathlib yet), should we rather remove this linter?

Batteries/Tactic/Lint/Misc.lean
```diff
@@ -45,6 +47,7 @@ their value, and allow arguments starting with `_` to be unused. -/
   test declName := do
     if ← isAutoDecl declName then return none
     if ← isProjectionFn declName then return none
+    if Linter.isDeprecated (← getEnv) declName then return none
```
grunweg, line 49: +1 to this one

grunweg (commented):
I think we should rather remove than fix the `dupNamespace` linter. Otherwise, LGTM!

Batteries/Tactic/Lint/Misc.lean
thorimur: Sorry, what do you mean by +1 here?

Batteries/Tactic/Lint/Misc.lean
thorimur: #1895 :)

(I've reverted the changes in this PR so that it doesn't interfere)

Batteries/Tactic/Lint/Misc.lean
grunweg: Sorry - I mean "I agree with this change".

grunweg:
Thanks, LGTM now.
CC @fgdorais

eric-wieser:
> could modifying my declaration's code to satisfy the linter result in deprecating that declaration?

I'm struggling to parse this, could you perhaps just enumerate the reasoning for each one?

wrenna-robson:
LGTM

thorimur:
@eric-wieser

> > could modifying my declaration's code to satisfy the linter result in deprecating that declaration?
> 
> I'm struggling to parse this, could you perhaps just enumerate the reasoning for each one?

Does it help if I make my intended bracketing explicit?

> could {modifying my declaration's code to satisfy the linter} result in {deprecating that declaration}?

eric-wieser:
I think I'm after an example of what that actually means in practice

fgdorais:
@eric-wieser Comments after merge are not that useful. Raise an issue or a correction PR.

eric-wieser:
@fgdorais: i expect the resolution here is that @thorimur made the right decision, so there is no issue or PR to make; I'm simply trying to understand the rationale. I can take this to DMs if you'd prefer.

fgdorais:
@eric-wieser: Sincere apologies, I overreacted to your comment.

thorimur:
@eric-wieser For posterity, I'll explain what I mean here 😁 Consider the following story:
- I have a pre-existing, established definition `foo`
- I now start linting with `myLinter` (maybe I wasn't using that linter before, or its scope has recently expanded...)
- `myLinter` complains about `foo`
- I fix `foo` to comply with `myLinter`

If the "fixing" in the last step likely requires deprecation of the original `foo` (e.g. it suggests changing the type or name), then the next step I take might reasonably be something like

- I attach `@[deprecated]` to `foo` and proceed into the future with my linter-compliant `foo'`

In such a situation `myLinter` should no longer complain about the deprecated `foo` despite it otherwise failing the linter, since the problem has been solved.

---

Note, this state of affairs just means `myLinter` should *definitely* ignore deprecated declarations. There may still be situations where fixing `foo` to comply with `myLinter` does *not* break `foo` in a way necessitating deprecation, but we still want to ignore deprecated declarations anyway. As long as my PR description didn't wind up drifting away from my actual choices in this PR, I took the minimal-exclusion approach here (maximal complaining). 😁

PR #1896, opened 2026-07-10
chore(Tactic/SeqFocus): rewrite `map_tacs` and `<;> [...]` tactic docstring

Vierkantor:
This PR rewrites the docstrings for the `map_tacs` and `<;> [...]` tactics, to consistently match the [official style guide](https://github.com/leanprover/lean4/blob/master/doc/style.md#tactics), to make sure they are complete while not getting too long.

In particular, add a remark about what happens if the number of goals mismatches, and make the `<;> [...]` an alternative to the plain `<;>` notation, so we don't get a duplicate entry in the doc-gen tactic overview.

### DIFF
--- a/Batteries/Tactic/SeqFocus.lean
+++ b/Batteries/Tactic/SeqFocus.lean
@@ -14,7 +14,8 @@ open Lean Elab Meta Tactic
 namespace Batteries.Tactic
 
 /-- Assuming there are `n` goals, `map_tacs [t1; t2; ...; tn]` applies each `ti` to the respective
-goal and leaves the resulting subgoals. -/
+goal and leaves the resulting subgoals. It raises an error if the number of subgoals does not match
+-/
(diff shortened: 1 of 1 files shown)


Batteries/Tactic/SeqFocus.lean
```diff
@@ -14,7 +14,8 @@ open Lean Elab Meta Tactic
 namespace Batteries.Tactic
 
 /-- Assuming there are `n` goals, `map_tacs [t1; t2; ...; tn]` applies each `ti` to the respective
-goal and leaves the resulting subgoals. -/
+goal and leaves the resulting subgoals. It raises an error if the number of subgoals does not match
```
Vierkantor:
```suggestion
goal and leaves the resulting subgoals. It raises an error if the number of subgoals does not match.
```

fgdorais (approved):

PR #1906, opened 2026-07-16
fix: alias respects noncomputable section and opens namespace prefix

### LINKED ISSUE #810: alias does not open namespaces
As opposed to `def` and `abbrev`, `alias` does not open the namespaces of the declaration it is in:
`````
import Batteries.Tactic.Alias

def Foo.bar : Nat := 0

alias Foo.baz := bar -- unknown constant 'bar'
def Foo.qux := bar -- works
abbrev Foo.quux := bar -- works
```

### LINKED ISSUE #1097: alias ignores noncomputable section
`alias` correctly handles `noncomputable` on a `def`, but not `noncomputable` on a `section`:
```
import Batteries.Tactic.Alias

noncomputable def foo : Nat := Classical.choice sorry

alias bar := foo -- works fine

noncomputable section

def baz : Nat := Classical.choice sorry

alias quux := baz
-- failed to compile definition, consider marking it as 'noncomputable' because it depends on 'baz', and it does not have executable code
```

Chessing234:
## Summary
- `alias` in a `noncomputable section` inherits noncomputable and marks the synonym (`#1097`).
- Namespaced `alias Foo.bar := x` opens `Foo` when resolving `x`, matching `def` (`#810`).

Fixes #810
Fixes #1097

## Test plan
- [x] `lake build BatteriesTest.alias`
- [x] `lake build BatteriesTest.alias_module`
- [ ] CI (awaiting fork workflow approval)

awaiting-review


Chessing234:
awaiting-review

Chessing234:
Pushed a follow-up: `resolveAliasTarget` no longer always uses `withoutExporting`, so public aliases of private defs still get rejected (fixes the `alias_module` guard). CI is waiting on workflow approval for the fork push.

YaelDillies:
Can you mark this as fixing #810 and #1097?

Chessing234:
Done — PR body now has Fixes #810 and Fixes #1097.

fgdorais:
awaiting-author

Chessing234:
fixed the mathlib break: alias now resolves the target in the ambient scope first, and only opens the namespace prefix (the #810 behavior) when that fails. that keeps `alias Std.Refl.reflexive := refl` unambiguous while still allowing `alias Foo.baz := bar`.

`lake build BatteriesTest.alias` is green locally. awaiting-review

PR #1907, opened 2026-07-16
Flake.nix

jmikedupont2:
creating  a flake.nix for batteries.
I hope this helps someone someday.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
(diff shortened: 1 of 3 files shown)


fgdorais:
Sorry for the delay, I'm not a nix user. I'm worried about supporting this PR in the future, it seems there's nobody on staff that could maintain this on a regular basis. With that constraint, we can only merge if there is some guarantee that this will not require maintenance more than once a year, maybe twice.

PR #1908, opened 2026-07-17
fix: sampleVariance definition for n=2

Chessing234:
## Summary
- Guard was `count ≤ 2`, so two samples always returned variance 0.
- Unbiased `S/(n-1)` needs `n ≥ 2` → guard `≤ 1`.

## Test plan
- [ ] `lake build Batteries.Data.RunningStats`

Made with [Cursor](https://cursor.com)


fgdorais (approved):

Batteries/Data/RunningStats.lean
fgdorais:
```suggestion
  if s.count = 0 then 0.0 else s.var / s.count.toFloat
```

fgdorais (approved):

PR #1916, opened 2026-07-19
feat: expand `Float` API

### LINKED ISSUE #13369: Missing inverse lemmas for Float.ofBits / Float.toBits
## Summary

`Float.ofBits` and `Float.toBits` are declared as `@[extern]` opaque primitives in `Init.Data.Float` but have no accompanying lemmas establishing their inverse relationship:

```lean
@[extern "lean_float_of_bits"] opaque Float.ofBits : UInt64 → Float
@[extern "lean_float_to_bits"] opaque Float.toBits : Float → UInt64
```

There are currently zero theorems about these functions in the standard library.

## Requested additions

```lean
axiom Float.ofBits_toBits (f : Float) : Float.ofBits f.toBits = f
axiom Float.toBits_ofBits (u : UInt64) : (Float.ofBits u).toBits = u
```

Both hold by the C implementation (`lean_float_of_bits` / `lean_float_to_bits` are `memcpy`-based reinterpret casts between `double` and `uint64_t`, which are exact inverses on all IEEE 754 platforms).

## Motivating use case

I maintain a production Lean 4 codebase that uses a typed data model with a positional JSON codec. The codec serializes `Float` values via their exact IEEE-754 bit pattern (`Float.toBits` → `UInt64` → `Nat` → `Json.num`) and decodes the reverse path. Proving the round-trip property (`fromPositionalJson dm (toPositionalJson v) = .ok v`) requires knowing that `Float.ofBits f.toBits = f`.

Without upstream support, I had to introduce a project-local axiom:

```lean
-- lean/Spec/Trusted/FloatBits.lean
axiom float_ofBits_toBits (f : Float) : Float.ofBits f.toBits = f
```

This is the only axiom in the project (714+ sorry-free theorems otherwise). Having this in the standard library would eliminate the need for downstream projects to each introduce their own trusted axiom for the same fact.

## Implementation question

Would the Lean team prefer these as:
1. `axiom` declarations in `Init.Data.Float` (simplest, matches the `opaque` status of the functions), or
2. Some other mechanism (e.g. `native_decide`-backed, or `@[implemented_by]`)?

Happy to open a PR once there's alignment on the approach.

Rob23oba:
This PR expands the `Float` API in several ways. Specifically, it adds:
- `Float32` versions of various `Float` definitions (e.g. `Float32.nan`)
- `Float.pi` and `Float32.pi` for the mathematical constant pi
- Basic lemmas about `Float` and `Float32`, in particular `Float.ofBits_toBits` and `PartialEquivBEq` instances

Furthermore, some modules were renamed, namely `Batteries.Lean.Float` to `Batteries.Data.Float.Basic` and `Batteries.Data.Rat.Float` to `Batteries.Data.Float.Rat`. The corresponding `deprecated_module`s will be added in a follow-up PR. In addition to that, two instances about `JsonNumber` already present in core were removed.

Closes leanprover/lean4#13369

### DIFF
--- a/Batteries/Data/Float/Basic.lean
+++ b/Batteries/Data/Float/Basic.lean
@@ -42,7 +42,7 @@ def toRatParts (f : Float) : Option (Int × Int) :=
   else none
 
 /-- Returns `v, exp` integers such that `f = v * 2^exp`.
-Like `toRatParts`, but `e` is guaranteed to be minimal (`n` is always odd), unless `n = 0`.
+Like `toRatParts`, but `e` is guaranteed to be minimal (`v` is always odd), unless `v = 0`.
 `n.abs` will be at most `2^53 - 1` because `Float` has 53 bits of precision.
 Returns `none` when `f` is not finite (i.e. `inf`, `-inf` or a `nan`). -/
 partial def toRatParts' (f : Float) : Option (Int × Int) :=
@@ -117,7 +117,7 @@ def toRatParts (f : Float32) : Option (Int × Int) :=
   else none
 
 /-- Returns `v, exp` integers such that `f = v * 2^exp`.
-Like `toRatParts`, but `e` is guaranteed to be minimal (`n` is always odd), unless `n = 0`.
+Like `toRatParts`, but `e` is guaranteed to be minimal (`v` is always odd), unless `v = 0`.
 `n.abs` will be at most `2^53 - 1` because `Float` has 53 bits of precision.
 Returns `none` when `f` is not finite (i.e. `inf`, `-inf` or a `nan`). -/
 partial def toRatParts' (f : Float) : Option (Int × Int) :=
--- a/Batteries/Data/Float/Basic.lean
+++ b/Batteries/Data/Float/Basic.lean
@@ -22,6 +22,11 @@ involving `nan` return `false`, including in particular `nan == nan`.
 -/
 def nan : Float := 0/0
 
+/--
+The floating point value closest to the mathematical constant `π`.
+-/
+def pi : Float := 3.141592653589793
+
 /-- Returns `v, exp` integers such that `f = v * 2^exp`.
 (`e` is not minimal, but `v.abs` will be at most `2^53 - 1`.)
 Returns `none` when `f` is not finite (i.e. `inf`, `-inf` or a `nan`). -/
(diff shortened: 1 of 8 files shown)


Batteries/Data/Float/Basic.lean
```diff
@@ -77,6 +82,83 @@ def toStringFull (f : Float) : String :=
...
+    let x := (2^24:Nat).toFloat32 * f'
+    let v := if x < 0 then
+      (-(-x).floor.toUInt32.toNat : Int)
+    else
+      (x.floor.toUInt32.toNat : Int)
+    some (v, exp - 24)
+  else none
+
+/-- Returns `v, exp` integers such that `f = v * 2^exp`.
+Like `toRatParts`, but `e` is guaranteed to be minimal (`n` is always odd), unless `n = 0`.
```
fgdorais:
```suggestion
Like `toRatParts`, but `e` is guaranteed to be minimal (`v` is always odd), unless `v = 0`.
```

Batteries/Data/Float/Basic.lean
```diff
@@ -77,6 +82,83 @@ def toStringFull (f : Float) : String :=
...
+    if n == 0 then (0, 0) else
+      let neg : Bool := n < 0
+      let v := n.natAbs.toUInt64
+      let c := trailingZeros v 0
+      let v := (v >>> c.toUInt64).toNat
+      (if neg then -v else v, e + c.toNat)
+where
+  /-- Calculates the number of trailing bits in a `UInt64`. Requires `v ≠ 0`. -/
+  -- Note: it's a bit dumb to be using a loop here, but it is hopefully written
+  -- such that LLVM can tell we are computing trailing bits and do good things to it
```
fgdorais: I'm curious if this is true. Check with [Compiler Explorer](https://godbolt.org/)?

Batteries/Data/Float/Lemmas.lean
```diff
@@ -0,0 +1,306 @@
...
+theorem Float.beq_self_eq_false_iff {f : Float} : (f == f) = false ↔ f = nan := by
+  grind [beq_self_iff]
+
+@[simp]
+theorem Float32.beq_self_eq_false_iff {f : Float32} : (f == f) = false ↔ f = nan := by
+  grind [beq_self_iff]
+
+/--
+This theorem is true in general but most other types have `LawfulBEq` where
+`bne_self_eq_false` or `bne_iff_ne` apply.
```
fgdorais: Do you really mean "true in general" or do you just mean "true"? The phrase "true in general" implies that there are edge cases where it is false. My understanding is that the theorem is true (by definition of `!=`) with the caveat that `f == f` is not always true, as one might expect were `==` a `LawfulBeq` for `Float`.

Please clarify this comment. Same for the next theorem.

Batteries/Data/Float/Lemmas.lean
```diff
@@ -0,0 +1,306 @@
...
+-- but!
+example : ¬ EquivBEq Float := by
+  intro h
+  have : Float.nan == Float.nan := BEq.rfl
+  contradiction
+
+example : ¬ EquivBEq Float32 := by
+  intro h
+  have : Float32.nan == Float32.nan := BEq.rfl
+  contradiction
```
fgdorais: These should either be theorems or in the module docs.

fgdorais (changes requested):

Batteries/Data/Float/Basic.lean
fgdorais: It can optimize the loop really well but it doesn't do anything smarter. https://godbolt.org/z/KEGY4K1WE

Rob23oba: To clarify, this is a copy of the code above for `Float32`.

fgdorais: And so? I don't understand.

Rob23oba: Oh that's just to say that I haven't written this code and you're reviewing code that's already in batteries; which isn't bad, having a fresh look at code can definitely help find ways to improve it (or find typos like the one you mentioned) but I just wanted to mention it in case it wasn't clear.

fgdorais: Got it!

Batteries/Data/Float/Basic.lean
```diff
@@ -77,6 +82,83 @@ def toStringFull (f : Float) : String :=
...
+  -- TODO: prove termination under suitable assumptions (only relevant if `Float` is not opaque)
```
fgdorais:
```suggestion
  -- TODO: optimize and make total
```

fgdorais (approved):

PR #1919, opened 2026-07-20
feat: prove List.take_flatten

MikeBoozer:
Completes the `proof_wanted` for `List.take_flatten`. The statement is unchanged.

The proof is by induction on `L` generalizing `i`, splitting on whether `i` falls inside the head sublist. The out-of-range case (`i` past the end, where `j = L.length` and `L[j]? = none`) is handled by the same branch, since `findIdx` returns the list length when nothing matches — so no separate case is needed.

Verified against current `main` on v4.33.0-rc1: `lake build --wfail`, `lake test`, and `lake lint` all pass. `#print axioms List.take_flatten` reports only `propext`, `Classical.choice`, `Quot.sound`.

Incidentally, the two `findIdx`/`partialSums_cons` index facts used here are also proved inline in `getElem_flatten` just above, with nearly identical proofs. Happy to factor them into shared lemmas in a follow-up if you'd like — I left that out to keep this PR to the one `proof_wanted`.

🤖 Proof generated by an LLM-based prover harness, verified and reviewed by me.


MikeBoozer:
awaiting-review

Batteries/Data/List/Scan.lean
```diff
@@ -224,7 +224,39 @@ The indices are computed as:
 - `j` is one less than where the cumulative sum first exceeds `i`
 - `k` is `i` minus the total length of the first `j` sublists
 -/
-proof_wanted take_flatten (L : List (List α)) (i : Nat) :
+theorem take_flatten (L : List (List α)) (i : Nat) :
     let j := (L.map length).partialSums.findIdx (· > i) - 1
     let k := i - (L.take j).flatten.length
```
fgdorais, line 256: Ideally, this should use: https://github.com/leanprover-community/batteries/blob/2c810760f0a0c4536b397dbe30ca9b2f2f467366/Batteries/Data/Fin/Coding.lean#L193-L213 Is that substitution a major problem here?

MikeBoozer: Not a major problem technically — `Fin/Coding.lean` isn't in `Scan.lean`'s import closure, so there's no cycle. The catch is the bound: `decodeSigma` takes `Fin (Fin.sum f)`, but `take_flatten` has no hypothesis bounding `i`, and the out-of-range case is real content — when `i` runs past the end, `j = L.length` and `L[j]? = none`, so nothing is appended. Restating it over `Fin` would mean adding `i < L.flatten.length`, which weakens it. `getElem_flatten` carries that hypothesis already, so the `Fin` form fits it but not this one.

The length side of that bridge nearly exists already — `List.length_flatten` gives `L.flatten.length = (L.map length).sum` — so the real work isn't a missing lemma but reindexing that `List.sum` as `Fin.sum` and threading it through `decodeSigma`'s bound. And the `partialSums.findIdx (· > i) - 1` idiom is shared by all four declarations here (`getElem_flatten_aux₁`/`aux₂`/`getElem_flatten` and this one), so converting only `take_flatten` would leave it out of step with its neighbours — the coherent version is to migrate all four.

Happy to go either way: merge this as-is matching the existing idiom, or do a follow-up that adds the `Fin.sum`/`flatten` bridge and migrates all four. If you'd prefer the latter, is losing the unbounded `i` acceptable, or would you want to keep a `Nat`-indexed version alongside?

wrenna-robson:
I find this proof quite hard to read - the inline lemmas are a large part of that. Just to start with, please could you work to factor those out? If they're useful in two places, that's a good argument for adding them. I would strongly recommend doing this manually rather than in an LLM-assisted way as this kind of delicate factoring out I find they can be somewhat unreliable at.

Chessing234:
take_flatten looks useful — curious if you also want the dual for drop_flatten in the same PR or leave it follow-up.

Batteries/Data/List/Scan.lean
wrenna-robson: It's not clear to me why using the Fin encodings is the right choice.

MikeBoozer:
@wrenna-robson Thanks — that's fair, and factoring them out did improve it.

It turned up more duplication than I expected: both `findIdx` facts were already proved inline in `getElem_flatten` too, with near-identical scripts, and `getElem_flatten_aux₁` proved a third (that the index is positive) the same way. So it's three lemmas rather than two, each used in two places:

- `findIdx_partialSums_map_length_pos` — `getElem_flatten_aux₁`, `take_flatten`
- `findIdx_partialSums_map_length_cons_of_lt_length` — `getElem_flatten`, `take_flatten`
- `findIdx_partialSums_map_length_cons_of_length_le` — `getElem_flatten`, `take_flatten`

On doing it manually: I should be straight with you — I did use an LLM, interactively and compiler-checked at each step, not the autonomous harness that produced the original proof. Since that doesn't answer your point on its own, I tried to make the change one you can *check* rather than trust. The tactic scripts of the three lemmas are character-for-character the inline `have`s they replace; nothing was re-derived or re-golfed. Only their statements changed, to take the branch hypothesis as an explicit argument.

Two call sites needed more than substituting the lemma in:

- `getElem_flatten_aux₁` used a bare `simp at hlt` that the following `omega` silently depended on, and which normalised `>` to `<` — so it stopped lining up once the fact came from a lemma stated with `>`. Now pinned to `simp only [length_partialSums, length_map] at hlt`.
- `take_flatten` opens with `simp only [flatten_cons]` instead of also unfolding `map_cons, partialSums_cons`, so the goal keeps the shape the new lemmas are stated in; its `hFpos` `have` is folded into `hFeq`.

`getElem_flatten` is a pure substitution. Net effect: 68 → 43 lines of proof script, `take_flatten` itself 32 → 19; the file grows 14 lines for the statements and docstrings.

Verified on `v4.34.0-rc1` (this branch now includes a merge of current `main`): `lake build --wfail`, `lake test` and `lake lint` pass, and `#print axioms` reports only `propext`, `Classical.choice`, `Quot.sound` for all seven declarations.

Two things I left alone: `getElem_flatten_aux₂` still open-codes the same case split, and two of the new lemmas keep a non-terminal `simp` inherited from the existing proofs. Both need scripts genuinely rewritten rather than moved, which would undercut the point above — happy to do either as a follow-up.

I made the three public to match `getElem_flatten_aux₁`/`aux₂`, but they're helper-shaped — happy to make them `private` or rename them.

You said "just to start with", so I'm assuming there's more — I'd like to get this properly readable, so point me at whatever reads worst next and I'll keep going.

MikeBoozer:
@Chessing234 Thanks — and yes, I think the general `drop` version is a real gap. Mathlib has `List.take_sum_flatten` and `List.drop_sum_flatten`, but those are the aligned case where the index is a sum of whole sublist lengths; I couldn't find the arbitrary-index `drop` form — the dual of this `take_flatten` — in Batteries, Mathlib or core.

I'd rather keep it out of this PR, which is scoped to the single `proof_wanted` and already has an open review thread. It should be a cheaper follow-up now, though: the three index lemmas factored out here (`findIdx_partialSums_map_length_pos` and the `_cons_of_lt_length`/`_cons_of_length_le` pair) compute the same `j` the `drop` side needs, so most of the index bookkeeping is already done. Happy to open that once this lands.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,53 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
     right
     simpa using ih
 
+/-!
```
fgdorais: Should be a regular comment.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,53 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+The next three lemmas are the index bookkeeping shared by `getElem_flatten` and `take_flatten`.
+In both, `(L.map length).partialSums.findIdx (· > i) - 1` is the index of the sublist of `L`
+that contains flat position `i`; these compute that `findIdx` for a `cons`.
+-/
+
+/--
```
fgdorais: Docstrings are not meant to be used to explain a proof. Use a regular comment or, preferably, make the proof self-explanatory.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,53 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+-/
+theorem findIdx_partialSums_map_length_pos (L : List (List α)) (i : Nat) :
+    0 < (L.map length).partialSums.findIdx (· > i) := by
+  by_contra w
+  simp at w
+
+/--
+If `i` indexes into the first sublist, the partial sums of the lengths of `l :: L` first exceed
+`i` at index `1`.
+-/
```
fgdorais: Delete. The theorem statement is self-explanatory.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,53 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+theorem findIdx_partialSums_map_length_cons_of_lt_length {l : List α} {L : List (List α)}
+    {i : Nat} (h : i < l.length) :
+    ((l :: L).map length).partialSums.findIdx (· > i) = 1 := by
+  simp [partialSums_cons, findIdx_cons]
+  rw [findIdx_eq] <;> grind
+
+/--
+If `i` is not an index into the first sublist `l`, the partial sums of the lengths of `l :: L`
+first exceed `i` exactly one index after those of `L` first exceed `i - l.length`.
+-/
```
fgdorais: Delete. The theorem statement is self-explanatory.

fgdorais (changes requested):

Batteries/Data/List/Scan.lean
MikeBoozer: Took the preferable option here — the proof is now plain `simp`, so there's nothing left to explain. The other three are done as asked.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,36 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
     right
     simpa using ih
 
+-- `(L.map length).partialSums.findIdx (· > i) - 1` is the index of the sublist of `L` containing
+-- flat position `i`; the next three lemmas compute that `findIdx`.
```
fgdorais, line 138: Multiple line comments are preferably in a `/- ... -/` block rather than `--` for every line.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,36 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+
+theorem findIdx_partialSums_map_length_pos (L : List (List α)) (i : Nat) :
+    0 < (L.map length).partialSums.findIdx (· > i) := by
+  simp
```
wrenna-robson, line 142: I think you can just leave it on the same line. This having been said I was quite surprised that this is just simp!

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,36 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+
+theorem findIdx_partialSums_map_length_cons_of_lt_length {l : List α} {L : List (List α)}
+    {i : Nat} (h : i < l.length) :
+    ((l :: L).map length).partialSums.findIdx (· > i) = 1 := by
+  simp [partialSums_cons, findIdx_cons]
```
wrenna-robson, line 147: Non-terminal simp.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,36 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+  rw [findIdx_eq] <;> grind
```
wrenna-robson, line 148: grind [findIdx_eq] or similar, surely?

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,36 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+
+theorem findIdx_partialSums_map_length_cons_of_length_le {l : List α} {L : List (List α)}
+    {i : Nat} (h : l.length ≤ i) :
+    ((l :: L).map length).partialSums.findIdx (· > i) =
+      (L.map length).partialSums.findIdx (· > i - l.length) + 1 := by
+  simp [partialSums_cons, findIdx_cons, Function.comp_def]
```
wrenna-robson, line 154: non-terminal simp.

Batteries/Data/List/Scan.lean
```diff
@@ -134,17 +134,36 @@ theorem length_flatten_mem_partialSums_map_length (L : List (List α)) :
...
+  congr
+  funext x
+  grind
```
wrenna-robson, line 157: This has bad code smell to me. I am surprised grind cannot do it.

Batteries/Data/List/Scan.lean
wrenna-robson: This is such an ugly theorem. Where would it ever come up? On considering this I realise therefore that @fgdorais is correct.

I do think there is something to be said for linking List (List α) and Fin (Fin.sum f) though. Let's take a step back: conceptually speaking, L : List (List α) is indexed by i < L.length , and then by j < L[i].length -
i.e. it is given by (i : Fin L.length) × (j : Fin L[i].length) OR instead, it is indexed by Fin (Fin.sum f), where f is defined by f i = L[i].length.

I agree therefore with @fgdorais that conceptually we simply must use the canonical encodings here. Actually should we not have the theorem that L.flatten.length = Fin.sum (List.length ∘ L.get)? I feel like surely we must have that but maybe we don't? Similarly I think getElem_flatten should then be re-stated in terms of these encodings...

wrenna-robson (commented):
I think there is work to do here but you need to pull up out of the LLM-driven tailspin to work out exactly what it is that is needed, rather than re-hashing.

wrenna-robson:
I have become convinced that there are deeper problems here - I think our entire approach to getElem_flatten is wrong. We probably want something like
```
theorem sum_get [Add α] [Zero α] (xs : List α) : Fin.sum xs.get = xs.sum := by
  induction xs with | nil => rfl | cons x xs IH =>
  rw [sum_cons, Fin.sum_succ]; exact congrArg (x + ·) IH

theorem _root_.Fin.sum_congr [Add α] [Zero α] {f : Fin n -> α} {g : Fin m -> α} 
    (h₁ : n = m) (h₂ : ∀ i, f i = g (i.cast h₁)) : Fin.sum f = Fin.sum g := by
  subst h₁; simp [<- funext_iff] at h₂; subst h₂; rfl

theorem sum_comp_get [Add α] [Zero α] (xs : List β) (f : β -> α) : 
    Fin.sum (f ∘ xs.get) = (xs.map f).sum := 
  sum_get (xs.map f) ▸ Fin.sum_congr (length_map f).symm (by simp)

proof_wanted get_flatten (L : List (List α)) :
    L.flatten.get = (fun x => (L.get x.1).get x.2) ∘ 
    Fin.decodeSigma (length ∘ L.get) ∘ Fin.cast (sum_comp_get L length ▸ length_flatten)

proof_wanted getElem_flatten (L : List (List α)) (i : Nat) (h : i < L.flatten.length) :
    let j : (i : Fin (L.length)) × (Fin (L[i].length)) := 
    Fin.decodeSigma (length ∘ L.get) ⟨i, lt_of_lt_of_eq h (sum_comp_get L length ▸ length_flatten)⟩
    L.flatten[i] = (L.get j.1).get j.2
```

Though I'm not certain of this. As a result I am not keen to add a proof for take_flatten in its current form as I think it only deepens this technical debt.

PR #1925, opened 2026-07-23
fix: correct Vector.scanrM fast path

### LINKED ISSUE #1924: `Vector.scanrM` `@[implemented_by] scanrMFast` is incorrect (off-by-one); `native_decide` disagrees with the kernel
### Summary

`Vector.scanrM`'s `@[implemented_by] scanrMFast` is incorrect: it produces different results from the (kernel-reducible) `scanrM` specification. Because `native_decide` and `#eval` run the compiled `scanrMFast`, they return wrong answers, and `native_decide` disagreeing with the kernel means **`native_decide` can prove `False`** on goals mentioning `Vector.scanr`/`scanrM`.

### Quick demonstration

```lean
import Batteries.Data.Vector.Basic
open Batteries

-- runs the compiled `scanrMFast`
#eval (Vector.scanr (· + ·) 0 (⟨#[1, 2, 3], rfl⟩ : Vector Nat 3)).toList
-- => [5, 3, 0, 0]     -- WRONG; the correct scan is [6, 5, 3, 0]
```

`scanr (·+·) 0 [1,2,3]` should be `[1+2+3+0, 2+3+0, 3+0, 0] = [6, 5, 3, 0]`.

### Root cause

In `Batteries/Data/Vector/Basic.lean`, `scanrMFast.loop`:

```lean
      let next ← f (as.uget pred (by omega)) cur
      loop next pred n_usize h_n (by omega) (acc.uset pred cur (by omega))
```

It writes `cur` (the accumulator **before** folding in the current element) into position `pred`, but it should write `next` (the value it just computed and is about to carry as the new `cur`). So every entry is off by one fold. The fix is:

```lean
      loop next pred n_usize h_n (by omega) (acc.uset pred next (by omega))
```

With this change the trace of `scanr (·+·) 0 [1,2,3]` becomes `[6,5,3,0]`, matching `scanrM` (and `List.scanr` via `Vector.toList_scanr`).

### Soundness reproduction (`native_decide` ≠ kernel ⟹ `False`)

This file elaborates with no `sorry`; `#print axioms boom` shows only `native_decide`'s axioms (`Lean.ofReduceBool`, `Lean.trustCompiler`), i.e. it is a genuine `native_decide`-driven proof of `False`.

```lean
import Batteries.Data.Vector.Basic
import Batteries.Data.Vector.Lemmas
import Batteries.Data.List.Scan
open Batteries

def mkV (l : List Nat) : Vector Nat l.length := ⟨l.toArray, by simp⟩

-- native_decide runs the (wrong) compiled scanrMFast and accepts [5,3,0,0]
theorem native_says : (Vector.scanr (· + ·) 0 (mkV [1,2,3])).toList = [5, 3, 0, 0] := by
  native_decide

-- the kernel, via Batteries' own `Vector.toList_scanr`, gets the correct [6,5,3,0]
theorem kernel_says : (Vector.scanr (· + ·) 0 (mkV [1,2,3])).toList = [6, 5, 3, 0] := by
  rw [Vector.toList_scanr]; decide

theorem boom : False := by
  have h : ([5,3,0,0] : List Nat) = [6,5,3,0] := native_says ▸ kernel_says
  exact absurd h (by decide)

#print axioms boom   -- propext, Quot.sound, Lean.ofReduceBool, Lean.trustCompiler  (no sorryAx)
```

### Notes

- `Vector.scanl`/`scanlM` and `Array`/`List` `scanl`/`scanr` all appear correct; only the `Vector.scanrM` fast path is affected. The `#guard`s in the source cover `Array.scanr` but not the `Vector` version, which is likely why it slipped through.
- Impact in practice is latent (a proof only becomes unsound if it pairs `native_decide` with `Vector.scanr`/`scanrM`), but the fast path is simply computing the wrong values, so it also affects `#eval`/`decide`-free compiled use.

### Version

- batteries `77d3cc514f987c1f42f2bbd8a8d56855012dc115` (2026-06-25), also present on `main`.
- Lean `leanprover/lean4:v4.32.0-rc1`.

gersh:
Fixes #1924.

`scanrMFast.loop` computed the next accumulator value but stored the previous value in the current output slot. Store `next` instead of `cur` so the compiled `@[implemented_by]` path agrees with the reference definition.

Add a compiled `#guard` for the reported suffix-sum example. This specifically exercises the fast path and would fail before the fix.

Checks run:
- `lake build --wfail`
- `lake test`
- `lake lint`
- `lake env lean scripts/check_imports.lean`
- `scripts/lintWhitespace.sh`
- remaining CI policy scans from `.github/workflows/build.yml`

No Mathlib adaptation is expected: this only corrects erroneous runtime behavior and adds a regression test.

### DIFF
--- a/Batteries/Data/Vector/Basic.lean
+++ b/Batteries/Data/Vector/Basic.lean
@@ -100,7 +100,7 @@ where
       let pred := i - 1
       have h_pred : pred.toNat = i.toNat - 1 := USize.pred_toNat h_gt
       let next ← f (as.uget pred (by omega)) cur
-      loop next pred n_usize h_n (by omega) (acc.uset pred cur (by omega))
+      loop next pred n_usize h_n (by omega) (acc.uset pred next (by omega))
     else
       return acc
   termination_by i.toNat
--- a/BatteriesTest/vector.lean
+++ b/BatteriesTest/vector.lean
@@ -1,5 +1,9 @@
 import Batteries.Data.Vector
 
+/-! ### Testing scans over `Vector`. -/
+
+#guard (Vector.scanr (· + ·) 0 (#v[1, 2, 3] : Vector Nat 3)).toList == [6, 5, 3, 0]
+
 /-! ### Testing decidable quantifiers for `Vector`. -/
 
 example : ∃ v : Vector Bool 6, v.toList.count true = 3 := by decide


gersh:
awaiting-review

Batteries/Data/Vector/Basic.lean
fgdorais, line 129: Same error happens here too?

gersh: No—the reference loop builds the result in reverse order. For `[1, 2, 3]`, `acc` evolves as `[] → [0] → [0, 3] → [0, 3, 5]`; at the base case it pushes the final `cur = 6` and reverses, producing `[6, 5, 3, 0]`. The fast loop instead preallocates the result in final order and writes directly at `pred`, so that slot must receive the freshly computed `next`.

fgdorais (approved):

PR #1932, opened 2026-07-31
feat: turn the `unusedHaveSuffices` linter back on

JovanGerb:
This PR turns the `unusedHavesSuffices` linter back on. Let's see what kind of (false) positives we get in mathlib!


JovanGerb:
Oh, it looks like the automation used an incompatible mathlib version :(

Chessing234:
turning unusedHaveSuffices back on seems overdue if the false positives are gone; any remaining noisy files?

JovanGerb:
It seems that unfortunately there are still quite a few false positives in mathlib (around 30). I don't see an easy way to filter these out unfortunately. I'm not sure if we want to go ahead with this PR if that means we need to add `no_lint` tags for the false positives.

I fixed the true positives in https://github.com/leanprover-community/mathlib4

PR #1945, opened 2026-08-15
chore: update process for Mathlib adaptations

fgdorais:
Added detailed steps for creating an adaptation PR on Mathlib.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,23 +73,29 @@ been proven.
 
 Batteries PRs often affect Mathlib, a key component of the Lean ecosystem.
 When Batteries changes in a significant way, Mathlib must adapt promptly.
-When necessary, Batteries contributors are expected to either create an adaptation PR on Mathlib, or ask for assistance for and to collaborate with this necessary process.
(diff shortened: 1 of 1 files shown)


wrenna-robson:
I think this is a perfectly serviceable explanation in prose. It would be even better, however, if git and gh commands were given where appropriate.

wrenna-robson:
In addition, in theory there should be zero reason to have a fork of both mathlib4-nightly and mathlib4. They are the same repository, or at least one is the fork of the other.

You could set up nightly as a separate remote on a given fork. That means your standard mathlib4 fork would have origin - your fork - upstream - leanprover-community/mathlib4, AND testing/nightly, which would be mathlib4-nightly-testing.
This means that an end user doesn't need to set up a new repository locally - they are just managing the state. git and the GitHub command line tool, gh, give quite flexible tools to help with this.

Basically I think want you want to do is
a) make a local branch that mirrors nightly/batteries-PR-testing-XX or whatever it is.
b) you can then work on that branch and push it to your fork.
c) when you are ready, you push it to nightly. 

I really think that encouraging people to think of mathlib4 and mathlib4-nightly-testing as different repositories, as opposed to different states of the same repository, is an error.

PR #1948, opened 2026-08-19
feat: support `+typeChanged` in `deprecated` attribute

wkrozowski:
This PR provides support to a newly added `+typeChanged` argument of `deprecated` argument. In `4.34.0-rc1` it's a no-op, but once a new release is cut, all aliases that change type (in particular "destructuring" aliases) will need this annotation.


fgdorais:
Are there actual instances of this? The forward/reverse splitting mode of `alias` is not meant to be used with the `deprecated` attribute. Sounds more reasonable to add a warning or error when that happens.

wkrozowski:
There are several instances of applying `deprecated` attribute to such restructuring aliases within mathlib. To name a few:


- `Mathlib/Algebra/Polynomial/Derivative.lean:453-454` - `alias ⟨natDegree_eq_zero_of_derivative_eq_zero, _⟩ :=
  derivative_eq_zero`
- `Mathlib/Data/ENNReal/Basic.lean:376-380` - `alias ⟨_, coe_le_coe_of_le⟩ := coe_le_coe and alias ⟨_, coe_lt_coe_of_lt⟩ :=
  coe_lt_coe`
- `Mathlib/Combinatorics/SimpleGraph/Coloring/Vertex.lean:507-508` - `alias ⟨isEmpty_of_chromaticNumber_eq_zero, _⟩ :=
  chromaticNumber_eq_zero_iff`
- `Mathlib/Data/Nat/MaxPowDiv.lean:269-270` - `alias ⟨_, maxPowDiv.le_of_dvd⟩ := pow_dvd_iff_le_padicValNat`

fgdorais:
That still seems to be unusual and bad style, but since it's already used...

It seems to me that a plain alias never changes the type and a forward/reverse alias always changes the type. Maybe `deprecated` when applied to alias should ignore `+typeChanged` and apply it where needed?

wkrozowski:
I think the intention of the `+typeChanged` annotations is that user becomes aware of the fact that the introduced deprecation changes types and is typed manually, rather than it is something that gets inserted automatically. 

cc: @TwoFX

TwoFX:
Yes, one of the ideas behind the `+typeChanged` marker is that it is an explicit documentation in the source code that the deprecation is not just a pure name change. From that point of view, it would make sense to require the `+typeChanged` in this situation, even if it is always required when deprecating and splitting at the same time.

Of course, this is just how core intended this feature, the Batteries maintainers can do what they want with their syntax.

fgdorais:
I'm still confused: Is `+typeChanged` purely ornamental? If not, what does it actually do?

TwoFX:
It silences the warning that the deprecation was in favor of a declaration with a different type. It might as well have been `set_option linter.deprecationTypeChanged false in` but disabling a linter usually indicates that you're doing something that is at least slightly unusual, which is not the case here.

fgdorais:
That makes sense.

fgdorais (approved):

wkrozowski:
@fgdorais It looks like I am not authorised to push to this branch - could you please do it?

PR #1953, opened 2026-08-27
refactor: split AssocList into Basic and Lemmas

kim-em:
This PR splits `Batteries/Data/AssocList.lean` into `Batteries/Data/AssocList/Basic.lean` and `Batteries/Data/AssocList/Lemmas.lean`, and leaves `Batteries.Data.AssocList` as an umbrella module importing both, so existing imports are unaffected.

`Basic` carries the inductive type, the operations, the instances, and the `@[csimp]` equivalence for `toListTR` together with the two `foldl` specifications its proof needs, following the pattern already used in `Batteries/Data/List/Basic.lean`. `Lemmas` carries the specifications relating each operation to the corresponding `List` operation along `toList`, and the `LawfulBEq` instance.

No declaration is added, removed, renamed or retagged. The set of names under `Batteries.AssocList`, and which of them carry `@[simp]`, is identical before and after.

🤖 Prepared with Claude Code


Batteries/Data/AssocList/Basic.lean
```diff
@@ -0,0 +1,196 @@
+/-
+Copyright (c) 2019 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Leonardo de Moura, Mario Carneiro
+-/
+module
+
+public import Batteries.Data.List.Basic
+
+@[expose] public section
```
fgdorais, line 10: Please refine. There are several `*TR` defs here that don't need that much visibility. If many defs need to be exposed, that suggests the `Lemmas` file needs improvement.

Batteries/Data/AssocList/Lemmas.lean
```diff
@@ -0,0 +1,146 @@
+/-
+Copyright (c) 2019 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Leonardo de Moura, Mario Carneiro
+-/
+module
+
+public import Batteries.Data.AssocList.Basic
+
+@[expose] public section
```
fgdorais: There should be no need for `@[expose]` in this file.

```suggestion
public section
```

Batteries/Data/AssocList/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+`AssocList α β` is a `List (α × β)` with the pair structure flattened into the list constructor.
+This file defines the type and its operations; the lemmas relating those operations to the
+corresponding `List` operations along `AssocList.toList` are in `Batteries.Data.AssocList.Lemmas`.
+-/
+
+namespace Batteries
+
+/--
+`AssocList α β` is "the same as" `List (α × β)`, but flattening the structure
+leads to one fewer pointer indirection (in the current code generator).
```
fgdorais, line 24: There's been a lot of improvements since this was written. Please check that this claim is still true. (I think it's probably still true.)

fgdorais (changes requested):

fgdorais (approved):

